PR #5, opened 2022-01-07
Lf sunspec modules

leofernandeso:


corneliusclaussen (changes requested):
In void sunspec_readerImpl::run_read_loop():

This while(1) loop runs in a child Everest::Thread of the main thread. On exit of the main thread the destructor of Everest::Thread() the child thread joins, waiting for infinite time here as far as I can see. You should query Thread::shouldExit() i your loop similar to this:

if (read_loop_thread.shouldExit()) break;

This will return from the infinite loop and ending the thread, allowing the join() in the Everest::Thread destructor to finish and finally ending the whole program normally.

PR #6, opened 2022-01-11
Add a OCPP-J 1.6 charge point module

hikinggrass:


corneliusclaussen (commented):
OCPP::init() is reading its own config file for charge point config from file system. Is it because of the limitation that we cannot store objects in config? Otherwise we should use overlays with the framework config here.

hikinggrass:
> OCPP::init() is reading its own config file for charge point config from file system. Is it because of the limitation that we cannot store objects in config? Otherwise we should use overlays with the framework config here.

Yes, that's the main reason for that. Once we have (properly typed) object types in the framework config I would change this.

corneliusclaussen (approved):

PR #8, opened 2022-01-12
Pionix juraed additional prerequisites

pionix-juraed:
added some prerequisites and ToDo is done, moved edm link to below prerequisites


LAD101work (approved):
I currently have no way to test on Ubuntu but the OpenSUSE part looks ok!

PR #9, opened 2022-01-17
Add JsEnergyManager status output

LAD101work:
Obtaining running status information on the Energy Manager has been impossible before. This change adds a status flag to the JsEnergyManager module which gets published both to the 'external' mqtt topic (for legacy reasons; will be removed in future) and the module-specific 'energy_manager:JsEnergyManager' topic. 

changes:
- add wrapper functions for enabling/disabling JsEnergyManager 'running' var to publish the var's state via mqtt
- add 'emgr_running' var to manifest

### DIFF
--- a/interfaces/energy_manager.json
+++ b/interfaces/energy_manager.json
@@ -4,6 +4,10 @@
         "logging":{
             "description": "Various data logging",
             "type": "object"
+        },
+        "emgr_running": {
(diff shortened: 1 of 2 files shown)


LAD101work:
DCO did not work again. Branch history is getting too dirty to proceed, abandoning branch and trying again.

PR #10, opened 2022-01-17
Add JsEnergyManager status output

LAD101work:
Obtaining running status information on the Energy Manager has been impossible before. This change adds a status flag to the JsEnergyManager module which gets published both to the `external` mqtt topic (for legacy reasons; will be removed in future) and the module-specific `energy_manager:JsEnergyManager` topic.

changes:
- add wrapper functions for enabling/disabling JsEnergyManager `running` var to publish the var's state via mqtt
- add `emgr_running` var to manifest


a-w50 (approved):

a-w50 (commented):
I wonder if in future the energy manager should be allowed to be non-running, right now the current implementation suggests so.  But I guess the energy manager is a central module which should be always running.  Maybe the variable name "running" is misleading ...

LAD101work:
Given that if "running" is false, it ceases its functionality and just returns on the periodic and event-based control functions, would you say "inactive" would be a better description? Or do you have any other suggestions?

a-w50:
Yes, that sounds more reasonable

PR #13, opened 2022-01-19
Fix number of phase publish in yeti driver

corneliusclaussen:
Signed-off-by: Cornelius Claussen <cc@cclaussen.de>


a-w50 (commented):
Did you run clang-format?

corneliusclaussen:
No, this would change most of the lines in this file and makes it really hard to find the actual change when reading the diffs later on. If we want to enforce clang-format we should reformat the files in a seperate commit that only does formatting changes and not actual code changes first.

a-w50 (approved):
Ok, you could still selectivly run clang-format, I guess

PR #14, opened 2022-01-24
Add force_unlock/powermeter/telemetry to evsemanager

corneliusclaussen:
The force_unlock command should be used by OCPP if requested by central
management system.
The powermeter interface is proxied to the actual power meter in use by
the evse to charge the car. Connect the OCPP powermeter requirement to
this one instead of directly connecting it to the actual hw driver.


modules/EvseManager/EvseManager.cpp
```diff
@@ -19,11 +21,16 @@ void EvseManager::ready() {
     });
 
     r_powermeter->subscribe_powermeter([this](json p) {
+        // Inform charger about current charging current. This is used for slow OC detection.
```
hikinggrass, line 24: Is current changging current a typo?

modules/EvseManager/evse_energy_control/evse_manager_energy_controlImpl.hpp
```diff
@@ -1,5 +1,3 @@
-// SPDX-License-Identifier: Apache-2.0
```
hikinggrass, line 1: How did the license header disappear here? Could this be a bug in ev-cli?

modules/EvseManager/powermeter/powermeterImpl.hpp
```diff
@@ -0,0 +1,63 @@
+#ifndef POWERMETER_POWERMETER_IMPL_HPP
```
hikinggrass, line 1: Again, no license header

modules/YetiDriver/ld-ev.hpp
```diff
@@ -1,5 +1,3 @@
-// SPDX-License-Identifier: Apache-2.0
```
hikinggrass, line 1: Missing license header.

hikinggrass (changes requested):
Looks good, just a few missing license headers and maybe a typo.

a-w50 (commented):
It seems, that part of the changes have been done with the old auto-generator and other parts with the new one.  Probably you want to do an force update with ev-cli.
Concerning the telemetry data, we can see here, that would we really should go on with global type definitions

modules/EvseManager/EvseManager.cpp
corneliusclaussen: No, english words just have too many meanings.

modules/EvseManager/powermeter/powermeterImpl.hpp
corneliusclaussen: YetiDriver and EvseManager updated with edm mod update --force to fix license tags

modules/YetiDriver/ld-ev.hpp
corneliusclaussen: solved by running edm mod update --force

modules/EvseManager/evse_energy_control/evse_manager_energy_controlImpl.hpp
corneliusclaussen: solved by running edm mod update --force

corneliusclaussen:
Currently this exposes a powermeter interface on evsemanager. This will need to be changed once we have json types, the powermeter var should be on the main interface instead of a seperate one. Seperating here creates a dependency mess: OCPP and friends should NEVER use a powermeter on its own, this would mean that e.g. start/end of session kwh values come from a different powermeter then the live values. Also with a 1:n dependency to evsemanagers it is even harder to match the correct power meters to the correct evsemanagers.

a-w50 (approved):
Licenses fixed now, perfect

hikinggrass (approved):

PR #19, opened 2022-02-21
Add complex type support

hikinggrass:
Add types folder, restructure generated headers, update modules.
Headers for interfaces, modules and types are now generated in their own subfolder of build/generated/generated.
Some modules have been updated to use types, all have been updated to use the new header includes.


types/CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
+file(GLOB TYPES_FILES
+    ${CMAKE_CURRENT_SOURCE_DIR}/*.json
+)
+
+add_custom_target(generate_type_files
+    DEPENDS
+        ${GENERATED_OUTPUT_DIR}/types
+)
+
+# FIXME (aw): right now, the header files are generated in-source for convenience
```
a-w50, line 10: Where does this comment comes from? I guess it's stale

hikinggrass: I found the exact same comment in interfaces/CMakeLists.txt which was the base for this file

a-w50 (approved):
Everything fine, though I don't know if everything will still be running as before?

hikinggrass:
> Everything fine, though I don't know if everything will still be running as before?

I just found a new bug, that I'll have to fix before being able to merge this

PR #21, opened 2022-03-09
Changed verbosity and dependencies for npm install

a-w50:
- npm install is now only issued if required and doesn't log any more on
  success


hikinggrass (approved):
Looks good, we can probably move (most of) this to a js_module_setup() function like for the c++ modules

PR #27, opened 2022-03-14
Add rebuild.sh that tries to build binaries from the source

corneliusclaussen:
tries to clear all caches etc and build binaries that actually match the source code


corneliusclaussen:
closing PR but keeping branch

PR #28, opened 2022-03-16
Combine changes from branch cc-ld-refac-energymanager and main

LAD101work:
This PR merges new functionality from main with the changes made to the way the control network of modules is built up in cc-ld-refac-energymanager.

The new system has a tree structure:
EnergyManager (root node) --[n:m]--> EnergyNode(s) (hierarchical nodes in the middle of the tree) --[n:m]--> EvseManager(s) (leaf nodes of the control tree)

1.) "User" settings/set points are entered into the leaf nodes and then propagated upwards through the tree. 
2.) EnergyNodes that are passed by this information add their limits and schedule information.
3.) The EnergyManager calculates the optimized schedule and sets the corresponding limits. It then calls `enforce_limits` functions addressed at each node individually with their respective limits and timeframes.
4.) The EvseManager(s) charge according to their limits obtained from the EnergyManager. (If an EvseManager cannot, for whatever reason, get a currently applicable limit and the time-based validity of its scheduled limits has run out, it stops charging and waits for further limits.)

This basic tree consists of:
- one EnergyManager
- one EnergyNode (which acts as grid connection point AND fuse)
- one EvseManager (with attached Yeti driver)

Price information gets collected at the `grid_connection_point` and merged into the upward-propagating `energy` data object.

Currently the EnergyManager only optimizes for `manual limits`.


corneliusclaussen:
The following modules can be removed and will not be readded as they are from the old arch:
JsPowerIn, Power, PowerIn

The following modules can be removed but will be readded as a C++ rewrite (probably not sharing code with the current implementation) in the new arch somehow:
Solar, JsSimPowerMeter

LAD101work:
> Looks like you committed your .vscode folder, please remove

Removed!

LAD101work:
Wrong action, reopening PR

corneliusclaussen (changes requested):
This PR cannot be rebased as it merged serveral times with different branches including main. A new PR will be provided.

corneliusclaussen:
Closing this PR, a new rebased one will be created.

PR #31, opened 2022-03-24
Fix a race condition in EnergyNode

corneliusclaussen:
Fix a race condition that happens if power meter reports an energy object before the first child does due to missing initialization.

Signed-off-by: Cornelius Claussen <cc@cclaussen.de>


LAD101work (approved):
Very good! Works.

PR #32, opened 2022-03-25
Update OCPP SIL config to new energy manager structure

hikinggrass:


LAD101work (approved):
Very good, from what I can tell this should work.

PR #33, opened 2022-03-25
Add a SQlite backed persistant key-value store module

hikinggrass:

### DIFF
--- a/modules/PersistentStore/main/kvsImpl.cpp
+++ b/modules/PersistentStore/main/kvsImpl.cpp
@@ -1,0 +1,234 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2022 - 2022 Pionix GmbH and Contributors to EVerest
+
+#include "kvsImpl.hpp"
+
+#include <boost/filesystem.hpp>
+#include <boost/variant.hpp>
+
+namespace module {
+namespace main {
+
+void kvsImpl::init() {
+    // open and initialize database
+    boost::filesystem::path sqlite_db_path =
+        boost::filesystem::absolute(boost::filesystem::path(mod->config.sqlite_db_file_path));
+    boost::filesystem::path database_directory = sqlite_db_path.parent_path();
+    if (!boost::filesystem::exists(database_directory)) {
+        boost::filesystem::create_directories(database_directory);
+    }
+
+    int ret = sqlite3_open(sqlite_db_path.c_str(), &this->db);
+
+    if (ret != SQLITE_OK) {
+        EVLOG(error) << "Error opening PersistentStore database '" << sqlite_db_path << "': " << sqlite3_errmsg(db);
+        throw std::runtime_error("Could not open PersistentStore database at provided path.");
+    }
+
+    EVLOG(debug) << "Using SQLite version " << sqlite3_libversion();
+
+    // prepare the database
+    std::string create_sql = "CREATE TABLE IF NOT EXISTS KVS ("
+                             "KEY   TEXT UNIQUE,"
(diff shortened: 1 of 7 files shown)


modules/PersistentStore/main/kvsImpl.cpp
```diff
@@ -0,0 +1,234 @@
...
+        boost::filesystem::absolute(boost::filesystem::path(mod->config.sqlite_db_file_path));
+    boost::filesystem::path database_directory = sqlite_db_path.parent_path();
+    if (!boost::filesystem::exists(database_directory)) {
+        boost::filesystem::create_directories(database_directory);
+    }
+
+    int ret = sqlite3_open(sqlite_db_path.c_str(), &this->db);
+
+    if (ret != SQLITE_OK) {
+        EVLOG(error) << "Error opening PersistentStore database '" << sqlite_db_path << "': " << sqlite3_errmsg(db);
```
a-w50, line 24: There is this combined EVLOG_AND_THROW macro, although we have no guidelines on when to use it

modules/PersistentStore/main/kvsImpl.cpp
```diff
@@ -0,0 +1,234 @@
...
+    if (sqlite3_finalize(create_statement) != SQLITE_OK) {
+        EVLOG(error) << "Error creating KVS table";
+        throw std::runtime_error("PersistentStore db access error");
+    }
+}
+
+void kvsImpl::ready() {
+}
+
+class TypeNameVisitor : public boost::static_visitor<std::string> {
```
a-w50, line 53: I would have put both visitors into one, returning a struct with containing type and text

modules/PersistentStore/main/kvsImpl.cpp
```diff
@@ -0,0 +1,234 @@
...
+
+    sqlite3_bind_text(select_statement, 1, key.c_str(), -1, NULL);
+
+    int res = sqlite3_step(select_statement);
+    if (res != SQLITE_ROW) {
+        // no key with that name exists in the database
+        return {};
+    }
+
+    boost::variant<boost::blank, Array, Object, bool, double, int, std::string> value;
```
a-w50, line 160: you could use decltype and result_of here for this function to deduce the type

modules/PersistentStore/main/kvsImpl.cpp
```diff
@@ -0,0 +1,234 @@
...
+            }
+        }
+    }
+
+    if (sqlite3_finalize(select_statement) != SQLITE_OK) {
+        EVLOG(error) << "Error selecting from KVS table";
+        throw std::runtime_error("PersistentStore db access error");
+    }
+
+    return value;
```
a-w50, line 195: concerning the interface definition we still have the problem here, that we can set a blank item, but we wont find out if it was set, because the blank return also means "not existent"

modules/PersistentStore/main/kvsImpl.cpp
```diff
@@ -0,0 +1,234 @@
...
+                value = std::stod(value_str);
+            } else if (type_str == "int") {
+                value = std::stoi(value_str);
+            } else if (type_str == "std::string") {
+                value = value_str;
+            }
+        }
+    }
+
+    if (sqlite3_finalize(select_statement) != SQLITE_OK) {
```
a-w50, line 190: could you move this statement above the if block? Then you can return directly from inside the if statement and don't need that value variable

a-w50 (approved):
I didn't unit test it and I'm not familiar with the sqlite api. But structure-wise it looks like it should.
The db_mutex can be dropped right?

PR #34, opened 2022-03-25
Use cmake PROJECT_SOURCE_DIR instead of CMAKE_SOURCE_DIR

hikinggrass:
This makes it possible to have this repository as a dependency and not break the interfaces and modules generation


a-w50 (approved):
Might be changed in future again, because the interfaces should be generated in a single place

PR #35, opened 2022-04-12
Add API module

hikinggrass:

### DIFF
--- a/modules/API/API.cpp
+++ b/modules/API/API.cpp
@@ -1,0 +1,191 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2020 - 2022 Pionix GmbH and Contributors to EVerest
+#include "API.hpp"
+#include <chrono>
+#include <date/date.h>
+
+namespace module {
+
+std::string to_rfc3339(std::chrono::time_point<std::chrono::system_clock> t) {
+    return date::format("%FT%TZ", std::chrono::time_point_cast<std::chrono::milliseconds>(t));
+}
+
+SessionInfo::SessionInfo() :
+    state("Unknown"), start_energy_wh(0), end_energy_wh(0), latest_total_w(0) {
+    this->start_time_point = std::chrono::system_clock::now();
+    this->end_time_point = this->start_time_point;
+}
+
+bool SessionInfo::is_state_charging(std::string current_state) {
+    if (current_state == "SessionStarted" || current_state == "AuthRequired" || current_state == "ChargingStarted" ||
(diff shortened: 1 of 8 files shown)


modules/API/manifest.json
```diff
@@ -0,0 +1,25 @@
+{
+    "description": "The EVerest API module, exposing some internal functionality on an external MQTT connection.",
+    "config": {
+    },
+    "provides": {
+        "main": {
+            "description": "EVerest API",
+            "interface": "empty"
```
a-w50, line 8: We could implement the complete logic in the module and don't provide anything

hikinggrass: That would be great, however our tooling doesn't really support modules that do not provide at least one empty interface at the moment

a-w50: Isn't that only the schema requirement?

hikinggrass: Yes, but isn't that just a technicality? I can't create a new module with

1. an empty provides object in the manifest
2. without a provides entry in the manifest

We could certainly remove this requirement from the schema, but we should check that it doesn't break any assumptions in everest-framework

a-w50: I see, i wouldn't impose this requirement - because we shouldn't force a module to provide something. On the other hand it should also be possible to only provide something and not to require something, which is already possible.
But if we don't require AND provide anything, then it might look a bit strange

hikinggrass: I agree, we should absolutely change this in everest-framework.
One could certainly imagine some module that doesn't interact with any other modules but still does some external mqtt communication. Such a module probably would not require AND provide anything. Might not be too useful, but maybe there are certain specialized use cases for that.

modules/API/API.cpp
```diff
@@ -0,0 +1,191 @@
...
+
+        count += 1;
+    }
+}
+
+void API::ready() {
+    invoke_ready(*p_main);
+}
+
+std::string API::sanitize_event(const std::string& event) {
```
corneliusclaussen, line 162: From my understanding this function converts the time series of events that comes in from EvseManager into an internal state of this module that is useful for display, so I would suggest to rename this function to e.g. update_state(event) or something better. I think it is important as the difference between events and states can be confusing anyway. This function may actually update several internal states in the future such as reservation_state as well as the events to control this one also arrive here.

corneliusclaussen (commented):
see inline comment about events vs states

modules/API/API.cpp
hikinggrass: Sounds good

corneliusclaussen (approved):

PR #36, opened 2022-04-12
Passing on reservation requests to EvseManager

chGoodchild:


corneliusclaussen:
Before we merge this it needs to work. I'm still working on implementing the evsemanager stuff. Since this branch is based on cc-evsemanager-add-reservations which is based on main we should wait until I'm done with the cc- branch and then rebase on that one.

chGoodchild:
will rebase onto Cornelius's branch and then open a pull request...

PR #38, opened 2022-04-19
Add support for building modules as binaries

hikinggrass:


a-w50 (approved):
See my comments on EVerest/everest-framework#28
There is quite large "code duplication" in this commit

PR #40, opened 2022-04-26
Add price limit optimization

LAD101work:
Optimizer in EnergyManager now can assign current while observing the following criteria:

- 2 modes are selectable: "Manual Limits" (always try to charge with up to a set current limit) and "Price Optimized" (always charge with up to a set current limit when import price for energy is below a set threshold)
- when not enough current is available for all charging goals, current will be equally scaled down until configured minimum current has been reached
- if not enough current is available to supply all endpoints even with only their configured minimum currents, start dropping current requests for random nodes and retry with the remainder (will be recalculated every 5 seconds but the order of nodes to be dropped will be the order in which nodes provide their UUID to the EnergyManager)


modules/EvseManager/Charger.cpp
```diff
@@ -121,6 +121,8 @@ void Charger::runStateMachine() {
         // retry on Low level etc until SLAC matched.
         ISO_IEC_Coordination();
 
+        // EVLOG(warning) << "getAuthorization(): " << getAuthorization() << "; powerAvailable(): " << powerAvailable();
+
```
hikinggrass, line 125: This is probably not needed anymore

modules/EvseManager/Charger.cpp
```diff
@@ -752,11 +754,22 @@ void Charger::checkSoftOverCurrent() {
...
 bool Charger::powerAvailable() {
+    
+    
+    // std::chrono::system_clock::time_point now = maxCurrentValidUntil;
+    // std::time_t now_c = std::chrono::system_clock::to_time_t(now);
+    // std::tm now_tm = *std::localtime(&now_c);
+    // std::stringstream ss;
+    // ss << std::put_time(&now_tm, std::string("UTC: %Y-%m-%d %H:%M:%S").c_str());
+    // EVLOG(warning) << "maxCurrentValidUntil: " << ss.str();
+
```
hikinggrass, line 765: This is probably not needed anymore

modules/EvseManager/Charger.cpp
```diff
@@ -752,11 +754,22 @@ void Charger::checkSoftOverCurrent() {
...
     if (maxCurrentValidUntil < std::chrono::system_clock::now()) {
         maxCurrent = 0.;
         signalMaxCurrent(maxCurrent);
+        //EVLOG(warning) << "maxCurrentValidUntil is in the past!";
```
hikinggrass, line 769: This is probably not needed anymore

modules/EvseManager/Charger.cpp
```diff
@@ -752,11 +754,22 @@ void Charger::checkSoftOverCurrent() {
...
     }
 
+    // EVLOG(warning) << "getMaxCurrent(): " << getMaxCurrent();
```
hikinggrass, line 772: This is probably not needed anymore

hikinggrass (changes requested):

modules/EvseManager/Charger.cpp
LAD101work: Right, that was old debug code. Good catch! Removed.

hikinggrass (approved):

modules/EnergyManager/manifest.json
```diff
@@ -8,7 +8,9 @@
     },
     "requires": {
         "energy_trunk": {
-            "interface": "energy"
+            "interface": "energy",
+            "min_connections": 1,
+            "max_connections": 12
```
corneliusclaussen, line 13: Not sure that we should support multiple energy_trunks here. That would mean totally independent house power connections. The danger is that someone just connects multiple evses to the energymanager here, which would not do what the user expects. I would suggest to leave it at a single energy_trunk and manage only one tree per energymanager. If someone really has two grid connections to his site he can still load two independent energy managers, which would be more explicit

corneliusclaussen:
Was not introduced in this PR but we could fix it here: static json global_energy_object is accessed by several threads but has no lock guards. We need to add (scoped) locks everywhere. This could be a source of the spurious segfaults

corneliusclaussen:
// error: wrong currency / currency not (yet) implemented
-> why do we need to add currencies explicitly? the idea was to just carry the currency string along with the numbers and show them in the UI

modules/EnergyManager/manifest.json
LAD101work: That is true. Reduced to single connection.

PR #41, opened 2022-04-26
Add ocpp firmware management

hikinggrass:


modules/OCPP/firmware_updater.sh
```diff
@@ -0,0 +1,16 @@
...
+sleep 2
+if [[ $curl_exit_code -eq 0 ]]
+then
+    echo "downloaded"
+else
+    echo "download_failed"
+fi
+sleep 2
+echo "installing"
+sleep 2
```
a-w50, line 15: these sleep 2 are just for emulating, right?

modules/OCPP/OCPP.cpp
```diff
@@ -57,6 +62,72 @@ void OCPP::init() {
...
+    });
+    this->charge_point->register_update_firmware_callback([this](std::string location) {
+        // // create temporary file
+        std::string date_time = "DATETIME"; // FIXME
+        std::string file_name = "firmware-" + date_time + "-%%%%-%%%%-%%%%-%%%%";
+
+        auto firmware_file_name = boost::filesystem::unique_path(boost::filesystem::path(file_name));
+        auto firmware_file_path = boost::filesystem::temp_directory_path() / firmware_file_name;
+
+        this->update_firmware_thread = std::thread([this, location, firmware_file_path]() {
```
a-w50, line 106: there's no detach or join for the threads, this might lead to hard to understand exits in case of exceptions

modules/OCPP/OCPP.cpp
```diff
@@ -57,6 +62,72 @@ void OCPP::init() {
...
     });
+    this->charge_point->register_upload_diagnostics_callback([this](std::string location) {
+        // create temporary file
+        std::string date_time = "DATETIME"; // FIXME
+        std::string file_name = "diagnostics-" + date_time + "-%%%%-%%%%-%%%%-%%%%";
+
+        auto diagnostics_file_name = boost::filesystem::unique_path(boost::filesystem::path(file_name));
+        auto diagnostics_file_path = boost::filesystem::temp_directory_path() / diagnostics_file_name;
+
+        auto diagnostics = json::object({{"diagnostics", json::object({{"key", "value"}})}});
```
a-w50, line 73: why not just json({"diagnostics", {"key", "value"}})?

modules/OCPP/OCPP.cpp
```diff
@@ -57,6 +62,72 @@ void OCPP::init() {
...
+        diagnostics_file << diagnostics.dump();
+        this->upload_diagnostics_thread = std::thread([this, location, diagnostics_file_name, diagnostics_file_path]() {
+            auto diagnostics_uploader = boost::filesystem::path("bin/diagnostics_uploader.sh");
+
+            boost::process::ipstream stream;
+            std::vector<std::string> args = {location, diagnostics_file_name.string(), diagnostics_file_path.string()};
+            boost::process::child cmd(diagnostics_uploader, boost::process::args(args),
+                                      boost::process::std_out > stream);
+            std::string temp;
+            while (std::getline(stream, temp)) {
```
a-w50, line 84: that's a nice solution. Probably I should have also used this boost process functionality in the manager.cpp

a-w50 (approved):
I can't test the code, but from logical point of view, looks very good

modules/OCPP/OCPP.cpp
corneliusclaussen: We should add the detach here before merging

hikinggrass: done

PR #45, opened 2022-05-02
Add command to modify a CarSimulator charging session whilst running

hikinggrass:
This isn't the most elegant solution, for that we should implement a proper interactive car simulator one day, but this change allows us to modify a running charging session by replacing all remaining simulation commands with new ones


corneliusclaussen (approved):
Looks good!

PR #46, opened 2022-05-05
Fix install of tibber module that could sporadically fail

hikinggrass:

### DIFF
--- a/modules/JsTibber/CMakeLists.txt
+++ b/modules/JsTibber/CMakeLists.txt
@@ -1,12 +1,13 @@
 get_filename_component(MODULE_NAME ${CMAKE_CURRENT_SOURCE_DIR} NAME)
 add_custom_target(${MODULE_NAME} ALL)
-add_dependencies(${MODULE_NAME} everestjs)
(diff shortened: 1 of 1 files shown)


a-w50 (approved):
Finally, hopefully this npm install stuff won't annoy us that much anymore.  I probably wouldn't call the target ${MODULE_NAME}_INSTALL_NODE_MODULES, because it doesn't install them to dist yet, it's more like a fetch?

hikinggrass:
> Finally, hopefully this npm install stuff won't annoy us that much anymore. I probably wouldn't call the target ${MODULE_NAME}_INSTALL_NODE_MODULES, because it doesn't install them to dist yet, it's more like a fetch?

This is pretty much a 1:1 copy of the [CMakeLists.txt of the jsPN532TokenProvider](https://github.com/EVerest/everest-core/blob/main/modules/JsPN532TokenProvider/CMakeLists.txt) hence the naming from there. However, looking at it we could maybe move all of this duplicated code into "js_module_setup()" function like the already existing "cpp_module_setup()" for C++ modules

PR #49, opened 2022-05-17
Add libcurl4-openssl-dev to documentation

LAD101work:
Not sure if curl is missing in the package list for OpenSuSE as well. Maybe someone else knows?


LAD101work:
Changes no longer needed as problem was solved another way. Branch closed

PR #51, opened 2022-05-17
Correctly set CMAKE_MODULE_PATH to subdirectory of PROJECT_SOURCE_DIR

hikinggrass:
This is the exact same bug that everest-framework had a while ago, this needs to be fixed to allow everest-core to be a dependency of another repository for out-of-tree modules.


a-w50 (approved):
:100:

PR #53, opened 2022-05-25
Add setup module for wifi connectivity

hikinggrass:
This module can setup wifi connectivity and might be extended to perform various other setup tasks as well


corneliusclaussen (approved):
Looks good!

PR #68, opened 2022-07-04
Add a C++ implementation of the PN532TokenProvider

hikinggrass:

### DIFF
--- a/modules/PN532TokenProvider/main/auth_token_providerImpl.cpp
+++ b/modules/PN532TokenProvider/main/auth_token_providerImpl.cpp
@@ -20,6 +20,7 @@ void auth_token_providerImpl::ready() {
     serial.run();
 
     serial.reset();
+    // configure Secure Access Module (SAM)
     auto configure_sam = serial.configureSAM();
     if (configure_sam.get()) {
         EVLOG_debug << "Configured SAM" << std::endl;
--- a/modules/PN532TokenProvider/main/auth_token_providerImpl.cpp
+++ b/modules/PN532TokenProvider/main/auth_token_providerImpl.cpp
@@ -1,0 +1,57 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2022 - 2022 Pionix GmbH and Contributors to EVerest
+
+#include "auth_token_providerImpl.hpp"
+
+namespace module {
+namespace main {
+
+void auth_token_providerImpl::init() {
+    // initialize serial driver
+    std::cerr << "port: " << config.serial_port << " baud: " << config.baud_rate;
+    if (!serial.openDevice(config.serial_port.c_str(), config.baud_rate)) {
+        EVLOG_AND_THROW(EVEXCEPTION(Everest::EverestConfigError, "Could not open serial port ", config.serial_port,
+                                    " with baud rate ", config.baud_rate));
+        return;
+    }
+}
+
+void auth_token_providerImpl::ready() {
+    serial.run();
+
+    serial.reset();
+    auto configure_sam = serial.configureSAM();
+    if (configure_sam.get()) {
+        EVLOG_debug << "Configured SAM" << std::endl;
+    }
+
+    auto firmware_version = serial.getFirmwareVersion().get();
+    if (firmware_version.valid) {
+        std::shared_ptr<FirmwareVersion> fv = std::dynamic_pointer_cast<FirmwareVersion>(firmware_version.message);
+        EVLOG_info << "PN532 firmware version: " << std::to_string(fv->ver) << "." << std::to_string(fv->rev)
+                  << std::endl;
+    }
+
+    while (true) {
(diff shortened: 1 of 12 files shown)


modules/PN532TokenProvider/main/auth_token_providerImpl.cpp
```diff
@@ -0,0 +1,57 @@
...
+    }
+}
+
+void auth_token_providerImpl::ready() {
+    serial.run();
+
+    serial.reset();
+    auto configure_sam = serial.configureSAM();
+    if (configure_sam.get()) {
+        EVLOG_debug << "Configured SAM" << std::endl;
```
a-w50, line 25: Maybe a bit more verbose? What is SAM?

modules/PN532TokenProvider/main/auth_token_providerImpl.cpp
```diff
@@ -0,0 +1,57 @@
...
+
+    while (true) {
+        auto in_list_passive_target_response = serial.inListPassiveTarget().get();
+        if (in_list_passive_target_response.valid) {
+            std::shared_ptr<InListPassiveTargetResponse> in_list_passive_target_response_message =
+                std::dynamic_pointer_cast<InListPassiveTargetResponse>(in_list_passive_target_response.message);
+            auto target_data = in_list_passive_target_response_message->target_data;
+            for (auto entry : target_data) {
+                EVLOG_info << "Got raw rfid/nfc tag: " << entry.getNFCID() << std::endl;
+                Object rfid_token;
```
a-w50, line 43: could use the json initializer list notation here

modules/PN532TokenProvider/main/auth_token_providerImpl.cpp
```diff
@@ -0,0 +1,57 @@
...
+                rfid_token["token"] = entry.getNFCID();
+                rfid_token["type"] = "rfid";
+                rfid_token["timeout"] = config.timeout;
+                EVLOG_info << "Publishing new rfid/nfc token: " << rfid_token;
+                this->publish_token(rfid_token);
+            }
+        }
+
+        std::this_thread::sleep_for(std::chrono::seconds(config.read_timeout));
```
a-w50, line 52: do we really sleep for seconds?

modules/PN532TokenProvider/pn532_serial/evSerial.h
```diff
@@ -0,0 +1,107 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2020 - 2022 Pionix GmbH and Contributors to EVerest
+#ifndef YETI_SERIAL
```
a-w50, line 3: wrong define here

a-w50 (changes requested):
I guess, this could be now refactored, using the evSerial from the framework, right?
Concerning the low-level code, there are quite a lot of constants / magic numbers which could be cleaned up.  I would also advise to use packed structs if possible, to decode the messages.  Is the heavy use of promises, shared ptrs, dynamic casts really necessary?

modules/PN532TokenProvider/main/auth_token_providerImpl.cpp
hikinggrass: Yes, that way we wont see the same token multiple times when its not immediately removed

hikinggrass:
> I guess, this could be now refactored, using the evSerial from the framework, right? Concerning the low-level code, there are quite a lot of constants / magic numbers which could be cleaned up. I would also advise to use packed structs if possible, to decode the messages. Is the heavy use of promises, shared ptrs, dynamic casts really necessary?

I'd merge it like this and then apply the general serial/date/threads PR afterwards

modules/PN532TokenProvider/main/auth_token_providerImpl.cpp
a-w50: ah ok, but that's quite hacky, isn't it? We would need some kind of debouncing.  But maybe this simple solution works best

hikinggrass: It is, but I'll have to look into the spec a little more to see if there's something the module does to help with debouncing

corneliusclaussen: Most implementations I have seen generate two events: the first when the card is seen and the second when the card leaves the RFID reader. The time between those two events is also sometimes used as a user input: some charging stations can be resetted by holding the card down for more then 10 seconds or so

a-w50 (approved):

modules/PN532TokenProvider/main/auth_token_providerImpl.cpp
hikinggrass: I just looked into the implementation of the js version we are using at the moment, they also use a polling interval: https://github.com/techniq/node-pn532/blob/master/src/pn532.js#L47

PR #69, opened 2022-07-05
Remove obsolete cross compile option, fix OCPP module targets

hikinggrass:


hikinggrass:
> Did the changes to the OCCP cmakefile slip in here? Otherwise the removal of the cross-compile option is ok

They are not strictly related to the removal of the cross compile option, but were something I stumbled upon whilst doing the cross compile fixes. I'll rename the PR accordingly

a-w50 (approved):

PR #71, opened 2022-07-05
Use Everest::Date/Serial/Thread where applicable

hikinggrass:


a-w50 (dismissed):
the refactoring is alright, but the addition of the rfid driver should not end up here

hikinggrass:
@a-w50 can you check if this still compiles/works. should be able to merge by now

corneliusclaussen (approved):

PR #73, opened 2022-07-08
Mapping and transmitting errors from evsemanager to OCPP errors now

Pietfried:
should be merged together with https://github.com/EVerest/libocpp/pull/31

### DIFF
--- a/modules/OCPP/OCPP.cpp
+++ b/modules/OCPP/OCPP.cpp
@@ -373,11 +373,7 @@ void OCPP::init() {
                 this->charge_point->plug_disconnected(connector);
             } else if (event == "Error") {
                 auto error = session_events["error"];
-                if (error == "OverCurrent") {
-                    this->charge_point->error(connector, ocpp1_6::ChargePointErrorCode::OverCurrentFailure);
-                } else {
-                    this->charge_point->vendor_error(connector, error);
-                }
+                this->charge_point->error(connector, error);
             } else if (event == "PermanentFault") {
                 this->charge_point->permanent_fault(connector);
             } else if (event == "ReservationStart") {
--- a/modules/OCPP/OCPP.cpp
+++ b/modules/OCPP/OCPP.cpp
@@ -371,8 +371,8 @@ void OCPP::init() {
                 this->charge_point->stop_session(connector, ocpp1_6::DateTime(timestamp), energy_Wh_import);
                 this->charge_point->plug_disconnected(connector);
             } else if (event == "Error") {
-                auto evse_error = session_events["error"];
-                ocpp1_6::ChargePointErrorCode ocpp_error_code = this->getOcppErrorCode(evse_error);
+                const auto evse_error = session_events["error"];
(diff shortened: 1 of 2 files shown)


modules/OCPP/OCPP.cpp
```diff
@@ -372,12 +371,9 @@ void OCPP::init() {
                 this->charge_point->stop_session(connector, ocpp1_6::DateTime(timestamp), energy_Wh_import);
                 this->charge_point->plug_disconnected(connector);
             } else if (event == "Error") {
-                auto error = session_events["error"];
-                if (error == "OverCurrent") {
-                    this->charge_point->error(connector, ocpp1_6::ChargePointErrorCode::OverCurrentFailure);
-                } else {
-                    this->charge_point->vendor_error(connector, error);
-                }
+                auto evse_error = session_events["error"];
```
a-w50, line 374: in general, always use const for constant variables.  With json dicts, please use .at(), because the bracket notation might just create an empty dict item, which is not what we want and also hard to debug. So the line should look like
const auto evse_error = session_events.at("error")

modules/OCPP/OCPP.cpp
```diff
@@ -414,4 +410,28 @@ void OCPP::ready() {
     invoke_ready(*p_auth_provider);
 }
 
+ocpp1_6::ChargePointErrorCode OCPP::getOcppErrorCode(json evse_error) {
```
a-w50, line 413: here you can pass by const reference, otherwise the object will be copied (which doesn't really hurt here)
furthermore we have this name alias (using Object = nlohmann::json), it's a bit easier to deduct the intended type.
so here one should write
const Object& evse_error

a-w50, line 413: furthermore, this is a completely static helper function, it doesn't need to be attached to the OCPP object, you could equally just write
static ocpp1_6::ChargePointErrorCode getOcppErrorCode(json evse_error)
also note, that we mostly use snake case for function names, that would be get_ocpp_error_code

modules/OCPP/OCPP.hpp
```diff
@@ -91,6 +91,8 @@ class OCPP : public Everest::ModuleBase {
     void init();
     void ready();
 
+    ocpp1_6::ChargePointErrorCode getOcppErrorCode(json evse_error);
```
a-w50, line 94: this function here will be deleted, the next time the ev-cli touches this file. in auto generated files you're only allowed to add code in between these // ev@... blocks
but if you make this function static and local to the OCPP.cpp, you don't need to declare it here at all

a-w50 (changes requested):
Logic-wise everything is ok.  C++ wise and w.r.t. to the framework, some minor changes are necessary

modules/OCPP/OCPP.cpp
Pietfried: I will make a separate PR for the syntactical and framework stuff, since that affects many places in the file

a-w50: ok

a-w50 (approved):
Perfect, looks good.

PR #77, opened 2022-07-18
Fixed error code conversion and parsing of user config

Pietfried:
- Fixed error code conversion and parsing of user config
- Added timeout to shell scripts


modules/OCPP/OCPP.cpp
```diff
@@ -32,6 +32,14 @@ static ocpp1_6::ChargePointErrorCode get_ocpp_error_code(const Object& evse_erro
     }
 }
 
+void create_user_config(const boost::filesystem::path& user_config_path) {
```
a-w50, line 35: should be rather called create_empty_user_config, because it creates an empty config

modules/OCPP/OCPP.cpp
```diff
@@ -283,7 +291,8 @@ void OCPP::init() {
         [this](ocpp1_6::SignedUpdateFirmwareRequest req, boost::filesystem::path file_path) {
             this->signed_update_firmware_thread = std::thread([this, req]() {
                 boost::process::ipstream install_stream;
-                auto firmware_installer = boost::filesystem::path("bin/signed_firmware_installer.sh");
+                auto firmware_installer =
```
a-w50, line 294: using const auto would be best practice here

modules/OCPP/OCPP.cpp
```diff
@@ -283,7 +291,8 @@ void OCPP::init() {
...
+                    boost::filesystem::path(this->config.ScriptsPath) / "signed_firmware_installer.sh";
```
a-w50, line 295: we used to use snake_case

modules/OCPP/diagnostics_uploader.sh
```diff
@@ -2,7 +2,7 @@
 
 echo "Uploading"
 sleep 2
-curl --progress-bar -T "${3}" "${1}"
+curl --progress-bar --connect-timeout 20 -T "${3}" "${1}"
```
a-w50, line 5: you're using the const literal 20 in 3 different files, maybe this could be refactored to use one single setting, like a constants.sh or similar which gets included

a-w50 (changes requested):

modules/OCPP/OCPP.cpp
```diff
@@ -50,17 +58,17 @@ void OCPP::init() {
...
 
-        auto user_config = json::parse(user_config_file);
-
-        EVLOG_info << "Augmenting chargepoint config with user_config entries";
-        json_config.merge_patch(user_config);
+        try {
+            auto user_config = json::parse(user_config_file);
+            EVLOG_info << "Augmenting chargepoint config with user_config entries";
+            json_config.merge_patch(user_config);
+        } catch (json::parse_error e) {
```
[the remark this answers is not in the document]
Pietfried: Yeah there was a version where a completely emtpy user config file was written elsewhere and that caused the parse_error. That should not happen anymore. 

I agree with logging, but just logging means that the libocpp cannot write into the user config later, this is why I preferred the reset with an empty user config.

a-w50 (approved):

PR #96, opened 2022-09-07
Extend setup module, add API documentation

hikinggrass:
- Only scan wifi when instructed, allow removal of individual networks
- Add a command to check if the host is "online"
- Add commands to enable/disable/select a network
- Add persistently stored language, mode and initialization status
The persistency functionality needs a PersistentStore module connected to the Setup module!

### DIFF
--- a/modules/Setup/README.md
+++ b/modules/Setup/README.md
@@ -147,5 +147,11 @@ To remove a wifi network a payload with the following format must be published t
 ### everest_api/setup/cmd/remove_all_networks
 If any arbitrary payload is published to this topic all wifi networks will be removed.
 
+### everest_api/setup/cmd/check_online_status
+If any arbitrary payload is published to this topic a ping will be sent to the host configured in the configuration key "online_check_host". Depending on the success of this ping a status of "online" or "offline" will be reported on the following topic:
+
+__everest_api/setup/var/online_status__
+
+
 ### everest_api/setup/cmd/reboot
 If any arbitrary payload is published to this topic the system will reboot.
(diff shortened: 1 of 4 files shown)


modules/Setup/README.md
```diff
@@ -0,0 +1,208 @@
...
+]
+```
+
+### everest_api/setup/cmd/add_network
+To add a wifi network a payload with the following format must be published to this topic:
+```json
+{
+    "interface": "wlan0",
+    "ssid": "Example",
+    "psk": "a_valid_pre_shared_key"
```
a-w50, line 130: I wonder if we should put the psk unencrypted onto mqtt

modules/Setup/Setup.cpp
```diff
@@ -298,6 +469,16 @@ std::vector<WifiList> Setup::list_configured_networks(std::string interface) {
             wifi_list.interface = interface;
             wifi_list.network_id = std::stoi(list_networks_results_columns.at(0));
             wifi_list.ssid = list_networks_results_columns.at(1);
+            wifi_list.connected = false;
+            if (wpa_cli_status_map.count("id") && wpa_cli_status_map.count("ssid") &&
```
a-w50, line 473: I would rather use a std::map<custom wpa_status_struct> for wpa_cli_status_map

a-w50 (approved):

modules/Setup/README.md
hikinggrass: Yeah, it would be great if we had a different way of doing this. But at the moment this mqtt connection is the only data link we have to the application

PR #97, opened 2022-09-07
Extend API module with limits and telemetry

hikinggrass:


modules/OCPP/OCPP.cpp
```diff
@@ -340,11 +340,11 @@ void OCPP::init() {
             this->charge_point->receive_power_meter(connector, powermeter_json); //
         });
 
-        evse->subscribe_limits([this, connector](Object limits) {
-            auto max_current = limits["max_current"].get<double>();
+        evse->subscribe_limits([this, connector](types::evse_manager::Limits limits) {
+            double max_current = limits.max_current;
```
a-w50, line 344: isnt const auto instead of double here possible?

a-w50 (approved):

PR #104, opened 2022-09-19
Added a auto choose net iface function to the SIL.

SebaLukas:
The first available iface with a link-local IPv6 address is then used.

### DIFF
--- a/modules/JsCarV2G/index.js
+++ b/modules/JsCarV2G/index.js
@@ -66,24 +66,22 @@ function JavaStartedDeferred(mqtt_base_path, module_name, mod) {
   });
 }
 
-const net_init = os.networkInterfaces();
+function choose_first_ipv6_local() {
+  const all_ifs = os.networkInterfaces();
 
-function choose_first_network_iface() {
-  for (let iface in net_init) {
-    for (let element of net_init[iface]) {
-      if (element.hasOwnProperty("address")) {
-        if (element.address.includes('fe80')) {
-          return iface
-        }
-      }
-    }
-  }
-  evlog.warning(`No necessary IPv6 link-local address was found!`)
-  return "eth0"
+  const found_if = Object.keys(all_ifs).find(
+    (iface) => all_ifs[iface].filter((info) => info.address.includes('fe80')).length > 0
+  );
+
+  if (found_if) return found_if;
+
+  evlog.warning('No necessary IPv6 link-local address was found!');
+  return 'eth0';
 }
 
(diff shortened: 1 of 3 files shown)


modules/JsCarV2G/index.js
```diff
@@ -66,9 +66,24 @@ function JavaStartedDeferred(mqtt_base_path, module_name, mod) {
...
+        if (element.address.includes('fe80')) {
+          return iface
+        }
+      }
+    }
+  }
+  evlog.warning(`No necessary IPv6 link-local address was found!`)
+  return "eth0"
+}
+
```
a-w50, line 84: Quite some codacy issues, here.  We need to get some setup for the js linting. There is already an .eslint.rc in the root folder of everest core.  Nevertheless, the following code would be more js-like, I guess:
```
function choose_first_ipv6_local() {
  const all_ifs = os.networkInterfaces();

  const found_if = Object.keys(all_ifs).find(
    (iface) => all_ifs[iface].filter((info) => info.address.includes('fe80')).length > 0
  );

  if (found_if) return found_if;

  evlog.warning('No necessary IPv6 link-local address was found!');
  return 'eth0';
}
```

a-w50 (changes requested):

modules/JsCarV2G/index.js
SebaLukas: Tried your suggestion. It is much better.

modules/JsRiseV2G/index.js
```diff
@@ -62,6 +62,19 @@ function JavaStartedDeferred(mqtt_base_path, module_name) {
...
+
+  if (found_if) return found_if;
+
+  evlog.warning('No necessary IPv6 link-local address was found!');
+  return 'eth0';
+}
+
 function check_network_interface(network_iface) {
   let networkfound = false;
   const net_init = os.networkInterfaces();
```
a-w50, line 80: this you won't need anymore

a-w50 (approved):

modules/JsRiseV2G/index.js
a-w50: I mean the net_init var

a-w50: if you use single quotes around "auto" codacy should be silent

SebaLukas: In the case of a HIL, it should still be checked whether the specified interface is available. Therefore I would leave the check_network_interface function with the net_init var in.

PR #105, opened 2022-09-20
Refactoring of Auth, OCPP and EvseManager

Pietfried:
- Implemented new Module Auth for authorization handling
- Implements authorization interface and reservation interface
- Supports different setups of readers and connectors
- Changed modules that implement token_provider interface to support new types

EvseManager:
- Divided Events into SessionEvents and TransactionEvents
- Module does not ask for authorization anymore, it is provided by cmd directly from Auth module

OCPP:
- Added requirement to system, auth and reservation interfaces
- Moved firmware update and log upload handling to System module
- Adjusted to new EvseManager types

System module:
- Added system module for log uploads and firmware updates

CMake refactoring
- cleaned up cmake files
- added cmake config style support for install target
- EDM compatibility
- added bundling support (bundle and bundle-tar targets)


Pietfried:
Should be merged with https://github.com/EVerest/libocpp/pull/41

corneliusclaussen (approved):

PR #114, opened 2022-09-26
Split powermeter types and interfaces into AC and DC

corneliusclaussen:


corneliusclaussen:
Splitting in AC and DC implementations is hard for modules such as API or OCPP, will create a new PR with a combined data structure

PR #117, opened 2022-09-26
Add examples for handwritten module documentation

andistorm:


LAD101work (approved):
Very good! A manual explanation (if done with a mind to help other devs) is always more helpful than autogenerated documentation, in my book. 
This will require quite some more work in the future to fill the gaps, but it is the right step!

PR #127, opened 2022-10-11
Use FetchContent to retreive everest-cmake if not present in parent path

hikinggrass:
This should allow the simple workflow of cloning just everest-core and building EVerest to work again

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,10 +6,24 @@ project(everest-core
 	LANGUAGES CXX C
 )
 
-find_package(everest-cmake 0.10 REQUIRED
+find_package(everest-cmake 0.10
     COMPONENTS bundling
     PATHS ../everest-cmake
 )
+
+if (NOT everest-cmake_FOUND)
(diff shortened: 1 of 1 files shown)


corneliusclaussen (approved):

CMakeLists.txt
```diff
@@ -6,11 +6,25 @@ project(everest-core
...
+    include(FetchContent)
+    FetchContent_Declare(
+        everest-cmake
+        GIT_REPOSITORY https://github.com/EVerest/everest-cmake.git
+        GIT_TAG        main
+    )
+    FetchContent_MakeAvailable(everest-cmake)
+    set(everest-cmake_DIR "${everest-cmake_SOURCE_DIR}")
+    set(everest-cmake_FIND_COMPONENTS "bundling")
+    include("${everest-cmake_SOURCE_DIR}/everest-cmake-config.cmake")
```
a-w50, line 25: It will be more robust, if find_package would be called here again.

PR #131, opened 2022-10-17
SessionLog: create log directory if it does not exist

corneliusclaussen:


SebaLukas (approved):
Looks good

PR #136, opened 2022-11-02
Adding setup and persistent_store to sil config

a-w50:


config/config-sil.json
```diff
@@ -64,6 +64,29 @@
     "yeti_driver": {
         "module": "JsYetiSimulator"
     },
+    "setup": {
+        "config_module": {
+            "localization": true,
+            "setup_simulation": false,
+            "setup_wifi": true
```
hikinggrass, line 71: I think these all of these options should be set to true.
Or at least simulation and localization, wifi might now be supported on all hosts

hikinggrass (changes requested):

corneliusclaussen (approved):

config/config-sil.json
a-w50: Thanks!

hikinggrass (approved):

PR #141, opened 2022-11-11
Feature/rapidyaml

a-w50:
Will rebase and recomment before merge


hikinggrass (approved):

hikinggrass:
Looks good, the only thing I would change is to bump the required ev-cli version to a higher number (>=0.0.16) and increase the version there too

PR #143, opened 2022-11-16
Remove boost::filesystem dependency

a-w50:
- using std::filesystem instead
- boost::filesystem::unique_file has no equivalent in std::filesystem for reasons.  Create replacement function called create_temp_file, which uses POSIX mkstemp (NOTE: mkstemp doesn't allow filename endings)
- changed some path objects to strings, because they're not used as an path at all

### DIFF
--- a/modules/Setup/CMakeLists.txt
+++ b/modules/Setup/CMakeLists.txt
@@ -17,5 +17,5 @@ target_sources(${MODULE_NAME}
 )
 
 # ev@c55432ab-152c-45a9-9d2e-7281d50c69c3:v1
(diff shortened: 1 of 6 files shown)


a-w50:
@hikinggrass @Pietfried can you do me a favour and checking if the refactored in the System module is still working correctly? Especially the creation of temporary files.  Thanks!

Pietfried:
> @hikinggrass @Pietfried can you do me a favour and checking if the refactored in the System module is still working correctly? Especially the creation of temporary files. Thanks!

yes will do

hikinggrass (approved):

PR #148, opened 2022-11-28
Fix out-of-tree support for interfaces and types

hikinggrass:
Out-of-tree repos that implement their own modules with a mix of custom interfaces/types and pre-defined types from everest-core can now use the ev_add_project()
function in their top-level CMakeLists.txt to make their interfaces and types known to ev-cli during compile time.

### DIFF
--- a/cmake/everest-generate.cmake
+++ b/cmake/everest-generate.cmake
@@ -16,12 +16,24 @@ set_target_properties(generate_cpp_files
 )
 
 #
+# out-of-tree interfaces/types/modules support
+#
+
+function(ev_add_project)
+    message(STATUS "APPENDING ${PROJECT_SOURCE_DIR} to EVEREST_PROJECT_DIRS")
+    set_property(TARGET generate_cpp_files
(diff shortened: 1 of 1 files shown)


a-w50 (approved):
I guess it is a matter of taste, but for everest-core I wouldn't explicitely call ev_add_project from the CMakeFile, you could call this function also at the end of everest-generate.cmake (so it will be called only one automatically when everest-core loads this file).
Furthermore I would rename EVEREST_DIR to something like EVEREST_PROJECT_DIRS or EVEREST_INC_DIRS or something similar (the ev-cli options everest-dir is also a bit misleading), because it may contain multiple dirs.
Other than that, everything looks good

hikinggrass:
> I guess it is a matter of taste, but for everest-core I wouldn't explicitely call ev_add_project from the CMakeFile, you could call this function also at the end of everest-generate.cmake (so it will be called only once automatically when everest-core loads this file). Furthermore I would rename EVEREST_DIR to something like EVEREST_PROJECT_DIRS or EVEREST_INC_DIRS or something similar (the ev-cli options everest-dir is also a bit misleading), because it may contain multiple dirs. Other than that, everything looks good

Great suggestions, I've immediately added them

PR #150, opened 2022-11-30
AuthModule: Clear Error on SessionFinished event 

Pietfried:
State machine needs error cleared event for the connector to become available for new tokens again


a-w50:
Is there any way to check in the session_finished handler, if there has been an error before? Could just resetting it always lead to race conditions?

Pietfried:
> Is there any way to check in the session_finished handler, if there has been an error before? Could just resetting it always lead to race conditions?

we could check this, but we dont have an ErrorCleared event yet in the EvseManager. Its just a workaround until we have that and better than not clearing the error, because in that case the authorization module will not authorize another id tag on that connector forever.

corneliusclaussen (approved):

PR #154, opened 2022-12-08
sync to remote belay box on install if BELAY_BOX=ip_address is set wi…

corneliusclaussen:
…th cmake

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -117,6 +117,10 @@ install(
     DESTINATION "${CMAKE_INSTALL_DATADIR}/everest"
 )
 
+if(BELAY_BOX)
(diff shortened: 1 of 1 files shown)


corneliusclaussen:
INSTALL_PREFIX is actually the wrong variable here

CMakeLists.txt
```diff
@@ -117,6 +117,10 @@ install(
     DESTINATION "${CMAKE_INSTALL_DATADIR}/everest"
 )
 
+if(BELAY_BOX)
+    install(CODE "execute_process(COMMAND /usr/bin/rsync -avh ${CMAKE_INSTALL_PREFIX}/ everest@${BELAY_BOX}:/mnt/user_data/opt/everest)")
```
a-w50, line 121: Is there any ordering in cmake concerning the install steps? Not that this command gets run somewhere in the middle and afterwards files get installed ...

a-w50: I also would prefer to separate these kind of tasks from cmake - but for now, we don't have anything else

PR #161, opened 2022-12-13
Change of token type in nodered flow

Pietfried:
- changed token type of nodered flow to RFID
- removed config option type of providers in configs so that default (RFID) is used


hikinggrass:
This PR should probably also change the "dummy" token types in the sil config(s)

a-w50 (approved):

PR #163, opened 2022-12-15
Use module info everest_prefix to access filesystem

Pietfried:
- Now using module info everest prefix to access filesystem paths in System and OCPP module
- replaced boost::filesystem by std::filesystem in OCPP module
- removed OcppMainPath from OCPP config options

### DIFF
--- a/config/config-sil-ocpp.yaml
+++ b/config/config-sil-ocpp.yaml
@@ -129,8 +129,7 @@ active_modules:
   ocpp:
     module: OCPP
     config_module:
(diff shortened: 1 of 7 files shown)


a-w50 (dismissed):
Well done,
there is only one issue, the dependency yaml refers to an older rev of everest-framework, where everest_prefix is not available.  I would suggest to put it's current commit hash into the dependency yaml.  Maybe @hikinggrass has an opinion?

hikinggrass (approved):

PR #164, opened 2022-12-19
Josev Integration

SebaLukas:
- Full Support for ISO15118-2 AC and DC
- Full Support for DIN70121
- Added power supply interface and SIL
- Improved IMD Interface and SIL
- Added type support for iso15118_charger, imd, power_supply and evse_manager interfaces
- Car simulator supports now full ISO15118-2 AC, partially ISO15118-2 DC & DIN70121
- Separate grid and car side powermeters in EvseManager
- Add ev_info to EvseManager for more details. The API module publishes this ev_info
- Increased the EV maximum physical value types, so charging is not interrupted when the car can do more than the standard specifies
- Configure NMKs at the end of the session to reduce time at the beginning of SLAC

To run the SIL correctly, please use the updated Node-Red flow "config-sil-two-evse-flow.json".

This pull request is dependent on this PR: [Integration Josev to EVerest #1](https://github.com/EVerest/ext-switchev-iso15118/pull/1)


SebaLukas:
TODO: Update Readme.md -> pip install python modules for pyjosev

SebaLukas:
TODO: Update single node-red flow with ISO15118-2 AC and DC

modules/EvseManager/Charger.hpp
```diff
@@ -187,9 +161,11 @@ class Charger {
         Disabled,
         Idle,
         WaitingForAuthentication,
+        PrepareCharging,
```
Pietfried, line 164: Right now a call to Charger::cancelTransaction will actually not stop the transaction while in state PrepareCharging. This should be added to the check of currentState (https://github.com/EVerest/everest-core/blob/main/modules/EvseManager/Charger.cpp#L747) because it should be possible to stop a transaction while in PrepareCharging.

modules/EvseManager/evse/evse_managerImpl.cpp
```diff
@@ -127,23 +127,20 @@ void evse_managerImpl::ready() {
             se.session_started = session_started;
         }
 
-        if (e == Charger::EvseEvent::SessionFinished) {
```
Pietfried, line 130: this and following should be else if or use switch case

types/evse_manager.yaml
```diff
@@ -31,26 +31,26 @@ types:
...
-      - HardReset
-      - Local
-      - Other
-      - PowerLoss
-      - Reboot
-      - Remote
-      - SoftReset
-      - UnlockCommand
-      - DeAuthorized
+    - EmergencyStop
```
hikinggrass, line 34: These changes look like formatter errors to me

hikinggrass (changes requested):
Nothin major to add to Piet and Sebastians comments, before the merge we should fix the formatting in the yaml and python files as well as the open TODOs

modules/EvseManager/Charger.hpp
corneliusclaussen: Good catch, looks like you already fixed that

corneliusclaussen: The link above points to main branch

modules/EvseManager/evse/evse_managerImpl.cpp
corneliusclaussen: ok

modules/EvseManager/VarContainer.hpp
```diff
@@ -0,0 +1,52 @@
...
+#include <chrono>
+#include <condition_variable>
+#include <mutex>
+
+/*
+ Simple helper class for a thread safe single producer / single consumer pattern
+ with a queue size of one.
+*/
+
+template <class T> class VarContainer {
```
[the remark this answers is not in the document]
corneliusclaussen: yes, the hardest part here is naming that thing. But the concept will probably be useful in other modules too.

modules/EvseManager/manifest.yaml
```diff
@@ -7,6 +10,10 @@ config:
     description: EVSE ID
     type: string
     default: DE*PNX*E1234567*1
+  evse_id_din:
+    description: EVSE ID DIN after DIN SPEC 91286
+    type: string
+    default: 49A80737A45678
```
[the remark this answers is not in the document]
corneliusclaussen: no idea

hikinggrass (approved):

dependencies.yaml
```diff
@@ -33,3 +33,8 @@ RISE-V2G:
 libocpp:
   git: https://github.com/EVerest/libocpp.git
   git_tag: v0.3.0
+
+# Josev
+Josev:
+  git: git@github.com:EVerest/ext-switchev-iso15118.git
+  git_tag: sl-integrate-to-everest
```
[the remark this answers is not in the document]
hikinggrass: I've tagged a 2022.12.0 version @ ext-josev so we do not have to point to master anymore, I've also added a new default branch "everest" there that can track our EVerest related changes

PR #166, opened 2023-01-04
Allow modules to be selectively excluded or collectively by programming language

hikinggrass:


cmake/everest-generate.cmake
```diff
@@ -127,171 +133,229 @@ function (ev_add_modules)
...
+
+    if(IS_DIRECTORY ${MODULE_PATH})
+        if(NOT ${EVEREST_ENABLE_JS_SUPPORT})
+            message(STATUS "Excluding JavaScript module ${MODULE_NAME} because EVEREST_ENABLE_JS_SUPPORT=${EVEREST_ENABLE_JS_SUPPORT}")
+            return()
+        elseif("${MODULE_NAME}" IN_LIST EVEREST_EXCLUDE_MODULES)
+            message(STATUS "Excluding module ${MODULE}")
+            return()
+        else()
+            message(STATUS "Setting up JavaScript module ${MODULE_NAME}")
```
a-w50, line 245: Don't know if you would agree with this style but in order to keep indentation low, one could invert the ifs like
```
if(NOT IS_DIRECTORY ..)
  warning ...
  return()
endif()

if (NOT ${EVEREST_ENABLE_JS_SUPPORT})
  warning ...
  return()
endif()

if ("${MODULE_NAME} IN_LIST ...)
  warning ...
  return()
endif()

# now the normal case ...
```
same would go the other add_ functions I guess

cmake/everest-generate.cmake
```diff
@@ -127,171 +133,229 @@ function (ev_add_modules)
...
-#
-
-macro(ev_setup_cpp_module)
-    get_filename_component(MODULE_NAME ${CMAKE_CURRENT_SOURCE_DIR} NAME)
+function (ev_add_cpp_module MODULE_NAME)
+    set(EVEREST_MODULE_INSTALL_PREFIX "${CMAKE_INSTALL_LIBEXECDIR}/everest/modules")
 
-    if(${EVEREST_EXCLUDE_CPP_MODULES})
-        message(STATUS "Excluding C++ module ${MODULE_NAME} because EVEREST_EXCLUDE_CPP_MODULES=${EVEREST_EXCLUDE_CPP_MODULES}")
+    set(MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/${MODULE_NAME}")
```
a-w50, line 143: these two `set` statements seem to appear in all three add_* functions.  This is a good spot for using a private macro like `macro (_ev_module_common_setup)` or so, which sets these variables, otherwise one would need to change all 3 of them, if there will be changes needed

a-w50 (approved):
Personally I would change the style as commented, but otherwise this looks like a good code reducing improvement

PR #167, opened 2023-01-11
Improve out-of-core build support

a-w50:
- creating proper cmake config file for everest-core so that it's cmake functionality for generation can be used by a depending package, which requires everest-core via find_package
- remove stale find_package(s)


a-w50:
Before merging, the branches need to be fixed!

hikinggrass (approved):

PR #170, opened 2023-01-16
Configurable timeout for CM_SET_KEY.REQ

corneliusclaussen:
The new default value supports Lumissil CK5317.
Also adds warnings if slac packages are skipped

### DIFF
--- a/modules/EvseSlac/main/evse_fsm.cpp
+++ b/modules/EvseSlac/main/evse_fsm.cpp
@@ -29,7 +29,7 @@ using EventMatchingFailed = EventTypeFactory::Derived<EventID::MatchingFailed>;
 using EventFinishSounding = EventTypeFactory::Derived<EventID::FinishSounding>;
 using EventAttenCharRspReceived = EventTypeFactory::Derived<EventID::AttenCharRspReceived>;
 
-EvseFSM::EvseFSM(SlacIO& slac_io) : slac_io(slac_io) {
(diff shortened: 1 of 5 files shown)


modules/EvseSlac/main/evse_fsm.cpp
```diff
@@ -355,6 +356,8 @@ void EvseFSM::sd_matching_hsm(FSMContextType& ctx, const EventSlacMessage& ev) {
     auto& start_atten = msg_in.get_payload<slac::messages::cm_start_atten_char_ind>();
 
     if (!matching_ctx.conforms(msg_in.get_src_mac(), start_atten.run_id)) {
+        EVLOG_warning << fmt::format("Ignoring CM_START_ATTEN_CHAR from {}, RunID {}",
```
a-w50, line 359: Please take a look at all these `const char*` casts, most of the the buffers are not required to be null-terminated, like `get_src_mac` for example, or even the `run_id`.  This could lead to segfaults.  Furthermore, the c++ way of casting pointers should be `reinterpret_cast<const char*>(pointer)`.  You probably need to write a separate printing helper for these values. The source mac should be represented as hex anyway

a-w50 (changes requested):
Looks good. Please have a look at the `const char*` casting.  From a design perspective, if there will me more options/settings coming for the EvseFSM class, we add a EvseFSMOptions struct or something similar in future.

modules/EvseSlac/main/evse_fsm.cpp
corneliusclaussen: I removed printing of the details, it is more important to see that something was dropped at all.

a-w50 (approved):

PR #174, opened 2023-01-23
Add PP ampacity reading interface

hikinggrass:


corneliusclaussen (approved):

corneliusclaussen:
I will add real support to Yeti driver later on, currently yeti HW does not support a connector anyway.

PR #176, opened 2023-01-24
Mitigate connector sometimes not transitioning to available in Auth 

hikinggrass:
When using libfsm in async mode events coming in quick succession can sometimes be lost.
By using blocking mode in submit_event this can be mitigated (at least most of the time) until a proper fix is implemented.

### DIFF
--- a/modules/Auth/lib/AuthHandler.cpp
+++ b/modules/Auth/lib/AuthHandler.cpp
@@ -413,8 +413,8 @@ void AuthHandler::handle_session_event(const int connector_id, const SessionEven
     case SessionEventEnum::SessionFinished:
         this->connectors.at(connector_id)->connector.is_reservable = true;
         this->connectors.at(connector_id)->connector.identifier = boost::none;
-        this->connectors.at(connector_id)->connector.submit_event(Event_Session_Finished());
-        this->connectors.at(connector_id)->connector.submit_event(Event_Error_Cleared());
+        this->connectors.at(connector_id)->connector.state_machine.controller->submit_event(Event_Session_Finished(), true);
+        this->connectors.at(connector_id)->connector.state_machine.controller->submit_event(Event_Error_Cleared(), true);
         this->connectors.at(connector_id)->timeout_timer.stop();
         break;
     case SessionEventEnum::PermanentFault:


a-w50 (commented):
For now this might work, but this behavior is a general flaw in libfsm.  I'll try to have a look if this can be handled differently

a-w50 (commented):
I had a look again at the implementation of submit_event.  There are two things you can try
1.  there is a second standard parameter, called blocking.  That will block the caller until the event can be accepted.  But it wouldn't make much sense to add this now to every call.
2. submit_event returns a bool, indicating if the event has been accepted or not, so you could implement your own submit function which blocks and retries as long as needed to submit the event.
The main problem is still, how to handle events, coming from multiple threads .. (including the state machine itself).
The fsm library definitely needs a rework soon.

hikinggrass:
> I had a look again at the implementation of submit_event. There are two things you can try
> 
>     1. there is a second standard parameter, called blocking.  That will block the caller until the event can be accepted.  But it wouldn't make much sense to add this now to every call.
> 
>     2. submit_event returns a bool, indicating if the event has been accepted or not, so you could implement your own submit function which blocks and retries as long as needed to submit the event.
>        The main problem is still, how to handle events, coming from multiple threads .. (including the state machine itself).
>        The fsm library definitely needs a rework soon.

I've now 

> I had a look again at the implementation of submit_event. There are two things you can try
> 
>     1. there is a second standard parameter, called blocking.  That will block the caller until the event can be accepted.  But it wouldn't make much sense to add this now to every call.
> 
>     2. submit_event returns a bool, indicating if the event has been accepted or not, so you could implement your own submit function which blocks and retries as long as needed to submit the event.
>        The main problem is still, how to handle events, coming from multiple threads .. (including the state machine itself).
>        The fsm library definitely needs a rework soon.

I've added the blocking parameter to the two events being submitted right after each other which mitigates this issue for now. I've also made it clear in the commit message and PR that this is only a mitigation and that a proper fix has to be implemented in the future

a-w50 (approved):
Still I would refactor the connector.submit_event function to allow for the second parameter, because you should access the connectors members directly.  Another idea would be to forward the call from connector.submit_event to the state machine ctrl with blocking=true, so that all events will be supplied in blocking mode

PR #178, opened 2023-01-30
Add PacketSniffer Module

SebaLukas:
With this module it is possible to capture and store the different ethernet packets on the PLC interface.
Testable only with hardware or a BelayBox, because admin rights are needed for capturing on the socket.

### DIFF
--- a/modules/PacketSniffer/PacketSniffer.cpp
+++ b/modules/PacketSniffer/PacketSniffer.cpp
@@ -1,0 +1,79 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Pionix GmbH and Contributors to EVerest
+#include "PacketSniffer.hpp"
+
+#include <fmt/core.h>
+
+namespace module {
+
+const bool PROMISC_MODE = true;
+const int PACKET_BUFFER_TIMEOUT_MS = 1000;
+const int ALL_PACKETS_PROCESSED = -1;
+const int WAIT_FOR_MS = 1;
+const int BUFFERSIZE = 8192;
+
+void PacketSniffer::init() {
+    invoke_init(*p_main);
+
+    capturing_stopped = false;
+    already_started = false;
+
+    p_handle = pcap_open_live(config.device.c_str(), BUFFERSIZE, PROMISC_MODE, PACKET_BUFFER_TIMEOUT_MS, errbuf);
+    if (p_handle == nullptr) {
+        EVLOG_AND_THROW(Everest::EverestConfigError(
+            fmt::format("Could not open device {}", config.device)));
+        return;
+    }
+
+    if (pcap_datalink(p_handle) != DLT_EN10MB) {
+        EVLOG_AND_THROW(Everest::EverestConfigError(
+            fmt::format("Device {} doesn't provide Ethernet headers - not supported", config.device)));
+        return;
+    }
+
+    r_evse_manager->subscribe_session_event([this](types::evse_manager::SessionEvent session_event) {
+
+        std::lock_guard<std::mutex> lock(this->capture_mutex);
+        if (session_event.event == types::evse_manager::SessionEventEnum::SessionStarted) {
+            if (!already_started) {
+                already_started = true;
+                capturing_stopped = false;
+                std::thread(&PacketSniffer::capture, this, config.session_logging_path, session_event.uuid).detach();
+            } else {
+                EVLOG_warning << fmt::format("Capturing already started. Ignoring this SessionStarted event");
+            }
+        } else if (session_event.event == types::evse_manager::SessionEventEnum::SessionFinished) {
+            capturing_stopped = true;
+            already_started = false;
+        }
+    });
+}
+
+void PacketSniffer::ready() {
+    invoke_ready(*p_main);
+}
+
+void PacketSniffer::capture(const std::string& logpath, const std::string& session_id) {
+
+    const std::string fn = fmt::format("{}/everest-session-{}.dump",
+                                        logpath, session_id);
+
+    if ((pdumpfile = pcap_dump_open(p_handle, fn.c_str())) == nullptr) {
+        EVLOG_error << fmt::format("Error opening savefile {} for writing: {}", 
+                                    fn, pcap_geterr(p_handle));
+        return;
+    }
+
+    while (!capturing_stopped) {
+        if (pcap_dispatch(p_handle, ALL_PACKETS_PROCESSED, &pcap_dump, (u_char *)pdumpfile) <= PCAP_ERROR) {
+            EVLOG_error << fmt::format("Error reading packets from interface: {}, error: {}",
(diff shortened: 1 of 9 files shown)


modules/PacketSniffer/cmake/FindPCAP.cmake
```diff
@@ -0,0 +1,75 @@
...
+#                            libpcap if the module has problems finding the
+#                            proper installation path.
+#
+# Variables defined by this module:
+#
+#  PCAP_FOUND                System has libpcap, include and library dirs found
+#  PCAP_INCLUDE_DIR          The libpcap include directories.
+#  PCAP_LIBRARY              The libpcap library (possibly includes a thread
+#                            library e.g. required by pf_ring's libpcap)
+#  HAVE_PF_RING              If a found version of libpcap supports PF_RING
```
a-w50, line 20: This file is copied from somewhere else I guess, this should be mentioned - and the license probably needs to be checked. Especially if its from wireshark (which is GPL)

modules/PacketSniffer/PacketSniffer.cpp
```diff
@@ -0,0 +1,63 @@
...
+
+const bool PROMISC_MODE = true;
+const int PACKET_BUFFER_TIMEOUT_MS = 1000;
+
+void PacketSniffer::init() {
+    invoke_init(*p_main);
+
+    capturing_stopped = false;
+
+    handle = pcap_open_live(config.device.c_str(), BUFSIZ, PROMISC_MODE, PACKET_BUFFER_TIMEOUT_MS, errbuf);
```
a-w50, line 17: where does BUFSIZ come from?

modules/PacketSniffer/PacketSniffer.cpp
```diff
@@ -0,0 +1,63 @@
...
+    if (handle == NULL) {
```
a-w50, line 18: in C++11 use nullptr instead of NULL

modules/PacketSniffer/PacketSniffer.cpp
```diff
@@ -0,0 +1,63 @@
...
+    r_evse_manager->subscribe_session_event([this](types::evse_manager::SessionEvent session_event) {
+        auto event = types::evse_manager::session_event_enum_to_string(session_event.event);
+        if (event == "SessionStarted") {
+            capturing_stopped = false;
+            std::thread(&PacketSniffer::capture, this, config.session_logging_path, session_event.uuid).detach();
+        } else if (event == "SessionFinished") {
+            capturing_stopped = true;
+        }
+    });
+}
```
a-w50, line 37: What if multiple SessionStarted without any SessionFinished happen? Then it is tried multiple times to attach via pcap.  I would add a mutex to this session event listener and check first if a session is already running

modules/PacketSniffer/PacketSniffer.cpp
```diff
@@ -0,0 +1,63 @@
...
+
+void PacketSniffer::ready() {
+    invoke_ready(*p_main);
+}
+
+void PacketSniffer::capture(const std::string& logpath, const std::string& session_id) {
+
+    const std::string fn = fmt::format("{}/everest-session-{}.dump", logpath, session_id);
+
+    if ((pdumpfile = pcap_dump_open(handle,fn.c_str())) == NULL) {
```
a-w50, line 47: this kind of notation is short, but sometimes hard to read, better make two lines of code

modules/PacketSniffer/PacketSniffer.cpp
```diff
@@ -0,0 +1,63 @@
...
+        EVLOG_error << fmt::format("Error opening savefile {} for writing: {}", fn, pcap_geterr(handle));
+        return;
+    }
+
+    while (!capturing_stopped) {
+        if (pcap_dispatch(handle, -1, &pcap_dump, (u_char *)pdumpfile) <= PCAP_ERROR) {
```
a-w50, line 53: what is -1? define a constant with a name, that clarifies what -1 means

modules/PacketSniffer/PacketSniffer.cpp
```diff
@@ -0,0 +1,63 @@
...
+            EVLOG_error << fmt::format("Error reading packets from interface: {}, error: {}", config.device, pcap_geterr(handle));
+            break;
+        }
+        std::this_thread::sleep_for(std::chrono::milliseconds(1));
```
a-w50, line 57: Don't use raw numbers here directly, again, define a constant

modules/PacketSniffer/PacketSniffer.hpp
```diff
@@ -0,0 +1,71 @@
...
+private:
+    friend class LdEverest;
+    void init();
+    void ready();
+
+    // ev@211cfdbe-f69a-4cd6-a4ec-f8aaa3d1b6c8:v1
+    // insert your private definitions here
+    void capture(const std::string& logpath, const std::string& session_id);
+    pcap_t *handle;
+    pcap_dumper_t *pdumpfile;
```
a-w50, line 59: Default initialize these values here, to make codacy silent
```
pcap_t *handle {nullptr};
pcap_dumper_t *pdumpfile {nullptr};
```

a-w50 (changes requested):
Just some minor things

modules/PacketSniffer/cmake/FindPCAP.cmake
SebaLukas: Done, the license is BSD.

modules/PacketSniffer/PacketSniffer.cpp
SebaLukas: Done

modules/PacketSniffer/PacketSniffer.cpp
SebaLukas: Done

modules/PacketSniffer/PacketSniffer.cpp
SebaLukas: Done

modules/PacketSniffer/PacketSniffer.cpp
SebaLukas: Done

modules/PacketSniffer/PacketSniffer.hpp
SebaLukas: Done

modules/PacketSniffer/PacketSniffer.cpp
SebaLukas: Good Point!

SebaLukas: Should now work like this

a-w50 (approved):
perfect, probably @corneliusclaussen  wants to test it

corneliusclaussen (approved):

PR #181, opened 2023-02-07
Plug'n Charge interface extension

SebaLukas:
Added an extension to the iso15118_charger interface for plug and charge support.

### DIFF
--- a/interfaces/ISO15118_charger.yaml
+++ b/interfaces/ISO15118_charger.yaml
@@ -121,12 +121,17 @@ cmds:
         type: boolean
   set_Auth_Okay_PnC:
     description: >-
-      Response on Require_Auth_PnC - Set this element if the authorization
-      was either successful or unsuccessful
+      PaymentDetails - With this message the SECC informs the EVCC whether the previously provided
+      payment details were accepted or not
     arguments:
-      auth_okay_pnc:
-        description: "True: Authentication is okay"
-        type: boolean
+      status:
+        description: Authorization status of the ID Token
+        type: string
+        $ref: /authorization#/AuthorizationStatus
+      certificateStatus:
+        description: Certificate status information
+        type: string
+        $ref: /authorization#/CertificateStatus
   set_FAILED_ContactorError:
     description: >-
       Any Time - Set this error when the contactor is opened or closed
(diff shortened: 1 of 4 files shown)


interfaces/ISO15118_charger.yaml
```diff
@@ -255,6 +255,44 @@ cmds:
...
         type: boolean
+  PnC_supporting_Cert_Installation_Update:
+    description: >-
+      ServiceDiscoveryRes - The charger supports the certificate installation/update
+      service and has a connection to an SA for this purpose
+    arguments:
+      status:
+        description: "True: The service is supported"
+        type: boolean
+  PnC_Response_Installation:
```
Pietfried, line 266: We should try to find a comprehensive naming for the async publish (request) and command (response) functionality for example:
PnC_Request_Certificate_Install <-> Set_PnC_Response_Certificate_Install
or even better imo (because it is not directly PnC related):
Install_Certificate_Request <-> Set_Install_Certificate_Response
Update_Certificate_Request <-> Set_Update_Certificate_Response
and 
PnC_Authorization_Request <-> Set_PnC_Authorization_Response

types/iso15118_charger.yaml
```diff
@@ -339,3 +339,72 @@ types:
         description: Contains the EXI stream as base64 string
         type: string
         minLength: 0
+  PnC_Response_Existream_Status:
```
Pietfried, line 342: I would name this Install_Certificate_Response

types/iso15118_charger.yaml
```diff
@@ -339,3 +339,72 @@ types:
...
+        enum:
+          - Accepted
+          - Failed
+      exiResponse:
+        description: >-
+          Raw CertificateInstallationRes or CertificateUpdateRes message
+          as exi stream (Base64 encoded)
+        type: string
+        maxLength: 5600
+  PnC_Request_ExiStream_Schema:
```
Pietfried, line 365: And this Install_Certificate_Request

Pietfried (changes requested):
The content is perfect, but we can improve the naming of cmds/vars/types

interfaces/ISO15118_charger.yaml
SebaLukas: I like that, done!

Pietfried (approved):

PR #182, opened 2023-02-08
Only use PP reading when connector type is a IEC 62196 Type 2 Socket

hikinggrass:


SebaLukas (approved):
Looks good :thumbsup:

PR #186, opened 2023-02-22
Add telemetry to EvseManager, YetiSimulator and YetiDriver

hikinggrass:


config/config-sil.yaml
```diff
@@ -1,9 +1,11 @@
+settings:
+  telemetry_enabled: true
```
Pietfried, line 2: Propably we want to enable this in the other configs as well

corneliusclaussen: Not sure if we need it everywhere. maybe one example on how to turn it on is enough

PR #187, opened 2023-02-28
Add test control module

LAD101work:
This branch adds a basic test control module for use in CI testing with the everest-testing framework.


tests/core_tests/everest_core_tests.py
```diff
@@ -0,0 +1,56 @@
...
+#     test_control_module_handle = test_control_module.start()
+#     logging.info("test_control_module started")
+
+#     test_control_module_handle.start_charging(mode = 'raw', charge_sim_string = "sleep 1;iec_wait_pwr_ready;sleep 1;draw_power_regulated 16,3;sleep 20;unplug")
+
+#     # assert that charged energy is > target
+#     assert await wait_for_and_validate_event(test_control_module, exp_event='transaction_started', exp_data=None, validation_function=set_test_memory, timeout=20)
+#     assert await wait_for_and_validate_event(test_control_module, exp_event='transaction_finished', exp_data=float(56.0), validation_function=validate_transaction_charged_amount, timeout=30)
+
+#     everest_core.stop()
```
a-w50, line 56: Delete if not needed

LAD101work: It is very much needed, but does not work at the moment, because changes in the everest-framework/in everestpy need to happen for this to work.

modules/PyTestControlModule/manifest.yaml
```diff
@@ -0,0 +1,24 @@
...
+provides:
+  main:
+    interface: empty
+    description: This module provides no outward-facing interface
+requires:
+  test_control:
+    interface: car_simulator
+  connector_1:
+    interface: evse_manager
+    min_connections: 1
```
Pietfried, line 18: min_connections should always be 0 in order to be able to execute tests with configs that do not load the module.

LAD101work: Changed min_connections and added debug messages on affected functions.

tests/core_tests/everest_core_tests.py
```diff
@@ -0,0 +1,56 @@
+
```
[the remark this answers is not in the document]
LAD101work: Changed name.

tests/core_tests/validations/validation_functions.py
```diff
@@ -0,0 +1,83 @@
+# SPDX-License-Identifier: Apache-2.0
```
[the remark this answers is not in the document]
LAD101work: Changed name.

tests/core_tests/everest_core_tests.py
LAD101work: I have split the tests into two test sets (and files) so that both of them can now run, even before everestpy/everest-framework have been modified.

Pietfried (approved):

tests/core_tests/startup_tests.py
```diff
@@ -0,0 +1,37 @@
+
+#!/usr/bin/env python3
+# SPDX-License-Identifier: Apache-2.0
+# Copyright 2020 - 2022 Pionix GmbH and Contributors to EVerest
+
+import logging
+import pytest
+
+from everest.testing.core_utils.fixtures import *
```
a-w50, line 9: `import *` is really deprecated, are there any fixtures used in this test at all?

tests/core_tests/startup_tests.py
```diff
@@ -0,0 +1,37 @@
...
+from validations.base_functions import wait_for_and_validate_event
+from everest.testing.core_utils.test_control_module import TestControlModule
+from everest.testing.core_utils.everest_core import EverestCore
+from validations.user_functions import *
+
+@ pytest.mark.asyncio
+async def test_000_startup_check(everest_core: EverestCore):
+    logging.info(">>>>>>>>> test_000_startup_check <<<<<<<<<")
```
a-w50, line 17: doesn't pytest by itself report on which test it is running, or why do we need to log this ourself?

tests/core_tests/validations/base_functions.py
```diff
@@ -0,0 +1,111 @@
...
+import asyncio
+from typing import Callable
+from typing import Any
+
+from everest.testing.core_utils.test_control_module import TestControlModule
+
+## helper functions
+
+## ---------------------------------------------------
+def get_key_if_exists(dataset:dict, keyset:list):
```
a-w50, line 15: syntax should be `dataset: dict, keyset: list`
the naming is a bit misleading,  dataset could be called just dictionary
keyset is really only a key.
Actually this function could be just replaced by `dict.get(key, None)`, of course if the value is allowed to be None, then one would need to check for `key in dict`

tests/core_tests/validations/base_functions.py
```diff
@@ -0,0 +1,111 @@
...
+        return False, None
+    else:
+        if keyset is not []:
+            if len(keyset) == 1:
+                if keyset[0] in dataset:
+                    return True, dataset[keyset[0]]
+                else:
+                    return False, None
+            else:
+                return get_key_if_exists(dataset.get(keyset[0]), keyset[1:])
```
a-w50, line 37: have a look at **https://stackoverflow.com/questions/43491287/elegant-way-to-check-if-a-nested-key-exists-in-a-dict**, this is pretty close to what is needed here

tests/core_tests/validations/user_functions.py
```diff
@@ -0,0 +1,83 @@
...
+    event_found, received_event_value = get_key_if_exists(event, ['transaction_finished', 'energy_Wh_import'])
+
+    if event_found == False:
+        return False, str("Error: Expected data is missing from event!")
+
+    if (float(received_event_value) - float(pre_event_value_from_memory)) > float(exp_data):
+        validation_memory = []
+        return True, str("Ok")
+
+    return False, str(f"Error: Event value ({received_event_value}) too low! "
```
a-w50, line 38: the return values here are pretty close to exceptions, maybe we can use exceptions instead?

PR #193, opened 2023-03-20
Update README.md

pionix-juraed:
added more prerequisirtes


SebaLukas:
Not needed, because the Python packages are installed with the josev/requirements.txt

PR #197, opened 2023-03-24
Link URL missed the nightly part

krealyt:


README.md
```diff
@@ -122,7 +122,7 @@ Done!
 
 In order to test your build of Everest you can simulate the code on your local machine! Check out the different configuration files to run EVerest and the corresponding nodered flows in the [config folder](config/).
 
- Check out this [guide for using EVerest SIL](https://everest.github.io/tutorials/run_sil/index.html
+ Check out this [guide for using EVerest SIL](https://everest.github.io/nightly/tutorials/run_sil/index.html)
```
andistorm, line 125:
```suggestion
 Check out this [guide for using EVerest SIL](https://everest.github.io/latest/tutorials/run_sil/index.html)
```

FYI: With https://github.com/EVerest/EVerest/pull/60 you can also link to the latest released version. Which is in the moment same as nightly.  But maybe it makes more sense to link to latest nightly, as you did, on the main branch in everest-core?

krealyt: @andistorm I think, we should go for a proper linking strategy in a separate step, yes. For now, let's fix those broken links. Is that ok for you?

andistorm (approved):

PR #203, opened 2023-04-03
Add TLS v1.2 support to EvseV2G module

FaHaGit:
Now the module supports TLS v1.2. The TLS support can be configured in the manifest.yaml file of the module. Session key logging can also be activated in the manifest.yaml file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,6 +97,8 @@ cd ~/checkout/everest-workspace/Josev/iso15118/shared/pki
 
 Note that this will generate an example PKI setup that can only be used for testing and simulation purposes. It will not work and is not recommended for production!
 
+The script for setting up PKI can also be used with the EvseV2G module.
+
 Now we can build EVerest!
 
 ```bash
--- a/modules/EvseV2G/charger/ISO15118_chargerImpl.cpp
+++ b/modules/EvseV2G/charger/ISO15118_chargerImpl.cpp
@@ -5,6 +5,8 @@
 #include "log.hpp"
 #include "v2g_ctx.hpp"
 
+const std::string EVEREST_ETC_CERTS_PATH = "etc/everest/certs"; // relativ path of the certs
+
 using namespace std::chrono_literals;
 
 namespace module {
@@ -16,6 +18,7 @@ void ISO15118_chargerImpl::init() {
         return;
     }
 
+    /* Configure if_name and auth_mode */
     v2g_ctx->if_name = mod->config.device.data();
     dlog(DLOG_LEVEL_DEBUG, "if_name %s", v2g_ctx->if_name);
     dlog(DLOG_LEVEL_DEBUG, "auth_mode %s", mod->config.highlevel_authentication_mode.data());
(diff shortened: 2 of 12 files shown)


modules/EvseV2G/charger/ISO15118_chargerImpl.cpp
```diff
@@ -47,6 +47,14 @@ void ISO15118_chargerImpl::init() {
...
         dlog(DLOG_LEVEL_DEBUG, "tls_security prohibit");
     }
+
+    v2g_ctx->tls_key_logging = mod->config.tls_key_logging;
+    v2g_ctx->tls_key_logging_path = mod->config.tls_key_logging_path;
+
+    if (mod->config.tls_key_logging == true) {
+        dlog(DLOG_LEVEL_DEBUG, "tls-key-logging enabled (path: %s)", mod->config.tls_key_logging_path.c_str());
+    }
+    v2g_ctx->certs_path = mod->config.certs_path;
```
SebaLukas, line 57: The certificates from the everest-core/config/certs folder are copied over to the everest-core/build/dist/etc/everest/certs with CMake during installation.

The path info.everest_prefix points to this everest-core/build/dist folder.

```cpp
const std::string EVEREST_ETC_CERTS_PATH = "etc/everest/certs";
v2g_ctx->certs_path = mod->info.everest_prefix + "/" + EVEREST_ETC_CERTS_PATH;
```

Please do not declare the const string in the function, but at the beginning of the file

README.md
```diff
@@ -97,6 +97,8 @@ cd ~/checkout/everest-workspace/Josev/iso15118/shared/pki
 
 Note that this will generate an example PKI setup that can only be used for testing and simulation purposes. It will not work and is not recommended for production!
 
+The script for setting up PKI can also be used with the EvseV2G module, but make sure to place the generated PKI in the right path configured in the manifest.yaml.
```
SebaLukas, line 100: Possibly delete the second sentence, because it can lead to confusion.

modules/EvseV2G/tools.cpp
```diff
@@ -329,3 +331,13 @@ uint8_t get_dir_numbered_file_names(char file_names[MAX_PKI_CA_LENGTH][MAX_FILE_
 
     return num_of_files + offset;
 }
+
+std::string convert_to_hex_str(const uint8_t* data, int len) {
+    std::stringstream sting_stream;
```
SebaLukas, line 336: string_stream instead sting_stream

SebaLukas (changes requested):
Looking good, only some small changes.
And a update to the current everest-core main.

SebaLukas:
These codacy errors can be ignored

modules/EvseV2G/charger/ISO15118_chargerImpl.cpp
FaHaGit: Where is this prefix defined in cmake file? I get "/usr" if I print out the string of mod->info.everest_prefix on charge byte platform.

modules/EvseV2G/manifest.yaml
```diff
@@ -35,10 +35,29 @@ config:
...
+  tls_key_logging_path:
+    description:  >-
+      Output directory for the TLS key log file
+    type: string
+    default: /tmp
+  certs_path:
+    description: >-
+      Path to the V2G PKI
+    type: string
+    default: /etc/everest/certs
```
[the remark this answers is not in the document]
FaHaGit: This is currently used, I am wondering why it should not be used.

modules/EvseV2G/charger/ISO15118_chargerImpl.cpp
FaHaGit: That would result to the file path "/usr/etc/everest/certs" which currently not exists.

SebaLukas: Here is my suggestion for a quick solution:
```cpp
const std::string EVEREST_ETC_CERTS_PATH = "etc/everest/certs";
[...]
if (mod->info.everest_prefix != "/usr") {
    v2g_ctx->certs_path = mod->info.everest_prefix + "/" + EVEREST_ETC_CERTS_PATH;
} else {
    v2g_ctx->certs_path = "/" + EVEREST_ETC_CERTS_PATH;
}
´´´

modules/EvseV2G/manifest.yaml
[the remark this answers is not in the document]
SebaLukas: No longer used when using the everest_prefix.

SebaLukas (approved):
👍

PR #205, opened 2023-04-04
SLAC based Autocharge support

corneliusclaussen:


modules/EvseManager/EvseManager.cpp
```diff
@@ -350,21 +366,26 @@ void EvseManager::ready() {
...
+            r_hlc[0]->subscribe_EVCCIDD([this](const std::string& _token) {
+                types::authorization::ProvidedIdToken autocharge_token;
+
+                std::string token = _token;
+                token.erase(remove(token.begin(), token.end(), ':'), token.end());
+                autocharge_token.id_token = "VID:" + token;
+                autocharge_token.type = types::authorization::TokenType::Autocharge;
+                autocharge_token.connectors.emplace(config.connector_id, 1);
+
+                p_token_provider->publish_provided_token(autocharge_token);
```
SebaLukas, line 379: Was changed by the Plug'n Charge commit and is no longer executed here!

corneliusclaussen: should be fixed now

SebaLukas: 👍

modules/EvseSlac/main/evse_fsm.cpp
```diff
@@ -472,6 +475,10 @@ void EvseFSM::sd_wait_for_slac_match_hsm(TransitionType& trans, const EventSlacM
 
     slac_io.send(msg_out);
 
+    signal_ev_mac_address_match_cnf(
+        std::string{reinterpret_cast<const char*>(match_cnf.pev_mac), sizeof(match_cnf.pev_mac)});
+    EVLOG_info << "SLAC_MATCH.CNF send.";
```
SebaLukas, line 480: Debug instead of info

modules/EvseSlac/main/slacImpl.cpp
```diff
@@ -135,6 +156,7 @@ void slacImpl::run() {
                 //             be in principle available now
                 // FIMXE (aw): this needs to be implemented in a higher level ->
                 //             stipulate link establishment if appropriate
+                EVLOG_info << "State::ReceivedSlacMatch";
```
SebaLukas, line 159: Debug instead of info

SebaLukas (changes requested):
Looks all right to me. I would suggest only three small changes 😃

modules/EvseManager/EvseManager.cpp
```diff
@@ -29,6 +29,22 @@ void EvseManager::init() {
     if (!(config.ac_hlc_enabled || config.ac_with_soc || config.charge_mode == "DC")) {
         slac_enabled = false;
     }
+
+    // Use SLAC MAC address for Autocharge if configured.
+    if (config.autocharge_use_slac_instead_of_hlc && slac_enabled) {
+        r_slac[0]->subscribe_ev_mac_address([this](const std::string& _token) {
```
a-w50, line 35: not sure, here, but you could change the signature into std::string token, than you'll capture directly a copy

modules/EvseManager/EvseManager.cpp
```diff
@@ -29,6 +29,22 @@ void EvseManager::init() {
...
+            types::authorization::ProvidedIdToken autocharge_token;
+
+            std::string token = _token;
+            token.erase(remove(token.begin(), token.end(), ':'), token.end());
```
a-w50, line 39: this line is really hard to read, I would rename this function into something like
inline static void trim_colons_from_string(std::string& text) { ... } and also reuse it in its second occurence

modules/EvseManager/EvseManager.cpp
```diff
@@ -29,6 +29,22 @@ void EvseManager::init() {
...
+            autocharge_token.id_token = "VID:" + token;
```
a-w50, line 40: you could also write a factory function like
```
inline static types::authorization::ProvidedIdToken create_autocharge_token(const std::string& token) {
  types::authorization::ProvidedIdToken autocharge_token;
  autocharge_token.type = types::authorization::TokenType::Autocharge;
  autocharge_token.id_token = "VID:" + trim_colons_from_string(token);
  return autocharge_token;
}
```
this would be just more explicit

modules/EvseManager/EvseManager.cpp
```diff
@@ -349,22 +365,22 @@ void EvseManager::ready() {
...
-            std::vector<int> referenced_connectors = {this->config.connector_id};
-            autocharge_token.connectors.emplace(referenced_connectors);
+        if (!config.autocharge_use_slac_instead_of_hlc) {
+            r_hlc[0]->subscribe_EVCCIDD([this](const std::string& _token) {
+                std::string token = _token;
+                token.erase(remove(token.begin(), token.end(), ':'), token.end());
+                autocharge_token.id_token = "VID:" + token;
+                autocharge_token.type = types::authorization::TokenType::Autocharge;
+                std::vector<int> referenced_connectors = {this->config.connector_id};
+                autocharge_token.connectors.emplace(referenced_connectors);
```
a-w50, line 375: `autocharge_token.connectors.emplace({this->config.connector_id});`

modules/EvseSlac/main/evse_fsm.hpp
```diff
@@ -96,6 +97,9 @@ class EvseFSM {
 
     explicit EvseFSM(SlacIO& slac_io, int _set_key_timeout);
 
+    sigslot::signal<std::string> signal_ev_mac_address_match_cnf;
```
a-w50, line 100: function callbacks should be also fine instead of sigslot here, but hopefully with the new fsm lib, there is a better way to communicate to the outside via the fsm

a-w50 (approved):

modules/EvseSlac/main/evse_fsm.cpp
corneliusclaussen: fixed

modules/EvseSlac/main/slacImpl.cpp
corneliusclaussen: fixed

modules/EvseSlac/main/slacImpl.cpp
```diff
@@ -121,6 +141,7 @@ void slacImpl::run() {
                 break;
             case State::Matched:
                 matched = true;
+                EVLOG_info << "publish MATCHED";
```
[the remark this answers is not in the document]
corneliusclaussen: fixed

modules/EvseManager/EvseManager.cpp
corneliusclaussen: done

modules/EvseSlac/main/evse_fsm.hpp
corneliusclaussen: basically the signals are also just registering callback handlers

modules/EvseManager/EvseManager.cpp
```diff
@@ -349,22 +370,18 @@ void EvseManager::ready() {
...
-        r_hlc[0]->subscribe_EVCCIDD([this](const std::string& _token) {
-            std::string token = _token;
-            token.erase(remove(token.begin(), token.end(), ':'), token.end());
-            autocharge_token.id_token = "VID:" + token;
-            autocharge_token.type = types::authorization::TokenType::Autocharge;
-            std::vector<int> referenced_connectors = {this->config.connector_id};
-            autocharge_token.connectors.emplace(referenced_connectors);
+        if (!config.autocharge_use_slac_instead_of_hlc) {
+            r_hlc[0]->subscribe_EVCCIDD([this](const std::string& token) {
+                p_token_provider->publish_provided_token(create_autocharge_token(token, config.connector_id));
```
SebaLukas, line 375:
```cpp
autocharge_token = create_autocharge_token(token, config.connector_id);
````
autocharge_token is published in subscribe_Require_Auth_EIM()

SebaLukas (changes requested):
Only this small thing

SebaLukas (approved):
👍

PR #215, opened 2023-04-14
stop_charging: try graceful shutdown before failed

corneliusclaussen:


modules/EvseV2G/charger/ISO15118_chargerImpl.cpp
```diff
@@ -195,7 +195,31 @@ void ISO15118_chargerImpl::handle_set_RCD_Error(bool& RCD) {
...
+            int i;
+            // allow 10 seconds for graceful shutdown
+            for (i = 0; i < 10; i++) {
+                if (v2g_ctx->is_connection_terminated)
+                    break;
+                sleep(1);
+            }
+
+            // If it did not stop within timeout, stop session using FAILED reply
+            if (i == 9)
```
SebaLukas, line 217: Basically, it's fine. I dont't like the raw number in the if condition.
My suggestion:
```cpp
bool timeout_reached = true;
for (int i = 0; i < 10; i++) {
    if (v2g_ctx->is_connection_terminated) {
        timeout_reached = false;
        break;
    }
    sleep(1);
}

if (timeout_reached) {
```

SebaLukas (changes requested):
One small suggestion :)

modules/EvseV2G/charger/ISO15118_chargerImpl.cpp
```diff
@@ -195,7 +195,31 @@ void ISO15118_chargerImpl::handle_set_RCD_Error(bool& RCD) {
...
+            v2g_ctx->evse_v2g_data.evse_notification = iso1EVSENotificationType_StopCharging;
+            memset(v2g_ctx->evse_v2g_data.evse_status_code, iso1DC_EVSEStatusCodeType_EVSE_Shutdown,
+                   sizeof(v2g_ctx->evse_v2g_data.evse_status_code));
+
+            int i;
+            // allow 10 seconds for graceful shutdown
+            for (i = 0; i < 10; i++) {
+                if (v2g_ctx->is_connection_terminated)
+                    break;
+                sleep(1);
```
hikinggrass, line 213: std::this_thread::sleep_for with std::chrono::seconds might be a bit more readable here

modules/EvseV2G/charger/ISO15118_chargerImpl.cpp
corneliusclaussen: yes that was a little ugly, changed it

SebaLukas (approved):
👍

PR #217, opened 2023-04-19
Integrate shutdown handler into modules

hikinggrass:


a-w50 (commented):
Could you please not on the commit, if the changes are purely due to the autogenerator or if there is anything additional on top of it

hikinggrass:
> Could you please not on the commit, if the changes are purely due to the autogenerator or if there is anything additional on top of it

I'll split this into multiple commits, first one with the autogenerated code, followed by module specific changes

modules/API/API.cpp
```diff
@@ -249,6 +249,11 @@ void API::ready() {
 
 void API::shutdown() {
     invoke_shutdown(*p_main);
+
+    this->running = false;
+    for(auto& api_thread : this->api_threads) {
```
a-w50, line 254: not clang formatted

modules/EnergyManager/EnergyManager.cpp
```diff
@@ -37,6 +37,7 @@ void EnergyManager::ready() {
 
 void EnergyManager::shutdown() {
     invoke_shutdown(*p_main);
+    this->optimizer_thread.stop();
```
a-w50, line 40: I'm not sure how the `utils/thread.hpp` should be used, but `stop()` is also called in its destructor, so is it really needed here?

modules/EnergyManager/EnergyManager.cpp
```diff
@@ -46,7 +46,12 @@ void EnergyManager::enforce_limits(const std::vector<types::energy::EnforcedLimi
             EVLOG_info << fmt::format("\033[1;92m{} Enforce limits {}A {}W \033[1;0m", it.uuid,
                                       it.limits_root_side.value().ac_max_current_A.value_or(-9999),
                                       it.limits_root_side.value().total_power_W.value_or(-9999));
-        r_energy_trunk->call_enforce_limits(it);
+        try {
+            r_energy_trunk->call_enforce_limits(it);
+        } catch (const EverestShuttingDown& e) {
+            // shutting down
+            break;
+        }
```
a-w50, line 54: this looks fishy, because if it would be really necessary to wrap that call into a try/catch, then why wouldn't we need to wrap all 'rpcs' into try/catch?

modules/EvseManager/Charger.cpp
```diff
@@ -45,7 +45,10 @@ Charger::Charger(const std::unique_ptr<board_support_ACIntf>& r_bsp, const std::
 }
 
 Charger::~Charger() {
-    r_bsp->call_pwm_F();
+    // FIXME: this cannot work since there is no guarantee that r_bsp is still around in shutdown
+    // TODO: do we need a pre-shutdown phase that does cleanup?
```
a-w50, line 49: exactly, we really need to think this throught, I guess like in the init, we should only shutdown if all modules successfully signaled, that they are shutdown although they might still receive signals

modules/EvseManager/Charger.cpp
```diff
@@ -71,12 +71,17 @@ void Charger::mainThread() {
...
+            // update power limits
+            powerAvailable();
+
+            // Run our own state machine update (i.e. run everything that needs
+            // to be done on regular intervals independent from events)
+            runStateMachine();
+        } catch (const EverestShuttingDown& e) {
+            // shutting down
+            break;
+        }
```
a-w50: When do we need try/catch and when not?

a-w50 (changes requested):
Only one minor and some fundamental problems :laughing:

PR #220, opened 2023-04-25
Support for multiple SLAC sessions

corneliusclaussen:
Due to cross-talk packets from other SLAC sessions might be received. EvseSlac now handles multiple sessions until one matches. Also adds minimal CM_VALIDATE handling.


a-w50 (approved):
Should be good now after testing has been done successful.

PR #221, opened 2023-04-25
Simple modbus RTU implementation for SerialCommHub

corneliusclaussen:
- Removes libmodbus dependency
- remove sleep time support from serial comm hub interface and all modules

tmp

### DIFF
--- a/modules/EvseManager/EvseManager.cpp
+++ b/modules/EvseManager/EvseManager.cpp
@@ -354,7 +354,6 @@ void EvseManager::ready() {
 
         // implement Auth handlers
         r_hlc[0]->subscribe_Require_Auth_EIM([this]() {
-            // EVLOG_info << "--------------------------------------------- Require_Auth_EIM called";
             p_token_provider->publish_provided_token(autocharge_token);
             //  Do we have auth already (i.e. delayed HLC after charging already running)?
             if ((config.dbg_hlc_auth_after_tstep && charger->Authorized_EIM_ready_for_HLC()) ||
--- a/modules/SerialCommHub/main/serial_communication_hubImpl.cpp
+++ b/modules/SerialCommHub/main/serial_communication_hubImpl.cpp
@@ -4,152 +4,36 @@
 #include "serial_communication_hubImpl.hpp"
 
 #include <chrono>
-#include <connection/exceptions.hpp>
 #include <date/date.h>
 #include <date/tz.h>
 #include <fmt/core.h>
-#include <modbus/exceptions.hpp>
-#include <modbus/utils.hpp>
-
 #include <typeinfo>
 
 namespace module {
 namespace main {
-
-static everest::connection::SerialDeviceConfiguration::BaudRate baudrate_from_int(int baudrate) {
-    switch (baudrate) {
-    case 1200:
-        return everest::connection::SerialDeviceConfiguration::BaudRate::Baud_1200;
-        break;
-    case 2400:
-        return everest::connection::SerialDeviceConfiguration::BaudRate::Baud_2400;
-        break;
-    case 4800:
-        return everest::connection::SerialDeviceConfiguration::BaudRate::Baud_4800;
-        break;
-    case 9600:
-        return everest::connection::SerialDeviceConfiguration::BaudRate::Baud_9600;
-        break;
-    case 19200:
-        return everest::connection::SerialDeviceConfiguration::BaudRate::Baud_19200;
-        break;
-    case 38400:
-        return everest::connection::SerialDeviceConfiguration::BaudRate::Baud_38400;
-        break;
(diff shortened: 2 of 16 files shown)


corneliusclaussen:
I added support for ignoring echo but we need to test this before merging.

modules/EvseManager/EvseManager.cpp
```diff
@@ -1105,6 +1105,9 @@ bool EvseManager::powersupply_DC_set(double _voltage, double _current) {
             // switching from import from grid to export to grid
             session_log.evse(false, "DC power supply: switch ON in import mode");
             r_powersupply_DC[0]->call_setMode(types::power_supply_DC::Mode::Import);
+            EVLOG_critical << "----------------------------------------------------------------------------------------"
+                              "--------------------------";
+            EVLOG_info << "switched ON  import by calling setMode";
```
a-w50, line 1110: same as above

modules/EvseManager/EvseManager.cpp
```diff
@@ -1146,6 +1149,9 @@ bool EvseManager::powersupply_DC_set(double _voltage, double _current) {
             // switching from export to grid to import from grid
             session_log.evse(false, "DC power supply: switch ON in export mode");
             r_powersupply_DC[0]->call_setMode(types::power_supply_DC::Mode::Export);
+            EVLOG_critical << "----------------------------------------------------------------------------------------"
+                              "--------------------------";
+            EVLOG_info << "switched ON export by calling setMode";
```
a-w50, line 1154: same as above

modules/SerialCommHub/main/serial_communication_hubImpl.cpp
```diff
@@ -4,452 +4,160 @@
...
-        if (time_since_last_message < pause_time)
-            std::this_thread::sleep_for(pause_time - time_since_last_message);
 
-        bool done{false};
         uint8_t retry_counter{this->num_resends_on_error};
-        do {
-            if (retry_counter < 1) {
-                return result;
-            }
+        while (retry_counter-- > 0) {
```
a-w50, line 55: only style issue,  readability is a bit harder because of `(retry_counter-- > 0)` is not the same as `(--retry_counter > 0)`, therefor I prefer to put the `retry_counter--` into the while loop itself.
I guess you could also write `auto retry_counter = this->num_resends_on_error;`

modules/SerialCommHub/tiny_modbus_rtu.cpp
```diff
@@ -0,0 +1,327 @@
...
+    This function transmits a modbus request and waits for the reply.
+    Parameter request is optional and is only used for writing multiple registers.
+*/
+std::vector<uint16_t> TinyModbusRTU::txrx(uint8_t device_address, FunctionCode function,
+                                          uint16_t first_register_address, uint16_t register_quantity,
+                                          bool wait_for_reply, std::vector<uint16_t> request) {
+    {
+        // size of request
+        int req_len = (request.size() == 0 ? 0 : 2 * request.size() + 1) + MODBUS_BASE_PAYLOAD_SIZE;
+        uint8_t* req = new uint8_t[req_len];
```
a-w50, line 276: why don't you just use a vector (there is also vector.data() if you want to access read-only contiguess memory)? Problem with `new` is that you need to take care for exceptions until the corresponding delete call, otherwise memory leaks.  Or just use a `unique_ptr` which will also take care of releasing like
`auto req = std::make_unique<uint8_t[]>(req_len);`

modules/SerialCommHub/tiny_modbus_rtu.cpp
```diff
@@ -0,0 +1,327 @@
...
+        // write to serial port
+        write(fd, req, req_len);
+        tcdrain(fd);
+
+        if (ignore_echo) {
+            // read back echo of what we sent and ignore it
+            read_reply(req, req_len);
+        }
+
+        delete req;
```
a-w50, line 315: if `new[]`, you need to use `delete[]`.  Interesting that it didn't break already

modules/EvseManager/EvseManager.cpp
corneliusclaussen: removed

modules/SerialCommHub/tiny_modbus_rtu.cpp
corneliusclaussen: changed to unique_ptr

corneliusclaussen:
Before merging ignore_echo should be fixed

a-w50 (approved):
:+1: 
should be squashed

hikinggrass:
Last commit fixed the GenericPowermeter, looks good :+1:

hikinggrass (approved):

PR #224, opened 2023-04-25
Driver for Pionix micro mega watt charger

corneliusclaussen:


modules/MicroMegaWattBSP/umwc_comms/protobuf/lo2hi.pb.c
```diff
@@ -1,5 +1,5 @@
 /* Automatically generated nanopb constant definitions */
-/* Generated by nanopb-0.4.6 */
+/* Generated by nanopb-0.4.5 */
```
a-w50, line 2: uMWC originally used nanopb 0.4.6, can we make it common with the Yeti one (0.4.5) or upgrade Yeti to 0.4.6

modules/MicroMegaWattBSP/umwc_comms/protobuf/generate_nanopb.sh
```diff
@@ -1,2 +1,2 @@
 #!/bin/sh
-nanopb_generator -I . -D . *.proto
+nanopb_generator.py -L "#include <nanopb/%s>" -I . -D . lo2hi.proto hi2lo.proto
```
a-w50, line 2: The `common.proto` file is still missing

modules/EvseManager/evse/evse_managerImpl.cpp
```diff
@@ -60,6 +60,16 @@ void evse_managerImpl::init() {
     mod->mqtt.subscribe(fmt::format("everest_external/nodered/{}/cmd/resume_charging", mod->config.connector_id),
                         [&charger = mod->charger](const std::string& data) { charger->resumeCharging(); });
 
+    mod->mqtt.subscribe(fmt::format("everest_external/nodered/{}/cmd/stop_transaction", mod->config.connector_id),
```
a-w50, line 63: the fmt::format should be a framework helper function which automatically adds the current correct external prefix

modules/MicroMegaWattBSP/umwc_comms/evSerial.cpp
```diff
@@ -0,0 +1,417 @@
...
+    forced_reset = false;
+    cobsDecodeReset();
+}
+
+evSerial::~evSerial() {
+    if (fd)
+        close(fd);
+}
+
+bool evSerial::openDevice(const char* device, int _baud) {
```
a-w50, line 38: again this `openDevice` function, it might be preferable to even share large parts of the `evSerial.cpp` between Yeti and uMWC

a-w50 (changes requested):
Looks alright, but the `common.proto` file is missing.  Rebase is also necessary, doesn't apply clean anymore

PR #229, opened 2023-04-27
Give some more time for modem and slac to be ready (for Tesla)

corneliusclaussen:


modules/EvseManager/Charger.cpp
```diff
@@ -166,6 +166,9 @@ void Charger::runStateMachine() {
             }
 
             if (hlc_use_5percent_current_session) {
+                // FIXME: wait for SLAC to be ready. Teslas are really fast with sending the first slac packet after
+                // enabling PWM.
+                sleep(1);
```
SebaLukas, line 171: Instead of sleep, use sleep_for:
```cpp
std::this_thread::sleep_for(std::chrono::seconds(1));
```

SebaLukas (changes requested):
Use sleep_for instead of sleep :)

modules/EvseManager/Charger.cpp
```diff
@@ -64,7 +66,7 @@ void Charger::mainThread() {
             break;
         }
 
-        std::this_thread::sleep_for(std::chrono::milliseconds(50));
+        std::this_thread::sleep_for(50ms);
```
a-w50, line 69: I would prefer to have these as constants,
like static const auto SOME_CONSTANT_NAME = std::chrono::milliseconds(50);

modules/EvseManager/Charger.cpp
```diff
@@ -15,6 +15,8 @@
 
 #include <fmt/core.h>
 
+using namespace std::chrono_literals;
```
a-w50, line 18: although it doesn't really hurt here, but extern namespaces should be made completely global, especially if there not used that much (only two times here)

modules/EvseManager/Charger.cpp
```diff
@@ -166,6 +168,9 @@ void Charger::runStateMachine() {
             }
 
             if (hlc_use_5percent_current_session) {
+                // FIXME: wait for SLAC to be ready. Teslas are really fast with sending the first slac packet after
+                // enabling PWM.
+                std::this_thread::sleep_for(1s);
```
a-w50, line 173: same here, could be named like `DELAY_FOR_SLAC_TO_BECOME_READY` or whatever

a-w50 (approved):

SebaLukas (approved):

PR #232, opened 2023-05-02
Refactor/libfsm ng

a-w50:


a-w50:
will be squashed before merging

a-w50:
@corneliusclaussen would be helpful if you could do a quick test with SLAC

corneliusclaussen:
slac works with a real car

corneliusclaussen (approved):

PR #234, opened 2023-05-09
Re-enable SHA1 as supported hash algorithm

FaHaGit:
SHA1 is in mbedtls 2.28 disabled by default and thus TLSv1.2 support is incomplete if the application does not allow it explicitly.

### DIFF
--- a/modules/EvseV2G/connection.cpp
+++ b/modules/EvseV2G/connection.cpp
@@ -50,6 +50,29 @@ mbedtls_ssl_cache_context cache;
 static const int v2g_cipher_suites[] = {MBEDTLS_TLS_ECDH_ECDSA_WITH_AES_128_CBC_SHA256,
                                         MBEDTLS_TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA256, 0};
 
+/* list of allowed hashes which our TLS server supports; this list is
+ * copied from mbedtls' ssl_preset_default_hashes (ssl_tls.c) with one
+ * nitpick: the MBEDTLS_MD_SHA1 is enabled unconditionally since TLSv1.2
+ * uses it as fallback when the client don't send 'signature_algorithms'
+ * extension; this means that we require this algo and including it without
+ * define results in a compile error in case we try to build against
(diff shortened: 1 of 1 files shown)


modules/EvseV2G/connection.cpp
```diff
@@ -50,6 +50,29 @@ mbedtls_ssl_cache_context cache;
 static const int v2g_cipher_suites[] = {MBEDTLS_TLS_ECDH_ECDSA_WITH_AES_128_CBC_SHA256,
                                         MBEDTLS_TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA256, 0};
 
+/* list of allowed hashes which our TLS server supports; this list is
+ * copied from mbedtls' ssl_preset_default_hashes (ssl_tls.c) with one
+ * nitpick: the MBEDTLS_MD_SHA1 is enabled unconditionally since TLSv1.2
+ * uses it as fallback when the client don't send 'signature_algorithms'
+ * extension; this means that we require this algo and including it without
+ * define results in a compile error in cause we try to build against
```
barsnick, line 58: spelling nitpíck: `in case`

modules/EvseV2G/connection.cpp
```diff
@@ -50,6 +50,29 @@ mbedtls_ssl_cache_context cache;
...
+ * a library where it is disabled (by error).
+ */
+static const int v2g_ssl_allowed_hashes[] = {
+#if defined(MBEDTLS_SHA512_C)
+    MBEDTLS_MD_SHA512,
+#endif
+#if defined(MBEDTLS_SHA512_C) && !defined(MBEDTLS_SHA512_NO_SHA384)
```
barsnick, line 65: This could be embedded in the previous block, but that's probably a question of style and preference. Something like:

```
#if defined(MBEDTLS_SHA512_C)
    MBEDTLS_MD_SHA512,
#  if !defined(MBEDTLS_SHA512_NO_SHA384)
    MBEDTLS_MD_SHA384,
#  endif
#endif
```

corneliusclaussen:
I just rebased it to latest main

PR #238, opened 2023-05-10
change EVCCID/MAC addresses to uppercase

corneliusclaussen:
right now it is not specified whether mac address strings should be upper or lower case. Defining this to upper case everywhere makes it a bit easier to use.


a-w50 (approved):
Not sure about the code formatting of EvseV2G files, but other than that - :100:

SebaLukas (approved):
👍
Not sure about the format changes either.

PR #239, opened 2023-05-11
Disable pnc if a rfid card was swiped before an ev was plugged in

SebaLukas:


modules/EvseManager/EvseManager.cpp
```diff
@@ -634,6 +634,26 @@ void EvseManager::ready() {
             ev_info = types::evse_manager::EVInfo();
             p_evse->publish_ev_info(ev_info);
         }
+
+        if (s == types::evse_manager::SessionEventEnum::SessionStarted &&
+            charger->getSessionStartedReason() == types::evse_manager::StartSessionReason::Authorized) {
+            Array payment_options;
+            payment_options.insert(payment_options.end(), "ExternalPayment");
+            r_hlc[0]->call_set_PaymentOptions(payment_options);
```
Pietfried, line 642: we should check if hlc is enabled / connection is available

Pietfried (changes requested):

modules/EvseManager/EvseManager.cpp
SebaLukas: Added

Pietfried (approved):

PR #245, opened 2023-05-12
Adapt to changes in EVerest/everest-utils#49

a-w50:
- disabled PyTestController in tests, this is broken

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -4,9 +4,5 @@
 import os
 
 def pytest_addoption(parser):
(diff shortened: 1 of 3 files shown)


LAD101work (approved):
This PR effectively disables all higher-level tests (those that require the TestControlModule). 

I agree, that for a clean solution, the everestpy funtionality should be reworked. Since this blocks all basic charging tests (and higher), this should only be done, if waiting for a change in everestpy is acceptable. 

If so, then I approve.

PR #247, opened 2023-05-13
Initial ci based on EVerest/everest-ci

a-w50:


.ci/build-kit/compile.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+
```
LAD101work, line 2: Currently, the steps to build utils, which this workflow depends upon is being done in the base image generation. If for example `ev-cli` needs to be updated for changes in this PR to work, there should be a local option/step to also build and install `ev-cli` and/or other tools.

Maybe it would not hurt to add a line to do that here?

.ci/e2e/scripts/tests.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+
+cd tests
+pytest --everest-prefix ../dist core_tests/startup_tests.py
```
LAD101work, line 4: There is currently no results artifact generation being done here.

I would advise to add this, even if it does currently only provide basic error output. This makes report retrieval much easier and can later be expanded to include other results (as discussed: build-logs and other reports).

.github/workflows/build_and_test.yaml
```diff
@@ -0,0 +1,51 @@
...
+          --volume "$(pwd):/ext" \
+          --name test-container \
+          build-image run-script test_and_install
+      - name: Run integration tests
+        run: |
+          docker commit test-container integration-image
+          pushd source/.ci/e2e
+          docker-compose run e2e-test-server run-script tests
+          
+
```
LAD101work, line 48: There is currently no results upload for artifacts generated during the workflow run being done.

I would advise to add this to obtain easily retrievable results which can be stored for longer than this workflow-run-log will be kept by Github (which I believe is currently 90 days?).

LAD101work (approved):
Please review comments and consider adding suggested steps.

Other than that, I approve.

PR #249, opened 2023-05-22
Adapted PyJosev to EVerest/everest-framework#85

a-w50:


modules/PyJosev/utilities.py
```diff
@@ -0,0 +1,104 @@
...
+    # handler.setFormatter(formatter)
+
+    logging.getLogger().addHandler(handler)
+
+def choose_first_ipv6_local() -> str:
+    for iface in netifaces.interfaces():
+        if netifaces.AF_INET6 in netifaces.ifaddresses(iface):
+            for netif_inet6 in netifaces.ifaddresses(iface)[netifaces.AF_INET6]:
+                if 'fe80' in netif_inet6['addr']:
+                    return iface
```
sdrabb, line 51: Single return woul be better

a-w50: Why would a single return be better?

PR #250, opened 2023-05-22
Feature/serialcommhub support rxtx gpio

corneliusclaussen:


lib/staging/gpio/gpio.cpp
```diff
@@ -0,0 +1,112 @@
...
+            line = chip.get_line(line_number);
+            is_open = true;
+        } catch (const std::system_error& e) {
+            is_open = false;
+        } catch (const std::out_of_range& e) {
+            is_open = false;
+        }
+    }
+
+    inverted = _inverted;
```
sdrabb, line 26: We can consider `this` keyword here

lib/staging/gpio/gpio.cpp
```diff
@@ -0,0 +1,112 @@
...
+        gpiod::line_request c;
+        c.consumer = "EVerest";
+        c.request_type = gpiod::line_request::DIRECTION_INPUT;
+        c.flags = gpiod::line_request::FLAG_BIAS_PULL_UP;
+        line.request(c);
+    } catch (const std::system_error& e) {
+        is_open = false;
+        return false;
+    }
+    return true;
```
sdrabb, line 84: https://stackoverflow.com/a/733858

lib/staging/gpio/gpio.cpp
```diff
@@ -0,0 +1,112 @@
...
+        is_open = false;
+    } else {
+        try {
+            chip.open(chip_name);
+            line = chip.get_line(line_number);
+            is_open = true;
+        } catch (const std::system_error& e) {
+            is_open = false;
+        } catch (const std::out_of_range& e) {
+            is_open = false;
```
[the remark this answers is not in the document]
corneliusclaussen: renamed is_open to ready, which describes better what the flag does and refactored its usage, have a look

lib/staging/gpio/gpio.cpp
corneliusclaussen: see above comment, now there is only a single return

lib/staging/gpio/gpio.cpp
corneliusclaussen: we could use this->inverted=inverted here, I think it is a matter of personal taste :-) I personally don't like shadowing member variables with local variables that much as it always causes confusion.

sdrabb (approved):

corneliusclaussen:
converted to draft, dependency to libgpiod will be removed beofre merging. This would add another LGPL 3 dependency which is a bit unclear since the C++ bindings may generate code in the header files. Will be replaced by direct ioctl calls without external dependency.

hikinggrass (approved):

PR #255, opened 2023-05-25
Increase EvseV2G TLS stability

Pietfried:
- rolled back mbed tls dependency to temporarily remove support for the trusted_ca_key extension
- increased tls read timeout from 2s to 5s


SebaLukas (approved):
👍

PR #257, opened 2023-05-30
Adopt to EVerest/everest-framework#87

a-w50:


config/config-sil.yaml
```diff
@@ -103,9 +103,8 @@ active_modules:
   iso15118_charger:
     config_module:
       device: auto
-      tls_security: allow
     connections: {}
-    module: EvseV2G
```
hikinggrass, line 108: How is that related to the rest?

hikinggrass (approved):

PR #260, opened 2023-06-13
Fix issue with multiple service list items

FaHaGit:
The mechanism for adding new service list items has been reworked. Now there is a separate function which is responsible to add new services to the service list. Now it is no longer possible to add multiple service list items of the same service ID to the service list. In case the function is called with a service ID which has been already configured, the service list item will now be overwritten.

### DIFF
--- a/modules/EvseV2G/charger/ISO15118_chargerImpl.cpp
+++ b/modules/EvseV2G/charger/ISO15118_chargerImpl.cpp
@@ -354,19 +354,16 @@ void ISO15118_chargerImpl::handle_cableCheck_Finished(bool& status) {
 void ISO15118_chargerImpl::handle_set_Certificate_Service_Supported(bool& status) {
     if (status == true) {
         // For setting "Certificate" in ServiceList in ServiceDiscoveryRes
+        struct iso1ServiceType cert_service;
         const std::string service_name = "Certificate";
-        uint8_t len = service_name.length();
-        v2g_ctx->evse_v2g_data.evse_service_list[v2g_ctx->evse_v2g_data.evse_service_list_len].FreeService = 1;
-        v2g_ctx->evse_v2g_data.evse_service_list[v2g_ctx->evse_v2g_data.evse_service_list_len].ServiceID = 2;
-        v2g_ctx->evse_v2g_data.evse_service_list[v2g_ctx->evse_v2g_data.evse_service_list_len].ServiceCategory = iso1serviceCategoryType_ContractCertificate;
-        memcpy(v2g_ctx->evse_v2g_data.evse_service_list[v2g_ctx->evse_v2g_data.evse_service_list_len].ServiceName.characters, reinterpret_cast<const char*>(service_name.data()), len);
-        v2g_ctx->evse_v2g_data.evse_service_list[v2g_ctx->evse_v2g_data.evse_service_list_len].ServiceName.charactersLen = len;
-        v2g_ctx->evse_v2g_data.evse_service_list_len += 1;
+        const int16_t cert_parameter_set_id[] = {1}; // parameter-set-ID 1: "Installation" service. TODO: Support of the "Update" service (parameter-set-ID 2)
 
-        // For setting "Installation" in the ServiceParameterList in ServiceDetailRes
-        configure_parameter_set(&v2g_ctx->evse_v2g_data.service_parameter_list[0], 1);
-        // TODO: support "Update" in the ServiceParameterList in ServiceDetailRes
-        // configure_parameter_set(&v2g_ctx->evse_v2g_data.service_parameter_list[0], 2);
+        cert_service.FreeService = 1; // true
+        cert_service.ServiceID = 2; // as defined in ISO 15118-2
+        cert_service.ServiceCategory = iso1serviceCategoryType_ContractCertificate;
+        memcpy(cert_service.ServiceName.characters, reinterpret_cast<const char*>(service_name.data()), service_name.length());
+
+        add_service_to_service_list(v2g_ctx, cert_service, cert_parameter_set_id, sizeof(cert_parameter_set_id) / sizeof(cert_parameter_set_id[0]));
     }
 }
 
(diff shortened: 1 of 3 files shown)


sdrabb (approved):

modules/EvseV2G/charger/ISO15118_chargerImpl.cpp
```diff
@@ -354,19 +354,16 @@ void ISO15118_chargerImpl::handle_cableCheck_Finished(bool& status) {
...
-        // For setting "Installation" in the ServiceParameterList in ServiceDetailRes
-        configure_parameter_set(&v2g_ctx->evse_v2g_data.service_parameter_list[0], 1);
-        // TODO: support "Update" in the ServiceParameterList in ServiceDetailRes
-        // configure_parameter_set(&v2g_ctx->evse_v2g_data.service_parameter_list[0], 2);
+        const int16_t cert_parameter_set_id = 1; // parameter-set-ID 1: "Installation" service. TODO: Support of the "Update" service (parameter-set-ID 2)
+
+        cert_service.FreeService = 1; // true
+        cert_service.ServiceID = 2; // as defined in ISO 15118-2
+        cert_service.ServiceCategory = iso1serviceCategoryType_ContractCertificate;
+        memcpy(cert_service.ServiceName.characters, reinterpret_cast<const char*>(service_name.data()), service_name.length());
```
sdrabb, line 364: Mix camelcase with snake case, is what we want?

sdrabb (approved):

modules/EvseV2G/charger/ISO15118_chargerImpl.cpp
FaHaGit: sometimes not really avoidable because external libraries, like openv2g, use camel case but everest framework uses snake case

modules/EvseV2G/charger/ISO15118_chargerImpl.cpp
```diff
@@ -354,19 +354,16 @@ void ISO15118_chargerImpl::handle_cableCheck_Finished(bool& status) {
...
-        v2g_ctx->evse_v2g_data.evse_service_list[v2g_ctx->evse_v2g_data.evse_service_list_len].ServiceCategory = iso1serviceCategoryType_ContractCertificate;
-        memcpy(v2g_ctx->evse_v2g_data.evse_service_list[v2g_ctx->evse_v2g_data.evse_service_list_len].ServiceName.characters, reinterpret_cast<const char*>(service_name.data()), len);
-        v2g_ctx->evse_v2g_data.evse_service_list[v2g_ctx->evse_v2g_data.evse_service_list_len].ServiceName.charactersLen = len;
-        v2g_ctx->evse_v2g_data.evse_service_list_len += 1;
-
-        // For setting "Installation" in the ServiceParameterList in ServiceDetailRes
-        configure_parameter_set(&v2g_ctx->evse_v2g_data.service_parameter_list[0], 1);
-        // TODO: support "Update" in the ServiceParameterList in ServiceDetailRes
-        // configure_parameter_set(&v2g_ctx->evse_v2g_data.service_parameter_list[0], 2);
+        const int16_t cert_parameter_set_id = 1; // parameter-set-ID 1: "Installation" service. TODO: Support of the "Update" service (parameter-set-ID 2)
```
mooraby, line 359: so the idea in the future will be creating an array of const int16_t to pass it to add_service_to_service_list? Why not finish the implementation now?

modules/EvseV2G/v2g_ctx.cpp
```diff
@@ -479,49 +479,90 @@ void log_selected_energy_transfer_type(int selected_energy_transfer_mode) {
...
+            }
+        }
+
+        if (service_found == false && 
+            (v2g_ctx->evse_v2g_data.evse_service_list_len - 1) != iso1ServiceListType_Service_ARRAY_SIZE) {
+            write_idx = v2g_ctx->evse_v2g_data.evse_service_list_len;
+            v2g_ctx->evse_v2g_data.evse_service_list_len++;
+        }
+        else if (v2g_ctx->evse_v2g_data.evse_service_list_len == iso1ServiceListType_Service_ARRAY_SIZE) {
+            dlog(DLOG_LEVEL_WARNING, "Maximum service list size reached. Unable to add service ID %u", evse_service.ServiceID);
```
mooraby, line 502: Is it only a warning?

modules/EvseV2G/v2g_ctx.cpp
```diff
@@ -479,49 +479,90 @@ void log_selected_energy_transfer_type(int selected_energy_transfer_mode) {
...
+            return false;
```
mooraby, line 503: does the return of the function has to be bool? You do not use the return in the calling function

FaHaGit: not used now but maybe someone in the future ;)

modules/EvseV2G/charger/ISO15118_chargerImpl.cpp
FaHaGit: because we are currently only supporting one parameter-set-id. Why we should create an array for one element?

modules/EvseV2G/v2g_ctx.cpp
```diff
@@ -479,49 +479,100 @@ void log_selected_energy_transfer_type(int selected_energy_transfer_mode) {
...
 
-        parameterSet->ParameterSetID = parameterSetId;
+void configure_parameter_set(struct iso1ServiceParameterListType* parameterSetList, int16_t parameterSetId,
+                             uint16_t serviceId) {
 
+    bool parameter_set_id_found = false;
+    uint8_t write_idx = 0;
+    for (uint8_t idx = 0; idx < parameterSetList->ParameterSet.arrayLen; idx++) {
+        if (parameterSetList->ParameterSet.array[idx].ParameterSetID == parameterSetId) {
+            parameter_set_id_found = true;
```
mooraby, line 526: Is the whole for loop and the check needed?

PR #261, opened 2023-06-20
Adding the PyEvJosev module to replace the JsCarV2G module

SebaLukas:
Please test with this PR: [Integration of the EVCC from Josev into EVerest](https://github.com/EVerest/ext-switchev-iso15118/pull/13)

### DIFF
--- a/modules/PyEvJosev/module.py
+++ b/modules/PyEvJosev/module.py
@@ -1,0 +1,113 @@
+# SPDX-License-Identifier: Apache-2.0
+# Copyright 2020 - 2023 Pionix GmbH and Contributors to EVerest
+import asyncio
+import sys
+from pathlib import Path
+import threading
+
+from everest.framework import Module, RuntimeSession, log
+
+# fmt: off
+JOSEV_WORK_DIR = Path(__file__).parent / '../../3rd_party/josev'
+sys.path.append(JOSEV_WORK_DIR.as_posix())
+
+from iso15118.evcc import EVCCHandler
+from iso15118.evcc.controller.simulator import SimEVController
+from iso15118.evcc.evcc_config import EVCCConfig
+from iso15118.evcc.everest import context as JOSEV_CONTEXT
+from iso15118.shared.exificient_exi_codec import ExificientEXICodec
+from iso15118.shared.settings import set_PKI_PATH
+
+from utilities import (
+    setup_everest_logging,
+    determine_network_interface,
+    patch_josev_config
+)
+
+setup_everest_logging()
+
+EVEREST_CERTS_SUB_DIR = 'certs'
+
+async def evcc_handler_main_loop(module_config: dict):
+    """
+    Entrypoint function that starts the ISO 15118 code running on
+    the EVCC (EV Communication Controller)
+    """
+    iface = determine_network_interface(module_config['device'])
+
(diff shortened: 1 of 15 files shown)


SebaLukas:
Missing: The private contract cert key is not saved in the correct format (.pem) during the certificate installation process.

modules/PyEvJosev/module.py
```diff
@@ -0,0 +1,113 @@
...
+            m.implement_command(
+                'ev', cmd, getattr(self, f'_handler_{cmd}'))
+            
+        # init ready event
+        self._ready_event = threading.Event()
+
+        self._mod = m
+        self._mod.init_done(self._ready)
+
+    def start_evcc_hander(self):
```
sdrabb, line 80: typo : )

sdrabb:
@SebaLukas what about EVSlac side? Are you working on it?

SebaLukas:
@sdrabb we are working on it :)

SebaLukas:
For PnC testing I will provide a extra pnc config

sdrabb (approved):

modules/PyEvJosev/module.py
```diff
@@ -0,0 +1,113 @@
...
+# fmt: off
+JOSEV_WORK_DIR = Path(__file__).parent / '../../3rd_party/josev'
+sys.path.append(JOSEV_WORK_DIR.as_posix())
+
+from iso15118.evcc import EVCCHandler
+from iso15118.evcc.controller.simulator import SimEVController
+from iso15118.evcc.evcc_config import EVCCConfig
+from iso15118.evcc.everest import context as JOSEV_CONTEXT
+from iso15118.shared.exificient_exi_codec import ExificientEXICodec
+from iso15118.shared.settings import set_PKI_PATH
```
a-w50, line 19: I would prefer having this solved differently / we need to come up with a best practice how external python modules are distributed to the system
- either by copying them directly
- or by pip
- wheels?

modules/PyEvJosev/utilities.py
```diff
@@ -0,0 +1,93 @@
...
+
+    def emit(self, record):
+        msg = self.format(record)
+
+        log_level: int = record.levelno
+        if log_level == logging.CRITICAL:
+            log.critical(msg)
+        elif log_level == logging.ERROR:
+            log.error(msg)
+        elif log_level == logging.WARNING:
```
a-w50, line 24: This is, I guess, exactly the same as in the EVSE part.  I'd prefer removing this duplication by creating a shared library for this inside 'everest-core/lib' - again we would need to think on how to package and distribute/install that later on

a-w50 (commented):
See my comments, depending on if we want to postpone these issues or not, this PR can be approved or not

modules/PyEvJosev/utilities.py
SebaLukas: It's the same code

SebaLukas:
Those are both good points. However, I would not solve them in this PR, but first discuss them and then possibly solve them in a separate PR.

a-w50 (approved):
Fine, please rebase and squash

corneliusclaussen:
Converted to draft, SIL is not working at the moment for ISO. This branch should fix SIL before merging

PR #263, opened 2023-06-27
EvseV2G: Do not stay long in authorization loop

mooraby:
An optional timeout is added to prevent an authorziation loop. Timeouts can be ignored if set to zero.

For PnC, 55 seconds is the default value for timeout according to V2G_SECC_Ongoing_Performance_Time.

For EIM, the 5 minutes value is not derived from ISO standard.

### DIFF
--- a/modules/EvseV2G/iso_server.cpp
+++ b/modules/EvseV2G/iso_server.cpp
@@ -1090,6 +1090,7 @@ static enum v2g_event handle_iso_payment_service_selection(struct v2g_connection
         /* Set next expected req msg */
         conn->ctx->state = (int)
             iso_dc_state_id::WAIT_FOR_AUTHORIZATION; // [V2G-551] (iso specification describes only the ac case... )
+        conn->ctx->session.auth_start_timeout = getmonotonictime();
     }
 
     return next_event;
@@ -1260,7 +1261,7 @@ static enum v2g_event handle_iso_payment_details(struct v2g_connection* conn) {
         ProvidedIdToken.id_token = std::string(cert_emaid);
         ProvidedIdToken.type = types::authorization::TokenType::PlugAndCharge;
         if (contract_cert_chain_pem.empty() == false) {
-           ProvidedIdToken.certificate = contract_cert_chain_pem;
+            ProvidedIdToken.certificate = contract_cert_chain_pem;
         }
         conn->ctx->p_charger->publish_Require_Auth_PnC(ProvidedIdToken);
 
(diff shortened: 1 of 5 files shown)


modules/EvseV2G/iso_server.cpp
```diff
@@ -1293,6 +1295,9 @@ static enum v2g_event handle_iso_authorization(struct v2g_connection* conn) {
     struct iso1AuthorizationReqType* req = &conn->exi_in.iso1EXIDocument->V2G_Message.Body.AuthorizationReq;
     struct iso1AuthorizationResType* res = &conn->exi_out.iso1EXIDocument->V2G_Message.Body.AuthorizationRes;
     enum v2g_event next_event = V2G_EVENT_NO_EVENT;
+    struct timespec ts_abs_timeout;
+    int rv = 0;
```
SebaLukas, line 1299: Delete rv, see codacy issue

SebaLukas (approved):
Looking good 👍 Before the merge please delete the unused variable 'rv'

modules/EvseV2G/iso_server.cpp
```diff
@@ -1329,12 +1334,24 @@ static enum v2g_event handle_iso_authorization(struct v2g_connection* conn) {
     }
     res->EVSEProcessing = (iso1EVSEProcessingType)conn->ctx->evse_v2g_data.evse_processing[PHASE_AUTH];
 
+    if (conn->ctx->evse_v2g_data.evse_processing[PHASE_AUTH] != iso1EVSEProcessingType_Finished) {
+        if (((is_payment_option_contract == false) && (conn->ctx->session.auth_timeout_eim == 0)) ||
+            ((is_payment_option_contract == true) && (conn->ctx->session.auth_timeout_pnc == 0))) {
+            dlog(DLOG_LEVEL_DEBUG, "Waiting for authorization forever!");
+        } else if ((time(NULL) - conn->ctx->session.auth_start_timeout) >=
+                   (is_payment_option_contract ? (time_t)conn->ctx->session.auth_timeout_pnc
+                                               : (time_t)conn->ctx->session.auth_timeout_eim)) {
```
a-w50, line 1343: I would advise to use `clock_gettime` using the monotonic clock here, otherwise when the system time changes, this will lead to hard to debug bugs

a-w50 (changes requested):

modules/EvseV2G/iso_server.cpp
mooraby: I do not entirely understand your point.. when would the system time change? In the period between start of counting (e.g PaymentDetails in PnC) and the check in Authorization? Can you describe more the corner case?

PR #264, opened 2023-06-29
Re-apply "Fix issue with multiple service list items (#260)"

corneliusclaussen:
This re-applies PR #260. The regression found was independent of this PR and will be fixed seperately.


SebaLukas (approved):
👍

PR #265, opened 2023-06-29
Driver for SCU Power DPM1000/30 DC power supply

corneliusclaussen:
This adds a driver module for the DPM1000/30 DC power supply module. For now it only supports one module, stacked configurations will be added later on.


a-w50 (approved):
Ok

PR #266, opened 2023-07-04
Slac implementation for EV

a-w50:


lib/staging/slac/fsm/ev/src/states/sounding.cpp
```diff
@@ -0,0 +1,156 @@
...
+#include <slac/fsm/ev/states/others.hpp>
+
+namespace slac::fsm::ev {
+
+SoundingState::SoundingState(Context& ctx, SessionParamaters session_parameters_) :
+    FSMSimpleState(ctx), session_parameters(std::move(session_parameters_)) {
+}
+
+void SoundingState::enter() {
+    ctx.signal_state("SOUNDING");
```
corneliusclaussen, line 21: "SOUNDING" is not a valid state enum. Are the correct states (MATCHED etc) also published somewhere?

modules/EvseSlac/main/slacImpl.cpp
```diff
@@ -61,7 +61,7 @@ void slacImpl::run() {
     }
 
     // setup callbacks
-    ContextCallbacks callbacks;
+    slac::fsm::evse::ContextCallbacks callbacks;
     callbacks.send_raw_slac = [&slac_io](slac::messages::HomeplugMessage& msg) { slac_io.send(msg); };
 
     callbacks.signal_dlink_ready = [this](bool value) { publish_dlink_ready(value); };
```
corneliusclaussen, line 67: This is never called. We could implement the proprietary extension to find out whether the PLC actually linked with a peer, but not sure if this is portable to other PLC chips. So for now it would be nice to signal this when we go to MATCHED and when we leave MATCHED

corneliusclaussen (changes requested):

modules/EvseSlac/main/slacImpl.cpp
corneliusclaussen: The comment was about the signal_dlink_ready

lib/staging/slac/fsm/ev/src/states/sounding.cpp
a-w50: Right, this should be corrected.  I will have a look at the interface definitions.

modules/EvseSlac/main/slacImpl.cpp
a-w50: Correct, I need to review the interface

lib/staging/slac/fsm/ev/src/states/sounding.cpp
corneliusclaussen: Added the correct state and dlink_ready publishes

corneliusclaussen (approved):

PR #268, opened 2023-07-05
Fix josev dependency for main

corneliusclaussen:


dependencies.yaml
```diff
@@ -42,8 +42,7 @@ libocpp:
 # Josev
 Josev:
   git: https://github.com/EVerest/ext-switchev-iso15118.git
-  git_tag: 59f6956
-
+  git_tag: everest
```
hikinggrass, line 45: can we not pin this to a specific version/rev/tag ?

hikinggrass (approved):

PR #276, opened 2023-07-07
Allow Modules in sub directories

corneliusclaussen:
everest-core has many different modules now and it is hard to find out the type of a module, e.g. if it is a hardware driver or a simulation module etc.
This PR allows to move modules into subfolders in the source tree. Upon install, the modules will still be installed flat into one folder. Two modules are moved to a deprected folder as an example. All other modules are not moved as this would create a lot of conflicts with other branches. Moving into a new structure needs to be a coordinated effort, but this adds the basis to be able to do that.


a-w50:
See #279

cmake/everest-generate.cmake
```diff
@@ -244,7 +244,7 @@ function (ev_add_cpp_module MODULE_NAME)
                         --disable-clang-format
                         --schemas-dir "$<TARGET_PROPERTY:generate_cpp_files,EVEREST_SCHEMA_DIR>"
                         --output-dir ${GENERATED_MODULE_DIR}
-                        ${MODULE_NAME}
+                        ${MODULE_SUBDIR}/${MODULE_NAME}
```
a-w50, line 247: This shouldn't work without modifications in ev-cli, please have a look at #279

corneliusclaussen:
Replaced by PR279

PR #280, opened 2023-07-10
first draft of directory structure for discussion

corneliusclaussen:
This is a first proposal for a sub directory structure for modules. This will break a lot of branches as all modules are moved to different folders. Let's discuss in the upcoming dev meeting!


modules/CMakeLists.txt
```diff
@@ -1,38 +1,13 @@
...
-ev_add_module(PowermeterBSM)
-ev_add_module(DummyV2G)
-ev_add_module(MicroMegaWattBSP)
-ev_add_module(DPM1000)
-
 add_subdirectory(deprecated)
 add_subdirectory(examples)
 add_subdirectory(simulation)
+add_subdirectory(evse)
+add_subdirectory(drivers)
```
a-w50, line 5: I would call it `hardware_drivers`

PR #286, opened 2023-07-13
EvseManager now signals Enabled event when ready

Pietfried:
EvseManager now signals Enabled event when ready for charging. OCPP listens to this event and only starts up when all EvseManagers it requires are ready. This was a problem when OCPP called cmds of EvseManager when the EvseManager was not done with his ready() function


modules/EvseManager/EvseManager.cpp
```diff
@@ -774,6 +774,7 @@ void EvseManager::ready() {
         }
     });
 
+    this->charger->signalEvent(types::evse_manager::SessionEventEnum::Enabled);
```
corneliusclaussen, line 777: This should not be neccessary. When charger->enable() is called it will also emit the event

corneliusclaussen: I'll push a commit removing this

PR #287, opened 2023-07-13
Fix startup tests

hikinggrass:


tests/core_tests/basic_charging_tests.py
```diff
@@ -8,7 +8,7 @@
 
 from everest.testing.core_utils.fixtures import *
 from everest.testing.core_utils.everest_core import EverestCore
-from validations.user_functions import *
+# from validations.user_functions import *
```
a-w50, line 11: Just remove it

tests/core_tests/startup_tests.py
```diff
@@ -76,17 +76,21 @@ def test(self, timeout: float) -> bool:
 
         return True
 
-
+@pytest.mark.everest_core_config('config-sil.yaml')
```
a-w50, line 79: Thats properly the correct way of passing this information into pytest.  Wondering if we should also pass the modules_to_test via these decorators

tests/core_tests/startup_tests.py
```diff
@@ -76,17 +76,21 @@ def test(self, timeout: float) -> bool:
...
     everest_core.start(standalone_module='probe_module', modules_to_test=[
-                       TestControlModuleConnection(evse_manager_id="connector_1", car_simulator_id="car_simulator")])
+                       TestControlModuleConnection(evse_manager_id="connector_1", car_simulator_id="car_simulator", ocpp_id=None)])
     logging.info("everest-core ready, waiting for probe module")
 
     session = RuntimeSession(str(everest_core.prefix_path), str(everest_core.everest_config_path))
 
     probe = ProbeModule(session)
 
+    if everest_core.status_listener.wait_for_status(10, ["ALL_MODULES_STARTED"]):
```
a-w50, line 92: The `everest_core` class' api should be made easier to use, for example it could have a public `wait_for_status` function, which it forwards to its private `status_listener` member

a-w50 (approved):

sdrabb (approved):

PR #289, opened 2023-07-14
Removed EvseV2G config option highlevel_authentication_mode

SebaLukas:
Since it can lead to a confusion where the payment methods are configured, the config option of the EvseV2G has been deleted.
The payment options are only now communicated via the EvseManager and the iso15118_charger interface.


FaHaGit (approved):
Looks ok for me. The configuration over the yaml would never work as long as it will be overwritten over the MQTT interface.

PR #290, opened 2023-07-14
Changed Authorization and EvseManager types for better support for OCPP201

Pietfried:
- support for RemoteStart / RemoteStop within OCPP201
- added Powermeter type to TransactionStarted and TransactionFinished event of EvseManager
- added some more StopReasons to StopReasonEnum
- renamed property type of ProvidedIdToken to authorization_type (RFID, OCPP, PnC, Autocharge)
- added id_token_type as optional param to ProvidedIdToken (Central, eMAID, ISO14443,...)
- changed types required changes in Auth, EvseManager, EvseV2G, OCPP, OCPP201 JsDummyTokenProvider, PN532TokenProvider, and JsPN532TokenProvider
- included new Powermeter and stopreason and remote start id  in OCPP201

### DIFF
--- a/modules/Auth/tests/auth_tests.cpp
+++ b/modules/Auth/tests/auth_tests.cpp
@@ -33,7 +33,7 @@ static ProvidedIdToken get_provided_token(const std::string& id_token,
                                           std::optional<bool> prevalidated = std::nullopt) {
     ProvidedIdToken provided_token;
     provided_token.id_token = id_token;
-    provided_token.type = types::authorization::TokenType::RFID;
+    provided_token.authorization_type = types::authorization::AuthorizationType::RFID;
     if (connectors) {
         provided_token.connectors.emplace(connectors.value());
     }
@@ -153,8 +153,8 @@ TEST_F(AuthTest, test_stop_transaction) {
     SessionEvent session_event2;
     session_event2.event = SessionEventEnum::TransactionStarted;
     TransactionStarted transaction_event;
-    transaction_event.energy_Wh_import = 0;
-    transaction_event.id_tag = provided_token.id_token;
+    transaction_event.meter_value.energy_Wh_import.total = 0;
+    transaction_event.id_tag = provided_token;
     transaction_event.timestamp = Everest::Date::to_rfc3339(date::utc_clock::now());
     session_event2.transaction_started = transaction_event;
 
@@ -385,16 +385,16 @@ TEST_F(AuthTest, test_transaction_finish) {
     SessionEvent session_event2;
     session_event2.event = SessionEventEnum::TransactionStarted;
     TransactionStarted transaction_event1;
-    transaction_event1.energy_Wh_import = 0;
-    transaction_event1.id_tag = provided_token_1.id_token;
+    transaction_event1.meter_value.energy_Wh_import.total = 0;
+    transaction_event1.id_tag = provided_token_1;
     transaction_event1.timestamp = Everest::Date::to_rfc3339(date::utc_clock::now());
     session_event2.transaction_started = transaction_event1;
 
     SessionEvent session_event3;
     session_event3.event = SessionEventEnum::TransactionStarted;
     TransactionStarted transaction_event2;
-    transaction_event2.energy_Wh_import = 0;
-    transaction_event2.id_tag = provided_token_2.id_token;
+    transaction_event2.meter_value.energy_Wh_import.total = 0;
+    transaction_event2.id_tag = provided_token_2;
     transaction_event2.timestamp = Everest::Date::to_rfc3339(date::utc_clock::now());
     session_event3.transaction_started = transaction_event2;
 
@@ -450,8 +450,8 @@ TEST_F(AuthTest, test_parent_id_finish) {
     SessionEvent session_event2;
     session_event2.event = SessionEventEnum::TransactionStarted;
     TransactionStarted transaction_event1;
-    transaction_event1.energy_Wh_import = 0;
-    transaction_event1.id_tag = provided_token_1.id_token;
+    transaction_event1.meter_value.energy_Wh_import.total = 0;
+    transaction_event1.id_tag = provided_token_1;
     transaction_event1.timestamp = Everest::Date::to_rfc3339(date::utc_clock::now());
     session_event2.transaction_started = transaction_event1;
 
@@ -549,8 +549,8 @@ TEST_F(AuthTest, test_parent_id_finish_because_no_available_connector) {
     SessionEvent session_event_3;
     session_event_3.event = SessionEventEnum::TransactionStarted;
     TransactionStarted transaction_event;
-    transaction_event.energy_Wh_import = 0;
-    transaction_event.id_tag = provided_token_1.id_token;
+    transaction_event.meter_value.energy_Wh_import.total = 0;
+    transaction_event.id_tag = provided_token_1;
     transaction_event.timestamp = Everest::Date::to_rfc3339(date::utc_clock::now());
     session_event_3.transaction_started = transaction_event;
     std::thread t4([this, session_event_3]() { this->auth_handler->handle_session_event(1, session_event_3); });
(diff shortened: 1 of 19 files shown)


corneliusclaussen:
Which libocpp PR needs to be merged for this?

modules/Auth/tests/auth_tests.cpp
```diff
@@ -385,16 +385,16 @@ TEST_F(AuthTest, test_transaction_finish) {
...
 
     SessionEvent session_event3;
     session_event3.event = SessionEventEnum::TransactionStarted;
     TransactionStarted transaction_event2;
-    transaction_event2.energy_Wh_import = 0;
-    transaction_event2.id_tag = provided_token_2.id_token;
+    transaction_event2.meter_value.energy_Wh_import.total = 0;
+    transaction_event2.id_tag = provided_token_2;
     transaction_event2.timestamp = Everest::Date::to_rfc3339(date::utc_clock::now());
     session_event3.transaction_started = transaction_event2;
```
a-w50, line 399: Just a comment on this style, it is more readable if you would create a small helper function for this like
```static auto init_empty_transaction_event(const ProvidedIdToken& token) { // name of the function up to discussion
    TransactionStarted te;
    te.meter_value.energy_Wh_import.total = 0;
    te.id_tag = token; // this line doesn't make sense due to naming, how can you assign a token to an id_tag?
    te.timestamp = Everest::Date::to_rfc3339(data::utc_clock::now()); // there should be a helper function in Everest::Date for the 'now" case
    return te;
}
```

Now you can simply write:
```
session_event1.transaction_started = init_empty_transaction_event(provided_token_1);
session_event3.transaction_started = init_empty_transaction_event(provided_token_2);
```
and reduce LOCs

modules/EvseManager/Charger.hpp
```diff
@@ -31,14 +31,16 @@
...
 #include <generated/types/evse_manager.hpp>
 #include <mutex>
 #include <queue>
 #include <sigslot/signal.hpp>
 
 namespace module {
 
-enum class ControlPilotEvent {
+enum class ControlPilotEvent
+{
```
a-w50, line 43: Is this correctly formatted?

modules/EvseManager/Charger.hpp
```diff
@@ -80,14 +82,16 @@ class Charger {
...
     sigslot::signal<float> signalMaxCurrent;
 
-    enum class ChargeMode {
+    enum class ChargeMode
+    {
         AC,
         DC
     };
 
     void setup(bool three_phases, bool has_ventilation, const std::string& country_code, bool rcd_enabled,
```
a-w50, line 91: this function has way too many parameters and gets like 5 bools in row which you can easily scramble up by accident, same for the two floats.  Should be refactored!

modules/EvseV2G/iso_server.cpp
```diff
@@ -174,18 +174,19 @@ static void convertIso1ToXmldsigSignedInfoType(struct xmldsigSignedInfoType* xml
     xmld_sig_signed_info->SignatureMethod.ANY_isUsed = 0;
 }
 
-static bool load_contract_root_cert(mbedtls_x509_crt *contract_root_crt, const char* V2G_file_path, const char* MO_file_path) {
+static bool load_contract_root_cert(mbedtls_x509_crt* contract_root_crt, const char* V2G_file_path,
+                                    const char* MO_file_path) {
     int rv = 0;
 
     if (((rv = mbedtls_x509_crt_parse_file(contract_root_crt, MO_file_path)) != 0) &&
         ((rv = mbedtls_x509_crt_parse_file(contract_root_crt, V2G_file_path)) != 0)) {
```
a-w50, line 182: This isn't part of this PR, but still could be flagged, because this checks if two loads fail at the same time, and just then logs an error. If the first load fails and the second succeeds, it returns true.  Don't think this is correct

modules/EvseV2G/iso_server.cpp
```diff
@@ -197,24 +198,22 @@ static bool load_contract_root_cert(mbedtls_x509_crt *contract_root_crt, const c
...
-        dlog(DLOG_LEVEL_INFO, "%s\n", buf );
+    if ((*flags) == 0)
+        dlog(DLOG_LEVEL_INFO, "  This certificate has no flags\n");
+    else {
+        mbedtls_x509_crt_verify_info(buf, sizeof(buf), "  ! ", *flags);
+        dlog(DLOG_LEVEL_INFO, "%s\n", buf);
     }
 
-    return( 0 );
+    return (0);
```
a-w50, line 216: why `(0)`

modules/EvseV2G/iso_server.cpp
```diff
@@ -197,24 +198,22 @@ static bool load_contract_root_cert(mbedtls_x509_crt *contract_root_crt, const c
  * \param flags
  * \return
  */
-static int debug_verify_cert( void *data, mbedtls_x509_crt *crt, int depth, uint32_t *flags )
-{
+static int debug_verify_cert(void* data, mbedtls_x509_crt* crt, int depth, uint32_t* flags) {
     char buf[1024];
-    ((void) data);
+    ((void)data);
```
a-w50, line 203: I guess this is because of compiler warnings, should add a comment to be more verbose, like `// unused variables`

modules/OCPP201/OCPP201.cpp
```diff
@@ -12,11 +12,57 @@ const std::string CERTS_DIR = "certs";
...
+    case ocpp::v201::ReasonEnum::TimeLimitReached:
+        return types::evse_manager::StopTransactionReason::TimeLimitReached;
+    case ocpp::v201::ReasonEnum::Timeout:
+        return types::evse_manager::StopTransactionReason::Timeout;
+    default:
+        return types::evse_manager::StopTransactionReason::Other;
+    }
+}
+
+ocpp::v201::SampledValue get_sampled_value(const ocpp::v201::ReadingContextEnum& reading_context,
```
a-w50, line 60: I can not imagine what a `ReadingContextEnum` is about, and it is usually good practice not to include a types type in its type name, like the `Enum` here. Types should be self-explaining if possible

modules/OCPP201/OCPP201.cpp
```diff
@@ -319,34 +398,35 @@ void OCPP201::init() {
...
-                sampled_value.value = meter_start;
-                meter_value.sampledValue.push_back(sampled_value);
-                meter_value.timestamp = timestamp;
+                const auto remote_start_id = transaction_started.id_tag.request_id;
 
                 ocpp::v201::IdToken id_token;
-                id_token.idToken = transaction_started.id_tag;
-                id_token.type = ocpp::v201::IdTokenEnum::ISO14443;
+                id_token.idToken = transaction_started.id_tag.id_token;
+                if (transaction_started.id_tag.id_token_type.has_value()) {
```
a-w50, line 410: I think, the naming here with `id_tag` and `id_token_type` is still broken. But this might be due to predifined OCCP types? At least for me, `transaction_started.id_tag.id_token_type` really reads unnaturally

modules/EvseManager/Charger.hpp
corneliusclaussen: This should be refactored, was not touched in this PR

modules/EvseV2G/iso_server.cpp
corneliusclaussen: Fix moved to seperate PR

modules/OCPP201/OCPP201.cpp
Pietfried: Agree. This is an auto generated ocpp::v201 type.

Pietfried:
> Which libocpp PR needs to be merged for this?

Changes for this in libocpp have already been merged

modules/EvseV2G/iso_server.cpp
SebaLukas: The if is correct. The contract leaf certificate can be based on the MO or on the V2G root certificate. Only if both root certificates can not be found, the leaf and sub ca certificates can not be verified.

corneliusclaussen (approved):

PR #291, opened 2023-07-18
Fix incorrect error handling when cert file not found

corneliusclaussen:


modules/EvseV2G/iso_server.cpp
```diff
@@ -177,7 +177,7 @@ static void convertIso1ToXmldsigSignedInfoType(struct xmldsigSignedInfoType* xml
 static bool load_contract_root_cert(mbedtls_x509_crt *contract_root_crt, const char* V2G_file_path, const char* MO_file_path) {
     int rv = 0;
 
-    if (((rv = mbedtls_x509_crt_parse_file(contract_root_crt, MO_file_path)) != 0) &&
+    if (((rv = mbedtls_x509_crt_parse_file(contract_root_crt, MO_file_path)) != 0) ||
```
mooraby, line 180: Can you explain more what caused the error? Some context maybe
The thing is, we first look internally for a MO root crt, if it is not found we look for a V2G root crt.. Only if we do not find both, we throw an error.

a-w50: Nothing caused an error, it was just that I commented on this code in a different PR.  The main problem with this `if`, is, that it is hard to reason about.
For example, is it intended, that there will be no error, if an MO crt is not found, but an V2G crt is found?
Furthermore, if an MO crt is found, then, the V2G crt won't be loaded - due to the implicit break in evaluating the `if` clause.
Finally, the error string, generated by `mbedtls_strerror` will only print the error produced by loading the V2G crt.
All of this is really implicit behaviour which is not clearly visible from the code.
This PR doesn't address these issues.  I guess @corneliusclaussen  interpretation of the correct behavior was, that the function should fail, if any loading fails.

corneliusclaussen: Looks like I misunderstood it and the current implementation of the if statement is correct. The error it prints however says:

"Unable to parse V2G (%s) or MO (%s) root certificate"

So I tried adapting the code to the printed error message (it says "or", so one of them failed). If it is an "and" we should fix the printed message to "and" as well to avoid confusion

mooraby: Okay I understand your comments.. If splitting the behavior in multiple conditions would make the could better readable for you, I can do this.. I will open another pull-request for this

corneliusclaussen: Ok so I close this one

PR #292, opened 2023-07-18
Forward dlink_ready signals from SLAC to ISO module

corneliusclaussen:
This forwards dlink_ready signals from SLAC to the ISO module. The ISO module should do something with this information, this is not implemented yet. @mooraby @FaHaGit Can you check if this is useful for the state machine? E.g. close TCP and reset module when a dlink_ready(false) is received?

### DIFF
--- a/interfaces/ISO15118_charger.yaml
+++ b/interfaces/ISO15118_charger.yaml
@@ -433,6 +433,15 @@ vars:
       Response will be reported async via  set_Get_Certificate_Response
     type: object
     $ref: /iso15118_charger#/Request_Exi_Stream_Schema
+  dlink_terminate:
+    description: Terminate the data link and become UNMATCHED.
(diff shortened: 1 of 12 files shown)


corneliusclaussen:
I will add dlink_terminate and friends as well in this branch

FaHaGit:
Yes, I think it is useful for the module. Now we can improve the behavior in case of an unexpected disconnect of the powerline connection. I have not tested it, but the changes looks ok for me.

corneliusclaussen:
The dlink_ready(false) will also always be sent at unplug so this may also help with all issues around TCP connections being still open waiting for timeout when the car unplugs etc

corneliusclaussen:
This PW will get bigger as initially planned, trying to add basic pause support for HLC as well. Will re-open once it is closer to being finished

PR #298, opened 2023-07-25
EvseV2G: Make the condition for loading root certificates more readable

mooraby:


modules/EvseV2G/iso_server.cpp
```diff
@@ -178,12 +178,18 @@ static bool load_contract_root_cert(mbedtls_x509_crt* contract_root_crt, const c
...
     int rv = 0;
 
-    if (((rv = mbedtls_x509_crt_parse_file(contract_root_crt, MO_file_path)) != 0) &&
-        ((rv = mbedtls_x509_crt_parse_file(contract_root_crt, V2G_file_path)) != 0)) {
+    if ((rv = mbedtls_x509_crt_parse_file(contract_root_crt, MO_file_path)) != 0) {
         char strerr[256];
         mbedtls_strerror(rv, strerr, 256);
-        dlog(DLOG_LEVEL_ERROR, "Unable to parse V2G (%s) or MO (%s) root certificate  (err: -0x%04x - %s)",
-             V2G_file_path, MO_file_path, -rv, strerr);
+        dlog(DLOG_LEVEL_ERROR, "Unable to parse MO (%s) root certificate. (err: -0x%04x - %s)",
```
lategoodbye, line 184: Since this is not critical i suggest to make it a DLOG_LEVEL_WARNING

modules/EvseV2G/iso_server.cpp
```diff
@@ -178,12 +178,18 @@ static bool load_contract_root_cert(mbedtls_x509_crt* contract_root_crt, const c
...
+             MO_file_path, -rv, strerr);
+        dlog(DLOG_LEVEL_INFO, "Attempting to parse V2G root certificate..");
+
+        if ((rv = mbedtls_x509_crt_parse_file(contract_root_crt, V2G_file_path)) != 0) {
+            mbedtls_strerror(rv, strerr, 256);
```
lategoodbye, line 189: Personally i would prefer to use sizeof(strerr)

lategoodbye (changes requested):

modules/EvseV2G/iso_server.cpp
FaHaGit: Yes, thats true. Maybe we should merge this one with the next log messages to one INFO message.

FaHaGit (changes requested):

modules/EvseV2G/iso_server.cpp
mooraby: Changed it to warning

modules/EvseV2G/iso_server.cpp
mooraby: Done..

FaHaGit (approved):

lategoodbye (approved):

SebaLukas (approved):
👍

PR #300, opened 2023-07-28
Fix typos in yaml: minumum -> minimum

SirVer:


SirVer:
I am in the process of understanding EVerest better, I am doing that by investing exactly 20 hours in writing a Rust implementation that makes it easy to write nodes in Rust. More on that on the mailing list soon.

While doing so, I had a need to understand the structure and content of the YAML files better and hence buildt a strongly typed parser for them that errors out on wrong or unknown options - a linter if you will. This tool found these typos - which makes me think these properties are probably not used currently?

hikinggrass (approved):

PR #301, opened 2023-07-28
Remove minProperty from type message

SirVer:
I believe this to be a copy & pasto from everest-framework/schemas/manifest.yaml where this shows in the context of outlining a schema. I would not understand what this defines for a data type, hence I believe it should be removed.

### DIFF
--- a/types/yeti.yaml
+++ b/types/yeti.yaml
@@ -4,6 +4,15 @@ types:
     description: simulation data
     type: object
     additionalProperties: false
+    required:
+      - cp_voltage
+      - pp_resistor
+      - rcd_current
+      - diode_fail
+      - error_e
+      - currents
+      - voltages
+      - frequencies
     properties:
       cp_voltage:
         description: CP Voltage level [V]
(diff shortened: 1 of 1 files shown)


types/yeti.yaml
```diff
@@ -4,7 +4,6 @@ types:
     description: simulation data
     type: object
     additionalProperties: false
-    minProperties: 8
```
a-w50, line 7: Hi, thanks for bringing this up again.  The reason for 'minProperties' here is, that all of the below properties should be required.  In my point of view this implicit style is not ideal.  It would be better to add all of the below properties to the 'required' list.  But just removing the 'minProperties' would make it possible to leave out all of the below properties.

a-w50 (changes requested):

types/yeti.yaml
SirVer: I at least did not understand that this was the intention here. I updated the PR to list out the required properties explicitly.

a-w50 (approved):
:100:

PR #307, opened 2023-08-07
Close TCP connection after dlink_ready(false)

corneliusclaussen:
As far as I can see this should be sufficient to close the TCP connection gracefully. It will do so after read timeout (1 sec in normal TCP and 5 sec in TLS) and ensures that the connection is always closed when SLAC reports a dlink_ready(false) on unplug (or during pause)


SebaLukas (approved):
👍

PR #313, opened 2023-08-14
Add zero-padding removal option to SerialCommHub

LAD101work:


modules/SerialCommHub/tiny_modbus_rtu.cpp
```diff
@@ -64,117 +64,16 @@ static void append_checksum(uint8_t* msg, int msg_len) {
 static bool validate_checksum(const uint8_t* msg, int msg_len) {
     if (msg_len < 5)
         return false;
+
+    EVLOG_info << hexdump(msg, msg_len);
```
corneliusclaussen: Looks like this debug needs to be removed before merge?

modules/SerialCommHub/manifest.yaml
```diff
@@ -24,6 +24,10 @@ provides:
         description: On some hardware every message that is sent is read back, this setting filters the sent message in the reply.
         type: boolean
         default: false
+      skip_zero_padding:
```
corneliusclaussen, line 27: This should probably not be a config option for the module. The problem is that it applies to all devices on the bus if you set it here, it probably is only needed for one of them? In that case it may be better to implement this as an argument for the read command (with a default disabled value) so that multiple devices can share the same bus. Also we should make sure this is really needed and is not a result of e.g. some misconfiguration of the serial port settings (in which case it would be work around code for hiding a bug)

modules/SerialCommHub/tiny_modbus_rtu.cpp
```diff
@@ -64,117 +64,16 @@ static void append_checksum(uint8_t* msg, int msg_len) {
...
-        case 0x0A:
-            EVLOG_error << "Modbus exception: Gateway path unavailable";
-            break;
-        case 0x0B:
-            EVLOG_error << "Modbus exception: Gateway target device failed to respond";
-            break;
-        }
-        return result;
-    }
+    EVLOG_info << "crc_sum: " << std::setfill('0') << std::setw(4) << std::hex << crc_sum;
```
corneliusclaussen: left over debug?

modules/SerialCommHub/tiny_modbus_rtu.cpp
```diff
@@ -354,11 +260,159 @@ std::vector<uint16_t> TinyModbusRTU::txrx(uint8_t device_address, FunctionCode f
...
+    }
+
+    if (function != function_code_recvd) {
+        EVLOG_error << fmt::format("Function code mismatch: expected: {} received: {}", function, function_code_recvd);
+        return result;
+    }
+
+    if (!validate_checksum(&buf.at(start_offset), len)) {
+        if (this->skip_zero_padding) {  // in case of zero-padding, remove trailing zero(es) and check again
+            EVLOG_info << "first checksum validation failed and skip_zero_padding set";
```
corneliusclaussen: remove debug?

corneliusclaussen (changes requested):

modules/SerialCommHub/tiny_modbus_rtu.cpp
LAD101work: Please make sure you comment only on the latest version, not on intermediary commits! 
This has already been fixed by the next commit.

modules/SerialCommHub/tiny_modbus_rtu.cpp
LAD101work: Please make sure you comment only on the latest version, not on intermediary commits! 
This has already been fixed by the next commit.

modules/SerialCommHub/manifest.yaml
LAD101work: This is actually the reason for this PR: 
A lot of times, when debugging new hardware, there is no clean bus, in which case one can set this option to remove timing issues introduced by non-clean bus connections. In this case it affects ALL communication on the bus and since the SerialCommHub is the module responsible for controlling the bus, I think this would be the best place to be able to set this option. 
Since I have run into this problem more than once, I think this option might add some value for other developers working on integrating new devices as well.
It has a default of "false", thus need not be considered in any situations that do not expressly need this option, but for those which need it there is no good other way...

corneliusclaussen: Do we know where these extra bytes are coming from? Are we sure they are always zero?

LAD101work: Yes, they are always zero. They seem to come from bus noise/too wide timing windows/too large capacitance on the bus. I would like to have measured with a scope but Julian is with you at the camp right now...

corneliusclaussen:
won't fix

PR #324, opened 2023-08-22
service name string length was not initialized which causes EXI encod…

corneliusclaussen:
…e errors if it is not 0 by sheer luck


SebaLukas (approved):
👍

PR #325, opened 2023-08-22
fix handling of external limits via API module

corneliusclaussen:
This fixes the handling of energy limits set via API module


Pietfried (approved):

modules/EvseManager/energy_grid/energyImpl.cpp
```diff
@@ -190,6 +190,7 @@ void energyImpl::ready() {
                 }
 
                 publish_energy_flow_request(energy_flow_request);
+                // EVLOG_info << "Local limits " << mod->getLocalEnergyLimits();
```
Pietfried, line 193: can just be removed

corneliusclaussen: fixed

PR #326, opened 2023-08-22
EvseSlac: log NMK

corneliusclaussen:


lib/staging/slac/fsm/evse/src/misc.cpp
```diff
@@ -13,6 +13,16 @@ std::string format_mac_addr(const uint8_t* mac) {
     return string_buffer;
 }
 
+std::string format_nmk(const uint8_t* nmk) {
+    char string_buffer[slac::defs::NMK_LEN * 3 + 1];
+    for (int i = 0; i < slac::defs::NMK_LEN; i++) {
+        snprintf(string_buffer + (i * 3), sizeof(string_buffer) - (i * 3), "%02X:", nmk[i]);
+    }
+    // remove last colon
+    string_buffer[slac::defs::NMK_LEN * 3 - 1] = 0;
```
a-w50, line 22: you could just make the size `slac::defs::NMK_LEN * 3`, and the final colon will be the terminating zero.  I would prefer using some constant instead of the 3.

a-w50 (approved):

PR #329, opened 2023-08-23
Add option for low power DC charging: deal with cars that request int…

corneliusclaussen:
…eger ampere values

### DIFF
--- a/modules/EvseManager/manifest.yaml
+++ b/modules/EvseManager/manifest.yaml
@@ -169,6 +169,13 @@ config:
     description: Set current measurement noise. Added to limit as an offset to avoid false triggers.
     type: number
     default: 0.5
+  hack_fix_hlc_integer_current_requests:
+    description: >-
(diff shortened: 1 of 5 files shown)


modules/EvseManager/manifest.yaml
```diff
@@ -169,6 +169,13 @@ config:
     description: Set current measurement noise. Added to limit as an offset to avoid false triggers.
     type: number
     default: 0.5
+  hack_fix_hlc_integer_current_requests:
+    description: >-
+      Some cars request only integer ampere values during DC charging. For low power DC charging that 
+      means that they charge a few hundret watts slower then needed. If enabled, this will charge at full power if the difference
```
SebaLukas, line 175: hundret -> hundred

SebaLukas (approved):
Looks good 👍 there is only one small thing to change

modules/EvseManager/manifest.yaml
corneliusclaussen: fixed

PR #330, opened 2023-08-23
hack BPT: allow starting with discharging

corneliusclaussen:
This allows starting with discharging directly in hack bpt mode


SebaLukas (approved):
Looks good 👍

PR #333, opened 2023-08-25
always trigger DC power supply update when changing between charge an…

corneliusclaussen:
…d discharge


corneliusclaussen:
This fixes a bug when e.g. switching from -5000W to +5000W in hack_bpt mode

SebaLukas (approved):
👍

PR #336, opened 2023-09-01
Update libocpp dependency, increase version to 2023.8.0

hikinggrass:


hikinggrass:
libocpp still needs to be tagged at v0.8.7 before we can merge this and tag the 2023.8.0 release

PR #338, opened 2023-09-05
Added plug in timeout event within EvseManager and OCPP1.6

Pietfried:


Pietfried:
Should be merged after https://github.com/EVerest/libocpp/pull/162 with patched libocpp dependency

corneliusclaussen (approved):

PR #339, opened 2023-09-05
Refactor powermeter signed meter value interface

corneliusclaussen:


corneliusclaussen:
@Pietfried EvseManager does currently not forward the OCMF in the TransactionStop event. We should probably add that as well to make it work with OCPP

Pietfried (approved):

PR #340, opened 2023-09-05
`nodered-sil.sh`: run as `USER root` in Dockerfile to avoid permission issues

Dominik-K:
I've stumbled upon this while following the quick-start guide, [section "software-in-the-loop", running `nodered-sil.sh`](https://everest.github.io/nightly/general/03_quick_start_guide.html#software-in-a-loop). Other new colleagues experienced this, too. And also recently [Akhila on the mailing-list](https://lists.lfenergy.org/g/everest/topic/100734106).

### DIFF
--- a/cmake/assets/docker/Dockerfile
+++ b/cmake/assets/docker/Dockerfile
@@ -1,4 +1,13 @@
 FROM nodered/node-red:2.2.3
+
+# run as root because below commands (sometimes?) return:
+# 
+# npm ERR! Your cache folder contains root-owned files, due to a bug in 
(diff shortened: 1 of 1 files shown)


hikinggrass:
@Dominik-K can you rebase and fix the dco sign-off ?

Dominik-K:
Actually, it got stalled because @a-w50 discussed this with me internally that running as `root` isn't good practice, even that it runs in a container (see also the Codacy hint).

I'm closing it now as I didn't experience it again and so perhaps can't reproduce it anymore.

PR #341, opened 2023-09-08
Add connector_id to evse_manager.yaml interface

Pietfried:
added connector_id to evse_manager.yaml interface and adjusted the respective commands within the modules that require this interface

### DIFF
--- a/interfaces/evse_manager.yaml
+++ b/interfaces/evse_manager.yaml
@@ -9,6 +9,10 @@ cmds:
       type: integer
   enable:
     description: Enables the evse. EVSE is available for charging after this operation
+    arguments:
+      connector_id:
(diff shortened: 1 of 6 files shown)


corneliusclaussen (changes requested):
Interface changes are fine, but ev-cli mod update needs to be run on all modules that require or provide this interface. Probably needs code adaption in more modules.

Pietfried:
> Interface changes are fine, but ev-cli mod update needs to be run on all modules that require or provide this interface. Probably needs code adaption in more modules.

Adoptions have been made for OCPP and OCPP201, which are the only modules that use the commands that have been modified (enable, disable, force_unlock)

corneliusclaussen (approved):

PR #343, opened 2023-09-14
Add callback for pause charging and some stubs so libocpp does not crash when calling them

marcemmers:
This needs to be merged before https://github.com/EVerest/libocpp/pull/167 so we don't crash on startup of libocpp


Pietfried (approved):
Thanks Marc, I have also implemented full functionality for Log Uploads and Firmware Updates because we had everything ready

marcemmers:
Ah even better, thanks!

PR #344, opened 2023-09-15
First example of a Rust module implementation

SirVer:
First Example of Rust Modules

This is modeled after the C++ Example modules, but shows up some more rustisms.

Co-authored-by: Dima Dorezyuk <ddo@qwello.eu> (@dorezyuk)
Co-authored-by: Kai Hermann (@hikinggrass)
Signed-off-by: Holger Rapp <HolgerRapp@gmx.net>


cmake/everest-generate.cmake
```diff
@@ -494,7 +499,82 @@ function (ev_add_py_module MODULE_NAME)
...
+
+            find_program(
+                    RSYNC
+                    rsync
+                    REQUIRED
+            )
+
+            # Copy source of module ${MODULE_NAME} to build dir
+            execute_process(
+                COMMAND ${RSYNC} -avq --delete "${MODULE_PATH}/" "${MODULE_BUILD_PATH}"
```
hikinggrass, line 540: This is broken since it does not pickup changes that happened between configure and build time, needs to be fixed before merge

hikinggrass: The whole reason for rsyncing here is that I have to modify the Cargo.toml in the next step, but dno not want to write absolute paths of the user into the version controlled source directory...

cmake/everest-generate.cmake
```diff
@@ -13,8 +13,6 @@ string(ASCII 27 ESCAPE)
 set(FMT_RESET "${ESCAPE}[m")
 set(FMT_BOLD "${ESCAPE}[1m")
 
-message(STATUS "${COLOR_BOLD}Should be bold?${COLOR_RESET}")
-
```
a-w50, line 17: Oh no :smile:

cmake/everest-generate.cmake
```diff
@@ -494,7 +499,82 @@ function (ev_add_py_module MODULE_NAME)
...
+        if(NOT ${EVEREST_ENABLE_RS_SUPPORT})
+            message(STATUS "Excluding Rust module ${MODULE_NAME} because EVEREST_ENABLE_RS_SUPPORT=${EVEREST_ENABLE_RS_SUPPORT}")
+            return()
+        elseif("${MODULE_NAME}" IN_LIST EVEREST_EXCLUDE_MODULES)
+            message(STATUS "Excluding module ${MODULE_NAME}")
+            return()
+        elseif(EVEREST_INCLUDE_MODULES AND NOT ("${MODULE_NAME}" IN_LIST EVEREST_INCLUDE_MODULES))
+            message(STATUS "Excluding module ${MODULE_NAME}")
+            return()
+        else()
```
a-w50, line 529: I'm not really happy with this code duplication for all different `ev_add_*_module` function

cmake/everest-generate.cmake
```diff
@@ -494,7 +499,82 @@ function (ev_add_py_module MODULE_NAME)
...
+            find_program(
+                CARGO
+                cargo
+                REQUIRED
+            )
+
+            add_custom_target(
+                ${MODULE_NAME} ALL
+                COMMAND ${CMAKE_COMMAND} -E env EVEREST_RS_FRAMEWORK_SOURCE_LOCATION="${everest-framework_SOURCE_DIR}" EVEREST_RS_FRAMEWORK_BINARY_LOCATION="${everest-framework_BINARY_DIR}" ${CARGO} build
+                WORKING_DIRECTORY ${MODULE_BUILD_PATH}
```
a-w50, line 562: Here I would use the generator expressions, to directly link to the correct library

cmake/everest-generate.cmake
```diff
@@ -494,7 +499,82 @@ function (ev_add_py_module MODULE_NAME)
...
+            # Copy source of module ${MODULE_NAME} to build dir
+            execute_process(
+                COMMAND ${RSYNC} -avq --delete "${MODULE_PATH}/" "${MODULE_BUILD_PATH}"
+            )
+
+            find_program(
+                SED
+                sed
+                REQUIRED
+            )
```
a-w50, line 547: As discussed, but only for the cmake build flow:
I would create a `build_rust_modules` target, which accesses a list (this list gets appended to via the ev_add_rust_module function) of rust modules.  This list then gets serialized into a generated cargo workspace file, which then should be build.
General problems here are how to track changes in rust source code.  One possibility would be in a rs module folder to add the files manually to a dependency list, which then gets checked.  But this doesn't really scale ...
On the other side, we could always run cargo build and let cargo determine if something new needs to be compiled.
But there are still a lot of other issues, I guess.

SirVer: @hikinggrass I do not think this is the right approach here - the rsync is rather brutal and makes it hard to track changes. I think @a-w50 idea has merits. I also think I found a solution that works - it relies on the ability to specify dependencies in the Workspace Cargo.toml and have members use that. The idea is outlined here: https://github.com/SirVer/rust_cmake_build_outline

1. The idea is that the sources must all be in a cargo workspace. And they do not reference everstrs directly, instead they use the `everstrs = { workspace = true }`. The source workspace [Cargo.toml](https://github.com/SirVer/rust_cmake_build_outline/blob/main/source/Cargo.toml) references them. 
2. For build, we create a new workspace in the build dir, we symlink all the modules and the everstrs source dir into this and create a [Cargo.toml](https://github.com/SirVer/rust_cmake_build_outline/blob/main/build/Cargo.toml) that references the everstrs that we care about. 
3. We then use https://github.com/corrosion-rs/corrosion to build this workspace with cMake, this should give us perfect reproducability and dependency tracking.

What do you think? @a-w50 @hikinggrass

cmake/everest-generate.cmake
SirVer: I added a todo for now.

cmake/everest-generate.cmake
SirVer: @hikinggrass I found a better solution that constructs a new workspace Cargo.toml and symlinks all rust modules together into one directory. I think this is good, because it ensure all modules use the same version of all libraries.

cmake/everest-generate.cmake
SirVer: I do not understand what this means.

cmake/everest-generate.cmake
SirVer: I did implement this now. Seems to be working fine.

cmake/everest-generate.cmake
```diff
@@ -494,7 +507,97 @@ function (ev_add_py_module MODULE_NAME)
...
+            message(STATUS "Excluding module ${MODULE_NAME}")
+            return()
+        elseif(EVEREST_INCLUDE_MODULES AND NOT ("${MODULE_NAME}" IN_LIST EVEREST_INCLUDE_MODULES))
+            message(STATUS "Excluding module ${MODULE_NAME}")
+            return()
+        else()
+            message(STATUS "Setting up Rust module ${MODULE_NAME}")
+
+            # Symlink the source directory into the build directory
+            execute_process(COMMAND rm -rf "${MODULE_BUILD_PATH}")
```
hikinggrass, line 564: That feels incredibly dangerous

barsnick: How about
`file(REMOVE_RECURSE [<files>...])`
https://cmake.org/cmake/help/latest/command/file.html#remove-recurse
> Changed in version 3.15: Empty input paths are ignored with a warning.

barsnick: But then again, the actual deletion is the critical part, I guess.

SirVer: It is not even needed to be recursive, it is just a symlink. Ideally I want the equivalent of `rm {file} || true`.

barsnick: > It is not even needed to be recursive, it is just a symlink. Ideally I want the equivalent of `rm {file} || true`.

Them `rm -f {file}` would be correct. You don't need to recurse on deleting a symlink, and no danger of clobbering actual directories.

hikinggrass: then you could use
```cmake
execute_process(COMMAND rm ${MODULE_BUILD_PATH}")
```

or even

```cmake
execute_process(COMMAND unlink ${MODULE_BUILD_PATH}")
```

errors in execute_process do not stop cmake from continuing (at least by default), you would have to specify a RESULT_VARIABLE and check that if you would want that

SirVer: I went with `rm -f`, this has the correct semantics: delete a single file (symlink) if it is there or do nothing if nothing is there.

SirVer:
@hikinggrass I fixed the bug that the Rust modules did not properly initialize themselves, they were immediately dropped, hence no work was taking place. This should be working fine now, I think it is ready for merging if you are happy.

The sign-offs are messed up. Once we are all happy, I will rebase once and make sure this is all signed-off correctly by me, the co-authored are all mentioned in the PR description.

cmake/everest-generate.cmake
SirVer: resolving, since this is no longer valid.

cmake/everest-generate.cmake
```diff
@@ -97,7 +95,14 @@ function(ev_add_project)
     if (EXISTS "${MODULES_DIR}/CMakeLists.txt")
         # FIXME (aw): default handling of building all modules?
         if (EVC_MAIN_PROJECT OR NOT EVEREST_DONT_BUILD_ALL_MODULES)
+            # We need a list of all to-be-built rust modules to create our
+            # workspace Cargo.toml, but we cannot easily pass it through the
+            # ev_add_module calls without polluting all CMakeLists.txt files
+            # with it. We bypass this using a global.
+            set_property(GLOBAL PROPERTY RUST_MODULE_LIST "")
```
a-w50, line 102: Not sure about this one, but doesn't that reset the list all the time?

a-w50: Furthermore, unfortunately there is this global variable, but a bit cleaner might be to create a custom target with `add_custom_target` called something like `everest_core_build_rust_modules`, then, you can set a property on this target with `set_property( target ...)` (so we have at least some 'encapsulation')
furthermore, when running the `add_rust_module ...`, this function can add additional targets/dependencies to the `everest_core_build_rust_modules` if necessary, so we have proper dependency tracking

cmake/everest-generate.cmake
```diff
@@ -494,7 +507,102 @@ function (ev_add_py_module MODULE_NAME)
...
+        PROPERTY EVEREST_MODULES ${EVEREST_MODULES}
+    )
+endfunction()
+
+function (create_rust_cargo_toml RUST_MODULE_LIST)
+    if(NOT ${EVEREST_ENABLE_RS_SUPPORT})
+        return()
+    endif()
+
+    set(OUTPUT_FILE_PATH "${CMAKE_BINARY_DIR}/modules/Cargo.toml")
```
a-w50, line 526: I would probably choose not the the `modules` folder, rather the `generated/rust-whatever..`

cmake/everest-generate.cmake
```diff
@@ -494,7 +507,102 @@ function (ev_add_py_module MODULE_NAME)
...
+
+    # Write the static parts
+    file(WRITE "${OUTPUT_FILE_PATH}" "[workspace]\nresolver = \"2\"\nmembers = [\n")
+
+    # Append each symlink path
+    foreach(SYMLINK ${RUST_MODULE_LIST})
+        file(APPEND "${OUTPUT_FILE_PATH}" "    \"${SYMLINK}\",\n")
```
a-w50, line 533: cant we use just the full path here, instead of symlinking?

cmake/everest-generate.cmake
```diff
@@ -494,7 +507,102 @@ function (ev_add_py_module MODULE_NAME)
...
+    endforeach()
+
+    # Close the list
+    file(APPEND "${OUTPUT_FILE_PATH}" "]\n\n")
+
+    file(APPEND "${OUTPUT_FILE_PATH}" "[workspace.dependencies]\neverestrs = { path = \"")
+    file(APPEND "${OUTPUT_FILE_PATH}" "${everest-framework_SOURCE_DIR}/everestrs/everestrs")
```
a-w50, line 540: Problem with `${everest-framework_SOURCE_DIR}` is, that this only works with the integrated cpm build, not with yocto, it's also better here to use target properties

cmake/everest-generate.cmake
```diff
@@ -494,7 +507,102 @@ function (ev_add_py_module MODULE_NAME)
...
+            return()
+        elseif(EVEREST_INCLUDE_MODULES AND NOT ("${MODULE_NAME}" IN_LIST EVEREST_INCLUDE_MODULES))
+            message(STATUS "Excluding module ${MODULE_NAME}")
+            return()
+        else()
+            message(STATUS "Setting up Rust module ${MODULE_NAME}")
+
+            # Symlink the source directory into the build directory
+            execute_process(COMMAND rm -f "${MODULE_BUILD_PATH}")
+            execute_process(COMMAND ln -s "${MODULE_PATH}/" "${MODULE_BUILD_PATH}")
```
a-w50, line 565: don't use `execute_process`, first of all, this will always run during cmake configure stage, secondly it will not stop the cmake configure stage if it fails
we should use `add_custom_command` here, which we can tack onto the above mentioned target as a dependency

cmake/everest-generate.cmake
```diff
@@ -494,7 +507,102 @@ function (ev_add_py_module MODULE_NAME)
...
+            file(RELATIVE_PATH MODULE_PARENT_DIR ${EVEREST_MODULE_DIR} "${CMAKE_CURRENT_SOURCE_DIR}/${MODULE_NAME}")
+            get_property(RUST_MODULE_LIST_GLOBAL GLOBAL PROPERTY RUST_MODULE_LIST)
+            list(APPEND RUST_MODULE_LIST_GLOBAL "${MODULE_PARENT_DIR}")
+            set_property(GLOBAL PROPERTY RUST_MODULE_LIST ${RUST_MODULE_LIST_GLOBAL})
+
+            find_program(
+                CARGO
+                cargo
+                REQUIRED
+            )
```
a-w50, line 577: this should not run for every add module, furthermore, try to things (that can fail) as early as possible, meaning, move it up

cmake/everest-generate.cmake
```diff
@@ -494,7 +507,102 @@ function (ev_add_py_module MODULE_NAME)
...
+                set(BIN_PATH "${CMAKE_BINARY_DIR}/modules/target/debug/${MODULE_NAME}")
+            endif()
+
+            # Note: We are building in the directory in which the Workspace Cargo.toml resides, to make sure it is used.
+            add_custom_target(
+                ${MODULE_NAME} ALL
+                COMMAND ${CMAKE_COMMAND} -E env EVEREST_RS_FRAMEWORK_SOURCE_LOCATION="${everest-framework_SOURCE_DIR}" EVEREST_RS_FRAMEWORK_BINARY_LOCATION="${everest-framework_BINARY_DIR}" ${CARGO} build ${CARGO_RELEASE_FLAG} --bin ${MODULE_NAME}
+                WORKING_DIRECTORY ${CMAKE_BINARY_DIR}/modules
+                DEPENDS everest::framework
+            )
```
a-w50, line 593: exactly, add this target as a dependency to the global `everest_core_build_rust_modules` target, and then do the symlinks inside here using commands

cmake/everest-generate.cmake
```diff
@@ -494,7 +507,102 @@ function (ev_add_py_module MODULE_NAME)
...
+
+            install(FILES ${MODULE_PATH}/manifest.yaml
+                DESTINATION "${EVEREST_MODULE_INSTALL_PREFIX}/${MODULE_NAME}"
+            )
+
+            install(FILES ${BIN_PATH}
+                PERMISSIONS OWNER_READ OWNER_WRITE OWNER_EXECUTE GROUP_READ GROUP_EXECUTE WORLD_READ WORLD_EXECUTE
+                DESTINATION "${EVEREST_MODULE_INSTALL_PREFIX}/${MODULE_NAME}"
```
a-w50, line 601: I think if the created binary already has X flag set, then it will also be installed like this

modules/rust_examples/Cargo.toml
```diff
@@ -0,0 +1,11 @@
...
+resolver = "2"
+members = [
+   "RsExample",
+   "RsExampleUser",
+]
+
+[workspace.dependencies]
+# This is not used when building with cmake, but you can change this for local
+# development to point to a directory that works for you to build with cargo.
+everestrs = { path = "../everest-framework/everestrs/everestrs" }
```
a-w50, line 11: so this file would be managed by hand for the *cargo only* experience?

a-w50: would it be an option, if we supply a sample file here (like Cargo.toml.sample), which then can be copied and modified by the rust developer to Cargo.toml, and putting it into .gitignore, so it won't be uploaded?

a-w50 (changes requested):
I wrote down some comments, I'll try to fix them

cmake/everest-generate.cmake
hikinggrass: Sure, then we have to figure out why the created binary only has OWNER_READ and OWNER_WRITE set and no *_EXECUTE flags

cmake/everest-generate.cmake
a-w50: In cmake you have generator expressions, which can give you information about targets and other things.  Say for example you have the library `everest::framework` (which resolves to a target), then you can get the full path to it via `$<TARGET_FILE:everest::framework>` instead of using *guesses*

SirVer: I defer here, not sure how to implement that.

cmake/everest-generate.cmake
SirVer: I defer here, not sure how to implement that.

cmake/everest-generate.cmake
SirVer: The Cargo.toml of the workspace must be in the top level of all Cargo modules. Having it in modules also allows to defer the exact location of where the binary ends up and makes sure that all rust modules use a consistent set of dependencies. I advise leaving it at its current place.

cmake/everest-generate.cmake
SirVer: no, every member of a Cargo workspace must be in a subdirectory of a cargo workspace, hence symlinking is the only thing that works in the general case.

cmake/everest-generate.cmake
SirVer: I defer here, not sure how to implement that, especially since I have no idea how you build this stuff in Yocto. Would be interesting to see though, since we need to build our own Yocto layers too.

cmake/everest-generate.cmake
SirVer: I defer here, not sure how to implement that.

cmake/everest-generate.cmake
SirVer: to where?

cmake/everest-generate.cmake
SirVer: On my system it is executable, but I have not tested what happens if the flags are removed.

hikinggrass (approved):

SirVer:
@hikinggrass What is this still waiting on?

hikinggrass:
> @hikinggrass What is this still waiting on?

Had to track down a bug in the Cargo.toml generation for the rust_workspace, that's fixed now and I'm on my way to squashing it :)

SirVer:
❤️ ❤️ 🎆 H O O R A Y  🚀 ❤️ ❤️

PR #352, opened 2023-09-21
Updating README.md packages list

SebaLukas:


krealyt (approved):
Nice.

Dominik-K:
@krealyt These installation instructions should be in sync with the everest.github.io . How about storing this part in an extra file and including it in both files? To have one single source of truth.

krealyt:
> @krealyt These installation instructions should be in sync with the everest.github.io . How about storing this part in an extra file and including it in both files? To have one single source of truth.

@Dominik-K Yes! :) This is one topic that we have already discussed and I want to do that on several locations where documentation is currently both on GitHub.com and on GitHub.io.

Should you want to start with this one, please drop me a line first, to talk about a "change strategy".

PR #356, opened 2023-09-26
don't build tests in `./lib`  by default

Dominik-K:
fixes EV-284


Dominik-K:
Perhaps `BUILD_DEV_TESTS` is more appropriate:

https://github.com/EVerest/everest-core/blob/lib_tests_cmake_flag/CMakeLists.txt#L34-L38

a-w50 (approved):

a-w50 (approved):

PR #361, opened 2023-09-27
API: prevent iterator invalidation

dorezyuk:
When defining more than 1 container the API node crahes.

The cause it that as the vector grows its iterators may get invalidated. The std::list offers stable iterators. 
The overhead between the vector and list in this case should not affect us since we're never iterating over `info` or `hw_capabilities_json`.

### DIFF
--- a/modules/API/API.hpp
+++ b/modules/API/API.hpp
@@ -18,11 +18,13 @@
 
 // ev@4bf81b14-a215-475c-a1d3-0a484ae48918:v1
 // insert your custom include headers here
-#include <date/date.h>
-#include <date/tz.h>
+#include <list>
 #include <memory>
 #include <mutex>
 #include <sstream>
+
+#include <date/date.h>
+#include <date/tz.h>
 
 namespace module {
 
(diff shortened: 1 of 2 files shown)


corneliusclaussen:
Could you sign off the commit (see DCO guidelines from the Linux foundation)?

modules/API/API.hpp
```diff
@@ -23,6 +23,7 @@
 #include <memory>
 #include <mutex>
 #include <sstream>
+#include <list>
```
a-w50, line 26: to make clang-format happy, the includes need to be sorted alphabetically, so `<list>` needs to be above `<memory>`

modules/API/API.hpp
```diff
@@ -104,8 +105,10 @@ class API : public Everest::ModuleBase {
     // insert your private definitions here
     std::vector<std::thread> api_threads;
     bool running = true;
-    std::vector<std::unique_ptr<SessionInfo>> info;
-    std::vector<json> hw_capabilities_json;
+    // We don't care here about locality since we're never iterating but must
+    // ensure that our iterators don't get invalidated.
```
a-w50, line 109: I wouldn't leave this command here, this should be evident for this piece of code (it could also be moved into the commit message)

a-w50 (changes requested):
Hi @dorezyuk 

thanks for this find, quite some "tricky" bug ...
Please fix the clang format

a-w50 (approved):

PR #284, opened 2023-07-13
Feature/error handling

andistorm:

PR #379, opened 2023-10-09
Refactor/bsp interface

corneliusclaussen:


lategoodbye:
Hi @corneliusclaussen
thanks for sharing this pull request. Could you please elaborate the motivation of this pull request? Is it to move as much as possible IEC 61851 specific stuff out of the BSP?

corneliusclaussen:
This is a refactoring that was planned for quite some time. As it breaks the BSP interfaces in an incompatible way, all BSP drivers will need to be changed. It should be relatively easy though.
Main motivations:
1) board_support_AC was used for DC as well for a while now but some things were missing for proper DC support. Also naming was wrong

2) To synchronize with the charging state machine in evsemanager abstract events were used (such as CarPluggedIn). This proved to be complex to use as the hardware typically knows states ABCDEF and the translation was non-trivial. To simplify BSP development, remove complexity from BSP drivers and avoid code duplication this code was moved into EvseManager. We decided to still use those abstract events internally in EvseManager as they hide some nasty details of IEC61851-1 (such as simplified charging mode). Now the BSP needs to report only the states ABCDEF. Most BSP drivers now only contain protocol translation code to the MCU and no logic anymore.

3) Factor out AC RCD. The RCD was in the board_support_AC interface before, in order to make the design a little cleaner it was moved to a separate interface (just like the IMD for DC charging)

In addition a lot of small improvements were made.

As this breaks code for many users we will also announce this on the mailing list as well as in todays community dev meeting

corneliusclaussen:
Another note: EvseManager itself will need internal refactoring as well at some point as it got quite complex, but for now this is more about improving the interfaces to the outside so that we hopefully do not need to break BSP compatibility soon again.

corneliusclaussen:
Please especially review the following:

interfaces/evse_board_support.yaml
types/board_support_common.yaml
types/evse_board_support.yaml

mooraby (commented):
General notes: 
- Not sure what guidelines do you follow, but what about using "this" to improve readability?
- Code documentation

types/board_support_common.yaml
```diff
@@ -0,0 +1,64 @@
...
+      ampacity:
+        description: Ampacity value of the cable assembly
+        type: string
+        enum:
+          - None
+          - A_13
+          - A_20
+          - A_32
+          - A_63
+  CPMeasurement:
```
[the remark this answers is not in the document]
corneliusclaussen: It is used by the EV implementation, probably I can move it into the upcoming ev_board_support instead

modules/EvseManager/BspStateMachine.cpp
```diff
@@ -0,0 +1,362 @@
...
+    while (true) {
+
+        if (main_thread_handle.shouldExit()) {
+            break;
+        }
+
+        std::this_thread::sleep_for(STATEMACHINE_UPDATE_RATE);
+
+        {
+            std::lock_guard l(state_mutex);
```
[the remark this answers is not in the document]
corneliusclaussen: fixed

types/board_support_common.yaml
```diff
@@ -0,0 +1,64 @@
...
+      - event
+    properties:
+      event:
+        type: string
+        enum:
+          - A
+          - B
+          - C
+          - D
+          - 'E'
```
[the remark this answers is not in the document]
corneliusclaussen: yaml will otherwise treat E as a number (from 1E0 etc)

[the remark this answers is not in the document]
corneliusclaussen: I will put all of them in quotes so it looks less awkward

interfaces/evse_board_support.yaml
```diff
@@ -70,29 +72,41 @@ cmds:
...
       value:
         description: Time in ms for the duration of the replug sequence
         type: integer
-  read_pp_ampacity:
+  ac_read_pp_ampacity:
     description: >-
-      Read the current carrying capacity of the connected cable in ampere.
-      In case of a fixed cable the configured maximum current the hardware can handle will be reported.
+      Read the current carrying capacity of the connected cable in ampere for AC charging.
+      This has no meaning for DC or AC charging with a fixed attached cable.
```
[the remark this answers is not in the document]
mooraby: But it does not get evaluated as you always call the read_pp_amapcity not matter what the value is, right? I though that you should only call read_pp_ampacity if the EVSE does not have a fixed cable

types/evse_board_support.yaml
```diff
@@ -111,9 +84,27 @@ types:
...
+      - allow_power_on
+      - reason
+    properties:
+      allow_power_on:
+        description: Allow switching on (true) or force switching off (false)
+        type: boolean
+      reason:
+        description: Reason for switching on/off
+        type: string
+        enum:
```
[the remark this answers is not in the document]
corneliusclaussen: The primary reason for adding this extra information was actually the CableCheck and Precharge phase where it is sometimes needed to know (e.g. if the DC output contactors should first close via a resistor before applying full power later on). While adding this I thought adding more reasons might be useful, but I'm not sure it actually is. Maybe remove "EmergencyStop" here?

modules/EvseManager/BspStateMachine.cpp
```diff
@@ -0,0 +1,362 @@
...
+            state_machine();
+        }
+    }
+}
+
+void BspStateMachine::process_bsp_event(types::board_support_common::BspEvent bsp_event) {
```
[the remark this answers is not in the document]
corneliusclaussen: I should not create a problem with the order of events, the worst case that I see is that PWM on/off is delayed by 50msecs for no reason. We could probably run the state machine whenever changing pwm as well to avoid the 50msec delay

types/board_support_common.yaml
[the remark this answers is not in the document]
corneliusclaussen: Removed here

modules/EvseManager/BspStateMachine.hpp
```diff
@@ -0,0 +1,110 @@
...
+    void set_pwm_F();
+
+    void enable(bool en);
+
+    // Signal for internal events type
+    sigslot::signal<CPEvent> event;
+
+private:
+    const std::unique_ptr<evse_board_supportIntf>& r_bsp;
+    void state_machine();
```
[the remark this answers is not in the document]
corneliusclaussen: Good idea, renamed

modules/EvseManager/BspStateMachine.cpp
```diff
@@ -0,0 +1,362 @@
...
+            // correctly, but simplified mode does not support HLC
+            // anyway.
+            if (!last_pwm_running && ev_simplified_mode) {
+                event(CPEvent::CarRequestedPower);
+            }
+        }
+
+        break;
+
+    case RawCPState::D:
```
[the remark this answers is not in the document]
corneliusclaussen: This will be added before it goes out of draft mode

modules/EvseManager/BspStateMachine.cpp
```diff
@@ -0,0 +1,362 @@
...
+        break;
+
+    case RawCPState::E:
+        if (last_cp_state != RawCPState::E) {
+            event(CPEvent::ErrorE);
+        }
+        break;
+
+    case RawCPState::F:
+        // this can only be set by evse, so we do not need to act here
```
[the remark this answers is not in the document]
corneliusclaussen: timeout stop added. F is handled by Charger fully, but I added power off here as well because it does not hurt. Locking/unlocking should be handled by the BSP, ideally on the MCU where the lowest level BSP code is running (you still want to unlock even if linux crashes)

modules/EvseManager/BspStateMachine.cpp
```diff
@@ -0,0 +1,362 @@
...
+    case RawCPState::D:
+        // not supported
+        event(CPEvent::ErrorVentilationNotAvailable);
+        break;
+
+    case RawCPState::E:
+        if (last_cp_state != RawCPState::E) {
+            event(CPEvent::ErrorE);
+        }
+        break;
```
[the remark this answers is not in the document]
corneliusclaussen: Added

types/evse_board_support.yaml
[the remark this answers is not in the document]
corneliusclaussen: I changed this to PowerOff only and no differentiation of the power off reason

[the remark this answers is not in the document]
corneliusclaussen: I could not come up with an idea why the BSP actually needs to know the power off reason. Do you think it is useful in the BSP driver code?

modules/EvseManager/BspStateMachine.cpp
```diff
@@ -0,0 +1,362 @@
...
+            event(CPEvent::CarRequestedPower);
+            if (power_on_allowed) {
+                internal_allow_power_on(true);
+            }
+        }
+
+        if (!pwm_running && last_pwm_running) { // X2->C1
+                                                // Table A.6 Sequence 10.2: EV does not stop drawing power
+                                                // even if PWM stops. Stop within 6 seconds (E.g. Kona1!)
+            timeout_state_c1.start(std::chrono::milliseconds(6000));
```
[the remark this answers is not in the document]
corneliusclaussen: yes but this would mean to globally pulling in the chrono literals namespace just for this line

modules/EvseManager/BspStateMachine.cpp
```diff
@@ -0,0 +1,362 @@
...
+    case RawCPState::C:
+        // Table A.6: Sequence 1.2 Plug-in
+        if (last_cp_state == RawCPState::A) {
+            event(CPEvent::CarPluggedIn);
+            EVLOG_info << "Detected simplified mode.";
+            ev_simplified_mode = true;
+        }
+        if (last_cp_state == RawCPState::B) {
+            event(CPEvent::CarRequestedPower);
+            if (power_on_allowed) {
```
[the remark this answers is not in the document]
corneliusclaussen: No it is not, I merged both of them

modules/EvseManager/BspStateMachine.cpp
```diff
@@ -0,0 +1,362 @@
...
+    pwm_running = false;
+    r_bsp->call_pwm_off();
+}
+
+void BspStateMachine::set_pwm_F() {
+    pwm_running = false;
+    r_bsp->call_pwm_F();
+}
+
+void BspStateMachine::allow_power_on(bool value, types::evse_board_support::Reason reason) {
```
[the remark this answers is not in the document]
corneliusclaussen: I tried to make it more clear

modules/EvseManager/BspStateMachine.cpp
```diff
@@ -0,0 +1,362 @@
...
+    if (std::holds_alternative<RawCPState>(event_or_error)) {
+        std::lock_guard l(state_mutex);
+        cp_state = std::get<RawCPState>(event_or_error);
+        state_machine();
+    } else {
+        event(std::get<CPEvent>(event_or_error));
+    }
+}
+
+void BspStateMachine::state_machine() {
```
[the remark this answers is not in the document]
corneliusclaussen: Disabled state: added. DF: Currently this is treated as a generic error (similar to e.g. over current error or generic CP error where no known CP state exists). This will put the Charger state machine into fault handling states which then will decide what to do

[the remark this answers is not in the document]
corneliusclaussen: They will be forwarded to charger in process_bsp_event() line 166:

    } else {
        event(std::get<CPEvent>(event_or_error));
    }

types/evse_board_support.yaml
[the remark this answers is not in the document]
mooraby: My though was to use the reason as information for logging, but I think this can be done already in EvseManager and not in BSP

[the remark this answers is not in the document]
corneliusclaussen: If you do not have additional hardware precharge relais you can ignore that in the BSP, but if you have the BSP needs to know the timing on when to switch on the first relais via resistor and when to switch on the second one shorting the resistor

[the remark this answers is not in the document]
corneliusclaussen: Yes logging can be done in EvseManager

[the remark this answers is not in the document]
lategoodbye: > If you do not have additional hardware precharge relais you can ignore that in the BSP, but if you have the BSP needs to know the timing on when to switch on the first relais via resistor and when to switch on the second one shorting the resistor

Okay, so `reason` is functional and should be kept

types/board_support_common.yaml
```diff
@@ -0,0 +1,54 @@
...
+          - 'ErrorOverTemperature'
+          - 'ErrorBrownOut'
+          - 'ErrorCablePP'
+          - 'ErrorEnergyManagement'
+          - 'ErrorNeutralPEN'
+          - 'ErrorCpDriver'
+          - 'PermanentFault'
+          - 'EvseReplugStarted'
+          - 'EvseReplugFinished'
+          - 'Disconnected'
```
FaHaGit, line 37: What is the difference between 'Disconnected' and 'A'?

types/board_support_common.yaml
```diff
@@ -0,0 +1,54 @@
...
+          - 'ErrorOverCurrent'
+          - 'ErrorOverVoltage'
+          - 'ErrorUnderVoltage'
+          - 'ErrorMotorLock'
+          - 'ErrorOverTemperature'
+          - 'ErrorBrownOut'
+          - 'ErrorCablePP'
+          - 'ErrorEnergyManagement'
+          - 'ErrorNeutralPEN'
+          - 'ErrorCpDriver'
```
FaHaGit, line 33: Should 'ErrorCpDriver' be sent in case of an invalid CP state?

interfaces/evse_board_support.yaml
```diff
@@ -14,31 +15,31 @@ cmds:
...
+    description: >-
+      Get Hardware capability/limits. For AC these are the limits of the power path (e.g. relais etc). 
+      For DC, these are the limits for the AC input of the ACDC converter stack, i.e. the complete AC input.
+      Note that DC output limits are reported by the DC power supply itself.
     result:
       description: Hardware capability/limits
       type: object
-      $ref: /board_support#/HardwareCapabilities
+      $ref: /evse_board_support#/HardwareCapabilities
   enable:
```
FaHaGit, line 27: Is it planned to introduce a cmd to enable/disable the CP (CP state E). Some hardware driver will be able to configure this state (e.g. Charge Control C hardware).

mooraby:
I will need to do another review round, but at least I can confirm that the SIL is working now, also in multiple subsequent charging sessions.

types/board_support_common.yaml
FaHaGit: any updates?

lategoodbye:
Hi,
on our [website](https://chargebyte.com/assets/Downloads/charge_control_c_user_guide_rev7.pdf) there is a detailed list (chapter 3.3.4 ff) of all error codes from our old charging stack. If i ignore all persistence related errors there are still some errors which are not available (e.g. lock capacitor not charged) above.

How should we handle these? Add them on demand later?

types/board_support_common.yaml
corneliusclaussen: Disconnected is only used on the EV side. In principle the EV will measure 0V when it is not connected to a charging station but there is a logical difference between e.g. the car setting E to indicate a fault and the car detecting it is not connected at all. This may be changed later on when the EV branch is being merged, but for now we need it for the EV sim. I'll clarify in the documentation that this should be ignored for EVSE side.

types/board_support_common.yaml
corneliusclaussen: This has been replaced with an MREC code with the following definition: "MREC_14_PilotFault: The control pilot voltage is out of range."

interfaces/evse_board_support.yaml
corneliusclaussen: The BSP has an enable(true/false) command that normally switches to state F to indicate EVSE not available. Yeti also has the capability to switch CP to high impedance. There are no valid procedures defined in IEC for that afaik, but we use it e.g. during virtual replugging during testing (along with PP replugging). This is done by evse_replug command. The implementation of how the replugging is done is left to the BSP (could e.g. be 0V on CP and a short interruption on PP or similar). Currently this is used in the "AC with SoC" mode which is start with DC until SoC is received, replug, charge AC PWM. Is there any other situation where this can be useful?

corneliusclaussen:
About the BspEvent error codes: I'm trying to collect all useful error codes in this enum at the moment. I'll have a look at your list as well and will include them, I think it is a good moment as we are breaking things anyway right now. Of course also more can be added later on. 
Before this PR gets undrafted, we want to also port it over to the new error handling framework, see PR #284. So the error list collected in the BspEvent enum is just a temporary list to collect all error codes that we want to be able to report.

corneliusclaussen:
I looked through the ChargeControl C documentation and added the lock motor failure details. I renamed them a bit to align better with the wording used in MREC codes. The other error codes such as RCD and powermeter do not belong to this interface, they should go into the ac_rcd and powermeter interfaces. Those will become reportable errors as well soon with the new error handling framework.

modules/EvseManager/IECStateMachine.cpp
```diff
@@ -0,0 +1,457 @@
...
+        return "MREC_15_PowerLoss";
+    case CPEvent::MREC_17_EVSEContactorFault:
+        return "MREC_17_EVSEContactorFault";
+    case CPEvent::MREC_18_CableOverTempDerate:
+        return "MREC_18_CableOverTempDerate";
+    case CPEvent::MREC_19_CableOverTempStop:
+        return "MREC_19_CableOverTempStop";
+    case CPEvent::MREC_20_PartialInsertion:
+        return "MREC_20_PartialInsertion";
+        return "MREC_19_CableOverTempStop";
```
mooraby, line 151: 2 returns but codacy caugh it anyway

interfaces/evse_board_support.yaml
```diff
@@ -0,0 +1,118 @@
...
+  ac_nr_of_phases_available:
+    description: Instantaneous phase count available to car
+    type: integer
+    minimum: 1
+    maximum: 3
+  telemetry:
+    description: Other telemetry
+    type: object
+    $ref: /evse_board_support#/Telemetry
+  capabilities:
```
mooraby, line 115: Should this var be also used to notify the EvseManager that the BSP has captured a change in PP value?

types/evse_manager.yaml
```diff
@@ -236,24 +236,38 @@ types:
...
+      - MREC_14_PilotFault
+      - MREC_15_PowerLoss
+      - MREC_16_EVContactorFault
+      - MREC_17_EVSEContactorFault
+      - MREC_18_CableOverTempDerate
+      - MREC_19_CableOverTempStop
+      - MREC_20_PartialInsertion
+      - MREC_21_CapacitanceFault
+      - MREC_22_ResistanceFault
+      - MREC_23_ProximityFault
```
mooraby, line 267: Do we need to differentiate between:
- ProximityFault (out of range proximity voltage)
- No cable detected during charing e.g. NoCableDuringCharging

types/board_support_common.yaml
```diff
@@ -0,0 +1,121 @@
...
+    properties:
+      ampacity:
+        description: Ampacity value of the cable assembly
+        type: string
+        enum:
+          - None
+          - A_13
+          - A_20
+          - A_32
+          - A_63
```
mooraby, line 121: A_70 missing?

interfaces/evse_board_support.yaml
corneliusclaussen: reinit also allows F instead of E, and most hw implementations cannot switch to E so I think F is the better choice for reinit

modules/EvseManager/IECStateMachine.cpp
```diff
@@ -0,0 +1,457 @@
...
+
+    case RawCPState::E:
+        if (last_cp_state != RawCPState::E) {
+            events.push(CPEvent::ErrorE);
+            timeout_state_c1.stop();
+            call_allow_power_on_bsp(false);
+        }
+        break;
+
+    case RawCPState::F:
```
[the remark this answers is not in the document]
corneliusclaussen: I do not really understand why we cannot tell if the car is connected, we see that from state A?
Yes, in the current type definition the BSP does not send any events such as CarUnplugged/requested power etc anymore, that is one of the goals of this PR. They are only used internally in the EvseManager module, and this may go away any time as it does not touch any external interfaces.

modules/EvseManager/IECStateMachine.cpp
```diff
@@ -0,0 +1,457 @@
...
+void IECStateMachine::call_allow_power_on_bsp(bool value) {
+    if (not value) {
+        power_on_allowed = false;
+        power_on_reason = types::evse_board_support::Reason::PowerOff;
+    }
+    r_bsp->call_allow_power_on({value, power_on_reason});
+}
+
+// High level state machine requests reading PP ampacity value.
+// We forware this request to the BSP driver. The high level state machine will never call this if it is not used
```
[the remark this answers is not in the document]
corneliusclaussen: fixed

modules/EvseManager/IECStateMachine.cpp
corneliusclaussen: fixed

interfaces/evse_board_support.yaml
corneliusclaussen: yes probably. At the moment there is just a getter (ac_read_pp_ampacity) which we need to read it at the right moment, but there should also be a var that publishes changes during the charging session (actually IEC only requires this to be read once on start, but we should also monitor during charging). Maybe it would be clearer to make it a separate var

types/evse_manager.yaml
corneliusclaussen: I would emit the ProximityFault in both cases (Trying to add as few custom error codes as possible, this maximizes the use of harmonized error codes)

types/board_support_common.yaml
corneliusclaussen: Or rename to A_63_70? It is the same resistor value

corneliusclaussen: or A_63_3ph_70_1ph?

mooraby: Yes good idea.. I think I A_63_70 is enough

lategoodbye:
@corneliusclaussen I have a question regarding the detection of an EV. It's possible to implement this via different approaches (only CP, only PP, mixed CP/PP).

Is it up to the implementor of the board support package to decide if an EV is connected or is there some kind of EVerest policy how to detected this?

mhei:
@corneliusclaussen: Would it be an option for you to cut the plug lock stuff out into a dedicated interface? This would make the bsp interface even more generic and minimalistic. BSPs with plug lock capabilities could easily implement both interfaces in a single module (e.g. Yeti if I'm right) and boards with no hardware support at all could easily just skip it. On platform like our Tarragon, we have for example two hardware plug lock interfaces. Our plan would be to create a module which cares about a single one, but customers could use another module instance for the second plug lock (e.g. for locking a fixed cable while a socket is in use...).

corneliusclaussen:
> @corneliusclaussen: Would it be an option for you to cut the plug lock stuff out into a dedicated interface? This would make the bsp interface even more generic and minimalistic. BSPs with plug lock capabilities could easily implement both interfaces in a single module (e.g. Yeti if I'm right) and boards with no hardware support at all could easily just skip it. On platform like our Tarragon, we have for example two hardware plug lock interfaces. Our plan would be to create a module which cares about a single one, but customers could use another module instance for the second plug lock (e.g. for locking a fixed cable while a socket is in use...).

In this case EvseManager would need to decide when to lock/unlock. In all HW that we have here this is actually done by an MCU (which ensures proper unlocking in case of Linux fails etc). It could be done though as the MCU can still do it on its own and not use the evsemanager supplied information. I factored out the RCD for the same reason of . I can make a draft, but probably after charin testival

corneliusclaussen:
> > @corneliusclaussen: Would it be an option for you to cut the plug lock stuff out into a dedicated interface? This would make the bsp interface even more generic and minimalistic. BSPs with plug lock capabilities could easily implement both interfaces in a single module (e.g. Yeti if I'm right) and boards with no hardware support at all could easily just skip it. On platform like our Tarragon, we have for example two hardware plug lock interfaces. Our plan would be to create a module which cares about a single one, but customers could use another module instance for the second plug lock (e.g. for locking a fixed cable while a socket is in use...).

@mhei : Another though on this: If we refactor out all motor lock things into a separate interface, would it make sense to also move all PP related things to this motor lock interface? If there is a socket it also requires a lock and only with a socket PP is useful?

If there is a fixed attached cable you don't need motor lock but you also don't need PP. What do you think?

types/board_support_common.yaml
corneliusclaussen: implemented

interfaces/evse_board_support.yaml
corneliusclaussen: implemented

interfaces/evse_board_support.yaml
```diff
@@ -0,0 +1,128 @@
...
+    type: object
+    $ref: /evse_board_support#/HardwareCapabilities
+  pp_ampacity:
+    description: >-
+      Current carrying capacity of the connected cable in ampere for AC charging
+      with a socket. Publish whenever it changes.
+      This has no meaning for DC or AC charging with a fixed attached cable, it does not 
+      need to be implemented and the returned value is not used in those cases.
+    type: object
+    $ref: /evse_board_support#/HardwareCapabilities
```
mooraby, line 128: /board_support_common#/ProximityPilot

corneliusclaussen: oops

corneliusclaussen: fixed

corneliusclaussen:
> > I agree that this is usually related, but as I mentioned, I'd like to (re-)use the lock motor interface for other purposes as well. Having then a stray PP interface in this interface would make things more complex on this side.
> 
> Agreed. Then we leave PP in the bsp interface and create a new one with just the lock/unlock commands and error reports for the lock

I pushed a draft for the interfaces/connector_lock.yaml, pls have a look. It is not used that, I'll implement in evsemanager once the interface is agreed

mhei:
@corneliusclaussen : motor lock interface LGTM

interfaces/ac_rcd.yaml
```diff
@@ -0,0 +1,27 @@
+description: >-
+  This interface defines an AC Residual Current Monitor (RCD). Actual emergency switch off is done in HW directly,
+  but this interface allows some control and telemetry.
```
dorezyuk, line 3: I think we should rather write *privides

interfaces/ac_rcd.yaml
```diff
@@ -0,0 +1,27 @@
...
+  rcd_current_mA:
+    description: Residual current in mA. Note that this does not trigger anything, it is merely for reporting.
+    type: number
+  fault_ac:
+    description: Indicates an AC residual current fault (informational only, hardware needs to shut down power on its own)
+    type: 'null'
+  fault_dc:
+    description: Indicates an DC residual current fault (on the AC wires, informational only, hardware needs to shut down power on its own).
+    type: 'null'
+  self_test_result:
```
dorezyuk, line 25: The `self_test_result` with a boolean is a little bit ambiguous - how about `self_test_success`?

interfaces/connector_lock.yaml
```diff
@@ -0,0 +1,13 @@
+description: >-
+  This interface defines one connector locking motor (e.g. for AC sockets with no fixed attached cable)
+cmds:
+  lock:
+    description: Lock connector lock
+  unlock:
+    description: unlock connector lock (e.g. normal unlock or enforced by OCPP)
+vars:
+  error:
+    description: Error events for locking motor
```
dorezyuk, line 10: what about unlocking - is this purposely left out?

interfaces/evse_board_support.yaml
```diff
@@ -0,0 +1,123 @@
...
+      on.
+    arguments:
+      value:
+        description: Flag and context
+        type: object
+        $ref: /evse_board_support#/PowerOnOff
+  ac_switch_three_phases_while_charging:
+    description: >-
+      Special command to force switching between one and three phases while
+      charging is active. HW must go through some special sequence to ensure safe
```
dorezyuk, line 59: Is it really *must* and IMHO its rather *a* special sequence or some special sequences

corneliusclaussen: IEC61851-1 does not support switching phases during charging. It is a non-standard feature that needs to be explicilty supported by the BSP driver and the BSP will need to implement a special sequence to do the switching. Many use state F for that but there are other implementations as well. Note that some cars will be destroyed permanently if you do not take special care here in the BSP.

corneliusclaussen: So I would not say that this is a must for any BSP to support, most BSP should not support this. For AC basic charging applications with solar where this is useful it can be implemented in the BSP and the BSP needs to then advertise it in its capabilities

corneliusclaussen: Looking into the future this should be entirely replaced with ISO-2 supporting 3ph/2A charging. But it will take some years until car manufacturers implement that if at all.

corneliusclaussen: Do you have a suggestion on how to rewrite the description to make it more clear?

interfaces/connector_lock.yaml
corneliusclaussen: No, the error events is a merge of the MRC codes and some additional ones found in the ChargeByte manual. Which ones would make sense to add here? The error handling will be refactored before this is merged once the new error handling is ready. But we are already collecting useful error types in this enum

interfaces/ac_rcd.yaml
corneliusclaussen: good point. If it is a bool it should be named _success. We could also make it return an enum, e.g. not_supported/failed/successful because probably it is more then a bool

interfaces/evse_board_support.yaml
dorezyuk: would it make sense to move this to a different interface then? So its clearer what is an optional interface and what not? This would also help to identify the interfaces which are rather for debugging and not for production

dorezyuk: This could also be true for `evse_replug`

interfaces/evse_board_support.yaml
```diff
@@ -0,0 +1,123 @@
...
+      need to be implemented and the returned value is not used in those cases.
+    result:
+      description: Returns the current carrying capacity of the connected cable
+      type: object
+      $ref: /board_support_common#/ProximityPilot
+  ac_set_overcurrent_limit_A:
+    description: >-
+      Sets the limit that should be used for over current detection in the BSP.
+      Do not use this to set PWM, use this value only for fast OC detection in the MCU.
+      This will set the actual charging value, so for 16A charging you will get 16A here,
```
dorezyuk, line 92: Not sure if I understand the this sentence - can you explain what you mean?

interfaces/evse_board_support.yaml
```diff
@@ -0,0 +1,123 @@
...
+    maximum: 3
+  telemetry:
+    description: Other telemetry
+    type: object
+    $ref: /evse_board_support#/Telemetry
+  capabilities:
+    description: Hardware capabilities/limits. This can be published during runtime whenever it changes to e.g. lower the limits when the hardware overheats.
+    type: object
+    $ref: /evse_board_support#/HardwareCapabilities
+  pp_ampacity:
```
dorezyuk, line 116: What is the difference to `ac_read_pp_ampacity` ? Why does this not have ac in the name?

types/board_support_common.yaml
```diff
@@ -0,0 +1,103 @@
+description: EV and EVSE board support types
+types:
+  BspEvent:
+    description: >-
+      Event stream from ControlPilot signal/Relais.
```
dorezyuk, line 5: why is relais capitalized?

types/board_support_common.yaml
```diff
@@ -0,0 +1,103 @@
...
+      MREC_26_CutCable: The output cable has been severed from the EVSE.
+    type: object
+    additionalProperties: false
+    required:
+      - event
+    properties:
+      event:
+        type: string
+        enum:
+          - 'A'
```
dorezyuk, line 55: How is this meant to be used - What happens if we have an error and the cp-state changes? Will the cp-state overwrite the error? If yes - how would we communicate if the error is resolved?

types/connector_lock.yaml
```diff
@@ -0,0 +1,37 @@
+description: EV and EVSE board support types
+types:
+  ConnectorLockError:
+    description: >-
+      Errors for Connector Lock
+
+      These following error types give more details about connector lock errors.
```
dorezyuk, line 7: Probably either "These errors" or "The following errors"

types/connector_lock.yaml
```diff
@@ -0,0 +1,37 @@
...
+      You should consider setting a MREC_1_ConnectorLockFailure to indicated a generic lock failure with an MREC error code as well
+      as one of the more detailed but custom error codes to specify the reason:
+
+      ConnectorLockCapNotCharged: The capacitor for connector lock motor failed to charge within expected time
+      ConnectorLockUnexpectedOpen: The connector lock feedback does return open while it should be close
```
dorezyuk, line 12: just `returns` and "while it should be closed".

Applies also for the line below

types/connector_lock.yaml
```diff
@@ -0,0 +1,37 @@
...
+      ConnectorLockUnexpectedClose: The connector lock feedback does return closed while it should be open
+      ConnectorLockFailedLock: The connector lock failed to lock (feedback still returns open)
+      ConnectorLockFailedUnlock: The connector lock failed to unlock (feedback still returns closed)
```
dorezyuk, line 15: What is the difference to the `ConnectorLockUnexpectedOpen`?

modules/MicroMegaWattBSP/board_support/evse_board_supportImpl.cpp
```diff
@@ -0,0 +1,129 @@
...
+
+#include "evse_board_supportImpl.hpp"
+
+namespace module {
+namespace board_support {
+
+/*
+
+
+static types::board_support::Event cast_event_type(const Event& e) {
```
dorezyuk, line 12: We probably don't want to commit this

modules/EvseManager/IECStateMachine.hpp
```diff
@@ -0,0 +1,141 @@
...
+    void set_pwm_off();
+    void set_pwm_F();
+
+    void enable(bool en);
+
+    // Signal for internal events type
+    sigslot::signal<CPEvent> signal_event;
+
+private:
+    const std::unique_ptr<evse_board_supportIntf>& r_bsp;
```
dorezyuk, line 115: Who has the ownership of this variable? If we suppose that it should be shared ownership why don't we use a shared_ptr?

modules/EvseManager/IECStateMachine.cpp
```diff
@@ -0,0 +1,451 @@
...
+    case CPEvent::MREC_23_ProximityFault:
+        return "MREC_23_ProximityFault";
+    case CPEvent::MREC_24_ConnectorVoltageHigh:
+        return "MREC_24_ConnectorVoltageHigh";
+    case CPEvent::MREC_25_BrokenLatch:
+        return "MREC_25_BrokenLatch";
+    case CPEvent::MREC_26_CutCable:
+        return "MREC_26_CutCable";
+    }
+    throw std::out_of_range("No known string conversion for provided enum of type CPEvent");
```
dorezyuk, line 156: The compiler should be able to check if we have covered all options for an enum class

types/board_support_common.yaml
corneliusclaussen: the errors will be refactored to the new error handling once that is merged, this will also have an interface to raise and clear errors

types/evse_board_support.yaml
```diff
@@ -56,52 +56,25 @@ types:
...
       - supply_voltage_12V
       - supply_voltage_minus_12V
-      - rcd_current
       - relais_on
     properties:
-      temperature:
+      evse_temperature_C:
         description: Current temperature of the EVSE in degree celsius
         type: number
+      gun_temperature_C:
```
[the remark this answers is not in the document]
corneliusclaussen: The plug you plug into the car for tethered charging

modules/EvseManager/IECStateMachine.cpp
```diff
@@ -0,0 +1,451 @@
...
+    // Subscribe to bsp driver to receive BspEvents from the hardware
+    r_bsp->subscribe_event([this](const types::board_support_common::BspEvent event) {
+        // feed into state machine
+        process_bsp_event(event);
+    });
+}
+
+void IECStateMachine::process_bsp_event(types::board_support_common::BspEvent bsp_event) {
+
+    auto event_or_error = from_bsp_event(bsp_event.event);
```
dorezyuk, line 172: *const

modules/EvseManager/IECStateMachine.cpp
corneliusclaussen: Some compilers will give a warning if ont enum value is missing. Still something needs to happen if it is the case anyway

modules/EvseManager/IECStateMachine.cpp
```diff
@@ -0,0 +1,451 @@
...
+
+    // If it was a CP change: feed state machine, else forward error
+    if (std::holds_alternative<RawCPState>(event_or_error)) {
+        {
+            std::lock_guard l(state_mutex);
+            cp_state = std::get<RawCPState>(event_or_error);
+        }
+        feed_state_machine();
+    } else {
+        signal_event(std::get<CPEvent>(event_or_error));
```
dorezyuk, line 182: can we use visit here to make sure the compilation crashes if someone extends our variant? Otherwise we would just go into this else and crash

modules/EvseManager/IECStateMachine.cpp
```diff
@@ -0,0 +1,451 @@
...
+// Main IEC state machine. Needs to be called whenever:
+// - CP state changes (both events from hardware as well as duty cycle changes)
+// - Allow power on changes
+// - The C1 6s timer expires
+std::queue<CPEvent> IECStateMachine::state_machine() {
+
+    std::queue<CPEvent> events;
+    std::lock_guard lock(state_mutex);
+
+    switch (cp_state) {
```
dorezyuk, line 205: would it make sense to get the copy of the cp-state here and just to release the mutex?

modules/EvseManager/IECStateMachine.cpp
```diff
@@ -0,0 +1,451 @@
...
+
+// enable/disable the charging port and CP signal
+void IECStateMachine::enable(bool en) {
+    r_bsp->call_enable(en);
+}
+
+// Forward the over current detection limit to the BSP. Many BSP MCUs monitor the charge current and trigger a fault
+// in case of over current. This sets the target charging current value to be used in OC detection. It cannot be
+// derived from the PWM duty cycle, use this value instead.
+void IECStateMachine::set_overcurrent_limit(double amps) {
```
dorezyuk, line 444: is there a reason amps is not protected by the mutex?

modules/EvseManager/IECStateMachine.cpp
```diff
@@ -0,0 +1,451 @@
...
+
+        // Table A.6: Sequence 1.1 Plug-in
+        if (last_cp_state == RawCPState::A || last_cp_state == RawCPState::Disabled) {
+            events.push(CPEvent::CarPluggedIn);
+            ev_simplified_mode = false;
+        }
+
+        if (last_cp_state == RawCPState::E || last_cp_state == RawCPState::F) {
+            // Triggers SLAC start
+            events.push(CPEvent::EFtoBCD);
```
dorezyuk, line 256: so in this case our events will have `[CarRequestedToStopPower, EFtoBCD]` - is this on purpose?

modules/EvseManager/IECStateMachine.cpp
```diff
@@ -0,0 +1,451 @@
...
+        }
+        break;
+
+    case RawCPState::D:
```
dorezyuk, line 260: why is D before C?

dorezyuk: ah understood

dorezyuk: https://en.cppreference.com/w/cpp/language/attributes/fallthrough

modules/EvseManager/IECStateMachine.cpp
```diff
@@ -0,0 +1,451 @@
...
+        }
+        // no break, intended fall through: If we support state D it is handled the same way as state C
+    case RawCPState::C:
+        // Table A.6: Sequence 1.2 Plug-in
+        if (last_cp_state == RawCPState::A) {
+            events.push(CPEvent::CarPluggedIn);
+            EVLOG_info << "Detected simplified mode.";
+            ev_simplified_mode = true;
+        }
+        if (last_cp_state == RawCPState::B) {
```
dorezyuk, line 276: *else if

modules/EvseManager/IECStateMachine.cpp
```diff
@@ -0,0 +1,451 @@
...
+            events.push(CPEvent::CarRequestedPower);
+        }
+
+        if (!pwm_running && last_pwm_running) { // X2->C1
+                                                // Table A.6 Sequence 10.2: EV does not stop drawing power
+                                                // even if PWM stops. Stop within 6 seconds (E.g. Kona1!)
+            timeout_state_c1.start(std::chrono::seconds(6));
+        }
+
+        if (timeout_state_c1.reached()) {
```
dorezyuk, line 286: Will this only be triggered if we receive another event from the bsp? Can't we use the timer to actually call `call_allow_power_on_bsp` directly?

modules/MicroMegaWattBSP/board_support/evse_board_supportImpl.cpp
corneliusclaussen: the bsps are not refactored yet as the interface continues to change. They are only compiling right now. Once the interfaces are done we'll work through the BSP modules using them

types/connector_lock.yaml
corneliusclaussen: their are two different errors that can happen: If you intent to lock/unlock it it may not work. The second error is: if it was locked for quite a while and there is no intent to open it but it still does.

modules/EvseManager/IECStateMachine.hpp
dorezyuk: you can always construct a shared_ptr from a unqiue_ptr by moving it into it https://en.cppreference.com/w/cpp/memory/shared_ptr/shared_ptr
You need probably somewhere 
```cpp
std::shared_ptr<evse_board_supportInt> my_shared{std::move(my_unique)};
```

types/connector_lock.yaml
dorezyuk: Ok, understood. I'm not sure if this is a nice interface since the `ConnectedLockFailedLock`(and unlock) are results from an rpc call. Maybe it would be better to communicate them as part of the rpc call?

types/evse_board_support.yaml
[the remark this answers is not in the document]
dorezyuk: Can we name it plug?

types/board_support_common.yaml
```diff
@@ -0,0 +1,103 @@
...
+      MREC_17_EVSEContactorFault: Contactors fail to open or close on EVSE's side. May also include welding related errors.
+      MREC_18_CableOverTempDerate: Temperature of charging cable or connector assembly is too high, resulting in reduced power operation.
+      MREC_19_CableOverTempStop: Temperature of charging cable or connector assembly is too high, resulting in a stopped charging session.
+      MREC_20_PartialInsertion: Cable latch is raised due to incomplete insertion into the vehicle charging port.
+      MREC_21/22 does not apply here
+      MREC_23_ProximityFault: The proximity voltage is out of range.
+      MREC_24_ConnectorVoltageHigh: The output voltage of EVSE is high before charging starts or after charging ends.
+      MREC_25_BrokenLatch: The latch on the connector is broken.
+      MREC_26_CutCable: The output cable has been severed from the EVSE.
+    type: object
```
FaHaGit, line 47: We think that object type is not correct and it should be of type "string" here with the event enumeration. Currently it is a object only with one property.

types/board_support_common.yaml
```diff
@@ -0,0 +1,103 @@
...
+          - 'MREC_18_CableOverTempDerate'
+          - 'MREC_19_CableOverTempStop'
+          - 'MREC_20_PartialInsertion'
+          - 'MREC_23_ProximityFault'
+          - 'MREC_24_ConnectorVoltageHigh'
+          - 'MREC_25_BrokenLatch'
+          - 'MREC_26_CutCable'
+  ProximityPilot:
+    description: Current capability (ampacity) of the cable
+    type: object
```
FaHaGit, line 90: Same here. We think type object is not correct here. Only one property defined in the object. Or is it planed to extend this object in the future?

interfaces/evse_board_support.yaml
```diff
@@ -0,0 +1,123 @@
...
+    description: Instantaneous phase count available to car
+    type: integer
+    minimum: 1
+    maximum: 3
+  telemetry:
+    description: Other telemetry
+    type: object
+    $ref: /evse_board_support#/Telemetry
+  capabilities:
+    description: Hardware capabilities/limits. This can be published during runtime whenever it changes to e.g. lower the limits when the hardware overheats.
```
mhei, line 113: I'd clarify that re-publishing is not required. For example, when the initial query announced 3 phases via `get_hw_capabilities` but then the `setup` call instructs to use a single phase only (due to configuration), then it should not be required to re-publish the capabilities, right?

FaHaGit:
Hello together,
we have found one possible bug plus one SIGABRT fault while testing the new interface and EvseManager module on our hardware.
The scenario:

The EV is plugged in, the EV is authorized and 26.76% duty cycle is signaled.
The EV starts charging (CP state C) and pauses charging (CP state B) after a few seconds.
When the EV tries to charge again, 100% is signaled for a short time, which is unexpected because the vehicle is already authorized. Furthermore, charging is unexpectedly interrupted with the log message:

> [WARN] connector_1:Evs std::queue<module::CPEvent> module::IECStateMachine::state_machine() :: Timeout of 6 seconds reached, EV did not go back to state B after PWM was switched off. Power off under load.

If now the vehicle goes back from CP state B to A (EV unplugged), then EVerest terminates with an exit code 134 "Abnormal Termination (SIGABRT)".


The following log message can be seen:

```
Oct 27 06:01:05 tarragon manager[13634]: 2023-10-27 06:01:05.375578 [INFO] cb_tarragon_dri  :: CP state change from B to A
Oct 27 06:01:05 tarragon manager[13634]: 2023-10-27 06:01:05.445374 [INFO] cb_tarragon_dri  :: handle_pwm_off: Setting new duty cycle of 100.00%
Oct 27 06:01:05 tarragon manager[13634]: 2023-10-27 06:01:05.630798 [INFO] cb_tarragon_dri  :: handle_allow_power_on: {
Oct 27 06:01:05 tarragon manager[13634]:     "allow_power_on": false,
Oct 27 06:01:05 tarragon manager[13634]:     "reason": "PowerOff"
Oct 27 06:01:05 tarragon manager[13634]: }
Oct 27 06:01:05 tarragon manager[13635]: terminate called after throwing an instance of 'std::future_error'
Oct 27 06:01:05 tarragon manager[13635]:   what():  std::future_error: Promise already satisfied
Oct 27 06:01:06 tarragon manager[13606]: 2023-10-27 06:01:06.157019 [CRIT] manager         int boot(const boost::program_options::variables_map&) :: Module connector_1 (pid: 13635) exited with status: 134. Terminating all modules.
```

corneliusclaussen:
> Hello together,
> 
> we have found one possible bug plus one SIGABRT fault while testing the new interface and EvseManager module on our hardware.
> The scenario:
> 
> The EV is plugged in, the EV is authorized and 26.76% duty cycle is signaled.
> The EV starts charging (CP state C) and pauses charging (CP state B) after a few seconds.
> When the EV tries to charge again, 100% is signaled for a short time, which is unexpected because the vehicle is already authorized. Furthermore, charging is unexpectedly interrupted with the log message:
> 
> > [WARN] connector_1:Evs std::queue<module::CPEvent> module::IECStateMachine::state_machine() :: Timeout of 6 seconds reached, EV did not go back to state B after PWM was switched off. Power off under load.
> 
> If now the vehicle goes back from CP state B to A (EV unplugged), then EVerest terminates with an exit code 134 "Abnormal Termination (SIGABRT)".
> 
> 
> The following log message can be seen:
> 
> ```
> Oct 27 06:01:05 tarragon manager[13634]: 2023-10-27 06:01:05.375578 [INFO] cb_tarragon_dri  :: CP state change from B to A
> Oct 27 06:01:05 tarragon manager[13634]: 2023-10-27 06:01:05.445374 [INFO] cb_tarragon_dri  :: handle_pwm_off: Setting new duty cycle of 100.00%
> Oct 27 06:01:05 tarragon manager[13634]: 2023-10-27 06:01:05.630798 [INFO] cb_tarragon_dri  :: handle_allow_power_on: {
> Oct 27 06:01:05 tarragon manager[13634]:     "allow_power_on": false,
> Oct 27 06:01:05 tarragon manager[13634]:     "reason": "PowerOff"
> Oct 27 06:01:05 tarragon manager[13634]: }
> Oct 27 06:01:05 tarragon manager[13635]: terminate called after throwing an instance of 'std::future_error'
> Oct 27 06:01:05 tarragon manager[13635]:   what():  std::future_error: Promise already satisfied
> Oct 27 06:01:06 tarragon manager[13606]: 2023-10-27 06:01:06.157019 [CRIT] manager         int boot(const boost::program_options::variables_map&) :: Module connector_1 (pid: 13635) exited with status: 134. Terminating all modules.
> ```
> 

Hi, thanks for the report. I will try to reproduce on our hardware early next week as well. On the promise satisfied crash: can you check the got revision of everest-framework? There was a quite recent commit that fixes one promise handling bug that showed up in another scenario. The short 100÷ in between looks more like a bug in the state machine that should be easy to find. If you still see the crash at the latest main of everest-framework it would be really helpful to do a gdb backtrace (run everest without the driver module (-s)) and then start the driver in gdb in a separate terminal)

FaHaGit:
> > Hello together,
> > we have found one possible bug plus one SIGABRT fault while testing the new interface and EvseManager module on our hardware.
> > The scenario:
> > The EV is plugged in, the EV is authorized and 26.76% duty cycle is signaled.
> > The EV starts charging (CP state C) and pauses charging (CP state B) after a few seconds.
> > When the EV tries to charge again, 100% is signaled for a short time, which is unexpected because the vehicle is already authorized. Furthermore, charging is unexpectedly interrupted with the log message:
> > > [WARN] connector_1:Evs std::queuemodule::CPEvent module::IECStateMachine::state_machine() :: Timeout of 6 seconds reached, EV did not go back to state B after PWM was switched off. Power off under load.
> > 
> > 
> > If now the vehicle goes back from CP state B to A (EV unplugged), then EVerest terminates with an exit code 134 "Abnormal Termination (SIGABRT)".
> > The following log message can be seen:
> > ```
> > Oct 27 06:01:05 tarragon manager[13634]: 2023-10-27 06:01:05.375578 [INFO] cb_tarragon_dri  :: CP state change from B to A
> > Oct 27 06:01:05 tarragon manager[13634]: 2023-10-27 06:01:05.445374 [INFO] cb_tarragon_dri  :: handle_pwm_off: Setting new duty cycle of 100.00%
> > Oct 27 06:01:05 tarragon manager[13634]: 2023-10-27 06:01:05.630798 [INFO] cb_tarragon_dri  :: handle_allow_power_on: {
> > Oct 27 06:01:05 tarragon manager[13634]:     "allow_power_on": false,
> > Oct 27 06:01:05 tarragon manager[13634]:     "reason": "PowerOff"
> > Oct 27 06:01:05 tarragon manager[13634]: }
> > Oct 27 06:01:05 tarragon manager[13635]: terminate called after throwing an instance of 'std::future_error'
> > Oct 27 06:01:05 tarragon manager[13635]:   what():  std::future_error: Promise already satisfied
> > Oct 27 06:01:06 tarragon manager[13606]: 2023-10-27 06:01:06.157019 [CRIT] manager         int boot(const boost::program_options::variables_map&) :: Module connector_1 (pid: 13635) exited with status: 134. Terminating all modules.
> > ```
> 
> Hi, thanks for the report. I will try to reproduce on our hardware early next week as well. On the promise satisfied crash: can you check the git revision of everest-framework? There was a quite recent commit that fixes one promise handling bug that showed up in another scenario. The short 100÷ in between looks more like a bug in the state machine that should be easy to find. If you still see the crash at the latest main of everest-framework it would be really helpful to do a gdb backtrace (run everest without the driver module (-s)) and then start the driver in gdb in a separate terminal)
> 
> [EVerest/everest-framework@e784a3c](https://github.com/EVerest/everest-framework/commit/e784a3c6a4aa99b8d70482af86a8246f51298045)

Currently we are using "7fcf0ac1e2d5362b25e967d48e163f0fbb4382b9" (0.7.0) of everest-framework. We can repeat the test as soon as we have updated our build.

interfaces/evse_board_support.yaml
FaHaGit: How does the bsp module know if it needs to send this var? After ac_read_pp_ampacity is received?

FaHaGit:
Hello together,

We at chargebyte would like to give you a brief update on the current draft of the BSP interface. We have tested the new interface for CP/PP and contactor handling with our hardware platform for the AC path. Everything seems to work as expected with the EvseManager. Some comments are still open from the code review, but once these are answered we see no problems to create a pull request from it.

Also the reported bug with resuming of the charging session by the EV (CP B -> C -> B -> C transition) works now after we updated everest framework.

We know that it will probably take a lot of effort on your side to prepare everything that it can be merged into erverest-core, but we are looking forward to working with the new interface soon :-).

Are there already plans when the new interface will be available?

Thank you and best regards

FaHaGit:
One additional note to the CP B -> C -> B -> C transition scenario. There is still 100% duty cycle for a short time after the second switch from B to C. I think this not expected.

corneliusclaussen:
Will look into the comments. Apart from that I would like to move all error logic over to the new error handling in framework. After that I will undraft the PR

modules/EvseManager/IECStateMachine.cpp
```diff
@@ -0,0 +1,370 @@
...
+        // If state D is not supported switch off.
+        if (not has_ventilation) {
+            call_allow_power_on_bsp(false);
+            timeout_state_c1.stop();
+            break;
+        }
+        // no break, intended fall through: If we support state D it is handled the same way as state C
+    case RawCPState::C:
+        // Table A.6: Sequence 1.2 Plug-in
+        if (last_cp_state == RawCPState::A || (!car_plugged_in && last_cp_state == RawCPState::F)) {
```
mooraby, line 179: What if we go from E > C? We expected the events CarPluggedIn (and CarRequestedPower?) and that we resume charging

modules/EvseManager/IECStateMachine.cpp
```diff
@@ -0,0 +1,370 @@
...
+
+    case RawCPState::D:
+        // If state D is not supported switch off.
+        if (not has_ventilation) {
+            call_allow_power_on_bsp(false);
+            timeout_state_c1.stop();
+            break;
+        }
+        // no break, intended fall through: If we support state D it is handled the same way as state C
+    case RawCPState::C:
```
mooraby, line 177: During testing with a Metrel tester, we made a fast switch between A -> B -> C before the authorization is done. We got the following: 
Ignoring CarRequestedPower in WaitingForAuth state. Probably a BCB toggle that does not make sense here.

What we were expecting is: After authoziation has succeeded and the duty cycle was set to nominal, we would still be able to start charging, but we never got the allow_power_on(true).

modules/EvseManager/IECStateMachine.cpp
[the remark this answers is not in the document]
corneliusclaussen: Lock unlock is not factored out into a separate interface and evse_manager will call lock/unlock commands

interfaces/ac_rcd.yaml
corneliusclaussen: changed

interfaces/ac_rcd.yaml
corneliusclaussen: this has been removed in the error refactoring. Now a self test error needs to be raised if it was not successful.

interfaces/connector_lock.yaml
corneliusclaussen: this var has been removed in the refactoring

interfaces/evse_board_support.yaml
corneliusclaussen: Clarfied the description

interfaces/evse_board_support.yaml
corneliusclaussen: renamed the var to ac_pp_ampacity to have more consistent naming. Updated the description to clarify when the var should be published.

types/board_support_common.yaml
corneliusclaussen: fixed

types/board_support_common.yaml
corneliusclaussen: refactoring done, the events do not have error codes anymore

types/connector_lock.yaml
corneliusclaussen: fixed

types/connector_lock.yaml
```diff
@@ -0,0 +1,37 @@
+description: EV and EVSE board support types
+types:
+  ConnectorLockError:
+    description: >-
+      Errors for Connector Lock
+
+      These following error types give more details about connector lock errors.
+      You should consider setting a MREC_1_ConnectorLockFailure to indicated a generic lock failure with an MREC error code as well
```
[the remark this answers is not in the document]
corneliusclaussen: fixed

types/connector_lock.yaml
corneliusclaussen: fixed

types/connector_lock.yaml
corneliusclaussen: refactored with the new error framework, so this type is gone

modules/EvseManager/IECStateMachine.cpp
```diff
@@ -0,0 +1,451 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2020 - 2022 Pionix GmbH and Contributors to EVerest
+/*
+ * Charger.cpp
+ *
+ *  Created on: 08.03.2021
```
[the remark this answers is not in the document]
corneliusclaussen: fixed

modules/EvseManager/IECStateMachine.cpp
corneliusclaussen: fixed

modules/EvseManager/IECStateMachine.cpp
corneliusclaussen: rewritten with a visitor

modules/EvseManager/IECStateMachine.cpp
corneliusclaussen: The mutex protects more then just the cp_state variable. I renamed it to state_machine_mutex to make that point more clear.

modules/EvseManager/IECStateMachine.cpp
```diff
@@ -0,0 +1,451 @@
...
+    last_power_on_allowed = power_on_allowed;
+
+    return events;
+}
+
+// High level state machine sets PWM duty cycle
+void IECStateMachine::set_pwm(double value) {
+    {
+        std::scoped_lock lock(state_mutex);
+        if (value > 0 && value < 1) {
```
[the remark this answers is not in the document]
corneliusclaussen: EvseManager uses 0..1 internally everywhere for historical reasons. Could be refactored later on.

modules/EvseManager/IECStateMachine.cpp
corneliusclaussen: yes

modules/EvseManager/IECStateMachine.cpp
corneliusclaussen: added

modules/EvseManager/IECStateMachine.cpp
corneliusclaussen: We would still need to check if we are in state C before etc, I think it is more readable to have it in the state machine

types/board_support_common.yaml
corneliusclaussen: This was removed anyway in the refactoring

interfaces/evse_board_support.yaml
corneliusclaussen: Right it should not, that is the maximum capabilities that the hw supports. It may still change over time if e.g. the hardware gets too warm. I'll clarify the description.

modules/EvseManager/IECStateMachine.cpp
corneliusclaussen: This has been fixed in main recently

modules/EvseManager/IECStateMachine.cpp
corneliusclaussen: Hmm technically this is not a plugin on EVSE side, a plugin is always A->B (or C or D). In my understanding E can be used by the EV to signal a (temporary) error. I think at the moment that stops charging and only recovers when you unplug. I can change it so that it restarts charging on E->C if it was authorized before etc.

types/board_support_common.yaml
corneliusclaussen: This is because of a limitation in the type support at the moment. If it is defined directly as an enum it will be a string in the generated C++ code. Having an object wrapper around this creates a real c++ enum. This may be changed if it changes in the framework.

modules/EvseManager/IECStateMachine.cpp
corneliusclaussen: Looking at the current code it is ok as far as I can see, last_amps is an atomic double

types/evse_board_support.yaml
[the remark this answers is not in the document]
corneliusclaussen: done

interfaces/evse_board_support.yaml
corneliusclaussen: most BSPs we came accross have this functionality, so I would keep it in the interface. I'll rework the description a bit

modules/EvseManager/IECStateMachine.cpp
lategoodbye: I agree with "this is not a CarPluggedIn event". But Tarragon (EVSE side) is able to generate 0 V, which we need for phase switching. So our question is: can we make it charge again for in this situation and how should the BSP behave in this case?

corneliusclaussen: Yes I would change it such that the charging session remains the same (since it is not plugin) but will simply start charging again (similar to the EV pausing within a session). Does that make sense?

lategoodbye: Yes

modules/EvseManager/IECStateMachine.hpp
corneliusclaussen: yes, but it would need to be changed in the auto generated code. I cannot have the same pointer in both a unique_ptr (in the auto generated header) as well as a shared pointer (in the added code)

corneliusclaussen:
All comments should be addressed now. It was tested both in SIL as well as on real hardware.

modules/EvseManager/Charger.cpp
```diff
@@ -446,15 +461,19 @@ void Charger::runStateMachine() {
 
                 if (!powerAvailable()) {
                     pauseChargingWaitForPower();
+                    // FIXME new in state breaks
```
hikinggrass, line 464: is this still relevant?

modules/EvseManager/Charger.cpp
```diff
@@ -524,26 +543,22 @@ void Charger::runStateMachine() {
...
+                if (initialize_state) {
                     signalEvent(types::evse_manager::SessionEventEnum::ChargingPausedEV);
-                    r_bsp->call_allow_power_on(true);
-                    // make sure we are enabling PWM
-                    if (hlc_use_5percent_current_session) {
-                        update_pwm_now(PWM_5_PERCENT);
-                    } else {
-                        update_pwm_now(ampereToDutyCycle(getMaxCurrent()));
-                    }
+                    // bsp->allow_power_on(true, types::evse_board_support::Reason::FullPowerCharging);
```
hikinggrass, line 548: Is there a reason these calls are commented out?

modules/EvseManager/Charger.cpp
```diff
@@ -1509,6 +1401,9 @@ void Charger::bcb_toggle_detect_start_pulse() {
 
 // call this on C->B transitions
 void Charger::bcb_toggle_detect_stop_pulse() {
+    if (!hlc_bcb_sequence_started)
+        return;
+
```
hikinggrass, line 1406:
```suggestion
    if (!hlc_bcb_sequence_started) {
        return;
    }

```

modules/EvseManager/Charger.hpp
```diff
@@ -178,8 +173,8 @@ class Charger {
 
     EvseState getCurrentState();
     sigslot::signal<EvseState> signalState;
-    sigslot::signal<types::evse_manager::ErrorEnum> signalError;
-    // /Deprecated
+    // sigslot::signal<types::evse_manager::ErrorEnum> signalError;
```
hikinggrass, line 176: Could this be removed entirely now?

modules/EvseManager/Charger.cpp
corneliusclaussen: removed the commented out commands

modules/EvseManager/Charger.hpp
corneliusclaussen: removed

corneliusclaussen: The rest of the "deprecated" things (EvseState) is still used in other places in EvseManager

modules/MicroMegaWattBSP/board_support/evse_board_supportImpl.cpp
hikinggrass: since the interface is done now, will this be updated in this PR or at a later date?

modules/YetiDriver/board_support/evse_board_supportImpl.cpp
```diff
@@ -0,0 +1,183 @@
...
+}
+
+void evse_board_supportImpl::ready() {
+    // Wait for caps to be received at least once
+    int i;
+    for (i = 0; i < 50; i++) {
+        {
+            std::lock_guard<std::mutex> lock(capsMutex);
+            if (caps_received)
+                break;
```
hikinggrass, line 127:
```suggestion
            if (caps_received) {
                break;
            }
```

hikinggrass (approved):

PR #381, opened 2023-10-11
OCPP persisting states during Reset

Pietfried:


modules/OCPP201/OCPP201.cpp
```diff
@@ -387,6 +387,54 @@ void OCPP201::init_evse_ready_map() {
     }
 }
 
+void OCPP201::init_evses() {
+
+    if (this->r_kvs->call_exists("0")) {
```
hikinggrass, line 392: This should have some kind of prefix, since the kvs modules (at the moment) do not do any sort of automatic prefixing based on their user (which we should probably add there soon)

modules/OCPP201/OCPP201.cpp
```diff
@@ -480,38 +528,48 @@ void OCPP201::ready() {
...
-                this->operational_connector_states[request.evse.value().id] = request.operationalStatus;
+            auto evse_id = request.evse.value().id;
+            auto connector_id = request.evse.value().connectorId;
+            if (request.operationalStatus == ocpp::v201::OperationalStatusEnum::Operative) {
+                // change to operative
+                if (connector_id.has_value()) {
+                    // connector is addressed
+                    this->evses.at(evse_id).connectors.at(connector_id.value()) =
+                        ocpp::v201::OperationalStatusEnum::Operative;
+                    auto connector_id_str = std::to_string(evse_id) + "." + std::to_string(connector_id.value());
```
hikinggrass, line 539: This could be moved into a dedicated function

corneliusclaussen (approved):

modules/System/main/systemImpl.cpp
```diff
@@ -375,7 +375,6 @@ systemImpl::handle_upload_logs(types::system::UploadLogsRequest& upload_logs_req
 };
 
 bool systemImpl::handle_is_reset_allowed(types::system::ResetType& type) {
-    // Right now we dont want to reject a reset ever
```
hikinggrass, line 378: This comment is still relevant, isn't it?

PR #394, opened 2023-10-20
Pg add security to evsev2g

AssemblyJohn:


modules/EvseV2G/manifest.yaml
```diff
@@ -71,6 +71,12 @@ provides:
     description: >-
       This module implements the ISO15118-2 implementation of
       an AC or DC charger
+requires:
+  security:
+    interface: evse_security
+    min_connections: 1
+    max_connections: 1
```
corneliusclaussen, line 78: min and max could be removed here, the default is always exaclty one

corneliusclaussen (changes requested):
Build in the CI should be fixed before merging

SebaLukas (changes requested):
Basically, I think the changes are good :+1: 
Since the requirement of EvseSecurity is not optional, please extend it for all config.yaml files that use EvseV2G. E.g. config-sil-dc, config-sil-two-evse etc.

Pietfried (approved):

SebaLukas (approved):
:+1:

corneliusclaussen (approved):

PR #395, opened 2023-10-22
EV-314: Add install/update request handling to OCPP module

klemmpnx:
- include handling of certificate update/install requests in ocpp module
- fix type "certificate_reponse" in evse_manager interface

### DIFF
--- a/modules/OCPP201/OCPP201.cpp
+++ b/modules/OCPP201/OCPP201.cpp
@@ -766,8 +766,9 @@ void OCPP201::ready() {
                 }
 
                 this->charge_point->on_transaction_started(
-                    evse_id, connector_id, session_id, timestamp, trigger_reason, meter_value, id_token, std::nullopt,
-                    reservation_id, remote_start_id,
+                    evse_id, connector_id, session_id, timestamp,
+                    ocpp::v201::TriggerReasonEnum::RemoteStart, // FIXME(piet): Use proper reason here
+                    meter_value, id_token, std::nullopt, reservation_id, remote_start_id,
                     ocpp::v201::ChargingStateEnum::EVConnected); // FIXME(piet): add proper groupIdToken +
                                                                  // ChargingStateEnum
                 break;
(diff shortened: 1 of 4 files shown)


modules/OCPP201/OCPP201.cpp
```diff
@@ -840,18 +841,50 @@ void OCPP201::ready() {
...
+                    switch (action) {
+                    case types::iso15118_charger::CertificateActionEnum::Install:
+                        return ocpp::v201::CertificateActionEnum::Install;
+                    case types::iso15118_charger::CertificateActionEnum::Update:
+                        return ocpp::v201::CertificateActionEnum::Update;
+                    }
+                }(certificate_request.certificateAction);
+
+                auto ocpp_response = this->charge_point->on_get_15118_ev_certificate_request(ocpp_request);
+                EVLOG_info << "received response: " << ocpp_response;
```
valentin-dimov, line 861: Is this a debug print?

modules/OCPP201/OCPP201.cpp
```diff
@@ -840,18 +841,50 @@ void OCPP201::ready() {
...
+                    switch (action) {
+                    case ocpp::v201::Iso15118EVCertificateStatusEnum::Accepted:
+                        return types::iso15118_charger::Status::Accepted;
+                    case ocpp::v201::Iso15118EVCertificateStatusEnum::Failed:
+                        return types::iso15118_charger::Status::Failed;
+                    }
+                }(ocpp_response.status);
+                const types::iso15118_charger::Response_Exi_Stream_Status everest_response{
+                    everest_response_status, certificate_request.certificateAction, ocpp_response.exiResponse};
+                EVLOG_info << "return response: " << everest_response;
```
valentin-dimov, line 873: Is this a debug print?

modules/OCPP201/OCPP201.cpp
```diff
@@ -840,18 +841,50 @@ void OCPP201::ready() {
...
     }
+    r_system->subscribe_firmware_update_status([this](const types::system::FirmwareUpdateStatus status) {
+        this->charge_point->on_firmware_update_status_notification(
+            status.request_id, get_firmware_status_notification(status.firmware_update_status));
+    });
+
+    r_system->subscribe_log_status([this](types::system::LogStatus status) {
+        this->charge_point->on_log_status_notification(get_upload_log_status_enum(status.log_status),
+                                                       status.request_id);
+    });
```
valentin-dimov, line 887: Question: These subscriptions don't seem related to M01-M05? It's probably not relevant to the review, but what functionality do they fulfill?

valentin-dimov (approved):
Looks good, I just requested a few clarifications.

modules/OCPP201/OCPP201.cpp
klemmpnx: yes indeed

modules/OCPP201/OCPP201.cpp
klemmpnx: yes (removed since more a dev comment)

modules/OCPP201/OCPP201.cpp
klemmpnx: hmm no this was just a minor "bug" (not even really) that is discovered: the r_system connection is independent of the evse, but due to an error the system subscriptions were included in the loop above (and thus done redundantly multiple times); but please verify this again in the code!

valentin-dimov: Hm, from what I can tell by glancing through the framework code, it seems that subscribing to the same var multiple times will cause EVerest to subscribe to the MQTT topic multiple times, there doesn't seem to be a check to prevent that. It's possible that the MQTT client simply uses the last subscription it's given, so this wouldn't be a bug, but I'd agree this is probably more like what we want.

Is it possible to exit early out of the loop over all EVSEs above without crashing?

PR #399, opened 2023-10-25
Initial draft for certificate generation

a-w50:


CMakeLists.txt
```diff
@@ -32,7 +32,6 @@ list(INSERT CMAKE_MODULE_PATH 0 "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
 option(CREATE_SYMLINKS "Create symlinks to javascript modules and auxillary files - for development purposes" OFF)
 option(CMAKE_RUN_CLANG_TIDY "Run clang-tidy" OFF)
 option(BUILD_TESTING "Run unit tests" OFF)
-option(ISO15118_2_GENERATE_AND_INSTALL_CERTIFICATES "Automatically generate and install certificates for development purposes" ON)
```
hikinggrass, line 35: It looks like there is no way to turn off certificate generation, but I think one should still be able to provide their own certificates if they so choose

config/certs/.gitignore
hikinggrass, line 1: Removing this directory structure does make it a bit harder to provide your own certificates in the required structure

CMakeLists.txt
```diff
@@ -76,11 +75,6 @@ include(ev-project-bootstrap)
 
 ev_add_project()
 
-# create MF_ROOT_CA if not available
-if (ISO15118_2_GENERATE_AND_INSTALL_CERTIFICATES)
-    file(TOUCH config/certs/ca/mf/MF_ROOT_CA.pem)
```
hikinggrass, line 81: I think this file is not generated in the generate_certificates.py but its absence breaks OCPP

hikinggrass (changes requested):

CMakeLists.txt
a-w50: Correct, this needs to be discussed - therefor I didn't at it yet

config/certs/.gitignore
a-w50: Good point, but I think it would be a bad idea to put your certificates as an external user into the source tree of everest-core.  This should be done differently and will be the responsibility of the external user

CMakeLists.txt
a-w50: Yes, thanks!  I need to check this, haven't tested yet, if the generated files are sufficient for running sil stuff.  But here I would rather at the generation of these necessary files into the generator script

AssemblyJohn (approved):

CMakeLists.txt
Pietfried: This should be mitigated with https://github.com/EVerest/libevse-security/tree/workaround/missing_certificates , which attempts to create the files that are configured but do not yet exist within the libevse-security. @AssemblyJohn

CMakeLists.txt
Pietfried: I agree certificate generation should be an option

corneliusclaussen:
What is the status of this one? Do we still need it?

Pietfried:
> What is the status of this one? Do we still need it?

yes, ill try to pick this up together with @a-w50

PR #403, opened 2023-10-30
Extensions for API and Setup module

hikinggrass:


modules/API/API.cpp
```diff
@@ -363,4 +379,264 @@ void API::ready() {
     invoke_ready(*p_main);
 }
 
+std::string API::limit_decimal_places(const types::powermeter::Powermeter& powermeter) {
```
a-w50, line 382: I'm quite unhappy with this solution:

- if used like this, I would put it into a separate cpp file, because it makes the `API.cpp` unreadable
- if would not couple the API module to libryml just because of the two functions `fmt::boolalpha` and `fmt::real`, these could be written by hand or  maybe there are also `std::*` functions for that
- problem seems to be, that nlohmann::json is not capable of custom number formatting
- in general this formatting is the responsibility of the presentation layer, why isn't the rounding done there
- there are so much config options now introduced now for the API module, which, if used, would just polute the config file - I doubt it, that the config file for everest should really care about configuring the precision, in which numbers are represented
- if it should be really done this way, I would introduce one string config option, which contains all these options in text format with a simple delimiter, so you can generically iterate of the options and set entries in the json accordingly (something like "energy_Wh_import.total:0.2:1", meaning, that the dict item at energy_Wh_import['total'] should be rounded to 0.2 with 1 place behind 0) . Don't know if config options support arrays, but with a simple string it would then look like
```
module_config:
  float_representation: >
     energy_Wh_import.total:0.2:1
     energy_Wh_import.L3:0.04:2
```
if done this way, you can write a loop in the code which automatically does all the work and don't have to reformat all the time
- isn't specifying round to and number of places somehow redunant?
- but *really*, this formatting should be done at the presentation layer, not here - there might be even different consumers of the API module which would like to represent it differently

corneliusclaussen (approved):

PR #404, opened 2023-10-30
Update everest-framework to 0.7.1 and require ev-cli 0.0.21

hikinggrass:
This fixes issues with integers in variants, which can be used by the PersistentStore module


a-w50 (approved):
:1234:

PR #405, opened 2023-10-31
Add legacy wakeup for PWM basic AC charging as per IEC61851-1 A.5.3

corneliusclaussen:


modules/EvseManager/Charger.cpp
```diff
@@ -83,12 +83,15 @@ void Charger::mainThread() {
 }
 
 void Charger::runStateMachine() {
-    bool new_in_state = lastState != currentState;
+    bool new_in_state = last_state_detect_state_change != currentState;
     if (new_in_state) {
-        session_log.evse(
-            false, fmt::format("Charger state: {}->{}", evseStateToString(lastState), evseStateToString(currentState)));
+        session_log.evse(false, fmt::format("Charger state: {}->{}", evseStateToString(last_state),
```
SebaLukas, line 88: `last_state_detect_state_change` instead `last_state`?

modules/EvseManager/Charger.cpp
```diff
@@ -421,8 +424,11 @@ void Charger::runStateMachine() {
                 break;
             }
 
-            if (new_in_state && lastState != EvseState::PrepareCharging) {
-                signalEvent(types::evse_manager::SessionEventEnum::ChargingResumed);
+            if (new_in_state) {
+                if (last_state not_eq EvseState::PrepareCharging) {
```
SebaLukas, line 428: Not sure if we should mix `!=` and `not_eq`. Better to stay here with `!=`.

modules/EvseManager/Charger.cpp
```diff
@@ -434,6 +440,18 @@ void Charger::runStateMachine() {
                 if (!hlc_use_5percent_current_session)
                     update_pwm_max_every_5seconds(ampereToDutyCycle(getMaxCurrent()));
             }
+
+            // If we are in charging for some time now, but relais did not close something is wrong.
+            // Try a T_step_EF to restart charging once (legacy wakeup according to IEC61851-1 A.5.3)
+            if (not hlc_charging_active and not legacy_wakeup_done and timeInCurrentState > legacy_wakeup_timeout and
+                not contactors_closed) {
```
SebaLukas, line 447: The same question as above

SebaLukas (approved):
Generally okay :+1:  just a few things to discuss.

modules/EvseManager/Charger.cpp
corneliusclaussen: fixed

modules/EvseManager/Charger.cpp
corneliusclaussen: ok, changed to old C++ style so that it is consistent within the file. On new files I prefer to use and/not etc.

modules/EvseManager/Charger.cpp
corneliusclaussen: changed

PR #407, opened 2023-11-01
OCPP201: using persist flag when operational state for ChargingStation,EVSE or Connector is changed

Pietfried:


modules/OCPP201/OCPP201.cpp
```diff
@@ -407,6 +407,51 @@ ocpp::v201::BootReasonEnum get_boot_reason(types::system::BootReason reason) {
     }
 }
 
+ocpp::v201::ReasonEnum get_reason(types::evse_manager::StopTransactionReason reason) {
+    switch (reason) {
+    case types::evse_manager::StopTransactionReason::DeAuthorized:
+        return ocpp::v201::ReasonEnum::DeAuthorized;
```
a-w50, line 413: For these enum convert functions, it saves some space and typing, if you're using `using` like
```
using EvseReason = types::evse_manager::StopTransactionReason;
using OcppReason = ocpp::v201::ReasonEnum;
```
inside the function. Then you can write:
```
case EvseReason::EmergencyStop:
    return OcppReason::EmergencyStop;
```
Furthermore, I would call the function `convert_reason`

modules/OCPP201/OCPP201.cpp
```diff
@@ -473,24 +518,24 @@ bool OCPP201::all_evse_ready() {
...
 
 void OCPP201::set_connector_operational_status(const ocpp::v201::OperationalStatusEnum operational_status,
-                                               const int32_t evse_id, const int32_t connector_id) {
+                                               const int32_t evse_id, const int32_t connector_id, const bool persist) {
     if (operational_status == ocpp::v201::OperationalStatusEnum::Operative) {
         this->evses.at(evse_id).connectors.at(connector_id) = ocpp::v201::OperationalStatusEnum::Operative;
         auto connector_id_str = std::to_string(evse_id) + "." + std::to_string(connector_id);
         this->r_kvs->call_delete(KVS_OCPP201_INOPERATIVE_KEY_PREFIX + connector_id_str);
-    } else {
+    } else if (persist) {
```
a-w50, line 526: Is it true, that the status will only be set to inoperative, when the persist flag is true?

hikinggrass (approved):

PR #408, opened 2023-11-01
Validating v201 Websocket connection URI in

Pietfried:
adresses: TC_B_43_CS

### DIFF
--- a/modules/OCPP201/OCPP201.cpp
+++ b/modules/OCPP201/OCPP201.cpp
@@ -4,6 +4,7 @@
 
 #include <fmt/core.h>
 #include <fstream>
+#include <websocketpp/uri.hpp>
 
 #include <evse_security_ocpp.hpp>
 namespace module {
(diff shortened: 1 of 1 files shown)


modules/OCPP201/OCPP201.cpp
```diff
@@ -689,7 +690,14 @@ void OCPP201::ready() {
     callbacks.validate_network_profile_callback =
         [this](const int32_t configuration_slot,
                const ocpp::v201::NetworkConnectionProfile& network_connection_profile) {
```
a-w50, line 692: Currently, `configuration_slot` is not used, is it necessary at all?
Furthermore I would drop the captured `this`, because it is also not used.

hikinggrass: this will be relevant in the future for other validations

hikinggrass (approved):

PR #410, opened 2023-11-02
Add option to disable auth in EvseManager

corneliusclaussen:
This allows EvseManager to be used without the Auth module for simple free charging applications.

### DIFF
--- a/modules/EvseManager/evse/evse_managerImpl.cpp
+++ b/modules/EvseManager/evse/evse_managerImpl.cpp
@@ -137,6 +137,15 @@ void evse_managerImpl::ready() {
 
             auto reason = mod->charger->getSessionStartedReason();
 
+            if (mod->config.disable_authentication && reason == types::evse_manager::StartSessionReason::EVConnected) {
+                // Free service, authorize immediately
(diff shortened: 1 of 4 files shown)


modules/EvseManager/evse/evse_managerImpl.cpp
```diff
@@ -137,6 +137,15 @@ void evse_managerImpl::ready() {
 
             auto reason = mod->charger->getSessionStartedReason();
 
+            if (mod->config.disable_authentication && reason == types::evse_manager::StartSessionReason::EVConnected) {
+                // Free service, authorize immediately
+                types::authorization::ProvidedIdToken provided_token;
+                provided_token.authorization_type = types::authorization::AuthorizationType::RFID;
+                provided_token.id_token = "FREESERVICE";
```
hikinggrass, line 144: Adding
```cpp
provided_token.prevalidated = true;
```
here might make the intention even clearer

Pietfried (approved):

modules/EvseManager/evse/evse_managerImpl.cpp
corneliusclaussen: added

PR #415, opened 2023-11-05
Add nanopb to exported targets

a-w50:


lib/3rd_party/nanopb/CMakeLists.txt
```diff
@@ -8,7 +8,34 @@ target_sources(nanopb
...
+)
+
+set_target_properties(nanopb PROPERTIES OUTPUT_NAME "everest_nanopb")
+
+install(TARGETS nanopb
+    EXPORT everest-core-targets
+    LIBRARY
+)
+
+# FIXME (aw): this would collide with the real nanopb library if
```
hikinggrass, line 34: Does this comment still apply? This will (only) collide if nanopb is installed into /<libdir>/3rd_party doesn't it? If it's still a problem why not fix it in this PR?

hikinggrass (changes requested):

lib/3rd_party/nanopb/CMakeLists.txt
a-w50: :+1: will remove it

hikinggrass (approved):

PR #416, opened 2023-11-06
Added missing cmd handler to PyJosev

SebaLukas:
The handler for the cmd `supporting_sae_j2847_bidi` was missing in PyJosev and caused a crash if the SAE Bidi was active in the config.


klemmpnx (approved):
lgtm!

PR #418, opened 2023-11-07
Rust: add config support

dorezyuk:
Adds rust support for config handling.

The interface for getting the config is a stand alone function allowing the user to get the config before instantiating the user's objects. The PR goes together with the PR into `everest-framework` https://github.com/EVerest/everest-framework/pull/113

### DIFF
--- a/modules/rust_examples/RsExample/src/eventually_generated.rs
+++ b/modules/rust_examples/RsExample/src/eventually_generated.rs
@@ -16,6 +16,71 @@ use std::collections::HashMap;
 use std::pin::Pin;
 use std::sync::Arc;
 
+/// Struct holding the config for the `foobar` interface.
+#[derive(Debug)]
+pub struct FoobarConfig {
+    /// An interface level bool config.
+    pub some_bool_config: bool,
+
+    /// An interface level integer config.
+    pub some_integer_config: i64,
+}
+
+/// Struct holding the module level config.
+#[derive(Debug)]
+pub struct ModuleConfig {
+    /// A module level string config.
+    pub some_string_config: String,
+
+    /// A module level number config   
+    pub some_number_config: f64,
+
+    /// The config for the `foobar` interface.
+    pub foobar_config: FoobarConfig,
+}
+
+/// Returns the config for the whole module. You can call this function at any
+/// time in your code.
+pub fn get_config() -> ModuleConfig {
+    let raw_config = everestrs::get_module_configs();
+    let foobar_config = FoobarConfig {
+        some_bool_config: raw_config
+            .get("foobar")
+            .unwrap()
+            .get("some_bool_config")
+            .unwrap()
+            .try_into()
+            .unwrap(),
+
+        some_integer_config: raw_config
+            .get("foobar")
+            .unwrap()
+            .get("some_integer_config")
+            .unwrap()
+            .try_into()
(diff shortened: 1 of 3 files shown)


modules/rust_examples/RsExample/src/eventually_generated.rs
```diff
@@ -16,6 +16,71 @@ use std::collections::HashMap;
 use std::pin::Pin;
 use std::sync::Arc;
 
+/// Struct holding the config for the `foobar` interface.
+#[derive(Debug)]
+pub struct FoobarConfig {
```
SirVer, line 21: Can you not somehow use Serde to deserialize the config structs? Because then you could use `#[serde(default=42)]` which would show up in docs and so on. That would be nice.

modules/rust_examples/RsExample/src/eventually_generated.rs
```diff
@@ -16,6 +16,71 @@ use std::collections::HashMap;
...
+    };
+
+    ModuleConfig {
+        some_string_config: raw_config
+            .get("!module")
+            .unwrap()
+            .get("some_string_config")
+            .unwrap()
+            .try_into()
+            .unwrap(),
```
a-w50, line 71: Isn't this a good place to use a rust macro?

dorezyuk: this code will be replaced by auto generated code - not sure if a macro is needed.. I think we should rather think how to handle errors

modules/rust_examples/RsExample/src/eventually_generated.rs
dorezyuk: I think this is not necessary - we aren't parsing a json blob here but get from c++ already one value (either default or an adjusted value)

modules/rust_examples/RsExample/src/eventually_generated.rs
```diff
@@ -16,6 +16,71 @@ use std::collections::HashMap;
...
+    /// The config for the `foobar` interface.
+    pub foobar_config: FoobarConfig,
+}
+
+/// Returns the config for the whole module. You can call this function at any
+/// time in your code.
+pub fn get_config() -> ModuleConfig {
+    let raw_config = everestrs::get_module_configs();
+    let foobar_config = FoobarConfig {
+        some_bool_config: raw_config
```
SirVer, line 47: I think `raw_config["foobar"]["get_some_bool_config"].try_into().unwrap()` should work and is equivalent.

however, since you are always anyways just `try_into().unwrap()`, I think you should instead in the other one just implement Into and panic there. That makes the code simpler and gets rid of an error that youl anyways never want the user to see.

SirVer (commented):
I feel we should not use TryInto, instead use `Into`, since we anyways expect the Config to be typesafe. Otherwise this looks nice to me.

modules/rust_examples/RsExample/src/eventually_generated.rs
dorezyuk: Its a little bit a question how we're perceiving the framework and code libraries. Since for now the framework is stand alone we should not panic inside the framework just because the consumer can't handle errors. If we will change this mind set in the future, we can reflect it in the interfaces

a-w50 (approved):
:+1:

PR #419, opened 2023-11-07
Improve build and test workflow

a-w50:
- disabling certificate generation in ci step, because creates "untarable" files, that should be fixed soon
- converting default runner back to free one
- allow selecting paid runner on manual dispatch
- improved github cache handling

### DIFF
--- a/.github/workflows/build_and_test.yaml
+++ b/.github/workflows/build_and_test.yaml
@@ -6,17 +6,17 @@ on:
       runner:
         description: Which runner to use
         type: choice
-        default: 'ubuntu-latest'
+        default: 'ubuntu-22.04'
         required: true
         options:
-          - 'ubuntu-latest'
+          - 'ubuntu-22.04'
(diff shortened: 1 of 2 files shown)


.github/workflows/build_and_test.yaml
```diff
@@ -1,21 +1,34 @@
 name: Build and test
-on: [workflow_dispatch, pull_request]
+on:
+  pull_request: {}
+  workflow_dispatch:
+    inputs:
+      runner:
+        description: Which runner to use
+        type: choice
+        default: 'ubuntu-latest'
```
andistorm, line 9:
```suggestion
        default: 'ubuntu-22.04'
```
I would prefer well defined runner version instead of latest

andistorm (approved):
Looks good to me.

.github/workflows/build_and_test.yaml
a-w50: Good point!

andistorm (approved):

PR #421, opened 2023-11-08
Allow edting nodered flows in place

andistorm:
* Using FLOWS arg in nodered run scripts
* Bind complete directory instead of flows.json file only
* Add other noderedfiles to .gitignore

### DIFF
--- a/config/nodered/.gitignore
+++ b/config/nodered/.gitignore
@@ -1,0 +1,6 @@
+.config*
+*.backup
+package.json
+settings.js
+lib/
+node_modules/
--- a/config/nodered/.gitignore
+++ b/config/nodered/.gitignore
@@ -1,6 +1,1 @@
-.config*
 *.backup
-package.json
-settings.js
-lib/
-node_modules/
--- a/cmake/assets/run_nodered_template.sh.in
+++ b/cmake/assets/run_nodered_template.sh.in
@@ -4,6 +4,6 @@ docker run \
     --rm \
     --network host \
     --name everest_nodered \
-    -e FLOWS=@FLOW_FILE_NAME@ \
-    --mount type=bind,source=@FLOW_FILE_DIRECTORY@,target=/data/ \
+    -e FLOWS=flows/@FLOW_FILE_NAME@ \
+    --mount type=bind,source=@FLOW_FILE_DIRECTORY@,target=/data/flows \
     everest-nodered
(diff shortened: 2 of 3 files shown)


config/nodered/.gitignore
```diff
@@ -0,0 +1,6 @@
+.config*
+*.backup
+package.json
+settings.js
+lib/
+node_modules/
```
a-w50, line 6: is this needed?

andistorm: To sync changes on the flows.json on the host system, it seems to be necessary to bind a complete data directory, because nodered doesn't update the file, but delete the old one and create a new one. This is also why links in a bind directory wouldn't work.
This is why I bind config/nodered directory as /data in the container, which means that nodered puts also other files there we don't want to track

andistorm: Ok, maybe it is not needed. I'll try something

andistorm:
We could think about using an plugin to manage flow storage.
Something like [this one](https://flows.nodered.org/node/node-red-contrib-js-storage) for  example, it promises more clear commits by splitting the flows.json up into multiple files. (no backup file anymore)

config/nodered/.gitignore
a-w50: I see, in generall I would try that nothing, if not otherwise possible gets created inside the source folder.  One common pattern would be to copy or install these node files to some subfolder of the build directory and then mount that
Do you think this is doable?

andistorm: My main goal was to edit the flows in place.
In the case you are starting ndoered to work on the flows, I see nodered - the ndoered container - as editor.

When we copy the files in build directory it wouldn't sync to the origin files. It would be encessary to move/copy files after editing, or to use the nodered export feature..

I also already tried to mount a temporary directory in the build directory with an symbolic link to the origin file, but nodered destroys this symbolic link and its again necessary to sync the file manually

a-w50: yes, symbolic links usually do not work through the mount/bind
I know that it is handy to be able to edit the files in the same place, we had this previously also with the config files (which btw. now are copied to dist), but the problem was that it happened from time to time, that people pushed their changes in the config files, but we don't want that

andistorm: I understand. What would be to distinguish between 'Start nodered to run sil' and 'Start nodered to edit flows'?
-> Deploying a script version that opens flows editable (in place) in addition to the current run scripts

corneliusclaussen: Maybe copying the flow files to build folder and have them editable there would be a good compromise. Then it is relativly easy to copy them back manually to commit them and it does not happen accidentially.

corneliusclaussen:
> We could think about using an plugin to manage flow storage. Something like [this one](https://flows.nodered.org/node/node-red-contrib-js-storage) for example, it promises more clear commits by splitting the flows.json up into multiple files. (no backup file anymore)

I think that is overkill for what we do with nodered

config/nodered/.gitignore
andistorm: Would be at least better as now :) I will implement this.

PR #422, opened 2023-11-08
Add command to system interface to explicity allow firmware updates

hikinggrass:
This is used by OCPP and OCPP201 to make sure that firmware updates are only installed when no transactions are active

This needs [#libocpp](https://github.com/EVerest/libocpp/pull/252)

### DIFF
--- a/modules/OCPP201/OCPP201.cpp
+++ b/modules/OCPP201/OCPP201.cpp
@@ -765,6 +765,11 @@ void OCPP201::ready() {
     callbacks.configure_network_connection_profile_callback =
         [this](const ocpp::v201::NetworkConnectionProfile& network_connection_profile) { return true; };
 
+    callbacks.all_connectors_unavailable_callback = [this]() {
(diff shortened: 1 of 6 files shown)


modules/OCPP201/OCPP201.cpp
```diff
@@ -765,6 +765,11 @@ void OCPP201::ready() {
     callbacks.configure_network_connection_profile_callback =
         [this](const ocpp::v201::NetworkConnectionProfile& network_connection_profile) { return true; };
 
+    callbacks.all_connectors_unavailable_callback = [this]() {
```
a-w50, line 768: Concerning naming: if the containing struct is already called `callbacks`, then I wouldn't add `_callback` at the end of all of its members.  But just by reading the name of this callback, `all_connectors_unavailable` I'm expecting that this callback is called, when all connectors are unavailable.   But then I'm not necessarily expecting that I should proceed with firmware installation

Pietfried (approved):

PR #424, opened 2023-11-10
Improving start of charge performance for AC BASIC charging

corneliusclaussen:
It took quite some time to enable PWM after plugin even in simple scenarios where no OCPP auth is neccessary and no power balancing is required.

This improves the following:
- Reducing time to PWM on by avoiding unneccesary calls to BSP to set PWM
- EvseManager now by default always requests the full amount of energy from the EnergyManager even if no car is plugged. In load balancing this can be disabled by setting config option "request_zero_power_in_idle" to true. This will add a slight delay to the start of charging as energy will need to be reserved by the EnergyManager first.

### DIFF
--- a/modules/EvseManager/manifest.yaml
+++ b/modules/EvseManager/manifest.yaml
@@ -193,7 +193,15 @@ config:
     enum:
       - V2H
       - V2G
-    default: V2G 
+    default: V2G
+  request_zero_power_in_idle:
+    description: >-
+      "true: In Idle mode (no car connected), request 0A from energy management. In installations with many charging stations this should be set"
+      "to allow the power to be distributed to the chargers that are connected to a car."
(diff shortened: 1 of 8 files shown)


modules/EvseManager/manifest.yaml
```diff
@@ -193,7 +193,15 @@ config:
     enum:
       - V2H
       - V2G
-    default: V2G 
+    default: V2G
+  request_zero_power_in_idle:
+    description: >-
+      "true: In Idle mode (no car connected), request 0A from energy management. In installations with many charging stations this should be set"
```
Pietfried, line 199: nit: quotes are not needed

Pietfried (approved):

SebaLukas (approved):
:+1:

modules/EvseManager/manifest.yaml
corneliusclaussen: aren't they because of the ":"?

PR #425, opened 2023-11-10
Interfaces needed to support bank payments

golovasteek:
Interfaces needed for the bank-transaction handlin.

The documentation about how does it work is in PR to [the main EVerest repo](https://github.com/EVerest/EVerest/pull/117)

### DIFF
--- a/interfaces/bank_transaction_summary_provider.yaml
+++ b/interfaces/bank_transaction_summary_provider.yaml
@@ -1,0 +1,10 @@
+description: >-
+  Provides information of the session that was committed to the bank.
+  This data may be needed for accounting purposes.
+vars:
+  bank_transaction_summary:
+    description: >-
+      Summary of a bank transaction. Depends on bank and the backend. Therefore
+      it's mostly opaque data at the moment.
+    type: object
+    $ref: /bank_transaction#/BankTransactionSummary
--- a/types/authorization.yaml
+++ b/types/authorization.yaml
@@ -159,6 +159,7 @@ types:
       - RFID
       - Autocharge
       - PlugAndCharge
+      - BankCard
   IdTokenType:
     description: >-
       IdTokenType of the provided token
--- a/types/bank_transaction.yaml
+++ b/types/bank_transaction.yaml
@@ -1,0 +1,32 @@
+description: >-
+  Types needed for bookkeeping bank transactions
+types:
+  BankSessionToken:
+    description: >-
+      In some setups the unique token needs to be provided in the bank statement.
+    type: object
+    additionalProperties: false
+    properties:
+      token:
+        description: >-
+          Arbitrary token string. Typically short string that is printable case insensitive
+          ascii.
+        type: string
+  BankTransactionSummary:
+    description: >-
+      Summary of a bank transaction. Depends on bank and the backend. Therefore
+      it's mostly opaque data at the moment.
+    type: object
+    additionalProperties: false
+    properties:
+      session_token:
+        description: >-
+          Token to identify the session. This token can be used to download the
+          invoices.
+        type: object
+        $ref: /bank_transaction#/BankSessionToken
+      transaction_data:
+        description: >-
+          Arbitrary data that is needed to identify the transaction. This is
+          mostly opaque data at the moment.
+        type: string
--- a/types/money.yaml
+++ b/types/money.yaml
@@ -1,0 +1,510 @@
+description: >-
+  The definition of the currency and price types.
+types:
+  CurrencyCode:
+    description: >-
+      Currency in 3 alphabetical characters according to ISO 4217.
+      The list of values is taken from the 
+      https://www.six-group.com/en/products-services/financial-information/data-standards.html
+      Technical values like "Testing Code", "No Currency"
+      as well as metal and bond-market codes are excluded.
+    type: string
+    minLength: 3
+    maxLength: 3
+    example: EUR
+    enum:
+      # 784, UAE Dirham, UNITED ARAB EMIRATES (THE)
+      - AED
+      # 971, Afghani, AFGHANISTAN
+      - AFN
+      # 008, Lek, ALBANIA
+      - ALL
+      # 051, Armenian Dram, ARMENIA
+      - AMD
+      # 532, Netherlands Antillean Guilder, CURAÇAO
+      # SINT MAARTEN (DUTCH PART)
+      - ANG
+      # 973, Kwanza, ANGOLA
+      - AOA
+      # 032, Argentine Peso, ARGENTINA
+      - ARS
+      # 036, Australian Dollar, AUSTRALIA
+      # CHRISTMAS ISLAND
+      # COCOS (KEELING) ISLANDS (THE)
+      # HEARD ISLAND AND McDONALD ISLANDS
+      # KIRIBATI
+      # NAURU
+      # NORFOLK ISLAND
+      # TUVALU
+      - AUD
+      # 533, Aruban Florin, ARUBA
+      - AWG
+      # 944, Azerbaijan Manat, AZERBAIJAN
+      - AZN
+      # 977, Convertible Mark, BOSNIA AND HERZEGOVINA
+      - BAM
+      # 052, Barbados Dollar, BARBADOS
+      - BBD
+      # 050, Taka, BANGLADESH
+      - BDT
+      # 975, Bulgarian Lev, BULGARIA
+      - BGN
+      # 048, Bahraini Dinar, BAHRAIN
+      - BHD
+      # 108, Burundi Franc, BURUNDI
+      - BIF
+      # 060, Bermudian Dollar, BERMUDA
+      - BMD
+      # 096, Brunei Dollar, BRUNEI DARUSSALAM
+      - BND
+      # 068, Boliviano, BOLIVIA (PLURINATIONAL STATE OF)
+      - BOB
+      # 984, Mvdol, BOLIVIA (PLURINATIONAL STATE OF)
+      - BOV
+      # 986, Brazilian Real, BRAZIL
+      - BRL
+      # 044, Bahamian Dollar, BAHAMAS (THE)
+      - BSD
+      # 064, Ngultrum, BHUTAN
+      - BTN
+      # 072, Pula, BOTSWANA
+      - BWP
+      # 933, Belarusian Ruble, BELARUS
+      - BYN
+      # 084, Belize Dollar, BELIZE
+      - BZD
+      # 124, Canadian Dollar, CANADA
+      - CAD
+      # 976, Congolese Franc, CONGO (THE DEMOCRATIC REPUBLIC OF THE)
+      - CDF
+      # 947, WIR Euro, SWITZERLAND
+      - CHE
+      # 756, Swiss Franc, LIECHTENSTEIN
+      # SWITZERLAND
+      - CHF
+      # 948, WIR Franc, SWITZERLAND
+      - CHW
+      # 990, Unidad de Fomento, CHILE
+      - CLF
+      # 152, Chilean Peso, CHILE
+      - CLP
+      # 156, Yuan Renminbi, CHINA
+      - CNY
+      # 170, Colombian Peso, COLOMBIA
+      - COP
+      # 970, Unidad de Valor Real, COLOMBIA
+      - COU
+      # 188, Costa Rican Colon, COSTA RICA
+      - CRC
+      # 931, Peso Convertible, CUBA
+      - CUC
+      # 192, Cuban Peso, CUBA
+      - CUP
+      # 132, Cabo Verde Escudo, CABO VERDE
+      - CVE
+      # 203, Czech Koruna, CZECHIA
+      - CZK
+      # 262, Djibouti Franc, DJIBOUTI
+      - DJF
+      # 208, Danish Krone, DENMARK
+      # FAROE ISLANDS (THE)
+      # GREENLAND
+      - DKK
+      # 214, Dominican Peso, DOMINICAN REPUBLIC (THE)
+      - DOP
+      # 012, Algerian Dinar, ALGERIA
+      - DZD
+      # 818, Egyptian Pound, EGYPT
+      - EGP
+      # 232, Nakfa, ERITREA
+      - ERN
+      # 230, Ethiopian Birr, ETHIOPIA
+      - ETB
+      # 978, Euro, ÅLAND ISLANDS
+      # ANDORRA
+      # AUSTRIA
+      # BELGIUM
+      # CROATIA
+      # CYPRUS
+      # ESTONIA
+      # EUROPEAN UNION
+      # FINLAND
+      # FRANCE
+      # FRENCH GUIANA
+      # FRENCH SOUTHERN TERRITORIES (THE)
+      # GERMANY
+      # GREECE
+      # GUADELOUPE
+      # HOLY SEE (THE)
+      # IRELAND
+      # ITALY
+      # LATVIA
+      # LITHUANIA
+      # LUXEMBOURG
+      # MALTA
+      # MARTINIQUE
+      # MAYOTTE
+      # MONACO
+      # MONTENEGRO
+      # NETHERLANDS (THE)
+      # PORTUGAL
+      # RÉUNION
+      # SAINT BARTHÉLEMY
+      # SAINT MARTIN (FRENCH PART)
+      # SAINT PIERRE AND MIQUELON
+      # SAN MARINO
+      # SLOVAKIA
+      # SLOVENIA
+      # SPAIN
+      - EUR
+      # 242, Fiji Dollar, FIJI
+      - FJD
+      # 238, Falkland Islands Pound, FALKLAND ISLANDS (THE) [MALVINAS]
+      - FKP
+      # 826, Pound Sterling, GUERNSEY
+      # ISLE OF MAN
+      # JERSEY
+      # UNITED KINGDOM OF GREAT BRITAIN AND NORTHERN IRELAND (THE)
+      - GBP
+      # 981, Lari, GEORGIA
+      - GEL
+      # 936, Ghana Cedi, GHANA
+      - GHS
+      # 292, Gibraltar Pound, GIBRALTAR
+      - GIP
+      # 270, Dalasi, GAMBIA (THE)
+      - GMD
+      # 324, Guinean Franc, GUINEA
+      - GNF
+      # 320, Quetzal, GUATEMALA
+      - GTQ
+      # 328, Guyana Dollar, GUYANA
+      - GYD
+      # 344, Hong Kong Dollar, HONG KONG
+      - HKD
+      # 340, Lempira, HONDURAS
+      - HNL
+      # 332, Gourde, HAITI
+      - HTG
+      # 348, Forint, HUNGARY
+      - HUF
+      # 360, Rupiah, INDONESIA
+      - IDR
+      # 376, New Israeli Sheqel, ISRAEL
+      - ILS
+      # 356, Indian Rupee, BHUTAN
+      # INDIA
+      - INR
+      # 368, Iraqi Dinar, IRAQ
+      - IQD
+      # 364, Iranian Rial, IRAN (ISLAMIC REPUBLIC OF)
+      - IRR
+      # 352, Iceland Krona, ICELAND
+      - ISK
+      # 388, Jamaican Dollar, JAMAICA
+      - JMD
+      # 400, Jordanian Dinar, JORDAN
+      - JOD
+      # 392, Yen, JAPAN
+      - JPY
+      # 404, Kenyan Shilling, KENYA
+      - KES
+      # 417, Som, KYRGYZSTAN
+      - KGS
+      # 116, Riel, CAMBODIA
+      - KHR
+      # 174, Comorian Franc , COMOROS (THE)
+      - KMF
+      # 408, North Korean Won, KOREA (THE DEMOCRATIC PEOPLE’S REPUBLIC OF)
+      - KPW
+      # 410, Won, KOREA (THE REPUBLIC OF)
+      - KRW
+      # 414, Kuwaiti Dinar, KUWAIT
+      - KWD
+      # 136, Cayman Islands Dollar, CAYMAN ISLANDS (THE)
+      - KYD
+      # 398, Tenge, KAZAKHSTAN
+      - KZT
+      # 418, Lao Kip, LAO PEOPLE’S DEMOCRATIC REPUBLIC (THE)
+      - LAK
+      # 422, Lebanese Pound, LEBANON
+      - LBP
+      # 144, Sri Lanka Rupee, SRI LANKA
+      - LKR
+      # 430, Liberian Dollar, LIBERIA
+      - LRD
+      # 426, Loti, LESOTHO
+      - LSL
+      # 434, Libyan Dinar, LIBYA
+      - LYD
+      # 504, Moroccan Dirham, MOROCCO
+      # WESTERN SAHARA
+      - MAD
+      # 498, Moldovan Leu, MOLDOVA (THE REPUBLIC OF)
+      - MDL
+      # 969, Malagasy Ariary, MADAGASCAR
+      - MGA
+      # 807, Denar, NORTH MACEDONIA
+      - MKD
+      # 104, Kyat, MYANMAR
+      - MMK
+      # 496, Tugrik, MONGOLIA
+      - MNT
+      # 446, Pataca, MACAO
+      - MOP
+      # 929, Ouguiya, MAURITANIA
+      - MRU
+      # 480, Mauritius Rupee, MAURITIUS
+      - MUR
+      # 462, Rufiyaa, MALDIVES
+      - MVR
+      # 454, Malawi Kwacha, MALAWI
+      - MWK
+      # 484, Mexican Peso, MEXICO
+      - MXN
+      # 979, Mexican Unidad de Inversion (UDI), MEXICO
+      - MXV
+      # 458, Malaysian Ringgit, MALAYSIA
+      - MYR
+      # 943, Mozambique Metical, MOZAMBIQUE
+      - MZN
+      # 516, Namibia Dollar, NAMIBIA
+      - NAD
+      # 566, Naira, NIGERIA
+      - NGN
+      # 558, Cordoba Oro, NICARAGUA
+      - NIO
+      # 578, Norwegian Krone, BOUVET ISLAND
+      # NORWAY
+      # SVALBARD AND JAN MAYEN
+      - NOK
+      # 524, Nepalese Rupee, NEPAL
+      - NPR
+      # 554, New Zealand Dollar, COOK ISLANDS (THE)
+      # NEW ZEALAND
+      # NIUE
+      # PITCAIRN
+      # TOKELAU
+      - NZD
+      # 512, Rial Omani, OMAN
+      - OMR
+      # 590, Balboa, PANAMA
+      - PAB
+      # 604, Sol, PERU
+      - PEN
+      # 598, Kina, PAPUA NEW GUINEA
+      - PGK
+      # 608, Philippine Peso, PHILIPPINES (THE)
+      - PHP
+      # 586, Pakistan Rupee, PAKISTAN
+      - PKR
+      # 985, Zloty, POLAND
+      - PLN
+      # 600, Guarani, PARAGUAY
+      - PYG
+      # 634, Qatari Rial, QATAR
+      - QAR
+      # 946, Romanian Leu, ROMANIA
+      - RON
+      # 941, Serbian Dinar, SERBIA
+      - RSD
+      # 643, Russian Ruble, RUSSIAN FEDERATION (THE)
+      - RUB
+      # 646, Rwanda Franc, RWANDA
+      - RWF
+      # 682, Saudi Riyal, SAUDI ARABIA
+      - SAR
+      # 090, Solomon Islands Dollar, SOLOMON ISLANDS
+      - SBD
+      # 690, Seychelles Rupee, SEYCHELLES
+      - SCR
+      # 938, Sudanese Pound, SUDAN (THE)
+      - SDG
+      # 752, Swedish Krona, SWEDEN
+      - SEK
+      # 702, Singapore Dollar, SINGAPORE
+      - SGD
+      # 654, Saint Helena Pound, SAINT HELENA, ASCENSION AND TRISTAN DA CUNHA
+      - SHP
+      # 925, Leone, SIERRA LEONE
+      - SLE
+      # 694, Leone, SIERRA LEONE
+      - SLL
+      # 706, Somali Shilling, SOMALIA
+      - SOS
+      # 968, Surinam Dollar, SURINAME
+      - SRD
+      # 728, South Sudanese Pound, SOUTH SUDAN
+      - SSP
+      # 930, Dobra, SAO TOME AND PRINCIPE
+      - STN
+      # 222, El Salvador Colon, EL SALVADOR
+      - SVC
+      # 760, Syrian Pound, SYRIAN ARAB REPUBLIC
+      - SYP
+      # 748, Lilangeni, ESWATINI
+      - SZL
+      # 764, Baht, THAILAND
+      - THB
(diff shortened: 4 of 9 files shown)


a-w50:
Hi, thanks for well written text.  Two things, I've seen:
- the everest-core type system unfortunately doesn't really stick to the `$ref` standard, I think - you can see in the ci action a log, stating, that you have to put a '/' in front of every ref.
- concerning the `AuthorizationType::BankCard`, I'm not :100:% sure but I thought that the `AuthorizationType::*` is for authorizing charging, and authorizing payment doesn't necessarily mean authorizing charging

types/authorization.yaml
```diff
@@ -137,6 +137,7 @@ types:
       - RFID
       - Autocharge
       - PlugAndCharge
+      - git Card
```
corneliusclaussen, line 140: extra git?

corneliusclaussen:
On the too expensive session: I do not see how this should be implemented here. Maybe we can extend the AuthToken to include limits as well? So that the AuthTokenProvider can emit a token that says "here is a new token, it is already prevalidated, but only valid for N kWh max or M minutes max". In this case the EvseManager would not need an external stop anymore

agardiol:
Hi,
Thanks for the proposal. I am also beginning a project which involve a payment terminal. Your implementation seems interesting and I agree that a common implementation should be a good idea.
I will take time later today to look at yours.

golovasteek:
> Nice work! One additional comment: I think we should use the "prevalidated" flag in the auth token in this case to indicate that the tokenprovider already reserved the money and the transaction is approved without the need for further validators

Right. This is sufficient for most of the cases. 
But this is hard to declare on the interface level. And we foresee use-cases, where we want to pass the token through validator. For example validators can dynamically limit which outlets can be used for this session.

golovasteek:
> On the too expensive session: I do not see how this should be implemented here. Maybe we can extend the AuthToken to include limits as well? So that the AuthTokenProvider can emit a token that says "here is a new token, it is already prevalidated, but only valid for N kWh max or M minutes max". In this case the EvseManager would not need an external stop anymore.

If you think it makes sense to make EvseManager aware of the price then it could be done this way.
I've seen that EvseManager cares about charging itself and in many cases price/cost information is not preset on the station at all. And therefore price-limitation should be part of the other module.

What do you think?

types/authorization.yaml
golovasteek: done

golovasteek:
> * concerning the `AuthorizationType::BankCard`, I'm not 💯% sure but I thought that the `AuthorizationType::*` is for authorizing charging, and authorizing payment doesn't necessarily mean authorizing charging

@a-w50 Maybe I misunderstood the semantic of `AuthorizationType` enum. If you can elaborate, I would interested to know more about it.

Maybe you can share as well what is the role of `IdTokenType` as well.

corneliusclaussen:
> If you think it makes sense to make EvseManager aware of the price then it could be done this way. I've seen that EvseManager cares about charging itself and in many cases price/cost information is not preset on the station at all. And therefore price-limitation should be part of the other module.
> 
> What do you think?

We also have pricing and various limits in the energy management. If e.g. the limits of the authorizations are known to the EvseManager we can also use those limits in the energy management. So if e.g. a max of 10kWh are authorized it will also not schedule more then that in the energy manager (although I'm not sure if this detail matters in the end)
You are right that in many cases the prices are unknown. But if they are known, they can also be used in the scheduled charging to e.g. optimize for the cheapest way to fulfill the target. Ideally they are only set in one place instead of two, so I guess we need to figure out which module should be the single source for pricing information in this use case. Currently we would have the session_cost_calculator module which calculates the actual price for a given charging session and the energy_price_information which gives pricing/kwh (with schedule for the future if available). I think we should try to unify this, so that payment pricing has the same source as the pricing for optimization/display. Maybe we have some ideas in tomorrow's dev meeting?

On the "prevalidated": You are right, there is no way to enforce one way or the other in the interface. But we can still document recommended guidelines how to use the interfaces in different use cases. 

If you do not set prevalidated in this case, the validator cannot really find out if it should be valid or not, this information is only available in the module that did the money reservation. If you use a validator that always says yes, you cannot really combine it with other token sources and OCPP for example that may not always be valid. So I think it would be cleaner to follow the design concept here and set the prevalidated flag.
If you want to change the ports that this is valid for you can also create a authtoken proxy module which modifies the auth token and e.g. adds connector limitations. This would be cleaner imho to achieve that as it can co-exist with other token providers/validators.

a-w50:
> > * concerning the `AuthorizationType::BankCard`, I'm not 💯% sure but I thought that the `AuthorizationType::*` is for authorizing charging, and authorizing payment doesn't necessarily mean authorizing charging
> 
> @a-w50 Maybe I misunderstood the semantic of `AuthorizationType` enum. If you can elaborate, I would interested to know more about it.
> 
> Maybe you can share as well what is the role of `IdTokenType` as well.

Hi, thanks for your response.  Unfortunately I'm not familiar with the design of the `Authorization` type and interface.  I based my comment solely on authorization logic, that I have seen so far in the code.   And concerning this logic I got the impression, that the `AuthorizationType` is used to fully authorize charging.  Therefore I just want to make sure, that these two different things (authorizing charging and authorizing payment) are coordinated correctly.

golovasteek:
> If you do not set prevalidated in this case, the validator cannot really find out if it should be valid or not, this information is only available in the module that did the money reservation. If you use a validator that always says yes, you cannot really combine it with other token sources and OCPP for example that may not always be valid. So I think it would be cleaner to follow the design concept here and set the prevalidated flag.

Can't validator look at the token `ProvidedToken::authorization_type` to distinguish between BankCard and the other tokens.

interfaces/bank_transction_summary_provider.yaml
agardiol, line 1: Typo in the file name: transction -> transaction

interfaces/bank_transction_summary_provider.yaml
```diff
@@ -0,0 +1,10 @@
+description: >-
+  Provides information of the session that was commited to the bank.
+  This date may be needed for accounting perposes.
+vars:
+  bank_transaction_summary:
+    description: >-
+      Summary of a bank transaction. Depends on bank and the backend. Therefore
+      it's mostly opque data at the moment.
```
agardiol, line 8: Typo: opque -> opaque ?

interfaces/bank_transction_summary_provider.yaml
```diff
@@ -0,0 +1,10 @@
+description: >-
+  Provides information of the session that was commited to the bank.
+  This date may be needed for accounting perposes.
```
agardiol, line 3: Typo: This **data** may be needed for accounting **purposes**.

types/bank_transaction.yaml
```diff
@@ -0,0 +1,34 @@
...
+    properties:
+      token:
+        description: >-
+          Arbitrary token string. Typically short string that is printable case insensitive
+          ascii.
+        type: string
+  BankTransactionSummary:
+    description: >-
+      Summary of a bank transaction. Depends on bank and the backend. Therefore
+      it's mostly opque data at the moment.
```
agardiol, line 20: Typo: opque -> opaque ?

types/session_price.yaml
```diff
@@ -0,0 +1,73 @@
...
+        type: object
+        $ref: /authorization#/ProvidedIdToken
+      total_cost:
+        description: >-
+          Total cost of the session in the currency specified in the session
+          price
+        type: number
+      currency:
+        description: >-
+          Currency in 3 digit ISO 4217. This is the currency used for all
```
dorezyuk, line 60: I'm confused is this the numerical code as `978` or the string as `EUR`? If you mean digit, then would it maybe make sense to change the type to string and make the constraints to minValue 100 and maxValue 999?

interfaces/session_cost_calculator.yaml
```diff
@@ -0,0 +1,11 @@
...
+  This interface publishes the running or finished session costs. This interface
+  provides cost of one session. If we have more than one outlet, we need to 
+  instantiate this interface for each outlet.
+vars:
+  session_cost:
+    description: >-
+      Session cost object containing the total cost of the session and a list
+      of chunks
+    type: object
+    $ref: session_cost#/SessionCost
```
dorezyuk, line 11: The file is called session_price - and here you refer to session_cost. I think it should be session_cost as we typically call the file name after the "main" message

interfaces/session_cost_calculator.yaml
```diff
@@ -0,0 +1,11 @@
+description: >-
+  This interface publishes the running or finished session costs. This interface
```
dorezyuk, line 2: I think the "calculator" is not necessary - it should be clear since its an interface

types/session_price.yaml
```diff
@@ -0,0 +1,73 @@
...
+        description: >-
+          Cost of the energy consumed during this chunk in the currency
+          specified in the session price
+        type: number
+      cost_time:
+        description: >-
+          Cost of the time spent during this chunk in the currency specified
+          in the session price
+        type: number
+  SessionStatus:
```
dorezyuk, line 30: where is this used?

types/session_price.yaml
```diff
@@ -0,0 +1,73 @@
...
+        description: >-
+          Absolute timestamp for the end of this chunk in RFC3339 UTC format
+        type: string
+        format: date-time
+      cost_energy:
+        description: >-
+          Cost of the energy consumed during this chunk in the currency
+          specified in the session price
+        type: number
+      cost_time:
```
dorezyuk, line 25: Can we make it more generic? How about we  call it additional_cost and users may include here what ever they want?

types/bank_transaction.yaml
```diff
@@ -0,0 +1,34 @@
...
+      session_token:
+        description: >-
+          Token to identify the session. This token can be used to download the
+          invoices.
+        type: object
+        $ref: /bank_transaction#/BankSessionToken
+      transaction_data:
+        description: >-
+          Arbitrary data that is needed to identify the transaction. This is
+          mostly opque data at the moment.
```
dorezyuk, line 33: same typo

types/bank_transaction.yaml
```diff
@@ -0,0 +1,34 @@
+description: >-
+  Types needed for bookkeeping bank transactions
+types:
+  BankSessionToken:
+    description: >-
+      In Germany (an probably other countries) we have to provide a token
```
dorezyuk, line 6: maybe we don't have to specify the country. We can just say its an optional token to be included in the invoice?

types/session_price.yaml
dorezyuk: or even more generic - something like
```cpp
struct Chunk{
   std::string timestamp_from;
   std::string timestamp_to;
  float cost;
  std::string purpose;
};
```

golovasteek:
I've added diagram how it supposed to work in current version.

Actually, I though it would be nice to document it somewhere int the code, but haven't found any good place for it.
(I think most of the PR description should go in the documentation).

golovasteek:
Summary of open topics for discussion on the tech call:

 * where to document model that is behind the presented interfaces
 * Which module should be tracking transaction costs/time/maximum energy? The reasons behind that are that limitations on the session can be provided by different components. Options
   * payment terminal
   * EVSE Manager
   * Separate module
 * Depends on the previous decision: How should the session be stopped?
 * Should we couple the energy prices and costs into one interface?

interfaces/session_cost_calculator.yaml
```diff
@@ -0,0 +1,11 @@
...
+  provides cost of one session. If we have more than one outlet, we need to
```
SirVer, line 3: nit: use EVSEID instead of outlet, that seems to be the lingo in EVerest.

types/bank_transaction.yaml
SirVer, line 6: typo: *and* probably other counties

types/session_price.yaml
```diff
@@ -0,0 +1,73 @@
...
+        description: >-
+          Absolute timestamp for the start of this chunk in RFC3339 UTC format
+        type: string
+        format: date-time
+      timestamp_to:
+        description: >-
+          Absolute timestamp for the end of this chunk in RFC3339 UTC format
+        type: string
+        format: date-time
+      cost_energy:
```
SirVer, line 20: I am a fan of describing currency as two integers, namely the value and the exponent. For example if you want to specify 10 eurocent, you could do it as { "EUR", 10, -2 }. The equivalent int bitcoin would be { "BitCoin", 3, -5 }. I learned that at Lyft where all currency are represented like that and in fact we do the same in the Qwello backend.

The advantage of course is you work around rounding errors and you can work with fractions of the actual currency safely. I suggest using this here too.

types/session_price.yaml
SirVer: never use floating point for money ever

https://www.linkedin.com/pulse/never-use-floating-point-double-data-types-monetary-elkhawaldeh/

types/session_price.yaml
```diff
@@ -0,0 +1,73 @@
...
+      - running
+      - finished
+  SessionCost:
+    description: >-
+      Session cost object containing the total cost of the session and a list
+      of chunks
+    type: object
+    additionalProperties: false
+    required:
+      - total_cost
```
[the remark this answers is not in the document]
SirVer: +1. The logic of summing all up is trivial.

types/session_price.yaml
```diff
@@ -0,0 +1,73 @@
...
+      id_tag:
+        description: The id tag assigned to this transaction
+        type: object
+        $ref: /authorization#/ProvidedIdToken
+      total_cost:
+        description: >-
+          Total cost of the session in the currency specified in the session
+          price
+        type: number
+      currency:
```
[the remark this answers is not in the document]
SirVer: +1, but see my comments about never using floating point values for money. I have been bitten before.

types/session_price.yaml
SirVer: I have a preference for using EUR, GBP instead of 978, 826 - it makes this more readable. I agree though that this should be pointed out in the example, since the ISO defines both it seems.

types/session_price.yaml
```diff
@@ -0,0 +1,73 @@
...
+          in the session price
+        type: number
+  SessionStatus:
+    description: >-
+      Session status enum. Session can be running or finished. Costs of the
+      running session are not final and can change.
+    type: string
+    enum:
+      - running
+      - finished
```
ju4nlu, line 37:
```suggestion
      - Running
      - Finished
```
Nit: first letter capitalized in enum values

types/session_price.yaml
```diff
@@ -0,0 +1,73 @@
...
+        type: string
+        minLength: 3
+        maxLength: 3
+      cost_chunks:
+        description: >-
+          List of cost chunks. The total cost of the session is the sum of
+          all chunks.
+        type: array
+        items:
+          description: One chunk of the session cost
```
ju4nlu, line 71: Since `cost_chunks` is a required field I think it would be a good idea to specify `minItems: 1` here so that empty arrays are not allowed. I feel like an empty array would defeat the purpose of the "required" flag

types/session_price.yaml
```diff
@@ -0,0 +1,73 @@
+description: Types to provide the session price
+types:
+  SessionCostChunk:
+    description: >-
+      A chunk of the session cost. The total cost of the session is the sum of
+      all chunks.
+    type: object
+    additionalProperties: false
+    properties:
+      timestamp_from:
```
wku12, line 10: One thing - for our use case (night tariff), the minutes we charged infrastructure fee for is not simply `t_stop - t_end`
Can we add this information here? Like infrastructure cost or something like that

interfaces/session_cost_calculator.yaml
golovasteek: Fixed

types/session_price.yaml
golovasteek: I've refactored it a bit. Can you take a look?

types/session_price.yaml
golovasteek: Done

types/session_price.yaml
golovasteek: Ok. I was thinking that not required property is what would be "Optional" in python or rust.

But here we can say that default value is an empty list. So that looks fine. I've changed. it.

interfaces/session_cost_calculator.yaml
golovasteek: Agree. Renames.

golovasteek:
The comments are addressed.
The description of functionality is moved to the documentation repository https://github.com/EVerest/EVerest/pull/117 

The cost/time/energy limitation is now missing in this interfaces. We agreed that it should become a metadata attached to the token. But it would be better to work on it separately. To limit scope of the change. What do others think?

SirVer (approved):
looks good to me.

agardiol (approved):
Seems good to me too. Good job !

types/session_cost.yaml
```diff
@@ -0,0 +1,79 @@
...
+        description: >-
+          Absolute timestamp for the end of this chunk in RFC3339 UTC format
+        type: string
+        format: date-time
+      cost:
+        description: >-
+          Cost of the energy consumed during this chunk in the currency
+          specified in the session price
+        type: object
+        $ref: /price#/Price
```
Dominik-K, line 25: The description is misleading resp. wrong. Actually, the currency should be part of the price, IMO. As its the unit of the price. Yeah, it will lead to more data, but not much as this is an enum.

Dominik-K (commented):
Found this good blog-post for a general inspiration: https://verraes.net/2016/02/type-safety-and-money/

types/price.yaml
```diff
@@ -0,0 +1,159 @@
...
+      - STD
+      - SUR
+      - TJR
+      - TMM
+      - TPE
+      - TRL
+      - UAK
+      - UGS
+      - UGW
+      - USS
```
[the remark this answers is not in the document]
golovasteek: Hi thanks for pointing out! I'll update the list.

What I've found now is [iso](https://www.iso.org/iso-4217-currency-codes.html) refers to [six-group](https://www.six-group.com/en/products-services/financial-information/data-standards.html) where they have an XLS-file with codes.

I don't see any license or copyright note around only:

> We maintain the code lists, update them and make them available online and free of charge.

Does it mean, that we can use it?

interfaces/bank_transaction_summary_provider.yaml
```diff
@@ -0,0 +1,10 @@
+description: >-
+  Provides information of the session that was commited to the bank.
+  This data may be needed for accounting perposes.
```
ju4nlu, line 3:
```suggestion
  This data may be needed for accounting purposes.
```

types/price.yaml
```diff
@@ -0,0 +1,159 @@
+description: >-
+  The definition of the currency and price types.
+types:
+  Currency:
+    description: >-
+      Currency in 3 alphabetical characters according to  ISO 4217.
```
ju4nlu, line 6:
```suggestion
      Currency in 3 alphabetical characters according to ISO 4217.
```

types/session_cost.yaml
```diff
@@ -0,0 +1,79 @@
...
+        items:
+          description: One chunk of the session cost
+          type: object
+          $ref: /session_cost#/SessionCostChunk  
+        default: []
+      status:
+        description: >-
+          Session status. Session can be running or finished. Costs of the
+          running session are not final and can change.
+          Only afeter the session is finished, the session cost can be commited
```
ju4nlu, line 76:
```suggestion
          Only after the session is finished, the session cost can be committed
```

ju4nlu (approved):
Left comments pointing out some typos

types/price.yaml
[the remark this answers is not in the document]
Dominik-K: Yeah, sounds so.

However, thinking again, I'm not sure if it really makes sense to maintain this list at all. Simply having a free-form, 3-letter string could also be enough. Or?

corneliusclaussen:
About the splitting of currency and value into two types: I think it makes sense here where we have an array and to not want to support arrays where every entry has a different currency. In other places it makes sense to have a combined type, e.g.

golovasteek:
> About the splitting of currency and value into two types: I think it makes sense here where we have an array and to not want to support arrays where every entry has a different currency. In other places it makes sense to have a combined type, e.g. price {money_value: int, money_currency: enum EUR} (ignore naming, they are bad). Then we can use money_value and money_curreny separately here but in other places we can use price which always carries both the currenty and the value. Apart from a good solution for this I think only the typos mentioned above are missing before we can merge this.

We have currently two types
  `/price/Currency` - it is an enum.
  `/price/Price` - it is a strut, that represent a sort of fixed-point value.
  
 and the price of 1.23 Eur for single item  can be expressed as:
 
 ```
 price:
     money_value:   # This is a type of /price/Price
        value: 123
        decimals: 2
     currency:  EUR
```

The "invoice" like object would look like the "SessionCost" object, with once currency and array of prices
 ```
 invoice:
     money_value_items:   # This is a type of /price/Price
         - value: 123
            decimals: 2
         - value: 456
            decimals: 2
     currency:  EUR
```

Does it cover your use-cases?

types/session_cost.yaml
golovasteek: So we need few types:
`Currency` - enum
`Price` - numeric value, that represent the amount of money in some currency. It is just an alternative to int/float but for money.

Cost - for single item
Costs - for multiple items.

Cost has both currency and price, probably what you refer as "price".
Costs represent a sort of invoice, where we have few separate Prices per item, and then we sum them up into the total amount. For this the currency is defined once, for entire object. Because we don't want to get accidentally different currencies in one "invoice".


Maybe my naming is bit confusing...
We may think about renaming `Price` to something like `MoneyAmount`, what do you think?

types/price.yaml
```diff
@@ -0,0 +1,159 @@
...
+      - value
+      - decimals
+    properties:
+      value:
+        description: >-
+          The amount of minor units of the currency. 
+        type: integer
+      decimals:
+        description: >-
+          The number of digits after the decimal separator.
```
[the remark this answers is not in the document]
golovasteek: Not really. It is an exponent, that should be applied to the number to get amount in the main currency units.

Or how many digits after the decimal separator are represented by the minor unit.

I've updated description, can you take a look?

types/price.yaml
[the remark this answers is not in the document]
golovasteek: Updated the list.

The question enum vs string is a good question.

I have no strong opinion in the tech call today we decided to try it out with the enum

types/price.yaml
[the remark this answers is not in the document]
agardiol: I think this exponent solution is excellent, as it avoids impossible values and rounding that would be present with floats. In most cases (EUR and CHF for example), the basic unit would be the centime.

types/session_cost.yaml
agardiol: For me the naming is OK and the fact that the currency is common is perfect, because it makes no sense for the amounts to be billed for a session to be in different currencies.

types/price.yaml
agardiol, line 1: This new list is not valid since some values are duplicated. By example:
```
      # 356, Indian Rupee, BHUTAN
      - INR
      # 356, Indian Rupee, INDIA
      - INR
```
The generated file price.hpp see such duplicated data as dual entry in a switch case which results in a build error.

agardiol (changes requested):

corneliusclaussen:
> Does it cover your use-cases?

I would have defined 3 types in the type file:


```
/money/Currency
/modey/Value
/money/Price

Currency is enum as is
Value is fixed point as is

Price:
  currency # of type money/Currency
  value # of type money/Value
```

Then in invoice type structs you can use Currency and Value separately as you described.

But if you have a total price (i.e. single value) you can use the predefined Price struct:

```
my_total_price:
  ref /money/Price
```

It is basically the same as you stated just that a combined type is added in the type file already so that total prices that carry the currency along is one object and it does not have to be passed as e.g. two parameters in the code later on.

An nameing is hard ;-)
But for me a price always includes a unit.

types/price.yaml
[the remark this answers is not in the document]
Dominik-K: Ok, let's see. 

But it seems that already still opens up a box of pandora :see_no_evil: https://github.com/EVerest/everest-core/pull/425#discussion_r1424925633

The ISO-standard gets updated from time to time. And it seems there are non-standard codes commonly used. See https://en.wikipedia.org/wiki/ISO_4217#Non-standard_codes

types/price.yaml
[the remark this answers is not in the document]
Dominik-K: > I've updated description, can you take a look?

Looks good. 

Still what about my added hint "Used only for displaying the price to the user."? Or yours: "should be applied to the number to get amount in the main currency units". It's basically the same that it should serve as a warning to not calculate with the major units, only on the minor units. Perhaps we could add this hint as warning, best at the beginning of the description.

types/session_cost.yaml
Dominik-K: Yeah, naming is hard :smile: 

> We may think about renaming Price to something like MoneyAmount, what do you think?

That's way better. As for me, [what also Cornelius mentioned](https://github.com/EVerest/everest-core/pull/425#issuecomment-1853436127),  a price has a value and a unit (= currency). 
Or we could name it also `MoneyValue` as "Cost is the value of money that has been used up to produce something or deliver a service," https://en.wikipedia.org/wiki/Cost. But that's up to you :smiley: 

Ok, I agree that one session has only one currency. Then it makes sense to save memory, sure.

Actually, then the `decimals` should also be part of the `Currency` (and not the `MoneyAmount`). It would save even an `integer` for each element in the `SessionCost.cost_chunks`-array.

Then `Currency` would become:
```
Currency {
  currency_code // the current `Currency`
  decimals  // the current `Price.decimals`
}
```

agardiol: ...and also speed up chunks sum to get total amount without having to check decimals consistency between chunks !

types/price.yaml
golovasteek: Fixed

types/session_cost.yaml
golovasteek: Applied.

types/money.yaml
```diff
@@ -0,0 +1,515 @@
...
+      # 124, Canadian Dollar, CANADA
+      - CAD
+      # 976, Congolese Franc, CONGO (THE DEMOCRATIC REPUBLIC OF THE)
+      - CDF
+      # 947, WIR Euro, SWITZERLAND
+      - CHE
+      # 756, Swiss Franc, LIECHTENSTEIN
+      - CHF
+      # 756, Swiss Franc, SWITZERLAND
+      - CHF
```
agardiol, line 84:
```
      # 756, Swiss Franc, SWITZERLAND
      # LIECHTENSTEIN
      - CHF
```

types/money.yaml
```diff
@@ -0,0 +1,515 @@
...
+      # 971, Afghani, AFGHANISTAN
+      - AFN
+      # 008, Lek, ALBANIA
+      - ALL
+      # 051, Armenian Dram, ARMENIA
+      - AMD
+      # 532, Netherlands Antillean Guilder, CURAÇAO
+      - ANG
+      # 532, Netherlands Antillean Guilder, SINT MAARTEN (DUTCH PART)
+      - ANG
```
agardiol, line 25:
```
      # 532, Netherlands Antillean Guilder, CURAÇAO
      # SINT MAARTEN (DUTCH PART)
      - ANG
```

types/money.yaml
```diff
@@ -0,0 +1,515 @@
...
+      # 203, Czech Koruna, CZECHIA
+      - CZK
+      # 262, Djibouti Franc, DJIBOUTI
+      - DJF
+      # 208, Danish Krone, DENMARK
+      - DKK
+      # 208, Danish Krone, FAROE ISLANDS (THE)
+      - DKK
+      # 208, Danish Krone, GREENLAND
+      - DKK
```
agardiol, line 114:
```
      # 208, Danish Krone, DENMARK
      # FAROE ISLANDS (THE)
      # GREENLAND
      - DKK
```

types/money.yaml
```diff
@@ -0,0 +1,515 @@
...
+      # 430, Liberian Dollar, LIBERIA
+      - LRD
+      # 426, Loti, LESOTHO
+      - LSL
+      # 434, Libyan Dinar, LIBYA
+      - LYD
+      # 504, Moroccan Dirham, MOROCCO
+      - MAD
+      # 504, Moroccan Dirham, WESTERN SAHARA
+      - MAD
```
agardiol, line 245:
```
      # 504, Moroccan Dirham, MOROCCO
      # WESTERN SAHARA
      - MAD
```

types/money.yaml
```diff
@@ -0,0 +1,515 @@
...
+      # 965, ADB Unit of Account, MEMBER COUNTRIES OF THE AFRICAN DEVELOPMENT BANK GROUP
+      - XUA
+      # 886, Yemeni Rial, YEMEN
+      - YER
+      # 710, Rand, LESOTHO
+      - ZAR
+      # 710, Rand, NAMIBIA
+      - ZAR
+      # 710, Rand, SOUTH AFRICA
+      - ZAR
```
agardiol, line 454:
```
      # 710, Rand, LESOTHO
      # NAMIBIA
      # SOUTH AFRICA
      - ZAR
```

types/money.yaml
```diff
@@ -0,0 +1,515 @@
...
+          For example for Euro, Euro cents express 2 digits after the
+          decimal separator.
+          In other words, the price in the main currency units would be the
+          `value * 10^(-decimals)`.
+        type: integer
+        minimum: 0
+        default: 2
+  Value:
+    description: >-
+      Price is expressed as an amount of minor units of the currency.
```
agardiol, line 486: Cost value is expressed as...

types/session_cost.yaml
```diff
@@ -0,0 +1,79 @@
...
+        type: string
+        format: date-time
+      timestamp_to:
+        description: >-
+          Absolute timestamp for the end of this chunk in RFC3339 UTC format
+        type: string
+        format: date-time
+      cost:
+        description: >-
+          Cost of the energy consumed during this chunk in the currency
```
agardiol, line 22: Cost of the energy consumed (or other things like time, base price, ...) during this chunk in the currency

agardiol (changes requested):
Thanks for the changes.
There are some other currencies to concatenate and minor comments to update

types/money.yaml
Dominik-K: The type should be named `CostValue`. Or `MoneyAmount` or `MoneyValue`, as we [concluded in conversation above](https://github.com/EVerest/everest-core/pull/425#discussion_r1425308295). Also the description should be adapted, as agardiol pointed out.

Or is the `money` from the filename automatically used as namespace? :thinking: But a namespace isn't available in all programming-languages.

Dominik-K (approved):
[LGTM with comments](https://google.github.io/eng-practices/review/reviewer/speed.html#lgtm-with-comments) (in the meaning: I'm sure you'll address my comments.)

It's looking now quite proper :+1:

SirVer:
> Nice work! One additional comment: I think we should use the "prevalidated" flag in the auth token in this case to indicate that the tokenprovider already reserved the money and the transaction is approved without the need for further validators

This is 2c from our actual using this now in our new product: We found that the validator that talks to the backend (think OCPP) still needed to do a bunch of work for such a session, and hence we used a non prevalidated token so that this goes through the same flow. So I think this should be a configurable option or another place must be found for hooking in such action.

golovasteek:
I've addressed all the comments.

And I've added a small test for the currencies list.

The current test infrastructure seems to be tailored for integration test and builds specific environment for that. This is not needed for basic yaml sanity checks, so I've added a very basic test jobs, that runs yaml checks serparatelly in the basic environment.

golovasteek:
> > Nice work! One additional comment: I think we should use the "prevalidated" flag in the auth token in this case to indicate that the tokenprovider already reserved the money and the transaction is approved without the need for further validators
> 
> This is 2c from our actual using this now in our new product: We found that the validator that talks to the backend (think OCPP) still needed to do a bunch of work for such a session, and hence we used a non prevalidated token so that this goes through the same flow. So I think this should be a configurable option or another place must be found for hooking in such action.

Yes, this should be configuration parameter of the module, but this is not covered in this MR

PR #427, opened 2023-11-13
Add missing find_package for CURL

hikinggrass:


CMakeLists.txt
```diff
@@ -68,6 +68,7 @@ else()
     find_package(slac REQUIRED)
 
     find_package(pugixml REQUIRED)
+    find_package(CURL REQUIRED)
```
a-w50, line 71: Didn't we need a recent version?  find_package also supports version requirements.

a-w50 (approved):

CMakeLists.txt
hikinggrass: Added the minimum version we need

a-w50 (approved):

PR #434, opened 2023-11-15
Extend example by Py modules

andistorm:
* Add module PyExample
* Add module PyExampleUser
* Extend config config-example.yaml with new modules


andistorm:
needs https://github.com/EVerest/everest-framework/pull/122

hikinggrass (approved):

PR #437, opened 2023-11-17
Add tests that simply run all (SIL) configs in everest-core/config

hikinggrass:
This should help in finding regressions that need configuration changes or detect broken configs

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -28,7 +28,7 @@ Go to your "everest-core/**tests**" folder and execute *pytest*:
 
 ```bash
 cd ~/checkout/everest-workspace/everest-core/tests
-pytest --everest-prefix ../build/dist ./core_tests/startup_tests.py
+pytest --everest-prefix ../build/dist core_tests/*.py framework_tests/*.py
 ```
 
 After execution a "results.xml" file should be available in the "everest-core/**tests**" folder, which details the current test results.
--- a/tests/README.md
+++ b/tests/README.md
@@ -28,7 +28,7 @@ Go to your "everest-core/**tests**" folder and execute *pytest*:
 
 ```bash
 cd ~/checkout/everest-workspace/everest-core/tests
-pytest-3 -s ./core_tests/startup_tests.py --path .. --junitxml=results.xml
+pytest --everest-prefix ../build/dist ./core_tests/startup_tests.py
 ```
 
 After execution a "results.xml" file should be available in the "everest-core/**tests**" folder, which details the current test results.
--- a/tests/core_tests/config_tests.py
+++ b/tests/core_tests/config_tests.py
@@ -7,7 +7,6 @@ from copy import deepcopy
 import logging
 from pathlib import Path
 import pytest
-import subprocess
 from tempfile import mkdtemp
 from typing import Dict
 
@@ -75,7 +74,6 @@ async def test_start_config_sil_gen_pm(request, everest_core: EverestCore):
     config_adaptation = request.node.get_closest_marker(
         'everest_config_adaptations').args[0]
     logging.info(f"config_adaptation: {config_adaptation.serial_port_0}")
-    # serial_port_process = subprocess.Popen(['socat', 'PTY,link=/tmp/serial_port', 'PTY,link=/tmp/serial_port_1'])
     everest_core.start()
 
 
--- a/tests/core_tests/config_tests.py
+++ b/tests/core_tests/config_tests.py
@@ -1,7 +1,7 @@
 
 #!/usr/bin/env python3
 # SPDX-License-Identifier: Apache-2.0
-# Copyright 2020 - 2023 Pionix GmbH and Contributors to EVerest
+# Copyright Pionix GmbH and Contributors to EVerest
 
 from copy import deepcopy
 import logging
--- a/tests/core_tests/config_tests.py
+++ b/tests/core_tests/config_tests.py
@@ -5,14 +5,10 @@
 
 import logging
 import pytest
-import time
-import threading
-import queue
 
 from everest.testing.core_utils.fixtures import *
-from everest.testing.core_utils.everest_core import EverestCore, Requirement
+from everest.testing.core_utils.everest_core import EverestCore
 
-from everest.framework import Module, RuntimeSession
 
 @pytest.mark.everest_core_config('config-example.yaml')
 @pytest.mark.asyncio
--- a/tests/core_tests/config_tests.py
+++ b/tests/core_tests/config_tests.py
@@ -19,46 +19,12 @@ from everest.testing.core_utils.configuration.everest_configuration_visitors.eve
     EverestConfigAdjustmentVisitor
 
 
-@pytest.mark.everest_core_config('config-example.yaml')
-@pytest.mark.asyncio
-async def test_start_config_example(everest_core: EverestCore):
-    logging.info(">>>>>>>>> test_start_config_example <<<<<<<<<")
-    everest_core.start()
-
-
-@pytest.mark.everest_core_config('config-sil-dc-sae-v2g.yaml')
-@pytest.mark.asyncio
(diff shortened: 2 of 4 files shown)


tests/core_tests/config_tests.py
```diff
@@ -0,0 +1,129 @@
...
+
+from everest.testing.core_utils.configuration.everest_configuration_visitors.everest_configuration_visitor import \
+    EverestConfigAdjustmentVisitor
+
+
+@pytest.mark.everest_core_config('config-example.yaml')
+@pytest.mark.asyncio
+async def test_start_config_example(everest_core: EverestCore):
+    logging.info(">>>>>>>>> test_start_config_example <<<<<<<<<")
+    everest_core.start()
```
Dominik-K, line 25: Can you generalize this test and walk the directory for all `*.yaml` (except the one test which needs special handling)? That way every new config will automatically included, too. Besides removing a lot of boilerplate code.

tests/core_tests/config_tests.py
```diff
@@ -0,0 +1,129 @@
+
+#!/usr/bin/env python3
+# SPDX-License-Identifier: Apache-2.0
+# Copyright 2020 - 2023 Pionix GmbH and Contributors to EVerest
```
Dominik-K, line 4: BTW: Is it correct to backdate the copyright? :shrug:

.ci/e2e/scripts/tests.sh
Dominik-K, line 1: How about moving these scripts below the `.github`-directory? One directory less on root which helps to get a better overview. 

TBD in another PR, sure (it just triggered the question here).

Dominik-K (changes requested):
The multiplied, same-functionality code should be generalized (see comment).

Dominik-K:
How long will these tests take?  Tried to answer myself but the [tests are currently failing on initialisation](https://github.com/EVerest/everest-core/actions/runs/6928297304/job/18843815039#step:11:29).

If it takes too long, one option would be to run them in parallel. But unfortunately, GitHub Actions doesn't support this out of the box, AFAIK. [CircleCI has a neat helper for it](https://circleci.com/docs/parallelism-faster-jobs/). But perhaps there's a good tool for GitHub Actions, too.

Dominik-K:
Sorry, clicked wrong :see_no_evil:

hikinggrass:
> How long will these tests take? Tried to answer myself but the [tests are currently failing on initialisation](https://github.com/EVerest/everest-core/actions/runs/6928297304/job/18843815039#step:11:29).
> 
> If it takes too long, one option would be to run them in parallel. But unfortunately, GitHub Actions doesn't support this out of the box, AFAIK. [CircleCI has a neat helper for it](https://circleci.com/docs/parallelism-faster-jobs/). But perhaps there's a good tool for GitHub Actions, too.

Let's see how long they take, them failing right now is already being worked on (basically we use a version of everest-testing that's too old in the ci tests)

Dominik-K:
> If it takes too long, one option would be to run them in parallel. But unfortunately, GitHub Actions doesn't support this out of the box, AFAIK. [CircleCI has a neat helper for it](https://circleci.com/docs/parallelism-faster-jobs/). But perhaps there's a good tool for GitHub Actions, too.

Yeah, there are tools for GitHub Actions. Was quite simple to find searching for "split tests":
* https://imhoff.blog/posts/parallelizing-jest-with-github-actions  (mentioning an [undocumented](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategymatrix) field `jobs.<job-id>.strategy.matrix.test-files`
* Split done by test-timing: https://github.com/leonid-shevtsov/split_tests

tests/core_tests/config_tests.py
hikinggrass: removed

tests/core_tests/config_tests.py
hikinggrass: That was a surprisingly challenging thing to do with pytest, but I have a, arguably a bit hacky, solution implemented

hikinggrass:
All 15 tests that are implemented right now run sequentially and take ~67 seconds, so <5s/test. Codacy for example usually takes much longer than that, so I think this is negligible for now.

Dominik-K (approved):
[LGTM with comments](https://google.github.io/eng-practices/review/reviewer/speed.html#lgtm-with-comments) (in the meaning: I'm sure you'll address my comments.)

tests/core_tests/config_tests.py
hikinggrass: Looking at the codebase there are pretty much two schools of thought here:

1. Not setting it at all (which is the default for modules created by ev-cli module create btw.)
2. Setting it with a date range starting from project inception to today

I would lean towards 1) here because the date ranges are almost never correctly updated in PRs

tests/core_tests/config_tests.py
```diff
@@ -0,0 +1,61 @@
...
+
+        return EverestEnvironmentCoreConfiguration(
+            everest_core_path=everest_prefix,
+            template_everest_config_path=everest_config_path,
+        )
+
+    @pytest.mark.asyncio
+    async def test_config(self, everest_core: EverestCore):
+        logging.info(">>>>>>>>> test_config <<<<<<<<<")
+        everest_core.start()
```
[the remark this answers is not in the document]
hikinggrass: at the moment everest_core.start() just times out after a few seconds (30s I think)

tests/core_tests/config_tests.py
```diff
@@ -0,0 +1,61 @@
...
+from copy import deepcopy
+import logging
+import os
+from pathlib import Path
+import pty
+import pytest
+from tempfile import mkdtemp
+from typing import Dict
+
+from everest.testing.core_utils.fixtures import *
```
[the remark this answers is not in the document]
hikinggrass: Normally I would agree, however in this case the fixtures themselves use quite a lot of other fixtures internally that you would have to import here as well. That list is quite long and I'm not sure it makes it much clearer for the next developer adding new tests

hikinggrass:
For this PR to work we need to have these merged:
https://github.com/EVerest/everest-core/pull/475
https://github.com/EVerest/everest-ci/pull/12

tests/README.md
```diff
@@ -28,7 +28,7 @@ Go to your "everest-core/**tests**" folder and execute *pytest*:
 
 ```bash
 cd ~/checkout/everest-workspace/everest-core/tests
-pytest-3 -s ./core_tests/startup_tests.py --path .. --junitxml=results.xml
+pytest --everest-prefix ../build/dist ./core_tests/startup_tests.py
```
andistorm, line 31:
```suggestion
pytest --everest-prefix ../build/dist core_tests/*.py framework_tests/*.py
```

CMakeLists.txt
```diff
@@ -130,11 +130,13 @@ if(BUILD_TESTING)
...
-        EXECUTABLE test_config
-        DEPENDENCIES test_config everest
-    )
+    if (LCOV_PATH)
+        setup_target_for_coverage_lcov(
+            NAME lcov_coverage
+            EXECUTABLE test_config
+            DEPENDENCIES test_config everest
+        )
+    endif()
```
andistorm, line 139: Why is this change needed?

tests/README.md
andistorm: This is what we are doing in https://github.com/EVerest/everest-core/blob/main/tests/run-testing.sh and https://github.com/EVerest/everest-core/blob/main/.ci/e2e/scripts/tests.sh

CMakeLists.txt
hikinggrass: I removed the change, maybe that's something for another PR. For these integration tests we do not need lcov (and gcovr)

andistorm (approved):
:1st_place_medal:

PR #440, opened 2023-11-20
EvseManager: Add support for powermeter transactions

dorezyuk:
This pr adds support for starting and stopping transactions in the power meter from within the EvseManager. The transactions are not started if no power meter is registered with the evse manager. If the manager has a power meter registered and it fails to start a transaction on the power meter, the evse manager goes into an error state - the rationale is that in this case we can't do any billing and therefore are not going to start charging

### DIFF
--- a/modules/EvseManager/Charger.cpp
+++ b/modules/EvseManager/Charger.cpp
@@ -10,6 +10,7 @@
 
 #include "Charger.hpp"
 
+#include <generated/types/powermeter.hpp>
 #include <math.h>
 #include <string.h>
 
@@ -17,11 +18,11 @@
 
 namespace module {
 
-Charger::Charger(const std::unique_ptr<board_support_ACIntf>& r_bsp, const std::string& connector_type) :
-    r_bsp(r_bsp), connector_type(connector_type) {
-
+Charger::Charger(const std::unique_ptr<board_support_ACIntf>& r_bsp,
+                 const std::vector<std::unique_ptr<powermeterIntf>>& r_powermeter_billing,
+                 const std::string& connector_type, const std::string& evse_id) :
+    r_bsp(r_bsp), r_powermeter_billing(r_powermeter_billing), connector_type(connector_type), evse_id{evse_id} {
     connectorEnabled = true;
-
     maxCurrent = 6.0;
     maxCurrentCable = r_bsp->call_read_pp_ampacity();
     authorized = false;
@@ -220,7 +221,8 @@ void Charger::runStateMachine() {
 
                 // If we are restarting, the transaction may already be active
                 if (!transactionActive()) {
-                    startTransaction();
+                    if (!startTransaction())
+                        break;
                 }
 
                 const EvseState targetState(EvseState::PrepareCharging);
@@ -303,7 +305,8 @@ void Charger::runStateMachine() {
                 }
             } else if (AuthorizedPnC()) {
 
-                startTransaction();
+                if (!startTransaction())
+                    break;
 
                 const EvseState targetState(EvseState::PrepareCharging);
 
@@ -1015,6 +1018,7 @@ bool Charger::cancelTransaction(const types::evse_manager::StopTransactionReques
         if (request.id_tag) {
             stop_transaction_id_tag = request.id_tag.value();
         }
+
         signalEvent(types::evse_manager::SessionEventEnum::ChargingFinished);
         signalEvent(types::evse_manager::SessionEventEnum::TransactionFinished);
         return true;
@@ -1050,17 +1054,38 @@ void Charger::stopSession() {
     signalEvent(types::evse_manager::SessionEventEnum::SessionFinished);
 }
 
-void Charger::startTransaction() {
+bool Charger::startTransaction() {
     std::lock_guard<std::recursive_mutex> lock(stateMutex);
     stop_transaction_id_tag.clear();
     transaction_active = true;
+
+    // TODO(ddo) client_id, tariff_id, cable_id and user_data are currently not
+    // set.
+    const types::powermeter::TransactionReq req{evse_id, id_token.id_token, "", 0, 0, ""};
+    for (const auto& meter : r_powermeter_billing) {
+        const auto response = meter->call_start_transaction(req);
+        // If we want to start the session but fail, we stop the charging.
+        if (response.status == types::powermeter::TransactionRequestStatus::UNEXPECTED_ERROR) {
+            currentState = EvseState::Error;
+            return false;
+        }
+    }
+
     signalEvent(types::evse_manager::SessionEventEnum::TransactionStarted);
+    return true;
 }
 
 void Charger::stopTransaction() {
     std::lock_guard<std::recursive_mutex> lock(stateMutex);
     transaction_active = false;
     last_stop_transaction_reason = types::evse_manager::StopTransactionReason::EVDisconnected;
+
+    const std::string transaction_id{};
+    for (const auto& meter : r_powermeter_billing) {
+        // TODO publish an error together with TransactionFinished and ChargingFinished.
+        const auto _ = meter->call_stop_transaction(transaction_id);
+    }
+
     signalEvent(types::evse_manager::SessionEventEnum::ChargingFinished);
     signalEvent(types::evse_manager::SessionEventEnum::TransactionFinished);
 }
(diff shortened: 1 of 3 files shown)


dorezyuk:
Not sure regarding the static analysis - how do I get rid of the false positives @a-w50 ?

a-w50:
> Not sure regarding the static analysis - how do I get rid of the false positives @a-w50 ?

I'm not sure if it is possible at all.  Honestly the static code analysis doesn't really make sense for files that use auto-generated code because it cannot really reason about it.  Concerning the unused variables, just ignore them.

modules/EvseManager/Charger.hpp
```diff
@@ -201,7 +214,9 @@ class Charger {
     Everest::Thread mainThreadHandle;
 
     const std::unique_ptr<board_support_ACIntf>& r_bsp;
+    const std::vector<std::unique_ptr<powermeterIntf>>& r_powermeter_billing;
     const std::string& connector_type;
+    const std::string evse_id;
```
corneliusclaussen, line 219: could probably be a const&

modules/EvseManager/Charger.cpp
```diff
@@ -1050,17 +1054,46 @@ void Charger::stopSession() {
     signalEvent(types::evse_manager::SessionEventEnum::SessionFinished);
 }
 
-void Charger::startTransaction() {
+bool Charger::startTransaction() {
     std::lock_guard<std::recursive_mutex> lock(stateMutex);
     stop_transaction_id_tag.clear();
     transaction_active = true;
+
+    // TODO(ddo) client_id, tariff_id, cable_id and user_data are currently not
```
corneliusclaussen, line 1062: client_id we already get from Auth   via this command:

authorize_response:
    arguments:
      provided_token:

cable_id and tarrif_id I think we do not have yet somewhere

corneliusclaussen:
"If the manager has a power meter registered and it fails to start a transaction on the power meter, the evse manager goes into an error state": Clarifying the description further: If the power meter does not support transactions, that is also ok and the charger will not go to an error state if I read the code correctly. That is also what we want, because in many (if not most) usecases a power meter is connected and used but not in a German Eichrecht  Transaction based way.

corneliusclaussen:
Also note: this will break with the bsp refactoring as the error state has been removed. Problem was that multiple errors can exist at the same time and the state machine can still do state transitions even if errors are present. So we will need to add proper error definition for that in the bsp refactoring.

modules/EvseManager/Charger.cpp
```diff
@@ -1050,17 +1054,46 @@ void Charger::stopSession() {
...
     std::lock_guard<std::recursive_mutex> lock(stateMutex);
     transaction_active = false;
     last_stop_transaction_reason = types::evse_manager::StopTransactionReason::EVDisconnected;
+
+    const std::string transaction_id{};
+
+    for (const auto& meter : r_powermeter_billing) {
+        const auto response = meter->call_stop_transaction(transaction_id);
+        // If we fail to stop the transaction, we ignore since there is no
+        // path to recovery. Its also not clear what to do
```
corneliusclaussen, line 1090: I think we should at least print an EVLOG_error here to not make it a complete silent failure. With the bsp refactoring I will probably add a real error reporting here. As you said there is nothing we can do about it but we can at least log it.

modules/EvseManager/Charger.cpp
```diff
@@ -1050,17 +1054,46 @@ void Charger::stopSession() {
...
     std::lock_guard<std::recursive_mutex> lock(stateMutex);
     stop_transaction_id_tag.clear();
     transaction_active = true;
+
+    // TODO(ddo) client_id, tariff_id, cable_id and user_data are currently not
+    // set.
+    const types::powermeter::TransactionReq req{evse_id, id_token.id_token, "", 0, 0, ""};
+    for (const auto& meter : r_powermeter_billing) {
+        const auto response = meter->call_start_transaction(req);
+        // If we want to start the session but fail, we stop the charging.
```
SirVer, line 1067: nit: explain why (you state it in the MR description). Also, print an error.

modules/EvseManager/Charger.cpp
SirVer: +1

modules/EvseManager/Charger.hpp
```diff
@@ -191,6 +198,12 @@ class Charger {
     void set_hlc_charging_active();
     void set_hlc_allow_close_contactor(bool on);
 
+    std::optional<std::string> getOcmfData() {
```
SirVer, line 201: Add documentation, for example it is surprising that this is not idempotent to me.

Also I am not a fan of having some of the implementation in the header and some in the .cc file. Please put the implementation in the .cc file too, so that all logic is together.

modules/EvseManager/Charger.hpp
SirVer: It is not obvious to me what guarantees `evse_id` outliving the Charger object in all circumstances. Hence I think it is better to have a non-reference and guarantee ownership here like it is currently done.

modules/EvseManager/EvseManager.cpp
```diff
@@ -748,6 +750,7 @@ void EvseManager::ready() {
     });
 
     charger->signalEvent.connect([this](types::evse_manager::SessionEventEnum s) {
+        // TODO(ddo) this is not publishing the session event???
```
SirVer, line 753: That is a statement, not a TODO. What are you concerned about?

SirVer (approved):
a few nits, but overall looks good to me.

modules/EvseManager/Charger.cpp
dorezyuk: sorry I don't get it. None of the messages defines currently `client_id` - and the function call does not sound like it helps here 
>   authorize_response:
    description: >-
      Reports the result of an authorization request to the EvseManager. 
      Contains the provided_token for which authorization was requested and
      the validation_result

can you elaborate?

dorezyuk:
>"If the manager has a power meter registered and it fails to start a transaction on the power meter, the evse manager goes into an error state": Clarifying the description further: If the power meter does not support transactions, that is also ok and the charger will not go to an error state if I read the code correctly. That is also what we want, because in many (if not most) usecases a power meter is connected and used but not in a German Eichrecht Transaction based way.

Yes, if the transaction is not supported we won't emit an error and just carry on

modules/EvseManager/Charger.hpp
```diff
@@ -191,6 +198,12 @@ class Charger {
     void set_hlc_charging_active();
     void set_hlc_allow_close_contactor(bool on);
 
+    /// @brief Returns the OCMF data.
+    ///
+    /// The data is generated when stopping the transaction. The call resets the
+    /// internal variable and is thus not idem potent.
```
SirVer, line 204: nit: idempotent is one word.

SirVer (approved):
one more nit, otherwise lgtm.

modules/EvseManager/Charger.cpp
corneliusclaussen: Ah you are using the id_token.id_token already which is the same as client_id. So we just need to remove "client_id" from the TODO comment as it is already done.

modules/EvseManager/Charger.cpp
```diff
@@ -1050,21 +1054,62 @@ void Charger::stopSession() {
...
+    // TODO(ddo) client_id, tariff_id, cable_id and user_data are currently not
+    // set.
+    const types::powermeter::TransactionReq req{evse_id, id_token.id_token, "", 0, 0, ""};
+    for (const auto& meter : r_powermeter_billing) {
+        const auto response = meter->call_start_transaction(req);
+        // If we want to start the session but fail, we stop the charging since
+        // we can't bill the customer.
+        if (response.status == types::powermeter::TransactionRequestStatus::UNEXPECTED_ERROR) {
+            EVLOG_error << "Failed to start a transaction on the power meter " << response.error.value_or("");
+            currentState = EvseState::Error;
```
corneliusclaussen, line 1071: Now that the BSP refactoring was merged to main the EvseState::Error does not exist anymore. I will add an error for that in a PR to core.

corneliusclaussen: https://github.com/EVerest/everest-core/pull/488

This contains a new error for this case. You can use:

error_handling->raise_powermeter_transaction_start_failed_error("some description");

to raise the error instead of setting currentState. It will be cleared on unplug, I added that already to PR488.

corneliusclaussen:
Converted to draft as it does not work anymore with the latest main

golovasteek:
I've re created the merge request after merging to the current release.

@dorezyuk can you close this one.


Let's continue in the https://github.com/EVerest/everest-core/pull/573

PR #441, opened 2023-11-20
SerialCommHub: support longer transactions and chunking

dorezyuk:
Pr adds support for maximum packet sizes to the SerialCommHub. The [wikipedia](https://en.wikipedia.org/wiki/Modbus) says that the the packet size is between 253 and 256 bytes

> PDU max size is 253 bytes. ADU max size on RS232/RS485 network is 256 bytes, and with TCP is 260 bytes

This pr makes the packet size configurable. If the payload exceeds the packet size we will read/write it in chunks

### DIFF
--- a/modules/SerialCommHub/main/serial_communication_hubImpl.cpp
+++ b/modules/SerialCommHub/main/serial_communication_hubImpl.cpp
@@ -66,7 +66,7 @@ serial_communication_hubImpl::handle_modbus_read_holding_registers(int& target_d
             //                           (uint16_t)first_register_address, (uint16_t)num_registers_to_read);
 
             response = modbus.txrx(target_device_id, tiny_modbus::FunctionCode::READ_MULTIPLE_HOLDING_REGISTERS,
-                                   first_register_address, num_registers_to_read);
+                                   first_register_address, num_registers_to_read, config.max_packet_size);
             if (response.size() > 0) {
                 break;
             }
@@ -102,7 +102,7 @@ serial_communication_hubImpl::handle_modbus_read_input_registers(int& target_dev
             //                           (uint16_t)first_register_address, (uint16_t)num_registers_to_read);
 
             response = modbus.txrx(target_device_id, tiny_modbus::FunctionCode::READ_INPUT_REGISTERS,
-                                   first_register_address, num_registers_to_read);
+                                   first_register_address, num_registers_to_read, config.max_packet_size);
             if (response.size() > 0) {
                 break;
             }
@@ -140,7 +140,7 @@ types::serial_comm_hub_requests::StatusCodeEnum serial_communication_hubImpl::ha
                                        (uint16_t)data.size());
 
             response = modbus.txrx(target_device_id, tiny_modbus::FunctionCode::WRITE_MULTIPLE_HOLDING_REGISTERS,
-                                   first_register_address, data.size(), true, data);
+                                   first_register_address, data.size(), config.max_packet_size, true, data);
             if (response.size() > 0) {
                 break;
             }
--- a/modules/SerialCommHub/main/serial_communication_hubImpl.cpp
+++ b/modules/SerialCommHub/main/serial_communication_hubImpl.cpp
@@ -63,9 +63,9 @@ serial_communication_hubImpl::handle_modbus_read_holding_registers(int& target_d
         auto retry_counter = this->num_resends_on_error;
         while (retry_counter > 0) {
 
-            // EVLOG_info << fmt::format("Try {} Call modbus_client->read_holding_register(id {} addr {} len {})",
-            //                           (int)retry_counter, (uint8_t)target_device_id,
-            //                           (uint16_t)first_register_address, (uint16_t)num_registers_to_read);
+            EVLOG_debug << fmt::format("Try {} Call modbus_client->read_holding_register(id {} addr {} len {})",
+                                       (int)retry_counter, (uint8_t)target_device_id, (uint16_t)first_register_address,
+                                       (uint16_t)num_registers_to_read);
 
             response = modbus.txrx(target_device_id, tiny_modbus::FunctionCode::READ_MULTIPLE_HOLDING_REGISTERS,
                                    first_register_address, num_registers_to_read, config.max_packet_size);
@@ -99,9 +99,9 @@ serial_communication_hubImpl::handle_modbus_read_input_registers(int& target_dev
         uint8_t retry_counter{this->num_resends_on_error};
         while (retry_counter-- > 0) {
 
-            // EVLOG_info << fmt::format("Try {} Call modbus_client->read_input_register(id {} addr {} len {})",
-            //                           (int)retry_counter, (uint8_t)target_device_id,
-            //                           (uint16_t)first_register_address, (uint16_t)num_registers_to_read);
+            EVLOG_debug << fmt::format("Try {} Call modbus_client->read_input_register(id {} addr {} len {})",
+                                       (int)retry_counter, (uint8_t)target_device_id, (uint16_t)first_register_address,
+                                       (uint16_t)num_registers_to_read);
 
             response = modbus.txrx(target_device_id, tiny_modbus::FunctionCode::READ_INPUT_REGISTERS,
                                    first_register_address, num_registers_to_read, config.max_packet_size);
--- a/modules/SerialCommHub/main/serial_communication_hubImpl.cpp
+++ b/modules/SerialCommHub/main/serial_communication_hubImpl.cpp
@@ -31,6 +31,7 @@ static std::vector<int> vector_to_int(const std::vector<uint16_t>& response) {
 // Implementation
 
 void serial_communication_hubImpl::init() {
+    using namespace std::chrono;
     Everest::GpioSettings rxtx_gpio_settings;
 
     rxtx_gpio_settings.chip_name = config.rxtx_gpio_chip;
@@ -38,7 +39,8 @@ void serial_communication_hubImpl::init() {
     rxtx_gpio_settings.inverted = config.rxtx_gpio_tx_high;
 
     if (!modbus.open_device(config.serial_port, config.baudrate, config.ignore_echo, rxtx_gpio_settings,
-                            static_cast<tiny_modbus::Parity>(config.parity))) {
+                            static_cast<tiny_modbus::Parity>(config.parity), milliseconds(config.initial_timeout_ms),
+                            milliseconds(config.within_message_timeout_ms))) {
         EVLOG_AND_THROW(Everest::EverestConfigError(fmt::format("Cannot open serial port {}.", config.serial_port)));
     }
 }
--- a/modules/SerialCommHub/manifest.yaml
+++ b/modules/SerialCommHub/manifest.yaml
@@ -42,6 +42,14 @@ provides:
         minimum: 0
         maximum: 65536
         default: 256
+      initial_timeout_ms:
+        description: Timeout in ms for the first packet.
+        type: integer
+        default: 500
+      within_message_timeout_ms:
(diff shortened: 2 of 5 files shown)


modules/SerialCommHub/tiny_modbus_rtu.cpp
```diff
@@ -263,18 +269,19 @@ bool TinyModbusRTU::open_device(const std::string& device, int _baud, bool _igno
 
 int TinyModbusRTU::read_reply(uint8_t* rxbuf, int rxbuf_len) {
     struct timeval timeout;
-    timeout.tv_sec = 0;
-    timeout.tv_usec = MODBUS_RX_INITIAL_TIMEOUT_MS * 1000; // 500msec intial timeout until device responds
+    timeout.tv_sec = 6;
```
barsnick, line 272: Why an additional 6 seconds?

dorezyuk: We run into timeouts when waiting for larger datablobs over the modbus... We may do this as a parameter

barsnick: Okay - while going from 500 to 6500 ms seems extreme.
Please adapt the misleading comment one line below my remark.

corneliusclaussen: I would also be in favor of making this a config option with a default less then the 6 seconds. This will create problems on some systems if the whole bus hangs that long.

SirVer: +1 for making this configurable. It is a Qwello problem that we need 6 seconds.

modules/SerialCommHub/manifest.yaml
```diff
@@ -36,6 +36,12 @@ provides:
         description: GPIO direction, false means low for TX, true means high for TX
         type: boolean
         default: false
+      max_packet_size:
+        description: Maximum size of a packet to read/write in bytes. Payload exceeding the size will be chunked.
```
SirVer, line 40: You state in your MR description that the limits are < 253, why are you using a default of 256 and why do you allow a maximum of 65k? I think the limits from wikipedia should be in the description and you should explain the max and mins too.

modules/SerialCommHub/tiny_modbus_rtu.cpp
```diff
@@ -174,6 +177,9 @@ static std::vector<uint16_t> decode_reply(const uint8_t* buf, int len, uint8_t e
         case 0x0B:
             EVLOG_error << "Modbus exception: Gateway target device failed to respond";
             break;
+        default: {
```
SirVer, line 180: nit: EVerest does not seem to use blocks in the case statements by default, so remove them if you do not declare a variable.

modules/SerialCommHub/tiny_modbus_rtu.cpp
```diff
@@ -296,13 +303,51 @@ int TinyModbusRTU::read_reply(uint8_t* rxbuf, int rxbuf_len) {
...
+                                          uint16_t first_register_address, uint16_t register_quantity,
+                                          uint16_t max_packet_size, bool wait_for_reply,
+                                          std::vector<uint16_t> request) {
+    // This only supports chunking of the read-requests.
+    std::vector<uint16_t> out;
+
+    const uint16_t register_chunk = (max_packet_size - MODBUS_MIN_REPLY_SIZE) / 2;
+    size_t written_elements = 0;
+    while (register_quantity) {
+
```
SirVer, line 316: remove empty line. did you run clang-format?

dorezyuk: yes - but can remove it

modules/SerialCommHub/tiny_modbus_rtu.cpp
dorezyuk: removed

modules/SerialCommHub/manifest.yaml
dorezyuk: well I understand it that the pdu is 253 - the apu (which is what we send --> pdu with check sum and address)  is 256 - so I've defaulted to it. Maybe we should rename the variable to max_adu_size? Or we let the user specify the maximum payload size and treat the checksum as implementation detail - I'm fine with both. 

Regarding the max - I reasoned that there might be devices which can accept larger packets...

SirVer: Your explanation makes sense, but it should be here in a comment somewhere.

modules/SerialCommHub/tiny_modbus_rtu.cpp
```diff
@@ -262,19 +271,28 @@ bool TinyModbusRTU::open_device(const std::string& device, int _baud, bool _igno
 }
 
 int TinyModbusRTU::read_reply(uint8_t* rxbuf, int rxbuf_len) {
-    struct timeval timeout;
-    timeout.tv_sec = 0;
-    timeout.tv_usec = MODBUS_RX_INITIAL_TIMEOUT_MS * 1000; // 500msec intial timeout until device responds
+    // Lambda to convert std::chrono to timeval.
+    auto to_timeval = [](const auto& time) {
```
SirVer, line 275: This should be a stand alone function in the anonymous namespace in this file. It does not capture anything from its environment.

modules/SerialCommHub/tiny_modbus_rtu.hpp
```diff
@@ -57,9 +55,14 @@ class TinyModbusRTU {
     ~TinyModbusRTU();
 
     bool open_device(const std::string& device, int baud, bool ignore_echo,
-                     const Everest::GpioSettings& rxtx_gpio_settings, const Parity parity);
+                     const Everest::GpioSettings& rxtx_gpio_settings, const Parity parity,
+                     std::chrono::milliseconds initial_timeout, std::chrono::milliseconds within_message_timeout);
+    std::vector<uint16_t> txrx_impl(uint8_t device_address, FunctionCode function, uint16_t first_register_address,
```
SirVer, line 60: this function should be private now, right?

modules/SerialCommHub/tiny_modbus_rtu.cpp
dorezyuk: I disagree - We can only use auto as a function argument in c++20. The the extracted function therefore would either need to pin down the exact chrono type (milliseconds) making the code slightly less generic or be a template function making the code less readable

modules/SerialCommHub/main/serial_communication_hubImpl.cpp
```diff
@@ -97,12 +99,12 @@ serial_communication_hubImpl::handle_modbus_read_input_registers(int& target_dev
         uint8_t retry_counter{this->num_resends_on_error};
         while (retry_counter-- > 0) {
 
-            // EVLOG_info << fmt::format("Try {} Call modbus_client->read_input_register(id {} addr {} len {})",
-            //                           (int)retry_counter, (uint8_t)target_device_id,
-            //                           (uint16_t)first_register_address, (uint16_t)num_registers_to_read);
+            EVLOG_debug << fmt::format("Try {} Call modbus_client->read_input_register(id {} addr {} len {})",
+                                       (int)retry_counter, (uint8_t)target_device_id, (uint16_t)first_register_address,
+                                       (uint16_t)num_registers_to_read);
```
corneliusclaussen, line 104: Debug needs to be removed before merging

modules/SerialCommHub/tiny_modbus_rtu.cpp
```diff
@@ -289,20 +307,57 @@ int TinyModbusRTU::read_reply(uint8_t* rxbuf, int rxbuf_len) {
             int bytes_read = read(fd, rxbuf + bytes_read_total, rxbuf_len - bytes_read_total);
             if (bytes_read > 0) {
                 bytes_read_total += bytes_read;
-                // EVLOG_info << "RECVD: " << hexdump(rxbuf, bytes_read_total);
+                EVLOG_debug << "RECVD: " << hexdump(rxbuf, bytes_read_total);
```
corneliusclaussen, line 310: Remove debug

modules/SerialCommHub/tiny_modbus_rtu.cpp
```diff
@@ -331,7 +386,7 @@ std::vector<uint16_t> TinyModbusRTU::txrx(uint8_t device_address, FunctionCode f
         // set checksum in the last 2 bytes
         append_checksum(req.get(), req_len);
 
-        // EVLOG_info << "SEND: " << hexdump(req.get(), req_len);
+        EVLOG_debug << "SEND: " << hexdump(req.get(), req_len);
```
corneliusclaussen, line 389: Remove debug

modules/SerialCommHub/tiny_modbus_rtu.cpp
```diff
@@ -289,20 +307,57 @@ int TinyModbusRTU::read_reply(uint8_t* rxbuf, int rxbuf_len) {
...
 }
 
+std::vector<uint16_t> TinyModbusRTU::txrx(uint8_t device_address, FunctionCode function,
+                                          uint16_t first_register_address, uint16_t register_quantity,
+                                          uint16_t max_packet_size, bool wait_for_reply,
+                                          std::vector<uint16_t> request) {
+    // This only supports chunking of the read-requests.
+    std::vector<uint16_t> out;
+
+    const uint16_t register_chunk = (max_packet_size - MODBUS_MIN_REPLY_SIZE) / 2;
```
corneliusclaussen, line 324: If register_chunk is zero because max_packet_size is < MODBUS_MIN_REPLY_SIZE this loop never ends. Add input validation here and increase the min setting for max_packet_size in manifest to > MIN_REPLY_SIZE

modules/SerialCommHub/tiny_modbus_rtu.hpp
```diff
@@ -57,18 +55,27 @@ class TinyModbusRTU {
...
+
+    std::vector<uint16_t> txrx_impl(uint8_t device_address, FunctionCode function, uint16_t first_register_address,
+                                    uint16_t register_quantity, bool wait_for_reply = true,
+                                    std::vector<uint16_t> request = std::vector<uint16_t>());
+
     int read_reply(uint8_t* rxbuf, int rxbuf_len);
 
     Everest::Gpio rxtx_gpio;
+    std::chrono::milliseconds initial_timeout;
+    std::chrono::milliseconds within_message_timeout;
```
corneliusclaussen, line 78: Both could be const and initialized in the initializer list

corneliusclaussen (changes requested):

modules/SerialCommHub/tiny_modbus_rtu.hpp
golovasteek: This would be a better design. But in the way the tiny_modbus implemented and used it would require quite a refactoring: 
* there is no constructor defined for TinyModbusRTU
* the object `modbus` is default-constructed in SerialComHub, and the device is opened later on.
So we will need to figure out when and with which parameters to call the TinyModbusRTU constructor

Do you think it makes sense to add this refactoring into this MR?

modules/SerialCommHub/tiny_modbus_rtu.cpp
golovasteek: Addressed in #535

modules/SerialCommHub/tiny_modbus_rtu.cpp
golovasteek: Addressed in #535

modules/SerialCommHub/main/serial_communication_hubImpl.cpp
golovasteek: Addressed in #535

PR #442, opened 2023-11-20
GenericPowerMeter: Add support for calling specialized power meter implementations

dorezyuk:
This pr adds support for transactions in the generic power meter. 

The mechanics of transactions differ significantly between power meter vendors. Therefore we can't just extend the config in the same way we did for reading power meter values. This pr adds an optional configuration field which will delegate the transaction calls to a "specific" power meter. This allows the user to re-use the work done by the generic power meter and still have transactions.

### DIFF
--- a/modules/GenericPowermeter/main/powermeterImpl.cpp
+++ b/modules/GenericPowermeter/main/powermeterImpl.cpp
@@ -48,19 +48,19 @@ void powermeterImpl::ready() {
 }
 
 types::powermeter::TransactionStopResponse powermeterImpl::handle_stop_transaction(std::string& transaction_id) {
-    if (mod->r_specific_power_meter.empty())
+    if (mod->r_transactional_power_meter.empty())
         return {types::powermeter::TransactionRequestStatus::NOT_SUPPORTED,
                 {},
                 "No specific powermeter configured to start a transaction"};
-    return mod->r_specific_power_meter.front()->call_stop_transaction(transaction_id);
+    return mod->r_transactional_power_meter.front()->call_stop_transaction(transaction_id);
 };
 
 types::powermeter::TransactionStartResponse
 powermeterImpl::handle_start_transaction(types::powermeter::TransactionReq& value) {
-    if (mod->r_specific_power_meter.empty())
+    if (mod->r_transactional_power_meter.empty())
         return {types::powermeter::TransactionRequestStatus::NOT_SUPPORTED,
                 "No specific powermeter configured to stop a transaction"};
-    return mod->r_specific_power_meter.front()->call_start_transaction(value);
+    return mod->r_transactional_power_meter.front()->call_start_transaction(value);
 }
 
 void powermeterImpl::init_default_values() {
--- a/modules/GenericPowermeter/main/powermeterImpl.cpp
+++ b/modules/GenericPowermeter/main/powermeterImpl.cpp
@@ -48,15 +48,19 @@ void powermeterImpl::ready() {
 }
 
 types::powermeter::TransactionStopResponse powermeterImpl::handle_stop_transaction(std::string& transaction_id) {
-    return {types::powermeter::TransactionRequestStatus::NOT_SUPPORTED,
-            {},
-            "Generic powermeter does not support the stop_transaction command"};
+    if (mod->r_specific_power_meter.empty())
+        return {types::powermeter::TransactionRequestStatus::NOT_SUPPORTED,
+                {},
+                "Generic powermeter does not support the stop_transaction command"};
+    return mod->r_specific_power_meter.front()->call_stop_transaction(transaction_id);
 };
 
 types::powermeter::TransactionStartResponse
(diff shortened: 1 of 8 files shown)


modules/GenericPowermeter/main/powermeterImpl.cpp
```diff
@@ -48,15 +48,19 @@ void powermeterImpl::ready() {
...
+    return mod->r_specific_power_meter.front()->call_stop_transaction(transaction_id);
 };
 
 types::powermeter::TransactionStartResponse
 powermeterImpl::handle_start_transaction(types::powermeter::TransactionReq& value) {
-    return {types::powermeter::TransactionRequestStatus::NOT_SUPPORTED,
-            "Generic powermeter does not support the start_transaction command"};
+    if (mod->r_specific_power_meter.empty())
+        return {types::powermeter::TransactionRequestStatus::NOT_SUPPORTED,
+                "Generic powermeter does not support the start_transaction command"};
```
SirVer, line 62: same comment

modules/GenericPowermeter/manifest.yaml
```diff
@@ -23,6 +23,10 @@ provides:
 requires:
   serial_comm_hub:
     interface: serial_communication_hub
+  specific_power_meter:
```
SirVer, line 26: This needs a comment. It is not obvious why this is useful to a first time reader. Maybe the `implementation_id` could also be more specific. Something like "start_transaction_meter"? I do not love this either, but I have no better idea at the moment.

modules/GenericPowermeter/serialization.hpp
```diff
@@ -0,0 +1,178 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Qwello GmbH and Contributors to EVerest
+
+/// @brief Serialization and deserialization for the modbus interface.
+///
+/// The format has a major quirk: Even though our exchanged data is four bytes
```
SirVer, line 6: can you add a static_assert for this somehow?

modules/GenericPowermeter/serialization.hpp
```diff
@@ -0,0 +1,178 @@
...
+inline void check_message(const Result& _message) {
+    if (!_message.value.has_value())
+        throw std::invalid_argument("The `value` must be set");
+}
+
+/// @brief Deserialization for small integral types (less or equal to two bytes).
+/// @param _message The message read.
+/// @param _result The result which will contain the deserialized data.
+template <typename Output, std::enable_if_t<std::is_integral<Output>::value && (sizeof(Output) <= 2), bool> = true>
+inline void deserialize(const Result& _message, Output& _result) {
```
SirVer, line 38: why are you using `_` at the beginning of your variables? For one, that is confusing to read, for two in c++:

> Each identifier that begins with an underscore is reserved to the implementation for use as a name in the global namespace.

From: https://eel.is/c++draft/lex.name#3.2.

So you should never start your identfiers with an underscore - unless you implement a c++ standard library or a c++ compiler. Same comment for all variables that start with `_` in your code here.

modules/GenericPowermeter/serialization.hpp
```diff
@@ -0,0 +1,178 @@
+// SPDX-License-Identifier: Apache-2.0
```
SirVer, line 1: I see that this code is tested, but where is it used?

dorezyuk: Its (for now) used in https://github.com/EVerest/everest-core/pull/443/files

modules/GenericPowermeter/serialization.hpp
dorezyuk: Would be nice but I don't think so.. its not expressed through the type system

modules/GenericPowermeter/serialization.hpp
SirVer: Ok, fine by me.

modules/GenericPowermeter/manifest.yaml
SirVer: friendly ping on this.

modules/GenericPowermeter/main/powermeterImpl.cpp
```diff
@@ -48,15 +48,19 @@ void powermeterImpl::ready() {
 }
 
 types::powermeter::TransactionStopResponse powermeterImpl::handle_stop_transaction(std::string& transaction_id) {
-    return {types::powermeter::TransactionRequestStatus::NOT_SUPPORTED,
-            {},
-            "Generic powermeter does not support the stop_transaction command"};
+    if (mod->r_transactional_power_meter.empty())
+        return {types::powermeter::TransactionRequestStatus::NOT_SUPPORTED,
+                {},
+                "No specific powermeter configured to start a transaction"};
```
barsnick, line 54: This should be "No specific powermeter configured to **stop** a transaction"

corneliusclaussen:
See feedback on https://github.com/EVerest/everest-core/pull/443

corneliusclaussen:
Can we close this? I think this should be solved differently

PR #443, opened 2023-11-20
WIP: PowermeterBSM_WS36A: initial commit

dorezyuk:
This pr adds another support for Bauer meters.

Contrary to the PowermeterBSM code - this uses the SerialCommHub interface and adds transaction support for it. 
The pr needs https://github.com/EVerest/everest-core/pull/442


modules/PowermeterBSM_WS36A/doc.rst
```diff
@@ -0,0 +1,22 @@
+.. _everest_modules_handwritten_PowermeterBSM_WS36A:
+
+..  This file is a placeholder for an optional single file handwritten documentation for
```
SirVer, line 3: It seems like you did neither :). I think you should put a few lines of documentation here: Why this module, how it differers vs the existing and how it interacts with the Serial one.

modules/PowermeterBSM_WS36A/main/powermeterImpl.cpp
```diff
@@ -0,0 +1,171 @@
...
+using namespace types::serial_comm_hub_requests;
+
+void powermeterImpl::init() {
+}
+
+void powermeterImpl::ready() {
+    std::thread([this]() {
+        while (true) {
+
+            TransactionReq req;
```
SirVer, line 54: probably want to add a TODO, since this is hardly what we actually want to do in the end, right?

modules/PowermeterBSM_WS36A/main/powermeterImpl.cpp
```diff
@@ -0,0 +1,171 @@
...
+    const auto response = read_register_impl(_register);
+    if (response.status_code != StatusCodeEnum::Success)
+        throw std::runtime_error("Failed to send the message");
+    return deserialize<Output>(response);
+}
+
+template <typename Input> void powermeterImpl::write_register(const Register& register_data, const Input& data) const {
+
+    // Serialize the data into a vector.
+    auto ser_data = serialize(data);
```
SirVer, line 156: this makes me think that `serialize` should really be in a namespace that makes it clear that it has to do with modbus. That has to be done in #442 though.

modules/PowermeterBSM_WS36A/manifest.yaml
```diff
@@ -0,0 +1,25 @@
+description: Powermeter driver for BAUER-BSM-WS36A
```
SirVer, line 1: using the serial_comm_hub as interface.

modules/PowermeterBSM_WS36A/manifest.yaml
```diff
@@ -0,0 +1,25 @@
...
+        minimum: 0
+        maximum: 65535
+        default: 30001
+requires:
+  serial_comm_hub:
+    interface: serial_communication_hub
+metadata:
+  license: https://opensource.org/licenses/Apache-2.0
+  authors:
+    - Dima Dorezyuk
```
SirVer, line 25: Probably: 'Qwello GmbH <embedded-team@qwello.eu>`, since that will be true forever - hopefully.

modules/PowermeterBSM_WS36A/main/powermeterImpl.cpp
dorezyuk: don't really think that this is necessary since the whole function is about calling the serial-comm-hubs' modbus methods but will do

corneliusclaussen:
This modules seems to implement the powermeter interface, but it looks like it actually does not. It only implements the transaction related commands but not the rest, where it requires a generic powermeter module to proxy.
I don't think that is a good architecture.
The powermeter interface is a contract that is not fulfilled by this implementation, and the Generic powermeter module with the PR proxies any powermeter (even those implementing everything). This means it is very easy to configure things that do not make sense, e.g. connect a full powermeter implementation via a generic powermeter proxy (which measurement values should be used in that case?) or use this BSM_WS36A as a powermeter (but then it does not do the measurements).

I think there are two possible solutions:

1) Make the implementation of BSM_WS36A complete. It is not much duplication and the much cleaner approach imho. I'm much in favor of this solution.

2) Create a new interface that is used to connect the incomplete WS36A to the generic powermeter. This new interface only has the transaction commands but not the rest. This also prevents the incomplete power meter to be used as a normal powermeter.

corneliusclaussen:
> Your second option would not tackle the problem with the generic powermeter being "incomplete". If we want to split the interfaces should split them in "reading" and "transactions"... wdyt?

I don't like the second approach either. But it would at least prevent mis usage of this incomplete (and hence non-working) power meter implementation.

Not all interfaces specify all details of the contract (such as frequency of publishes) but they should not just be ignored because of that. If someone is using this power meter driver and connects it to EvesManager or somewhere else it would obviously not work as expected, hence the workaround to create a new interface for this specific add-on.

In any case as discussed in the weekly I think the solution should be to factor out code duplication into a "lib" under lib/staging and have a driver that provides the complete functionality instead of relying on generic power meter.

corneliusclaussen:
Can we close this? I think this needs to be reworked as a module that works stand-alone without the need for #442. If there is too much code duplication we can factor this out under /lib/staging to share code between different power meter implementations.

PR #448, opened 2023-11-23
Implement ocpp_data_transfer interface in OCPP201

hikinggrass:


modules/OCPP201/manifest.yaml
```diff
@@ -51,6 +54,10 @@ requires:
     interface: kvs
     min_connections: 1
     max_connections: 1
+  data_transfer:
+    interface: ocpp_data_transfer
+    min_connections: 0
+    max_connections: 1
```
valentin-dimov, line 60: Maybe add a comment here that this is the handler for data transfer from the CSMS to the charging station, just to avoid confusion.

PR #451, opened 2023-11-23
Throw outside conversion switch/case to pacify compiler warnings

hikinggrass:
Otherwise throws a "control reaches end of non-void function [-Wreturn-type]" warning in gcc

### DIFF
--- a/modules/OCPP201/OCPP201.cpp
+++ b/modules/OCPP201/OCPP201.cpp
@@ -566,9 +566,8 @@ ocpp::v201::CertificateActionEnum get_certificate_action(const types::iso15118_c
         return ocpp::v201::CertificateActionEnum::Install;
     case types::iso15118_charger::CertificateActionEnum::Update:
         return ocpp::v201::CertificateActionEnum::Update;
-    default:
-        throw std::out_of_range("Could not convert CertificateActionEnum"); // this should never happen
     }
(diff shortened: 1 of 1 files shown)


modules/OCPP201/OCPP201.cpp
```diff
@@ -566,6 +566,8 @@ ocpp::v201::CertificateActionEnum get_certificate_action(const types::iso15118_c
         return ocpp::v201::CertificateActionEnum::Install;
     case types::iso15118_charger::CertificateActionEnum::Update:
         return ocpp::v201::CertificateActionEnum::Update;
+    default:
+        throw std::out_of_range("Could not convert CertificateActionEnum"); // this should never happen
```
a-w50, line 570: This is debatable, but have a look at https://stackoverflow.com/questions/57532056/switch-may-fall-through-no-it-may-not
For these switch statements, you wont get warnings, if you add a new value to the enum list.  If you do a return at the end of the function (instead of using default), you will.

a-w50 (approved):

modules/OCPP201/OCPP201.cpp
hikinggrass: The problem in these two cases is that none of the two returns make sense as a default return value :/

However removing the default and just throw()-ing outside of the switch would be the best option I guess...

PR #453, opened 2023-11-27
Proper shutdown of the TCP connection

SebaLukas:
Currently, `dlink_terminate` is sent to the SLAC module directly after the `SessionStopRes` message is sent. This is too fast. As a result, the EV has no chance to close its TCP/TLS connection. Most EVs have problems with this.

Now the EV is informed that the EVSE intends to close the TCP/TLS server. After waiting 5 seconds, the Charger closes the connection and only then sends `dlink_terminate`.

### DIFF
--- a/modules/EvseV2G/connection.cpp
+++ b/modules/EvseV2G/connection.cpp
@@ -476,9 +476,11 @@ static void connection_teardown(struct v2g_connection* conn) {
         dlog(DLOG_LEVEL_TRACE, "d_link/error");
         break;
     case MQTT_DLINK_ACTION_TERMINATE:
+        conn->ctx->p_charger->publish_dlink_terminate(nullptr);
         dlog(DLOG_LEVEL_TRACE, "d_link/terminate");
         break;
     case MQTT_DLINK_ACTION_PAUSE:
+        conn->ctx->p_charger->publish_dlink_pause(nullptr);
         dlog(DLOG_LEVEL_TRACE, "d_link/pause");
         break;
     }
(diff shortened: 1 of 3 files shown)


modules/EvseV2G/connection.cpp
```diff
@@ -673,6 +675,10 @@ static void* connection_handle_tcp(void* data) {
     if (shutdown(conn->conn.socket_fd, SHUT_RDWR) == -1) {
         dlog(DLOG_LEVEL_ERROR, "shutdown() failed: %s", strerror(errno));
     }
+
+    // Waiting for client closing the connection
+    std::this_thread::sleep_for(std::chrono::seconds(5));
```
corneliusclaussen, line 680: We tested this a lot in Cleveland. I think we should also wait before we send the shutdown and not only the close. Some cars did not really like that we send a FIN first even if we keep the connection running. They insisted that the EV should send the first FIN. Apart from that it is good.

corneliusclaussen (changes requested):

modules/EvseV2G/connection.cpp
SebaLukas: Good point, then I'll add a sleep of 2 seconds before shutdown. And change the 5 seconds sleep after shutdown to 3 seconds

SebaLukas: And I'll also add a sleep before the TLS shutdown. I also suspect that it is shutting down too quickly.

corneliusclaussen (approved):

PR #456, opened 2023-11-29
feat: Add configuration option to delay message queue resume

valentin-dimov:
Companion to https://github.com/EVerest/libocpp/pull/283

### DIFF
--- a/modules/OCPP201/manifest.yaml
+++ b/modules/OCPP201/manifest.yaml
@@ -24,6 +24,10 @@ config:
     description: If true websocket can be disconnected and connected externally
     type: boolean
     default: false
+  MessageQueueResumeDelay:
+    description: Time (seconds) to delay resuming the message queue after reconnecting
+    type: integer
+    default: 0
 provides:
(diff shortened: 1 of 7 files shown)


modules/OCPP201/manifest.yaml
```diff
@@ -24,6 +24,10 @@ config:
     description: If true websocket can be disconnected and connected externally
     type: boolean
     default: false
+  MessageQueueResumeDelay:
```
hikinggrass, line 27: Probably makes sense to have this in the OCPP (1.6) module as well

valentin-dimov: We would need to extend the libocpp PR a bit as well, in order to allow passing that setting to the 1.6 chargepoint. Should we do that as part of this PR?

valentin-dimov: Done

hikinggrass (approved):
please update libocpp rev then this should build and be able to merge

PR #458, opened 2023-11-30
Readme: Changed config file from json to yaml

krealyt:


krealyt:
@pionix-juraed Please also check if this is still the correct place to put those user config files.

config/user-config/README.md
```diff
@@ -1,15 +1,16 @@
 # User provided configuration
-You can add user provided configuration files in this folder that can augment a loaded config of the same name.
-For example, if you want to change the ***serial_port*** of the ***yeti_driver*** in the [config-hil.json](../config-hil.json) to a more appropriate value on you machine you can achieve this by adding a [config-hil.json](config-hil.json) file in this user-config folder with the following content:
+You can add user provided configuration files in this folder that can augment
+a loaded config of the same name.
+For example, if you want to change the ***serial_port*** of the
+***yeti_driver*** in the [config-hil.yaml](../config-hil.yaml) to a more
+appropriate value on your machine you can achieve this by adding a
+*config-hil.yaml* file in this user-config folder with the following content:
```
valentin-dimov, line 7: `config-hil.yaml` will probably be removed very soon. Could we use a different example here?

krealyt: @valentin-dimov Could you make a very simple example? Or who could?

valentin-dimov: Maybe we could suggest changing the `evse_id` of `connector_1` in `config-sil.yaml` to `DE*YOURCOMPANY*E12345*1`?

krealyt: @valentin-dimov Changed as you suggested.

valentin-dimov (approved):
lgtm!

PR #460, opened 2023-11-30
Extend API with enable disable cmd

hikinggrass:


modules/API/API.cpp
```diff
@@ -301,6 +301,34 @@ void API::init() {
...
+            if (!data.empty()) {
+                try {
+                    connector_id = std::stoi(data);
+                } catch (const std::exception& e) {
+                    EVLOG_error << "Could not parse connector id for enable connector, using " << connector_id
+                                << ", error: " << e.what();
+                }
+            }
+            evse->call_enable(connector_id);
+        });
```
klemmpnx, line 316: in case the user sends data but this cannot be parsed, do we really want to enable/disable connector 0? We might rather just ignore the command then / log an error?

```suggestion
        this->mqtt.subscribe(cmd_enable, [&evse](const std::string& data) {
            if (!data.empty()) {
                int connector_id{};
                try {
                    connector_id = std::stoi(data);

                } catch (const std::exception& e) {
                    EVLOG_error << "Could not parse connector id for enable connector, ignoring command; error: "
                                << e.what();
                    return;
                }
                evse->call_enable(connector_id);
            } else {
                evse->call_enable(0);
            }
        });
```

klemmpnx (approved):
lgtm, but consider the one comment

modules/API/API.cpp
hikinggrass: That's a good point, I'll change that for enable/disable and update the readme accordingly

PR #463, opened 2023-12-01
Add generic OCPP interface

klemmpnx:
- Introduces the ocpp_generic interface to be shared by 1.6 and 2.0.1. 
- Implements this for OCPP 1.6

Requires:
- https://github.com/EVerest/everest-utils/pull/79 (ev-cli version 0.2.4)


interfaces/ocpp.yaml
```diff
@@ -0,0 +1,111 @@
...
+      description: >-
+        List of SetVariableResult containing the result for every requested set operation
+      type: array
+      items:
+        type: object
+        $ref: /ocpp#/SetVariableResult
+  monitor_variables:
+    description: >-
+      Command to start monitoring the given ComponentVariable(s). In case this configuration keys is 
+      changed by the CSMS the configuration key the changed configuration key will be published.
```
barsnick, line 79: "the configuration key the changed configuration key" -> some words seem to be duplicated. Probably meant: "the changed configuration key"

modules/OCPP/ocpp_generic/ocppImpl.hpp
```diff
@@ -0,0 +1,69 @@
...
+private:
+    const Everest::PtrContainer<OCPP>& mod;
+    const Conf& config;
+
+    virtual void init() override;
+    virtual void ready() override;
+
+    // ev@3370e4dd-95f4-47a9-aaec-ea76f34a66c9:v1
+    // insert your private definitions here
+    std::mutex m;
```
hikinggrass, line 58: This could be a bit more descriptive

hikinggrass (approved):

modules/OCPP/ocpp_generic/ocppImpl.hpp
klemmpnx: indeed, adapted it

hikinggrass:
Since you require a newer version of ev-cli you should also update this in https://github.com/EVerest/everest-core/blob/5be1ef7d24f5eff6230ce5d8fc3078ae2b175def/cmake/ev-project-bootstrap.cmake#L8 (this already throws an error quite early in the build process, not just during compilation)

PR #465, opened 2023-12-05
umwc: Add config option for max voltage

corneliusclaussen:


modules/MicroMegaWattBSP/manifest.yaml
```diff
@@ -16,6 +16,12 @@ config:
     minimum: -1
     maximum: 1000
     default: -1
+  dc_max_voltage:
+    description: Maximum voltage to support
+    type: integer
+    minimum: 150
```
a-w50, line 22: this is smaller than the default 180V, is this intentional?

a-w50 (approved):

modules/MicroMegaWattBSP/manifest.yaml
corneliusclaussen: True. I changed both to 50V, that makes more sense. Will merge now.

PR #466, opened 2023-12-05
Increase TLS timeout for really slow EV implementations

corneliusclaussen:
We have seen really slow EV implementations in Cleveland testival failing initial setup at 5s and it should not hurt to increase the TLS timeout


modules/EvseV2G/v2g_ctx.cpp
```diff
@@ -309,7 +309,7 @@ struct v2g_context* v2g_ctx_create(ISO15118_chargerImplBase* p_chargerImplBase,
     ctx->if_name = "eth1";
 
     ctx->network_read_timeout = 1000;
-    ctx->network_read_timeout_tls = 5000;
+    ctx->network_read_timeout_tls = 30000;
```
mooraby, line 312: As this was changed before due to errors in another testing event (https://github.com/EVerest/everest-core/commit/b30916d99da197ff4f709dd73088cda59cd8965b), should this better be made configurable in the manifest?

SebaLukas: I like the idea of making this configurable

SebaLukas: @mooraby The TLS timeout is now configurable

corneliusclaussen: Should we increase the config default of 5000 a bit? 30s was needed for a debugging build on a vehicle, but even the production code took 10-11 seconds. So maybe default to 15s?

SebaLukas: Yes 15 sec are safer. I change that

SebaLukas (approved):

PR #473, opened 2023-12-10
allow B->C transition in waiting for auth

corneliusclaussen:


corneliusclaussen:
This should never happen with a compliant car, but can be triggered during manual testing. In this case it is better to just accept it as it not easy to understand for the tester what happend and why it does not charge.

corneliusclaussen:
Relais will of course only close when authorization etc is done.

SebaLukas (approved):
:+1:

PR #475, opened 2023-12-11
Move step install everest-testing to github workflow

andistorm:
* Remove dependency to iso15118_pip_install_dist, since it isn't needed in most cases
* Refactor name fields of github workflow
* Use `install_everest_testing` instead of `everestpy_pip_install_dist` to make sure using the corresponding everest-testing version in ci

### DIFF
--- a/.ci/build-kit/test_and_install.sh
+++ b/.ci/build-kit/test_and_install.sh
@@ -3,8 +3,8 @@
 # ninja -j$(nproc) -C build tests
 ninja -j$(nproc) -C build install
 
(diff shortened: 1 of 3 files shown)


andistorm:
Installation of everest-testing starts somewhere here: https://github.com/EVerest/everest-core/actions/runs/7165366279/job/19507175718?pr=475#step:10:1119

Right now it uninstalls the version deployed in the docker image before installing with https://github.com/EVerest/everest-ci/pull/12 this will change

hikinggrass (approved):

PR #476, opened 2023-12-11
Create custom sil config to run integration tests

ju4nlu:
Integration tests are disabled since some of the modules used in the default `config-sil.yaml` are no longer built in our pipeline.

This PR:

- Creates a custom config that doesn't use the non-built modules
- Marks the ignored test as ready again


ju4nlu:
My bad, pushed this into the wrong place.

PR #478, opened 2023-12-11
Bugfix/fix legacy wakeup sequence

corneliusclaussen:
The previous version was implemented wrong. In AC basic charging, if PWM is enabled in PrepareCharging (CP state B) and the EV does not transition to state C within 30s, a T_step_EF legacy wake up sequence is tried once. We cannot distinguish whether the EV does not want to start charging of it is stuck in a deep sleep mode. So we try to wake it up once and if that does not help we do not try again until the EV is unplugged (we assume it does not want to start charging then)

### DIFF
--- a/modules/EvseManager/Charger.cpp
+++ b/modules/EvseManager/Charger.cpp
@@ -414,6 +414,15 @@ void Charger::runStateMachine() {
                 }
             }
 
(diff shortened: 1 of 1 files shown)


SebaLukas (approved):
:+1:

PR #482, opened 2023-12-19
CI: modify to reusable workflow

Dominik-K:
Tested directly in
- https://github.com/EVerest/libocpp/pull/356


.github/workflows/build_and_test.yaml
```diff
@@ -23,47 +44,67 @@ jobs:
...
-          build-image run-script test_and_install
-      - name: Run integration tests
+          unit-tests-image run-script unit_tests
+      - name: Integration tests
+        if: ${{ inputs.integration_tests || true }}
         run: |
-          docker commit test-container integration-image
-          pushd source/.ci/e2e
-          docker-compose run e2e-test-server run-script tests
+          docker commit compile-container integration-tests-image
```
Dominik-K, line 102: Do the integration-tests rely on the unit-tests somehow?

.github/workflows/build_and_test.yaml
```diff
@@ -16,54 +37,110 @@ on:
...
-        uses: actions/checkout@v3
+      - name: Checkout `${{ env.CI_REPO }}` to `${{ env.CI_REPO_DIR }}/` (for reusable workflow mode, see https://stackoverflow.com/a/74123003/1168315)
+        uses: actions/checkout@v4
         with:
-          path: source
-      - name: Run clang-format
-        uses: everest/everest-ci/github-actions/run-clang-format@v1.0.0
+          path: ${{ env.CI_REPO_DIR }}
+          repository: ${{ env.CI_REPO }}
+          ref: CI/reusable-workflow  # TODO get reference in general, see https://stackoverflow.com/questions/74784735 resp. https://github.com/actions/toolkit/issues/1264
```
Dominik-K, line 66: @andistorm Do you know a solution for this? Otherwise, I would just set it to `main` for now.

.github/workflows/build_and_test.yaml
```diff
@@ -16,54 +37,110 @@ on:
...
         with:
-          source-dir: source
-          extensions: hpp,cpp
-          exclude: cache
-      - name: Setup run scripts
-        run: |
-          mkdir scripts
-          rsync -a source/.ci/build-kit/ scripts
+          path: ${{ env.SOURCE_DIR }}
+          repository: ${{ inputs.repository }}
```
christopher-davis-afs, line 71: This doesn't seem like the right way to set up a reusable workflow. When someone makes a fork, the CI instructions on their forked branches will point to whatever original repository is passed in, for example, [EVerest/libocpp](https://github.com/EVerest/libocpp/pull/356/files#diff-868425f4c6dd9927d83974d334ea46b9ec96854cdac9bb15b15f43bc4ad17820R9). What *should* happen is that the current branch that initialized the workflow should run, fetching the branch from the forked repository.

In the current state, this [fails at the checkout stage](https://github.com/US-JOET/libfsm/actions/runs/7660200585/job/20876978782#step:5:46) because we can't check out the original repository from our fork.

.github/workflows/build_and_test.yaml
andistorm: I would suggest to not checkout the ci repo in the workflow, but to use custom github actions. There are multiple options for github actions: `composite actions` that are similiar to `reusable workflows`, `Javascript actions` and `Docker container actions`. The last one would be an option to include additional filesto the action.

I could also imagine a solution with having the bash scripts in each repo with "custom" commands and cmake flags..

.github/workflows/build_and_test.yaml
andistorm: I don't see the problem in the way the `reusable workflow` is implemented, but the way it is called.

In the[ example of `libocpp` ](https://github.com/EVerest/libocpp/pull/356/files#diff-868425f4c6dd9927d83974d334ea46b9ec96854cdac9bb15b15f43bc4ad17820R13)the parameter `repository: EVerest/libocpp` is used. You adapted it similiar in[ your example ](https://github.com/US-JOET/libfsm/actions/runs/7660200585/workflow#L12) with `repository: EVerest/libfsm`

To enable workflows running on forks I suggest the use of the[ env variables 
](https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables).

-> In both examples I would replace the line with `repository: ${{ github.repository }}`.

.github/workflows/build_and_test.yaml
andistorm: I don't think so :)

.github/workflows/build_and_test.yaml
christopher-davis-afs: Shouldn't the reusable workflow use that by default?

andistorm: yeah, maybe this parameter is obsolete, not sure about this.. I will spend some time on this in the next weeks to figure those things out

golovasteek:
The everest-core is an umbrella-repository that depends on all other EVerest components. If the reusable workflow will be part of everest-core, then all components will depend on the everest-core.
This looks like a cyclic dependency. Do we see a problem with it? Did we consider other places to "host" the reusable workflow?

andistorm:
> The everest-core is an umbrella-repository that depends on all other EVerest components. If the reusable workflow will be part of everest-core, then all components will depend on the everest-core. This looks like a cyclic dependency. Do we see a problem with it? Did we consider other places to "host" the reusable workflow?

Thank you for the notice. I planned to shift the workflow to everest-ci.

andistorm:
Close in favor of upcomming reusable workflow in everest-ci https://github.com/EVerest/everest-ci/pull/54

PR #483, opened 2023-12-19
Add additional information to SessionFinished events

ju4nlu:
Up until now the `SessionFinished` events have no timestamp. This makes it harder for other modules to keep track of how long a session has been running for.

This PR adds a specific session_finished property to the `SessionEvent` type which just contains a timestamp. Changes should be fine for previous builds since the new field `session_finished` for type `SessionEvent` is optional.

This comes from a mailing list topic ([source](https://lists.lfenergy.org/g/everest/topic/102826025#995)).


modules/EvseManager/evse/evse_managerImpl.cpp
```diff
@@ -168,12 +168,16 @@ void evse_managerImpl::ready() {
 
             se.session_started = session_started;
         } else if (e == types::evse_manager::SessionEventEnum::SessionFinished) {
+            types::evse_manager::SessionFinished session_finished;
+            session_finished.timestamp =
+                date::format("%FT%TZ", std::chrono::time_point_cast<std::chrono::milliseconds>(date::utc_clock::now()));
```
corneliusclaussen, line 173: Please use the following to format timestamps in everest:

Everest::Date::to_rfc3339(date::utc_clock::now());

corneliusclaussen (changes requested):

modules/EvseManager/evse/evse_managerImpl.cpp
ju4nlu: Done. I wildly copy-pasted what is done for `SessionStarted` events. I might as well create a different MR to use the format you specified as well in `SessionStarted`, `TransactionStarted` and `TransactionFinished` timestamps' events and unify them.

corneliusclaussen: That sounds like a very good idea!

corneliusclaussen (approved):

PR #495, opened 2024-01-13
OCPP: Adjusted argument of on_session_started from string to SessionStartedReason

Pietfried:
Adjusted changed argument type for reason of on_session_started from string to ocpp::SessionStartedReason.

This change is required by: https://github.com/EVerest/libocpp/pull/402

### DIFF
--- a/modules/OCPP/OCPP.cpp
+++ b/modules/OCPP/OCPP.cpp
@@ -116,6 +116,18 @@ static ErrorInfo get_error_info(const std::optional<types::evse_manager::Error> 
     }
 }
 
+ocpp::SessionStartedReason get_session_started_reason(const types::evse_manager::StartSessionReason reason) {
+    switch (reason) {
+    case types::evse_manager::StartSessionReason::EVConnected:
+        return ocpp::SessionStartedReason::EVConnected;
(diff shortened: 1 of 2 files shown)


hikinggrass (approved):

hikinggrass:
Looks good, just needs a version bump for the libocpp dependency once https://github.com/EVerest/libocpp/pull/402 is merged

modules/OCPP/OCPP.cpp
```diff
@@ -116,6 +116,18 @@ static ErrorInfo get_error_info(const std::optional<types::evse_manager::Error>
     }
 }
 
+ocpp::SessionStartedReason get_session_started_reason(const types::evse_manager::StartSessionReason reason) {
+    switch (reason) {
+    case types::evse_manager::StartSessionReason::EVConnected:
+        return ocpp::SessionStartedReason::EVConnected;
+    case types::evse_manager::StartSessionReason::Authorized:
+        return ocpp::SessionStartedReason::Authorized;
+    default:
```
a-w50, line 125: @Pietfried best practice here is to do the `throw` outside of the switch.  That way the compiler might see, that some switch cases are missing (for example, if new values are added to the enum).

PR #505, opened 2024-01-23
Add source to enable/disable commands/events

corneliusclaussen:
Add new enable_disable commands to EvseManager
    
    It replaces the old enable/disable commands and adds a source and
    priority. See README of the API module for more details.
    
    Compile time breaking change on the evse_manager interface.
    All enable/disable calls need to be switched over to the new
    enable_disable call, for everest-core this is done in this PR.
    
    In some uses cases it is useful if e.g. a LocalAPI source disables
    a connector that only the same local source can enable again.
    Think of a service technician who has an app to disable the connector
    locally. Then it should be prevented that it is re-enabled by the CSMS
    until the service technician enables it again via the same app.
    Using priorities this is quite flexible and a lot of use-cases can
    be implemented without a change in EVerest code.
    
    The API module also has a new enable_disable command. It also has
    compatibility implementations for the old enable/disable commands.

### DIFF
--- a/modules/API/README.md
+++ b/modules/API/README.md
@@ -37,22 +37,28 @@ This variable is published every second and contains the hardware capabilities i
 ### everest_api/evse_manager/var/session_info
 This variable is published every second and contains a json object with information relating to the current charging session in the following format:
 ```json
-    {
-        "charged_energy_wh": 0,
-        "charging_duration_s": 84,
-        "datetime": "2022-10-11T16:48:35.747Z",
-        "discharged_energy_wh": 0,
-        "latest_total_w": 0.0,
-        "state": "Unplugged",
-        "active_permanent_faults": [],
-        "active_errors": [],
-        "uk_random_delay": {
-            "remaining_s": 34,
-            "current_limit_after_delay_A": 16.0,
-            "current_limit_during_delay_A": 0.0,
-            "start_time": "2024-02-28T14:11:11.129Z"
-        }
-    }
+{
+    "charged_energy_wh": 0,
+    "charging_duration_s": 84,
+    "datetime": "2022-10-11T16:48:35.747Z",
+    "discharged_energy_wh": 0,
+    "latest_total_w": 0.0,
+    "state": "Unplugged",
+    "active_permanent_faults": [],
+    "active_errors": [],
+    "active_enable_disable_source": {
+        "source": "Unspecified",
+        "state": "Enable",
+        "priority": 5000
+    },
+    "uk_random_delay": {
+        "remaining_s": 34,
+        "current_limit_after_delay_A": 16.0,
+        "current_limit_during_delay_A": 0.0,
+        "start_time": "2024-02-28T14:11:11.129Z"
+    },
+    "last_enable_disable_source": "Unspecified"
+}
 ```
 
 Example with permanent faults being active:
@@ -77,7 +83,19 @@ Example with permanent faults being active:
   "datetime": "2024-01-15T14:58:15.172Z",
   "discharged_energy_wh": 0,
   "latest_total_w": 0,
-  "state": "Preparing"
+  "state": "Preparing",
+  "active_enable_disable_source": {
+    "source": "Unspecified",
+    "state": "Enable",
+    "priority": 5000
+  },
+  "uk_random_delay": {
+    "remaining_s": 34,
+    "current_limit_after_delay_A": 16.0,
+    "current_limit_during_delay_A": 0.0,
+    "start_time": "2024-02-28T14:11:11.129Z"
+  },
+  "last_enable_disable_source": "Unspecified"
 }
 ```
 
@@ -135,10 +153,13 @@ Example with permanent faults being active:
     - PermanentFault
     - PowermeterTransactionStartFailed
 
-- **active_errors** array of all active errors that do not block charging. This could be shown to the user but the current state should still be shown as it does not interfere with charging. The enum is the same as for active_permanent_faults.
+- **active_errors** array of all active errors that do not block charging.
+This could be shown to the user but the current state should still be shown
+as it does not interfere with charging. The enum is the same as for active_permanent_faults.
 
 ### everest_api/evse_manager/var/limits
-This variable is published every second and contains a json object with information relating to the current limits of this EVSE.
+This variable is published every second and contains a json object with information
+relating to the current limits of this EVSE.
 ```json
     {
         "max_current": 16.0,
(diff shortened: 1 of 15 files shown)


modules/API/README.md
```diff
@@ -46,12 +46,19 @@ This variable is published every second and contains a json object with informat
...
+        },
         "uk_random_delay": {
             "remaining_s": 34,
             "current_limit_after_delay_A": 16.0,
             "current_limit_during_delay_A": 0.0,
             "start_time": "2024-02-28T14:11:11.129Z"
-        }
+        },
+        "last_enable_disable_source": "Unspecified",
+        "enable_disable_allow_others_to_modify": true
```
TMilburnPodPoint, line 61: Outdated example

modules/API/README.md
```diff
@@ -77,7 +84,9 @@ Example with permanent faults being active:
   "datetime": "2024-01-15T14:58:15.172Z",
   "discharged_energy_wh": 0,
   "latest_total_w": 0,
-  "state": "Preparing"
+  "state": "Preparing",
+  "last_enable_disable_source": "Unspecified",
+  "enable_disable_other_sources_may_modify": true
```
TMilburnPodPoint, line 89: Outdated example

modules/OCPP/OCPP.cpp
```diff
@@ -624,15 +625,19 @@ void OCPP::ready() {
 
     this->charge_point->register_disable_evse_callback([this](int32_t connector) {
         if (this->connector_evse_index_map.count(connector)) {
-            return this->r_evse_manager.at(this->connector_evse_index_map.at(connector))->call_disable(0);
+            return this->r_evse_manager.at(this->connector_evse_index_map.at(connector))
+                ->call_enable_disable(
+                    0, {types::evse_manager::Enable_source::CSMS, types::evse_manager::Enable_state::Disable, 5000});
```
TMilburnPodPoint, line 630: The magic 5000 priority for CSMS should be documented somewhere.

modules/EvseManager/Charger.cpp
```diff
@@ -1167,31 +1167,131 @@ bool Charger::deauthorize_internal() {
...
+    if (not replaced) {
+        // Add to the table
+        enable_disable_source_table.push_back(source);
+    }
+
+    // Find out if we are actually enabled or disabled at the moment
+    bool is_enabled = parse_enable_disable_source_table();
+
+    if (is_enabled) {
+        if (connector_id not_eq 0) {
```
TMilburnPodPoint, line 1192: Shouldn't there be a check that the connector ID does correspond to this charger's connector ID?

modules/API/README.md
```diff
@@ -201,16 +214,96 @@ This variable is published every second and contains powermeter information of t
...
 ### everest_api/evse_manager/cmd/enable
-Command to enable a connector on the EVSE. They payload should be a positive integer identifying the connector that should be enabled. If the payload is 0 the whole EVSE is enabled.
+Legacy command to enable a connector on the EVSE kept for compatibility reasons.
+They payload should be a positive integer identifying the connector that should be enabled.
+If the payload is 0 the whole EVSE is enabled.
+It will actually call the following command on everest_api/evse_manager/cmd/enable_enable:
+```json
+    {
+        "connector_id": 1,
+        "source": "LocalAPI",
```
brett-swann, line 288: If for compatibility, should this not map on to "CSMS"? Any previous users of this API might rely on it and the CSMS mutating the same state?

modules/OCPP/OCPP.cpp
```diff
@@ -624,15 +625,19 @@ void OCPP::ready() {
 
     this->charge_point->register_disable_evse_callback([this](int32_t connector) {
```
brett-swann, line 626: I think this callback might be used by both the OCPP `ChangeAvailability.req` and firmware update? If so it might be necessary to extend the callback to cater for source.

modules/API/README.md
corneliusclaussen: It is true that this is a breaking change, but if we map it to CSMS the feature cannot be used

modules/EvseManager/Charger.cpp
corneliusclaussen: The EvseManager should only get calls for connector_id(s) that it owns. This is a bit confusing at the moment since right now EvseManager supports only one connector per Evse. OCPP2 already supports multiple connectors per evse. EvseManager will need to change architecture a bit, then this will be cleaner

modules/API/README.md
james-ctc: updated

modules/OCPP/OCPP.cpp
james-ctc: Charge Availability, Update Firmware, and Signed Update Firmware are all requests from the CSMS. i.e. the same source. (note firmware update only disables availability - enabled following restart of EVerest)

modules/OCPP/OCPP.cpp
james-ctc: added readme

modules/OCPP/OCPP.cpp
brett-swann: I'd thought it would be necessary for the following two use cases:

- If a firmware upgrade is ongoing (the charge point is unavailable), I would not expect the CSMS to be able to make the charge point available. 
- Similarly, if the CSMS has made the charge point unavailable, I would not expect a failed firmware upgrade to make it available. 

I don't think they're specifically covered in OCPP 1.6 (unless one infers from pg42 that the state of internal unavailability sources is distinct).

james-ctc: When a FW upgrade is ongoing libocpp could/should reject change availability requests. That wouldn't require a change the disable EVSE callback.
Failure of a firmware upgrade doesn't cause the charge point to be come available (from libocpp). It is only made unavailable once the update has been successfully downloaded and verified.

corneliusclaussen:
@james-ctc Anything missing here still or are we good?

james-ctc:
All good to review and merge

Pietfried (changes requested):

Pietfried (approved):

Pietfried:
@SebaLukas your review is also required for this one.

types/evse_manager.yaml
```diff
@@ -324,6 +324,39 @@ types:
...
+        type: string
+        enum:
+        - Unassigned
+        - Disable
+        - Enable
+      enable_priority:
+        description: Priority of this entry. The highest priority is 0.
+        type: integer
+        minimum: 0
+        maximum: 10000
```
mhei, line 359: Any reason to use 10000 here? I'd stick to usual byte range boundaries here, e.g. 255 - should be far enough IMHO.

james-ctc (approved):
The changes look okay to me

SebaLukas (approved):
:+1:

PR #509, opened 2024-01-25
DCSupplySimulator: remove designated assignments

barsnick:
These are only officially supported from C++20 upwards.

While at it, also fix a description typo in the manifest(s).

### DIFF
--- a/modules/simulation/DCSupplySimulator/main/power_supply_DCImpl.cpp
+++ b/modules/simulation/DCSupplySimulator/main/power_supply_DCImpl.cpp
@@ -17,28 +17,33 @@ void power_supply_DCImpl::init() {
     this->power_supply_thread_handle = std::thread(&power_supply_DCImpl::power_supply_worker, this);
 }
 
+static auto get_capabilities_from_config(const Conf& config) {
+    types::power_supply_DC::Capabilities cap;
+
+    cap.bidirectional = config.bidirectional;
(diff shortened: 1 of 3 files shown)


modules/simulation/DCSupplySimulator/main/power_supply_DCImpl.cpp
```diff
@@ -13,32 +13,34 @@ namespace main {
 void power_supply_DCImpl::init() {
     this->connector_voltage = 0.0;
     this->connector_current = 0.0;
+    this->init_capabilities();
 
     this->power_supply_thread_handle = std::thread(&power_supply_DCImpl::power_supply_worker, this);
 }
 
+void power_supply_DCImpl::init_capabilities() {
```
a-w50, line 21: I wouldn't put this into a member function, as it potentially increases coupling.  I thought of something like
```
static auto create_default_capabilities_from_config(const Conf& config) {
    types::power_supply_DC::Capabilities c;
    c.bidirectional = config.bidirectional;
    c.current_regulation_tolarance_A = 2.0;
    ...
    return c;
}
```
and then
```
types::power_supply_DC::Capabilities power_supply_DCImpl::handle_getCapabilities() {
    return create_default_capabilities_from_config(this->config);
}
```
If necessary, you could also keep the capabilities as a member variable, but here I guess it is not.

corneliusclaussen (approved):
Needs to be rebased and squashed before merging

PR #512, opened 2024-01-25
Improve car manufacturer detection for Tesla

mhei:
Tesla has acquired multiple OUIs according to offical OUI list from IEEE. Additionally, Tesla has a /28 sub-range.

### DIFF
--- a/modules/EvseManager/CarManufacturer.cpp
+++ b/modules/EvseManager/CarManufacturer.cpp
@@ -1,16 +1,37 @@
 // SPDX-License-Identifier: Apache-2.0
 // Copyright 2020 - 2022 Pionix GmbH and Contributors to EVerest
 #include "CarManufacturer.hpp"
+#include <set>
 
 namespace module {
 
 types::evse_manager::CarManufacturer get_manufacturer_from_mac(const std::string& mac) {
-    if (mac.size() < 8)
+    // Tesla OUIs according to http://standards-oui.ieee.org/oui.txt
+    const std::set<std::string> tesla = {
+        "0C:29:8F",
+        "4C:FC:AA",
+        "54:F8:F0",
+        "98:ED:5C",
+    };
+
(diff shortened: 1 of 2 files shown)


modules/EvseManager/CarManufacturer.cpp
```diff
@@ -1,16 +1,30 @@
 // SPDX-License-Identifier: Apache-2.0
 // Copyright 2020 - 2022 Pionix GmbH and Contributors to EVerest
 #include "CarManufacturer.hpp"
+#include <set>
 
 namespace module {
 
 types::evse_manager::CarManufacturer get_manufacturer_from_mac(const std::string& mac) {
+    // Tesla OUIs according to http://standards-oui.ieee.org/oui.txt
+    std::set<std::string> tesla = {
```
corneliusclaussen, line 10: This could be "const"

modules/EvseManager/CarManufacturer.cpp
```diff
@@ -1,16 +1,30 @@
...
+        "4C:FC:AA",
+        "54:F8:F0",
+        "98:ED:5C",
+    };
+
     if (mac.size() < 8)
         return types::evse_manager::CarManufacturer::Unknown;
     if (mac.substr(0, 8) == "00:7D:FA")
         return types::evse_manager::CarManufacturer::VolkswagenGroup;
-    if (mac.substr(0, 8) == "98:ED:5C")
```
corneliusclaussen, line 12: All the if's here: Now that we are changing the file anyway it would be a good moment to update it to everest coding style: For if we should always use {} even if it is only one line if. We should update all the if's in this file (not just the new ones)

corneliusclaussen (changes requested):

corneliusclaussen (approved):

corneliusclaussen:
Please rebase so that we can merge

mhei:
rebase done

PR #516, opened 2024-01-27
Fix build on platforms which require libatomic

mhei:
On some platforms, e.g. armel or mipsel, std::atomic<...> requires linking with -latomic since such platforms do not support lock-free instruction sequences.

Credits:
The CheckAtomic.cmake part was shamelessly borrowed from KDE's krita application:
https://github.com/KDE/krita/raw/2d6521792cf24413511be49b82618baa3c47543c/cmake/modules/CheckAtomic.cmake


corneliusclaussen (approved):
lgtm

ruifaria (approved):

mhei:
If there are no objections, I'd merge this within the next 24h.

PR #522, opened 2024-02-01
Fix connector lock unlock too early before relais open

corneliusclaussen:
In the current implementation, the motor lock is controlled merely by the CP state. It should however not open the lock while the output contactors are still closed. This PR changes the behaviour so that the unlock command is only sent if both the CP state is appropriate and the opening of contactors has been confirmed by the BSP.


corneliusclaussen:
Leaving in draft until it has been tested

lategoodbye:
Successful tested

lategoodbye (approved):

PR #528, opened 2024-02-05
Adapt unit tests

andistorm:
Adapt unit tests according https://everest.github.io/nightly/general/faq.html#testing

See https://github.com/EVerest/EVerest/issues/129


hikinggrass:
There's a few other unit tests (eg. in the Auth module subdirectory) that should be similarly guarded

Pietfried (approved):

PR #531, opened 2024-02-07
Fix conversion of EVerest to OCPP KeyPair type

Pietfried:
Change required due to: https://github.com/EVerest/libocpp/pull/455

ocpp::KeyPair was extended so that certificate_single_path is added on conversion


Pietfried:
Requires this: https://github.com/EVerest/libocpp/pull/455 and + update of dependencies.yaml

AssemblyJohn (approved):

PR #533, opened 2024-02-07
EvseV2G: always assign GenChallenge and EVSETimeStamp in PaymentDetailsRes

barsnick:
When responding in PaymentDetailsRes with a FAILED ResponseCode, EVSETimeStamp and GenChallenge need to be assigned, even if unused. Especially GenChallenge needs to conform to the fixed size mandated by the schema.

Previously, GenChallenge was sent with size 0 in this case, which could lead to message validation errors on the receiving EV side.

Signed-off-by: Moritz Barsnick <moritz.barsnick@chargebyte.com>


SebaLukas (approved):
:+1:

PR #534, opened 2024-02-08
remove obsolete `rcd_enabled` config variable

barsnick:
Its implementation and uses were removed in commit e194737f06f7a5a90635046635f399fd4582cd46, but partly forgotten, or came back in with PR #357 in commit da8e936fc0bfb610fb96f22fec1af622ebb35bce.


SebaLukas (approved):
I think that the rcd_enabled has been forgotten in the bsp refactor PR. Thanks for noticing.

PR #535, opened 2024-02-09
Support modbus single register writes and long reads

### LINKED ISSUE #525: SerialCommHub doesn't support write_single_register command
As far as I can see the re is no way to send command WRITE_SINGLE_REGISTER to the Modbus device.

Theoretically, it it could be done by "write multiple registers" with count 1. But behaviour of the remote device can be different depending on command that is used. 

To my knowledge some functionality of [Iskra meter](https://www.iskra.eu/de/Energy-meters_2/Energy-meter-WM3M4-and-WM3M4C_1/) can be only used with WRITE_SINGLE_REGISTER command.

golovasteek:
Resolves #525 
Supersedes #441

### DIFF
--- a/interfaces/serial_communication_hub.yaml
+++ b/interfaces/serial_communication_hub.yaml
@@ -72,6 +72,30 @@ cmds:
       description: Status code of the transfer
       type: string
       $ref: /serial_comm_hub_requests#/StatusCodeEnum
+  modbus_write_single_register:
+    description: >-
+      Send a Modbus RTU 'write single register' command via serial interface
+      to the target hardware. (return value: response)
+    arguments:
+      target_device_id:
+        description: ID (1 byte) of the device to send the commands to
+        type: integer
+        minimum: 0
+        maximum: 255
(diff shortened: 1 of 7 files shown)


interfaces/serial_communication_hub.yaml
```diff
@@ -72,6 +72,30 @@ cmds:
       description: Status code of the transfer
       type: string
       $ref: /serial_comm_hub_requests#/StatusCodeEnum
+  modbus_write_signle_register:
```
barsnick, line 75: signle -> single

corneliusclaussen:
Supersedes https://github.com/EVerest/everest-core/pull/441: Should we close #441 then?

corneliusclaussen (approved):
Should be squashed before merging with proper commit description

interfaces/serial_communication_hub.yaml
golovasteek: Thanks. FIxed.

PR #537, opened 2024-02-09
Fix DC full power on reason and shutdown in PreCharge

corneliusclaussen:
In DC Charging, the power on reason was wrong for the actual charging in current demand. When stopping the sequence in PreCharge and the EV returns to state B the power supply was not immediately switched off. Both are fixed with this PR.


SebaLukas (approved):
Looks good :+1:

PR #542, opened 2024-02-13
Fix forced unlocking

corneliusclaussen:
Force unlocking was not working after the bsp refactoring. This PR fixes force unlocking in state B. Force unlocking in state C is not supported.

### DIFF
--- a/modules/EvseManager/IECStateMachine.cpp
+++ b/modules/EvseManager/IECStateMachine.cpp
@@ -415,7 +415,17 @@ void IECStateMachine::connector_unlock() {
 }
 
 void IECStateMachine::connector_force_unlock() {
-    force_unlocked = true;
+    RawCPState cp;
+
+    {
+        Everest::scoped_lock_timeout lock(state_machine_mutex, Everest::MutexDescription::IEC_force_unlock);
+        cp = cp_state;
+    }
+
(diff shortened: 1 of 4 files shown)


lategoodbye:
Thanks for addressing this issue. We also noticed this on Tarragon, but thought this was an issue in our BSP implementation.

Nevertheless i like to leave some comment regarding the implementation. From our understanding this command is the last resort for the EV driver who isn't able to get his cable back. So "force" should be considered as almost "unconditional", which from our understanding means always unlock the plug lock no matter what the CP state says. Since we cannot trust the mechanics of the plug lock and its sensors. The only exception should be that the EVSE contactor is closed.

In case this change is to big for this regression, i can open a new issue here.

corneliusclaussen:
It is just a one line change to support state C as well. Agreed that probably it is better to only forbid a force_unlock if the relais are closed as state C is beyond our control on the EVSE side.

corneliusclaussen:
Now force_unlock remains active until it is "normally" unlocked because state A/E/F is reached. Unlocking happens whenever relais are open and it is either force_unlocked or normally unlocked. It needs some testing still.

lategoodbye:
I will try to discuss this within our team so we can test this soon.

corneliusclaussen:
Tested on one HW now

SebaLukas (approved):
:+1:

PR #545, opened 2024-02-13
Charger: Fix relay welding handling

corneliusclaussen:
If the relais are welded, the BSP will not report a PowerOff event back. The Charger will continue to send allow_power_on(false) to the BSP at every mainloop run. This PR fixes this, if the relays are welded no further allow_power_on(false) is being sent.


SebaLukas (approved):
Looks good :+1:

PR #549, opened 2024-02-14
Adding the JsEvManager as a JsCarSimulator replacement

SebaLukas:
The aim is to rewrite the JsCarSimulator so that the new JsEvManager can also run on "real" hardware. The EV counterpart of the EvseManager module, so to speak.

For this purpose, a new ev_board_support interface and a suitable YetiEvDriver were written. The yeti_simulation_control was removed.

The new JsEvManager now also uses the ev_slac interface. The ev slac state machine and the JsSlacSimulator module have been adapted for this purpose.


interfaces/ev_board_support.yaml
```diff
@@ -0,0 +1,58 @@
+description: This defines the board support package for the EV side
+cmds:
+  enable:
+    description: Enable/disable the simulation
+    arguments:
+      value:
+        description: true to enable, false to disable
+        type: boolean
+  set_cp_state:
+    description: Sets the CP State that should be set by the EV Yeti driver (controlled by S2)
```
hikinggrass, line 10: Should probably just say "EV board support driver" instead of "Yeti"

modules/simulation/JsYetiSimulator/index.js
```diff
@@ -16,10 +14,9 @@ const STATE_F = 6;
 const Event_PowerOn = 8;
 const Event_PowerOff = 9;
 
-var module_id;
 let global_info;
 
-let active_errors = {
+const active_errors = {
```
hikinggrass, line 19: Is this really supposed to be const?

hikinggrass (approved):

corneliusclaussen (approved):
lgtm

SebaLukas:
Todo from my side before merge: 1. Fix integration test 2. squash all commits together

interfaces/ev_board_support.yaml
SebaLukas: Done :+1:

modules/simulation/JsYetiSimulator/index.js
SebaLukas: Yeah that should be not a const. Thank you for seeing that

SebaLukas:
I found a few things that still need to be changed here (node-red flows changes, PnC tests). Therefore I would suggest to move the merge after the release.

config/config-sil-ocpp201-pnc.yaml
```diff
@@ -80,33 +80,33 @@ active_modules:
       connector_id: 2
   slac:
     module: JsSlacSimulator
-  car_simulator_1:
-    module: JsCarSimulator
+  v_manager_1:
```
hikinggrass, line 83: typo: ev_manager_1

modules/simulation/JsEvManager/index.js
```diff
@@ -310,257 +203,274 @@ function simulation_loop(mod) {
...
-    mod.maxCurrent = c.args[0];
-    mod.numPhases = c.args[1];
+    mod.uses.ev_board_support.call.set_ac_max_current({ current: c.args[0] });
+    if (c.args[1] === 3) mod.uses.ev_board_support.call.set_three_phases({ three_phases: true });
+    else mod.uses.ev_board_support.call.set_three_phases({ three_phases: false });
     mod.state = 'charging_regulated';
     return true;
   });
 
+  // Todo(sl). Check power_fixed again
```
hikinggrass, line 309: Is that still relevant?

hikinggrass (changes requested):

modules/simulation/JsEvManager/index.js
SebaLukas: Is still relevant, but for a later PR. By moving the simulation parts from the CarSimulator to the YetiSimulator, the implementation of the power_fixed cmd is more difficult.

config/config-sil-ocpp201-pnc.yaml
SebaLukas: fixed

hikinggrass (approved):

modules/PySlacLoopback/manifest.yaml
```diff
@@ -0,0 +1,14 @@
...
+    description: Nothing to do here
+requires:
+  ev_slac:
+    interface: ev_slac
+  evse_slac:
+    interface: slac
+metadata:
+  license: https://opensource.org/licenses/Apache-2.0
+  authors:
+    - aw@pionix.de
```
a-w50, line 14: I can't remember, but I'm pretty sure I didn't write this module.

modules/PySlacLoopback/module.py
```diff
@@ -0,0 +1,31 @@
...
+        self._mod.call_command(evse_ff, 'enter_bcd', {})
+
+        time.sleep(WAIT_SECONDS)
+        log.info('Calling trigger_matching on EvSlac')
+        self._mod.call_command(ev_ff, 'trigger_matching', {})
+
+
+LoopbackTestModule()
+
+time.sleep(50)
```
a-w50, line 31: Have you tested this?  Doesn't this module terminate after 50 seconds and then everything crashes?

Pietfried (approved):

a-w50 (changes requested):
Some more remarks:
- would it be possible, to have the slac changes (`lib/staging/slac`) in a separate PR?
- naming could be a bit more consistent: YetiEVDriver and evyeti_comms
- moving the `evSerial` things into a separate library would reduce code duplication

hikinggrass:
> * moving the `evSerial` things into a separate library would reduce code duplication

There's [Everest::Serial in everest-framework](https://github.com/EVerest/everest-framework/blob/main/lib/serial.cpp) which was written with the intention to have a common implementation, but that hasn't been widely use (I think only in the pn532 rfid reader driver)

modules/PySlacLoopback/manifest.yaml
SebaLukas: I think you are the author of this. I only copied this module from the ccc-ev branch to this branch. But I can delete this module of course if we dont need that

modules/PySlacLoopback/module.py
SebaLukas: Not tested. But I can delete this

SebaLukas:
I think separating the lib/slac stuff from this PR into another PR is possible. I can do that right now

modules/PySlacLoopback/manifest.yaml
SebaLukas: I just deleted the PySlacLoopback module

modules/PySlacLoopback/module.py
SebaLukas: I just deleted the PySlacLoopback module

a-w50 (approved):

alpersefa:
2024.6.0 version of everest cannot use with auto generated EVerest Admin panel yaml files. It still generate as a JsCarSimulator.

PR #551, opened 2024-02-20
Bugfix/no wifi related tasks when not enabled

hikinggrass:
This fixes https://github.com/EVerest/everest-core/issues/486


james-ctc (approved):
ran a quick test and fix works fine.

PR #555, opened 2024-02-22
Setup module to support WPA3 Wi-Fi networks

james-ctc:
# background
`wpa_cli` supports the configuration of WPA2 and WPA3 networks. However WPA3 doesn't support a pre-shared key (the output from the wpa_passphrase command), it always requires the plain text passphrase.

By default the update will try and configure dual support for WPA2 and WPA3 based on the "psk": "" MQTT parameter
i.e.

psk|key_mgmt|configuration
---|-------|---------
""|NONE|open network
"&lt;length == 64&gt;"|WPA-PSK|WPA2 only with pre-shared key
"\"MyPassphrase\""|WPA-PSK WPA-PSK-SHA256 SAE|WPA2 and WPA3
"&lt;length &gt; 64&gt;"|SAE|WPA3 only

## additional
- unit tests now part of cmake build
- an additional MQTT parameter could be added to explicitly set the mode (WPA2, WPA3, both, none)<br>it is common for WiFi configuration to require the security being applied. Adding an additional parameter would allow EVerest to fully configure the network (and perhaps support more options: WEP, WPA2 Enterprise mode ...)

The WpaCliSetup has been updated to support network configuration with an additional parameter to indicate the network security to apply. This should make it easier to support new modes alongside an MQTT interface update.

### DIFF
--- a/modules/Setup/CMakeLists.txt
+++ b/modules/Setup/CMakeLists.txt
@@ -23,4 +23,9 @@ target_sources(${MODULE_NAME}
 
 # ev@c55432ab-152c-45a9-9d2e-7281d50c69c3:v1
 target_compile_features(${MODULE_NAME} PRIVATE cxx_std_17)
+if((${CMAKE_PROJECT_NAME} STREQUAL ${PROJECT_NAME} OR ${PROJECT_NAME}_BUILD_TESTING) AND BUILD_TESTING)
+    include(CTest)
+    add_subdirectory(tests)
+    set(CMAKE_BUILD_TYPE Debug)
+endif()
 # ev@c55432ab-152c-45a9-9d2e-7281d50c69c3:v1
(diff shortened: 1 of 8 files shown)


corneliusclaussen:
Not sure if related to this PR, but during testing I found that configured networks often publishes empty messages in between:
image

corneliusclaussen:
> Not sure if related to this PR, but during testing I found that configured networks often publishes empty messages in between: image

Not a bug, I had a second wireless device that had no configured networks

corneliusclaussen:
Tested on WPA3 and WPA2 and mixed WPA3/2 networks

james-ctc:
re-reading the README for the Setup module and the configured networks list should be a single list rather than a list per wifi network device.

corneliusclaussen:
> > Not sure if related to this PR, but during testing I found that configured networks often publishes empty messages in between: image
> 
> Not a bug, I had a second wireless device that had no configured networks

No a joined list of all interface is returned as described in the README

PR #556, opened 2024-02-22
EvseManager: Fix 6s timeout in C1

corneliusclaussen:
This fixes the logic when legacy wakeups are done after 30s and the 6s force power off under load


SebaLukas (approved):
:+1:

PR #561, opened 2024-02-23
Restructure OCPP and OCPP201 modules: Moved conversions

Pietfried:
* Added conversions headers and implementations for OCPP and OCPP201 modules
* Refactored OCPP and OCPP201 modules to use conversion functions instead of string conversions

### DIFF
--- a/modules/OCPP201/conversions.cpp
+++ b/modules/OCPP201/conversions.cpp
@@ -557,34 +557,6 @@ ocpp::v201::ReasonEnum to_ocpp_reason(types::evse_manager::StopTransactionReason
     }
 }
 
-ocpp::v201::ChargingStateEnum to_ocpp_charging_state_enum(types::evse_manager::StopTransactionReason reason) {
-    switch (reason) {
-    case types::evse_manager::StopTransactionReason::DeAuthorized:
-    case types::evse_manager::StopTransactionReason::EmergencyStop:
-    case types::evse_manager::StopTransactionReason::EnergyLimitReached:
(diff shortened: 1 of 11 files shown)


modules/OCPP201/conversions.cpp
```diff
@@ -0,0 +1,859 @@
...
+    case types::evse_manager::StopTransactionReason::TimeLimitReached:
+        return ocpp::v201::ReasonEnum::TimeLimitReached;
+    case types::evse_manager::StopTransactionReason::Timeout:
+        return ocpp::v201::ReasonEnum::Timeout;
+    default:
+        return ocpp::v201::ReasonEnum::Other;
+    }
+}
+
+ocpp::v201::ChargingStateEnum to_ocpp_charging_state_enum(types::evse_manager::StopTransactionReason reason) {
```
hikinggrass, line 560: That's a strange conversation, is that intentional?

hikinggrass (changes requested):
There's just this one strange conversion from types::evse_manager::StopTransactionReason to ocpp::v201::ChargingStateEnum. that's either accidental or needs a clearly state reason in a comment

modules/OCPP201/conversions.cpp
Pietfried: It was used to derive the Charging State from the StopTransactionReason for the TransactionEvent(Ended) event. But it looks like we have the charging state in libocpp internally, so we might want to remove this argument from `on_transaction_finished`.

Ill remove this function

hikinggrass (approved):

PR #563, opened 2024-02-25
Switch to CBV2G from OpenV2G for DIN, ISO-2, PnC and SAP

SiebrenW:
This is the initial work to switch from OpenV2G to CBV2G in the EvseV2g module.

Currently The following is done:

- [x] add cbv2g as a dependency
- [x] link cbv2g to EvseV2G module
- [x] switch SAP
- [x] switch DIN
- [x] switch ISO -2
- [x] switch xmldsig fragment encoding
- [x] Remove OpenV2G
- [x] Test din
- [x] Test iso2
- [x] Test PnC

### DIFF
--- a/modules/EvseV2G/CMakeLists.txt
+++ b/modules/EvseV2G/CMakeLists.txt
@@ -30,7 +30,9 @@ target_link_libraries(${MODULE_NAME}
         mbedcrypto
         mbedtls
         mbedx509
-        everest::openv2g
+        cbv2g::din
+        cbv2g::iso2
+        cbv2g::tp
 )
 
 target_sources(${MODULE_NAME}
--- a/modules/EvseV2G/din_server.cpp
+++ b/modules/EvseV2G/din_server.cpp
@@ -2,8 +2,9 @@
 // Copyright (C) 2023 chargebyte GmbH
 // Copyright (C) 2023 Contributors to EVerest
 
+#include <cbv2g/din/din_msgDefDatatypes.h>
+
 #include <inttypes.h>
-#include <openv2g/dinEXIDatatypes.h>
 #include <string.h>
 
 #include "din_server.hpp"
@@ -19,36 +20,36 @@
  * @param phy_value_dest is the destination struct.
  * @param phy_value_source is the source struct.
  */
-void load_din_physical_value(struct dinPhysicalValueType* const phy_value_dest,
-                             struct iso1PhysicalValueType* const phy_value_source) {
+void load_din_physical_value(struct din_PhysicalValueType* const phy_value_dest,
+                             struct iso2_PhysicalValueType* const phy_value_source) {
     phy_value_dest->Multiplier = phy_value_source->Multiplier;
     phy_value_dest->Value = phy_value_source->Value;
     phy_value_dest->Unit_isUsed = 1;
 
     switch (phy_value_source->Unit) {
-    case iso1unitSymbolType_h:
-        phy_value_dest->Unit = dinunitSymbolType_h;
+    case iso2_unitSymbolType_h:
+        phy_value_dest->Unit = din_unitSymbolType_h;
         break;
-    case iso1unitSymbolType_m:
-        phy_value_dest->Unit = dinunitSymbolType_m;
+    case iso2_unitSymbolType_m:
+        phy_value_dest->Unit = din_unitSymbolType_m;
         break;
-    case iso1unitSymbolType_s:
-        phy_value_dest->Unit = dinunitSymbolType_s;
+    case iso2_unitSymbolType_s:
+        phy_value_dest->Unit = din_unitSymbolType_s;
         break;
-    case iso1unitSymbolType_A:
-        phy_value_dest->Unit = dinunitSymbolType_A;
+    case iso2_unitSymbolType_A:
+        phy_value_dest->Unit = din_unitSymbolType_A;
         break;
-    case iso1unitSymbolType_V:
-        phy_value_dest->Unit = dinunitSymbolType_V;
+    case iso2_unitSymbolType_V:
+        phy_value_dest->Unit = din_unitSymbolType_V;
         break;
-    case iso1unitSymbolType_W:
-        phy_value_dest->Unit = dinunitSymbolType_W;
+    case iso2_unitSymbolType_W:
+        phy_value_dest->Unit = din_unitSymbolType_W;
         break;
-    case iso1unitSymbolType_Wh:
-        phy_value_dest->Unit = dinunitSymbolType_Wh;
+    case iso2_unitSymbolType_Wh:
+        phy_value_dest->Unit = din_unitSymbolType_Wh;
         break;
     default:
-        phy_value_dest->Unit = dinunitSymbolType_h;
+        phy_value_dest->Unit = din_unitSymbolType_h;
         break;
     }
 }
@@ -57,15 +58,15 @@ void load_din_physical_value(struct dinPhysicalValueType* const phy_value_dest,
  * \brief din_validate_state This function checks whether the received message is expected and valid at this
  * point in the communication sequence state machine. The current V2G msg type must be set with the current V2G msg
  * state. \param state is the current state of the charging session \param current_v2g_msg is the current handled V2G
- * message \param state of the actual session. \return Returns a dinResponseCode with sequence error if current_v2g_msg
+ * message \param state of the actual session. \return Returns a din_ResponseCode with sequence error if current_v2g_msg
  * is not expected, otherwise OK.
  */
-static dinresponseCodeType din_validate_state(int state, enum V2gMsgTypeId current_v2g_msg) {
+static din_responseCodeType din_validate_state(int state, enum V2gMsgTypeId current_v2g_msg) {
     /* if the request type has enabled (= set) bit in expected_requests, then this
      * message is ok at this stage; otherwise reject it.
      */
-    return (din_states[state].allowed_requests & (1 << current_v2g_msg)) ? dinresponseCodeType_OK
-                                                                         : dinresponseCodeType_FAILED_SequenceError;
+    return (din_states[state].allowed_requests & (1 << current_v2g_msg)) ? din_responseCodeType_OK
+                                                                         : din_responseCodeType_FAILED_SequenceError;
 }
 
 /*!
@@ -74,10 +75,10 @@ static dinresponseCodeType din_validate_state(int state, enum V2gMsgTypeId curre
  * external error has occurred. \param conn the structure with the external error information. \return Returns the next
  * v2g-event.
  */
-static v2g_event din_validate_response_code(dinresponseCodeType* const din_response_code,
+static v2g_event din_validate_response_code(din_responseCodeType* const din_response_code,
                                             struct v2g_connection const* conn) {
     enum v2g_event nextEvent = V2G_EVENT_NO_EVENT;
-    dinresponseCodeType response_code_tmp;
+    din_responseCodeType response_code_tmp;
 
     if (conn->ctx->is_connection_terminated == true) {
         dlog(DLOG_LEVEL_ERROR, "Connection is terminated. Abort charging");
@@ -86,7 +87,7 @@ static v2g_event din_validate_response_code(dinresponseCodeType* const din_respo
 
     /* If MQTT user abort or emergency shutdown has occurred */
     if ((conn->ctx->stop_hlc == true) || (conn->ctx->intl_emergency_shutdown == true)) {
-        *din_response_code = dinresponseCodeType_FAILED;
+        *din_response_code = din_responseCodeType_FAILED;
     }
 
     /* [V2G-DC-390]: at this point we must check whether the given request is valid at this step;
@@ -95,22 +96,22 @@ static v2g_event din_validate_response_code(dinresponseCodeType* const din_respo
      * error path, since it might not be set, yet!
      */
     response_code_tmp = din_validate_state(conn->ctx->state, conn->ctx->current_v2g_msg);
-    *din_response_code = (response_code_tmp >= dinresponseCodeType_FAILED) ? response_code_tmp : *din_response_code;
+    *din_response_code = (response_code_tmp >= din_responseCodeType_FAILED) ? response_code_tmp : *din_response_code;
 
     /* [V2G-DC-391]: check whether the session id matches the expected one of the active session */
     *din_response_code = ((conn->ctx->current_v2g_msg != V2G_SESSION_SETUP_MSG) &&
                           (conn->ctx->evse_v2g_data.session_id != conn->ctx->ev_v2g_data.received_session_id))
-                             ? dinresponseCodeType_FAILED_UnknownSession
+                             ? din_responseCodeType_FAILED_UnknownSession
                              : *din_response_code;
 
     if ((conn->ctx->terminate_connection_on_failed_response == true) &&
-        (*din_response_code >= dinresponseCodeType_FAILED)) {
+        (*din_response_code >= din_responseCodeType_FAILED)) {
         nextEvent = V2G_EVENT_SEND_AND_TERMINATE; // [V2G-DC-665]
     }
 
     /* log failed response code message */
-    if (*din_response_code >= dinresponseCodeType_FAILED &&
-        *din_response_code <= dinresponseCodeType_FAILED_WrongEnergyTransferType) {
+    if (*din_response_code >= din_responseCodeType_FAILED &&
+        *din_response_code <= din_responseCodeType_FAILED_WrongEnergyTransferType) {
         dlog(DLOG_LEVEL_ERROR, "Failed response code detected for message \"%s\", error: %s",
              v2g_msg_type[conn->ctx->current_v2g_msg], dinResponse[*din_response_code]);
     }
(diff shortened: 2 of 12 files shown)


modules/EvseV2G/CMakeLists.txt
```diff
@@ -30,7 +30,10 @@ target_link_libraries(${MODULE_NAME}
         mbedcrypto
         mbedtls
         mbedx509
-        everest::openv2g
+        cb_v2g::DIN
+        cb_v2g::ISO2
+        cb_v2g::ISO20
```
SebaLukas, line 35: The module does not require ISO20

.clangd
SebaLukas, line 1: You can add this file to .gitignore but dont commit this kind of file

SebaLukas (commented):
For now, the PR looks really good. I have just successfully started and finished a DIN session in the SIL. Now the code has to be tested a lot in the SIL and on real cars. I'll be doing that over the next few weeks

.clangd
SiebrenW: Sorry, that was by accident. I'll just move it down.

modules/EvseV2G/iso_server.cpp
```diff
@@ -2435,10 +2366,10 @@ enum v2g_event iso_handle_request(v2g_connection* conn) {
 
         /* Configure session id */
         memcpy(exi_out->V2G_Message.Header.SessionID.bytes, &conn->ctx->evse_v2g_data.session_id,
-               iso1MessageHeaderType_SessionID_BYTES_SIZE);
+               iso2_sessionIDType_BYTES_SIZE);
 
-        /* We always set bytesLen to iso1MessageHeaderType_SessionID_BYTES_SIZE */
-        exi_out->V2G_Message.Header.SessionID.bytesLen = iso1MessageHeaderType_SessionID_BYTES_SIZE;
+        /* We always set bytesLen to iso2_MessageHeaderType_SessionID_BYTES_SIZE */
+        exi_out->V2G_Message.Header.SessionID.bytesLen = iso2_sessionIDType_BYTES_SIZE - 1;
```
SebaLukas, line 2372: -1 causes a Failed_UnknowSession, as the byte length is set to 7. And the car sends a false SessionID with only 7 bytes. Therefore delete -1.

SiebrenW: Good catch.
Fixed. That creapt in because I wasn't sure about the null terminator initially. Removed it for the chars, but I must have added it to the bytes accidentally.

modules/EvseV2G/din_server.cpp
```diff
@@ -916,27 +916,25 @@ static enum v2g_event handle_din_session_stop(struct v2g_connection* conn) {
...
+    struct din_exiDocument* exi_in = conn->exi_in.dinEXIDocument;
+    struct din_exiDocument* exi_out = conn->exi_out.dinEXIDocument;
     enum v2g_event next_v2g_event = V2G_EVENT_TERMINATE_CONNECTION; // ERROR_UNEXPECTED_REQUEST_MESSAGE;
 
     /* check whether we have a valid EXI document embedded within a V2G message */
-    if (!exi_in->V2G_Message_isUsed) {
-        dlog(DLOG_LEVEL_ERROR, "V2G Message not used. Ignoring msg");
-        return V2G_EVENT_IGNORE_MSG;
-    }
+    // if (!exi_in->V2G_Message_isUsed) {
```
SiebrenW, line 924: Should remove

SebaLukas (commented):
I found a few little things

SebaLukas:
I was able to successfully test DIN and ISO-2 AC & DC EIM with the SIL. Only with PnC there seems to be a problem with the authorization. I'll have a look at that soon. Once this has been resolved, I would also test this PR on a few cars.

Please update/rebase your branch to eversest-core main. Then there should be no more conflict with the dependencies.yaml.

SiebrenW:
I had a few errors building with main, which is why this was rebased to the release initially. I didn't think it was going to cause merge conflicts in the actual module code.
I will rebase to main hopefully later today (and cleanup those comments).

Then the iso2_ParameterSetType_5_ARRAY_SIZE being insufficient, I would prefer to keep using the constants from cbv2g, so I will wait for that update.
I marked master as the dependency tag, so we will see a compile error once this number (and thus the name) has changed and fix it then and there.

SebaLukas:
You can also just clone everest-core and during cmake `edm` will take care of pulling the right dependencies with the right versions. Then building should not really be a problem with different dependency versions.

I totally agree with `iso2_ParameterSetType_5_ARRAY_SIZE`. This should be fixed in libcbv2g and not here. That was actually just a reminder for me not to forget this.

SiebrenW:
I will try again then when I get home. Last time I tried to build master as is it didn't like my effort.
Or do you mean without everest-workspace? The readme does not cite this as an option, so I never tried.

SebaLukas:
You have two options: 
Like the readme says, create a everest-workspace with edm init and get all necessary repos/libs in one folder.

Or you can clone only the repos you work on. For me it is usually:
```
.
├── everest-core
├── Josev
└── libiso15118
```
First, edm checks whether all necessary repos are stored locally in the same folder as everest-core. If not, then edm download the missing repos in the $CPM_SOURCE_CACHE folder or build folder.

The second option should be somewhere in our documentation.

modules/EvseV2G/v2g.hpp
```diff
@@ -262,15 +253,15 @@ struct v2g_context {
         uint8_t evse_processing[PHASE_LENGTH];
         struct v2g_evse_id evse_id;
         unsigned int date_time_now_is_used;
-        struct iso1ChargeServiceType charge_service;
-        struct iso1ServiceType evse_service_list[iso1ServiceListType_Service_ARRAY_SIZE];
-        struct iso1ServiceParameterListType service_parameter_list[iso1ServiceListType_Service_ARRAY_SIZE];
+        struct iso2_ChargeServiceType charge_service;
+        struct iso2_ServiceType evse_service_list[iso2_ServiceType_8_ARRAY_SIZE];
+        struct iso2_ServiceParameterListType service_parameter_list[iso2_ParameterSetType_5_ARRAY_SIZE];
```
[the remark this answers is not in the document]
SiebrenW: I see that I did use a variable that is not comparable directly now. I will fix that too, but won't rebase it for now.
Though before I don't see what the relation is between iso1ServiceListType_Service_ARRAY_SIZE and iso1ServiceParameterListType either.

The direct alternative to the OpenV2G iso1ServiceListType_Service_ARRAY_SIZE is iso2_ServiceType_8_ARRAY_SIZE.
Why? It has little to do with each other. A hardcoded 8 would make more sense, or am I missing some relation between the size of the array of an array of ServiceParameters versus an array of ParameterSets.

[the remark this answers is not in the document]
SiebrenW: I found the reason. in v2g_ctx.cpp add_service_to_service_list the same index is used for both the evse_service_list and service_parameter_list.
Weird choice, but okay.

SiebrenW:
Question: does it make sense to enforce the requirement of unused variables in the SignedInfo type?
For instance to set ANY_isUsed = 0; in SignatureMethod?
I made use of the fact that the object needs no conversion, it's the same one, but the conversion did set some of these fields to 0 explicitly.
I will add this in another commit for now. We can always drop it.

SebaLukas:
Yes, add that :+1:

SebaLukas:
I activated the pr actions (A maintainer have to approve this for forks :shrug:)
The clang-format check failed. Can you run clang-format on your changes (in your everest-core folder) with:
```
docker run \
--volume "$(pwd):/source
ghcr.io/everest/everest-clang-format:latest \
-r /source
```
This Docker image is also used here in the actions. This ensures that the same clang-format version is used.

SiebrenW:
> I activated the pr actions (A maintainer have to approve this for forks 🤷) The clang-format check failed. Can you run clang-format on your changes (in your everest-core folder) with:
> 
> ```
> docker run \
> --volume "$(pwd):/source
> ghcr.io/everest/everest-clang-format:latest \
> -r /source
> ```
> 
> This Docker image is also used here in the actions. This ensures that the same clang-format version is used.

The command has some issues. The quote isn't even closed and when I do add it docker can't find the image.

modules/EvseV2G/iso_server.cpp
```diff
@@ -277,27 +211,28 @@ static void printMbedVerifyErrorCode(int AErr, uint32_t AFlags) {
...
     int err = 0;
-    const struct iso1SignatureType* sig = iso1_signature;
+    const struct iso2_SignatureType* sig = iso2_signature;
     unsigned char buf[MAX_EXI_SIZE];
-    size_t buffer_pos = 0;
-    const struct iso1ReferenceType* req_ref = &sig->SignedInfo.Reference.array[0];
-    bitstream_t stream = {MAX_EXI_SIZE, buf, &buffer_pos, 0, 8 /* Set to 8 for send and 0 for recv */};
+    const struct iso2_ReferenceType* req_ref = &sig->SignedInfo.Reference.array[0];
+    exi_bitstream_t stream;
+    exi_bitstream_init(&stream, buf, MAX_EXI_SIZE, 8, NULL);
```
SebaLukas, line 227: Please change the 8 to a zero. That was the reason why PnC didn't work. The Exi fragment was encoded with an 8 byte offset, so of course the digest value was no longer correct.

SebaLukas:
Now PnC works also in the SIL. All that's left now are the tests with real cars.

SebaLukas:
> 
> The command has some issues. The quote isn't even closed and when I do add it docker can't find the image.
> 
> I did the clang check manually with git-clang-format HEAD~4. I fixed it in one of the commits.

Sorry about the incorrect clang-format cmd. I always use podman, so my cmd is slightly different.

That should work here:
```
docker run --user "1000:100" --volume "$(pwd):/source" http://ghcr.io/everest/everest-clang-format:latest -r /source --extensions "hpp,cpp" -e "/source/cache" -e "/source/build" -e "/source/build-cross" -e "/source/build-clang" -i
```

Docker should find the image, I was able to pull it normally with podman earlier.

modules/EvseV2G/iso_server.cpp
SiebrenW: Well. That's embarressing.
I didn't even copy paste that. I actually typed 8 for some reason.

SiebrenW:
The 8 has be turned into a 0 🤦 
If the SIL was better documented I would have tested this myself. I am already happy to be able to test DC BPT with nodered.

Docker is still complaining with docker: invalid reference format. I think my git-clang-tidy was effective anyways.

barsnick:
> ~Then the iso2_ParameterSetType_5_ARRAY_SIZE being insufficient, I would prefer to keep using the constants from cbv2g, so I will wait for that update.~ (or not [#563 (comment)](https://github.com/EVerest/everest-core/pull/563#discussion_r1516819337) )
> I marked master as the dependency tag, so we will see a compile error once this number (and thus the name) has changed and fix it then and there.

https://github.com/EVerest/cbexigen/pull/74

corneliusclaussen:
Will tag as post-release, I think we need a lot more testing and it should be merged after the march stable release

SebaLukas:
Todo from my side before merge: 1. Fix integration test 2. squash all commits together

SebaLukas:
Hi :)

unfortunately I haven't had time to test your PR with real cars yet.
In a week and a half the Charin Testival will take place in Poland and some of my Pionix colleagues will be there, too.

I would also like to test your PR with the cars on site.

I cloned your fork earlier and updated locally your branch to the current main. And created an extra module “EvseV2GLibCbV2G” with a copy of your EvseV2G module. I would then use this module for testing. Then it is easy to quickly change and test the modules through the config.

If there are no problems at the testival, then we can definitely merge your PR afterwards.

SebaLukas:
We were actually able to test everything well at the Testival (except PnC) and we only found one error. I was able to fix this error straight away. We were able to successfully test PnC with an Ioniq6 last week. So nothing speaks against a merge :+1: 

I would update this branch in the next few days and add the fixes we needed for the Testival. And merge this PR after that.

I hope that I can create another release for the libcbv2g this week, then we can use the new libcbv2g version `0.1.1` right away.

SebaLukas:
For me, nothing speaks against a merge. @corneliusclaussen & @barsnick could you take another look and see if you notice anything?

modules/EvseV2G/v2g_ctx.cpp
```diff
@@ -8,18 +8,19 @@
 #include <string.h>
 #include <unistd.h> // sleep
 
+#include "cbv2g/iso_2/iso2_msgDefDatatypes.h"
```
barsnick, line 11: Alternatively
```suggestion
#include <cbv2g/iso_2/iso2_msgDefDatatypes.h>
```

modules/EvseV2G/v2g_server.cpp
```diff
@@ -380,10 +377,11 @@ int v2g_handle_connection(struct v2g_connection* conn) {
 
     do {
         /* setup for receive */
-        conn->stream.buffer = 0;
-        conn->stream.capacity = 0; // Set to 8 for send and 0 for recv
-        conn->buffer_pos = 0;
+        conn->stream.data[0] = 0;
+        conn->stream.bit_count = 0;
+        conn->stream.byte_pos = 0;
```
barsnick, line 382: Might use `exi_bitstream_reset()`, but okay if it works.

barsnick (commented):
Thanks for the big effort!

It looks pretty good from what I can tell. I cannot judge on the fragment (signature) encoding, but fine since you tested it.

Only small remarks, mostly regarding the #include style.

modules/EvseV2G/v2g_server.cpp
SebaLukas: Good point :+1: But I think we can delete line 381 and 382 because in line 383 `exi_bitstream_init()` is called. In the init function `bit_count` and `byte_pos` is set to 0 anyway.

modules/EvseV2G/iso_server.cpp
```diff
@@ -2,22 +2,23 @@
...
-#include <openv2g/iso1EXIDatatypesEncoder.h>
-#include <openv2g/v2gtp.h> //for V2GTP_HEADER_LENGTHs
-#include <openv2g/xmldsigEXIDatatypes.h>
-#include <openv2g/xmldsigEXIDatatypesEncoder.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
 
+#include "cbv2g/common/exi_bitstream.h"
+#include "cbv2g/iso_2/iso2_msgDefDatatypes.h"
```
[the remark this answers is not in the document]
SebaLukas: Good point. I will change the three includes to `#include <`.

modules/EvseV2G/v2g_ctx.cpp
SebaLukas: Done

modules/EvseV2G/iso_server.cpp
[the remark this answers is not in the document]
SebaLukas: Done

modules/EvseV2G/v2g_server.cpp
SebaLukas: Done

modules/EvseV2G/iso_server.cpp
```diff
@@ -17,8 +17,8 @@
 #include <stdlib.h>
 #include <string.h>
 
-#include "cbv2g/common/exi_bitstream.h"
-#include "cbv2g/iso_2/iso2_msgDefDatatypes.h"
+#include <cbv2g/common/exi_bitstream.h>
+#include <cbv2g/iso_2/iso2_msgDefDatatypes.h>
```
barsnick: Move this up with the other includes? I would have thought clang-format might automatically re-order anyway (in a different way, BTW) - therefore the PR check will fail anyway. ;-)

modules/EvseV2G/v2g_ctx.cpp
```diff
@@ -8,7 +8,7 @@
 #include <string.h>
 #include <unistd.h> // sleep
 
-#include "cbv2g/iso_2/iso2_msgDefDatatypes.h"
+#include <cbv2g/iso_2/iso2_msgDefDatatypes.h>
```
barsnick: Move this up with the other includes? I would have thought clang-format might automatically re-order anyway (in a different way, BTW) - therefore the PR check will fail anyway. ;-)

modules/EvseV2G/iso_server.cpp
SebaLukas: I noticed that, too. This should be fixed with the recent commit

modules/EvseV2G/v2g_ctx.cpp
SebaLukas: I noticed that, too. This should be fixed with the recent commit

modules/EvseV2G/iso_server.cpp
```diff
@@ -1,24 +1,25 @@
...
-#include <openv2g/v2gtp.h> //for V2GTP_HEADER_LENGTHs
-#include <openv2g/xmldsigEXIDatatypes.h>
-#include <openv2g/xmldsigEXIDatatypesEncoder.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
 
-#include "iso_server.hpp"
 #include "log.hpp"
 #include "tools.hpp"
```
barsnick, line 24: Now I don't know what the h*ll clang-format is trying to do. :-D
#include "iso_server.hpp"
belongs further down, to where it used to be, above log.hpp.

barsnick (approved):
Just the minor #include nit.

modules/EvseV2G/iso_server.cpp
SebaLukas: Yes, that was me, I'll change it back.

SebaLukas (approved):
:+1:

corneliusclaussen (approved):
Looks good to me! We also did quite some funtional testing here.

SebaLukas:
I would now squash the commits together and then merge them. Thank you @SiebrenW  for creating this PR :smiley:

PR #569, opened 2024-02-29
feat: add MQTT topic to publish OCPP transaction state and ID

james-ctc:
Addition to the OCPP module (1.6) to provide:

- start transaction
- transaction ID (from the CSMS)
- end transaction

as a published topic over MQTT


james-ctc:
updated to include OCPP 2.0.1
depends on update in libocpp https://github.com/EVerest/libocpp/pull/509

modules/OCPP/ocpp_transaction/transactionImpl.cpp
hikinggrass, line 1: Looks like this isn't used anymore

modules/OCPP/ocpp_transaction/transactionImpl.hpp
hikinggrass, line 1: Looks like this isn't used anymore

hikinggrass (changes requested):

modules/OCPP/ocpp_transaction/transactionImpl.cpp
james-ctc: good spot, removed

modules/OCPP/ocpp_transaction/transactionImpl.hpp
james-ctc: good spot, removed

types/ocpp.yaml
```diff
@@ -451,13 +484,42 @@ types:
...
+    properties:
+      event:
+        description: >-
+          The transaction related event.
+        type: string
+        enum:
+          - Start
+          - Update
+          - End
+      connector:
```
hikinggrass, line 511: This should probably also include the evse since it's being used in OCPP 2.0.1 as well

modules/OCPP201/conversions.cpp
```diff
@@ -827,5 +827,27 @@ to_everest_certificate_status(const ocpp::v201::AuthorizeCertificateStatusEnum s
...
+        break;
+    case ocpp::v201::TransactionEventEnum::Updated:
+        ocpp_transaction_event.transaction_event = types::ocpp::TransactionEvent::Updated;
+        break;
+    case ocpp::v201::TransactionEventEnum::Ended:
+        ocpp_transaction_event.transaction_event = types::ocpp::TransactionEvent::Ended;
+        break;
+    }
+    ocpp_transaction_event.evse_id = transaction_event.evse.value().id;
+    ocpp_transaction_event.connector = 1;
```
hikinggrass, line 845: if the connectorId is available this should probably use the one from transaction_event.evse.value().connectorId

hikinggrass (changes requested):

types/ocpp.yaml
james-ctc: The evse id is included as an optional element in types/ocpp.yaml
perhaps there has been a merge issue?

modules/OCPP201/conversions.cpp
james-ctc: not sure where this change comes from. I don't remember changing this fine.
a merge issue perhaps?

hikinggrass (approved):

PR #573, opened 2024-03-03
Support starting transaction in EvseManager

golovasteek:
resolves https://github.com/EVerest/everest-core/issues/572

The solution presented here is suitable for basic use-case under German Measurement law.

Prior to starting charging on BSP, the EvseManager will request meter to start transaction.
Starting the transaction on the power meter is expected to return quicly (by either spawning some process in the meter driver, or initiating the transaction on the meter hardware).

The signed meter values for start transaction are not expected to be reported by the meter when transaction is requested.

There are different meter models, and signed meter values for transaction start, may be a separate "SignedMeterValue" object. in this case the `start_signed_meter_value` of "StopTransactionResponse" can be used. And the `signed_meter_value` would contain data for the session finish.
 
 In many cases the meter only issue one signed dataset for session, that contains multiple snapshots encoded, in this case only `signed_meter_value` is used.
 
 
 What is not covered by this PR: 
   * OCPP module as well as libOCPP need to be adjusted, to use "start_signed_meter_value" from `TransactionFinished` event
   * the meter values for the transaction are only transferred at the end of the session, there is no support for asynchronous sending of meter values introduced. Which needs to be designed spearately, once we have better requiremetns from the OCPP of different versions.

### DIFF
--- a/modules/EvseManager/Charger.cpp
+++ b/modules/EvseManager/Charger.cpp
@@ -10,6 +10,7 @@
 
 #include "Charger.hpp"
 
+#include <generated/types/powermeter.hpp>
 #include <math.h>
 #include <string.h>
 #include <thread>
@@ -23,8 +24,13 @@
 namespace module {
 
 Charger::Charger(const std::unique_ptr<IECStateMachine>& bsp, const std::unique_ptr<ErrorHandling>& error_handling,
-                 const types::evse_board_support::Connector_type& connector_type) :
-    bsp(bsp), error_handling(error_handling), connector_type(connector_type) {
+                 const std::vector<std::unique_ptr<powermeterIntf>>& r_powermeter_billing,
+                 const types::evse_board_support::Connector_type& connector_type, const std::string& evse_id) :
+    bsp(bsp),
+    error_handling(error_handling),
+    r_powermeter_billing(r_powermeter_billing),
+    connector_type(connector_type),
+    evse_id(evse_id) {
 
 #ifdef EVEREST_USE_BACKTRACES
     Everest::install_backtrace_handler();
@@ -296,7 +302,8 @@ void Charger::run_state_machine() {
 
                 // If we are restarting, the transaction may already be active
                 if (not shared_context.transaction_active) {
-                    start_transaction();
+                    if (!start_transaction())
+                        break;
                 }
 
                 const EvseState target_state(EvseState::PrepareCharging);
@@ -377,7 +384,8 @@ void Charger::run_state_machine() {
                 }
             } else if (shared_context.authorized and shared_context.authorized_pnc) {
 
-                start_transaction();
+                if (!start_transaction())
+                    break;
 
                 const EvseState target_state(EvseState::PrepareCharging);
 
@@ -1013,6 +1021,7 @@ bool Charger::cancel_transaction(const types::evse_manager::StopTransactionReque
             signal_hlc_stop_charging();
         } else {
             shared_context.current_state = EvseState::ChargingPausedEVSE;
+            pwm_off();
         }
 
         shared_context.transaction_active = false;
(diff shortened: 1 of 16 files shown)


cmake/everest-generate.cmake
```diff
@@ -177,6 +177,7 @@ if (EVEREST_ENABLE_RS_SUPPORT)
             EVEREST_CORE_ROOT="${CMAKE_CURRENT_SOURCE_DIR}"
             EVEREST_RS_FRAMEWORK_SOURCE_LOCATION="${everest-framework_SOURCE_DIR}"
             EVEREST_RS_FRAMEWORK_BINARY_LOCATION="${everest-framework_BINARY_DIR}"
+            EVEREST_CORE_ROOT="${CMAKE_SOURCE_DIR}"
```
corneliusclaussen, line 180: This is a re-definition of line 177

modules/EvseManager/Charger.cpp
```diff
@@ -993,11 +1001,32 @@ bool Charger::cancel_transaction(const types::evse_manager::StopTransactionReque
             shared_context.current_state = EvseState::ChargingPausedEVSE;
         }
 
+        // FIXME(evgeny): We disable pwm here since we need to get signed meter values.
+        // Maybe state machine should be refactored and cancelTransaction should trigger
+        // transition to EvseState::Finished, with active transaction.
+        // This way, the signed meter values will be retrieved there.
+        pwm_off();
```
corneliusclaussen, line 1008: I agree we should change the state machine flow here, also to make basic charging more aligned with ISO charging. But let's maybe do this in a separate PR, otherwise we cannot get this into the March release anymore as this requires a bit of thinking and testing to make sure we don't break anything else.

So to keep changes minimal for the moment I think it is enough to move the pwm_off() into the else line 1001. Then it would only stop PWM on basic charging and not trigger an emergency shutdown on ISO. On ISO we have sent the ISO message to the car that we would like to stop, but it of course takes a bit of time for the car to actually stop drawing power etc. The same is true for basic charging, worst case the car can draw power for 6s still in C1 until we force power off under load (IEC61851-1). I would accept that in this case though, the worst case means we give 6s away for free on non-cooperative cars. ISO shutdown should also be within few seconds.

With that in mind, the question is if PWM really needs to be switched off here at all. It will be switched off very quickly after the power meter reading anyway once we are in state ChargingPausedEVSE for basic charging. It does not hurt though as long as we only do it for basic charging and not for ISO.

modules/EvseManager/Charger.cpp
```diff
@@ -1023,20 +1052,66 @@ void Charger::stop_session() {
...
     shared_context.transaction_active = true;
+
+    const types::powermeter::TransactionReq req{evse_id, "", "", 0, 0, ""};
+    for (const auto& meter : r_powermeter_billing) {
+        const auto response = meter->call_start_transaction(req);
+        // If we want to start the session but the meter fail, we stop the charging since
+        // we can't bill the customer.
+        if (response.status == types::powermeter::TransactionRequestStatus::UNEXPECTED_ERROR) {
+            EVLOG_error << "Failed to start a transaction on the power meter " << response.error.value_or("");
+            set_faulted();
```
corneliusclaussen, line 1066: This should not set faulted directly, it should emit an error so that this can be handled by the error framework and e.g. forwarded to OCPP etc. There is an error definition already: PowermeterTransactionStartFailed in errors/evse_manager.yaml. Please rebase on RC1 before that as the error handling in EvseManager was changed in RC1.

modules/EvseManager/Charger.cpp
```diff
@@ -1023,20 +1052,66 @@ void Charger::stop_session() {
...
 }
 
+std::optional<types::units_signed::SignedMeterValue>
+Charger::take_signed_meter_data(std::optional<types::units_signed::SignedMeterValue>& in) {
+    std::optional<types::units_signed::SignedMeterValue> out;
+    std::swap(out, in);
+    return out;
+}
+
+std::optional<types::units_signed::SignedMeterValue> Charger::get_stop_signed_meter_value() {
```
corneliusclaussen, line 1107: The two getters may need a mutex locking depending on how they are used

corneliusclaussen (changes requested):

cmake/everest-generate.cmake
golovasteek: Removed

modules/EvseManager/Charger.cpp
golovasteek: Moved.

modules/EvseManager/Charger.cpp
golovasteek: Checked by trying to add locking, they don't need it, they are used only inside event handling, that is already under lock.

corneliusclaussen:
Next time if you change something during review, please do not squash commits. It is very hard to follow your changes and see if they are resolved now. The best way is to squash before opening the PR, then stack commits with changes during review on it and finally squash again as needed during the merging.

golovasteek:
> Next time if you change something during review, please do not squash commits. It is very hard to follow your changes and see if they are resolved now. The best way is to squash before opening the PR, then stack commits with changes during review on it and finally squash again as needed during the merging.

Makes sense.
Will do this way.

corneliusclaussen (approved):

SirVer (approved):

Pietfried (approved):

PR #580, opened 2024-03-07
RsIskraMeter: Initial commit

dorezyuk:
Pr adds support for Iskra's WM3M4 & WM3M4C power meters

- [x] Update the framework reference for bazel https://github.com/EVerest/everest-framework/pull/166


interfaces/powermeter.yaml
```diff
@@ -26,3 +26,6 @@ vars:
     description: Measured dataset
     type: object
     $ref: /powermeter#/Powermeter
+  public_key:
```
SirVer, line 29: @corneliusclaussen @hikinggrass @Pietfried Can we get an opinion from Pionix on this change? It seems straight forward enough for us, but just wanted to hear from you.

SirVer (approved):

SirVer:
@corneliusclaussen @Pietfried @hikinggrass This is supporting the Iskra Meter, including everything needed for german eichrecht (i.e. signed meter values). We use this one in production and would like to contribute to the eco system.

SirVer:
@dorezyuk this needs merging of main.

modules/RsIskraMeter/src/main.rs
```diff
@@ -0,0 +1,1034 @@
...
+        self.check_signature_status()?;
+        let signature = self.read_signature()?;
+        let signed_meter_values = self.read_signed_meter_values()?;
+        Ok(TransactionStartResponse {
+            error: Option::None,
+            signed_meter_value: Some(SignedMeterValue {
+                signed_meter_data: create_ocmf(signed_meter_values, signature),
+                signing_method: "OCMF".to_string(),
+                encoding_method: String::new(),
+                public_key: None,
```
hikinggrass, line 593: If possible you can include the public key that's published in ready() here as well

modules/RsIskraMeter/src/main.rs
```diff
@@ -0,0 +1,1034 @@
...
+        self.check_signature_status()?;
+        let signature = self.read_signature()?;
+        let signed_meter_values = self.read_signed_meter_values()?;
+        Ok(TransactionStopResponse {
+            error: Option::None,
+            signed_meter_value: Some(SignedMeterValue {
+                signed_meter_data: create_ocmf(signed_meter_values, signature),
+                signing_method: "OCMF".to_string(),
+                encoding_method: String::new(),
+                public_key: None,
```
hikinggrass, line 642: If possible you can include the public key that's published in ready() here as well

modules/RsIskraMeter/src/main.rs
```diff
@@ -0,0 +1,1034 @@
...
+use generated::{get_config, Module, SerialCommunicationHubClientPublisher};
+use std::fmt::Debug;
+use std::sync::{Arc, Mutex};
+use std::time::Duration;
+use utils::{
+    counter, create_ocmf, create_random_meter_session_id, from_t5_format, from_t6_format,
+    string_to_vec, to_8_string, to_hex_string,
+};
+
+/// Public key prefix for transparency software, defined under 6.5.14.
```
hikinggrass, line 61: Could these definitions and conversions be moved into another file for readability?

hikinggrass (changes requested):

modules/RsIskraMeter/src/main.rs
dorezyuk: will do -  shall we then actually drop the public key topic?

modules/RsIskraMeter/src/main.rs
```diff
@@ -0,0 +1,1034 @@
...
+            frequency_hz_signed: None,
+            power_w_signed: None,
+            signed_meter_value: None,
+            var_signed: None,
+            voltage_v_signed: None,
+        };
+        Ok(resp)
+    }
+
+    fn write_metadata(&self, evse_id: &str) -> Result<()> {
```
lategoodbye, line 490: From my understanding the billing info must be prepared with 4 parameters:
GI gateway identification
GS gateway serial
GV gateway version
CI charge point identification

So we need more parameter here

modules/RsIskraMeter/src/main.rs
```diff
@@ -0,0 +1,1034 @@
...
+        // Warning: JSON names must be in specified order and without whitespaces.
+        let message: String = format!("{{\"FV\":\"1.0\",\"GI\":\"\",\"GS\":\"\",\"PG\":\"\",\"MV\":\"\",\"MM\":\"\",\"MS\":\"\",\"MF\":\"\",\"IS\":true,\"IF\":[\"RFID_PLAIN\"],\"IT\":\"NONE\",\"ID\":\"{} {}\",\"CT\":\"EVSEID\",\"CI\":\"vestel\",\"RD\":[]}}", evse_id, create_random_meter_session_id());
```
lategoodbye, line 492: This string contains a lot hard coded values. Especially "vestel" doesn't make sense here.

modules/RsIskraMeter/src/utils.rs
```diff
@@ -0,0 +1,166 @@
...
+    input
+        .into_iter()
+        .flat_map(|value| value.to_be_bytes())
+        .map(|value| format!("{value:02X}"))
+        .collect::<Vec<_>>()
+        .concat()
+}
+
+pub fn create_ocmf(signed_meter_values: String, signature: String) -> String {
+    format!("OCMF|{}|{{\"SD\":\"{}\"}}", signed_meter_values, signature)
```
lategoodbye, line 72: A general comment because at this place everything seems to be put together:

`The sections within the data format are separated with the special pipe character "|". Within the sections, the format corresponds to the JSON notation. Thus, the use of the special pipe character "|" is not allowed within the sections.`

From my understanding we need to filter all the "|" from external input strings.

lategoodbye (changes requested):

modules/RsIskraMeter/src/utils.rs
dorezyuk: I don't think its like that - as long the `|` is in a escaped string within the JSON it should be fine

interfaces/powermeter.yaml
dorezyuk: We can revert this since the public key is now part of the signed meter message

modules/RsIskraMeter/src/main.rs
```diff
@@ -0,0 +1,1197 @@
...
+        let signature = self.read_signature()?;
+        let signed_meter_values = self.read_signed_meter_values()?;
+        Ok(TransactionStartResponse {
+            error: Option::None,
+            signed_meter_value: Some(SignedMeterValue {
+                signed_meter_data: create_ocmf(signed_meter_values, signature),
+                signing_method: "OCMF".to_string(),
+                encoding_method: String::new(),
+                public_key: self.read_public_key().ok(),
+                timestamp: None,
```
dorezyuk, line 730: @hikinggrass @golovasteek should we set also the tiemstamps?

modules/RsIskraMeter/src/main.rs
dorezyuk: below most stuff is derived from the iskra manual linked at the top of the file... splitting it up would probably cause readers to flip back and forth between the enums and the impl - but I don't have a strong opinion here

 @hikinggrass  how much does it bother you ?

modules/RsIskraMeter/src/main.rs
```diff
@@ -0,0 +1,1197 @@
...
+    /// signature.
+    pagination: String,
+
+    #[serde(rename = "MV")]
+    /// Manufacturer identification of the meter, name of the manufacturer.
+    meter_vendor: String,
+
+    #[serde(rename = "MM")]
+    /// Model identification of the meter.
+    meter_model: String,
```
dorezyuk, line 274: we should probably also meter_model and meter_serial (we read them) and also meter_vendor (iskra?) and probably also meter_firmware (read under register 30013) - opinions @golovasteek @hikinggrass ?

SirVer: we should what?

dorezyuk: we should not forget the verb in a sentence

modules/RsIskraMeter/src/main.rs
```diff
@@ -0,0 +1,1197 @@
...
+
+        self.check_signature_status()?;
+        let signature = self.read_signature()?;
+        let signed_meter_values = self.read_signed_meter_values()?;
+        Ok(TransactionStartResponse {
+            error: Option::None,
+            signed_meter_value: Some(SignedMeterValue {
+                signed_meter_data: create_ocmf(signed_meter_values, signature),
+                signing_method: "OCMF".to_string(),
+                encoding_method: String::new(),
```
golovasteek, line 728: If I understand correctly, in "lem_xxxx" meter here https://github.com/EVerest/everest-core/blob/a6e7ad88a753a2829749e1b1dee00ba3a6c6862e/modules/LemDCBM400600/main/lem_dcbm_400600_controller.cpp#L89

The signing method is "" and "OCMF" goes to `encoding_method`.

@hikinggrass what is right?

modules/RsIskraMeter/src/main.rs
golovasteek: Generally it makes sense for consistency, but only if we decode the OCMF string and take it from there.

modules/RsIskraMeter/src/main.rs
```diff
@@ -0,0 +1,1197 @@
...
+        };
+        Ok(resp)
+    }
+
+    fn write_metadata(&self, evse_id: &str) -> Result<()> {
+        let mut ocmf_data = self.ocmf_data.clone();
+        ocmf_data.identification_data = format!("{} {}", evse_id, create_random_meter_session_id());
+
+        let message = everest_serde_json::to_string(&ocmf_data)?;
+        println!("{message}");
```
dorezyuk, line 628: remove

modules/RsIskraMeter/src/utils.rs
dorezyuk: I've checked some sample data containing `|` and the ocmf transparentzsoftware tool and indeed it will not accept any data with `|` ...

However, here we can't remove the `|` since the signature will not fit anymore. To my understanding the only source for a `|` would be from user's config (e.x. when specifying `GV` or similar) - so I would add a check there and filter the `|` there

modules/RsIskraMeter/src/main.rs
hikinggrass: The encoding_method describes the format in which the signed meter data was encoded, so in this case this would be "OCMF" and the signing_method can be left empty, so it's exactly like you suggested: needs to be flipped around

modules/RsIskraMeter/src/main.rs
hikinggrass: Ok that makes sense, it's not a dealbreaker :wink:

PR #582, opened 2024-03-13
Added charging schedules definition

james-ctc:
## Describe your changes

Create YAML/JSON definition for charging schedules.

Note this is an API breaking change. Originally charging schedules were published:
```json    
    { "0": {...}, "1": {...}}
```

Using the connector number as the key. This is not supported in the YAML interface or type definitions.

Instead an array is used with an additional key added:
```json
    [ { "connector": 0, ...}, { "connector": 1, ...} ... ]
```

Unfortunately the generated code doesn't like arrays at that point, so the array has been wrapped in an object:
```json
    { "schedules": [ { "connector": 0, ...}, { "connector": 1, ...} ... ] }
```

Note is should not be assumed that the array index will match the connector number.
    
The existing types in libocpp have not been changed. A new mapping has been created in the OCPP module to convert from the internal format to the MQTT JSON format.

Note OCPP201 doesn't publish schedules at the moment and has not changed.

None


types/ocpp.yaml
```diff
@@ -2,6 +2,62 @@ description: >-
...
+      Element providing information on an OCPP charging schedule.
+    type: object
+    required:
+      - connector
+      - chargingRateUnit
+      - chargingSchedulePeriod
+    properties:
+      connector:
+        type: integer
+        minimum: 0
```
hikinggrass, line 33: I think it might be a good idea to already include the evse in this type definition in preparation for usage of this type with OCPP 2.0.1

Pietfried (changes requested):
Types and interfaces look good from my side.

Although the definitions in the OCPP spec are camelCase, we try to stick to snake_case in our internal interfaces. I have added some suggestions inline, but the code also needs to apply the changes.

@corneliusclaussen : Do we need changes on the EnergyManager side if we make these changes?

types/ocpp.yaml
james-ctc: added evseid as optional for OCPP 2.0.1

james-ctc:
> Do we need changes on the EnergyManager side if we make these changes?

It doesn't look like EnergyManager depends on the ocpp interface and I can't locate any MQTT subscribe for charging_schedules.

Pietfried (approved):

PR #590, opened 2024-03-19
EvseManager attempt to address errors seen with clearing events

james-ctc:
## Describe your changes
The following has been observed
```
Mar 06 08:31:16 manager[3415]: 2024-03-06 08:31:16.613914 [INFO] evse_manager:Ev  :: All errors cleared
Mar 06 08:31:16 manager[3415]: 2024-03-06 08:31:16.635325 [INFO] evse_manager:Ev  :: Cleared error evse_board_support/VendorError: Vendor specific error code. Will stop charging session. (FAULT_CT_CLAMP)
Mar 06 08:31:16 manager[3415]: 2024-03-06 08:31:16.636143 [INFO] evse_manager:Ev  :: All errors cleared
Mar 06 08:31:16 manager[3415]: 2024-03-06 08:31:16.669671 [INFO] evse_manager:Ev  :: Cleared error evse_board_support/VendorError: Vendor specific error code. Will stop charging session. (FAULT_CT_CLAMP)
Mar 06 08:31:16 manager[3415]: 2024-03-06 08:31:16.670540 [INFO] evse_manager:Ev  :: All errors cleared
```
This problem has been difficult to reproduce so unit tests were added and code reviewed to identify potential issues.

1. VendorError was missing from the all_cleared() check
2. the threads used to update shared_context in Charger.cpp. If both were blocked on state_machine_mutex then the ordering of events may not be maintained

Bit flags are used to replace individual boolean values since this makes the method all_cleared() trivial (is the value of the bit flags equal to 0) and should be easier to maintain should additional error types be added.

An event queue was used for pending updates to shared_context to ensure that event order is maintained.

Unit tests added for the new classes and to ensure correct operation of event handling.


modules/EvseManager/CMakeLists.txt
```diff
@@ -57,4 +57,9 @@ target_sources(${MODULE_NAME}
 
 # ev@c55432ab-152c-45a9-9d2e-7281d50c69c3:v1
 # insert other things like install cmds etc here
+if((${CMAKE_PROJECT_NAME} STREQUAL ${PROJECT_NAME} OR ${PROJECT_NAME}_BUILD_TESTING) AND BUILD_TESTING)
```
hikinggrass, line 60: This could also just check for EVEREST_CORE_BUILD_TESTING [which is set in the toplevel CMakeLists.txt](https://github.com/EVerest/everest-core/blob/aed024ba594d9842dacbc25355407c03ba7bffa7/CMakeLists.txt#L46)

modules/EvseManager/README.md
```diff
@@ -0,0 +1,17 @@
+# EvseManager documentation
+
+see `doc.rst`
+
+## Tests
+
+GTest is required for building the test cases target.
```
hikinggrass, line 7: GTest is actually [pulled in as a source-code dependency now](https://github.com/EVerest/everest-core/blob/aed024ba594d9842dacbc25355407c03ba7bffa7/dependencies.yaml#L85) so that comment probably isn't needed anymore

modules/EvseManager/tests/CMakeLists.txt
```diff
@@ -0,0 +1,33 @@
...
+    EventQueueTest.cpp
+    ../ErrorHandling.cpp
+)
+
+target_compile_definitions(${TEST_TARGET_NAME} PRIVATE
+    BUILD_TESTING_MODULE_EVSE_MANAGER
+)
+
+
+find_package(GTest REQUIRED)
```
hikinggrass, line 24: This find_package is not needed anymore

hikinggrass (changes requested):

modules/EvseManager/tests/CMakeLists.txt
james-ctc: it fails to build for me without it. Many link errors.

hikinggrass: If you replace the GTEST_LIBRARIES and GTEST_MAIN_LIBRARIES in the target_link_libraries with
 GTest::gtest_main
It should work, [this is done like this in other unit tests as well](https://github.com/EVerest/everest-core/blob/aed024ba594d9842dacbc25355407c03ba7bffa7/modules/Auth/tests/CMakeLists.txt#L17C4-L17C22)

hikinggrass (changes requested):

corneliusclaussen (commented):
This is a lot better then what we have right now. I added a commit to fix the three lines that it always prints on start up and unplug ("All errors cleared" along with an error that the error could not be cleared because it was not raised earlier). This was broken for a long time and I though if we touch this now anyway. @james-ctc Can you have a look at the commit I added?

james-ctc:
> Can you have a look at the commit I added?

The change looks good to me.

barsnick:
> I added a commit to fix the three lines that it always prints on start up and unplug ("All errors cleared" along with an error that the error could not be cleared because it was not raised earlier). This was broken for a long time and I though if we touch this now anyway.

This probably fixes https://github.com/EVerest/everest-framework/issues/156, right? I had tracked it down to everest-framework and assumed it needed to be fixed there.

In other words: Very welcome additional fix!

corneliusclaussen:
> > I added a commit to fix the three lines that it always prints on start up and unplug ("All errors cleared" along with an error that the error could not be cleared because it was not raised earlier). This was broken for a long time and I though if we touch this now anyway.
> 
> This probably fixes [EVerest/everest-framework#156](https://github.com/EVerest/everest-framework/issues/156), right? I had tracked it down to everest-framework and assumed it needed to be fixed there.
> 
> In other words: Very welcome additional fix!

yes it fixes that

hikinggrass (approved):
Looks good!

PR #599, opened 2024-03-25
Add QCA chip reset

corneliusclaussen:
## Describe your changes

In some scenarios, QCA7k requires a chip reset after setting up keys. This PR adds an option (not enabled by default) to send a RS_DEV.REQ custom vendor MMU to reset the chip. Works at least on QCA7k.

Requires add-qca-reset branch from libslac as well.


lategoodbye:
I heard about this problem, but never saw it by myself on a setup. So i'm interested in the details of the cause (QCA7K firmware version, Linux QCA driver, what's the symptom of the QCA7K which requires a reset).

corneliusclaussen:
> I heard about this problem, but never saw it by myself on a setup. So i'm interested in the details of the cause (QCA7K firmware version, Linux QCA driver, what's the symptom of the QCA7K which requires a reset).

I tried to understand for quite a while and gave up. It depends on the EV. Some EV never trigger the bug, some trigger it almost always (E.g. Volvo XC40 with a PLC sniffer attached, it does not do this without the sniffer though).
On most cars it happens quite seldomly (maybe once in 100 charging sessions or so).
 
It happens both with firmware 1.2.5 and 3.0.0 on the EVSE side, linux driver version:

qcaspi spi0.0: ver=0.2.7-i, clkspeed=12000000, burst_len=5000, pluggable=0

The symptoms are as follows:

After unplug, a new key is set using CM_SET_KEY.REQ. A matching CNF is received. I also verified that the correct key is set by reading it back using plctool. If the bug occurs, the chip seems to behave normally from a linux sw point of view, it replies to all requests, no errors are logged in debugfs/dmesg or ethtool. 

On the next plug in, SLAC is performed correctly. After the matching cnf, the EV tries to send an association request (CM_ASSOC.REQ or so) to the EVSE PLC and it does not reply to it (we saw this on an external PLC sniffer). Since this is PLC to PLC chip communication we cannot do much about it, it is a firmware bug in the QCA as far as I understand. The result is that the link does not come up and no SDP is possible etc.

The link is never reported as up (using plctool -L). EVerest does not verify this at the moment as it is a non-standard PLC message and hence gets stuck after slac. But even if, the chip never recovers, so even restarting SLAC would not help at all in this case.

I verified in a lot of testing on the Charin events in Valencia and Cleveland (where i had good reproducability) that resetting the chip after setting new keys after unplug works 100% of the time, this is what is implemented here.
This is a cleaned-up version of the hacks we did during the testival.

It is disabled by default though, I have not verified if other PLC chips (lumissil, vertexcom) implement the RS_DEV command (they also don't need it as I never had problems with stability with those).

Next step would be to implement link up detection in a separate PR and restart SLAC in that case as there are probably still cases which are not solved by this PR.

corneliusclaussen:
I also discussed this with several car OEMs and they said that also on the EV side they reset the QCA on every unplug due to some firmware bugs and said it is good practice to do so. I prefer to switch to Lumissil or VertexCom instead ;-)

lategoodbye:
Thanks for the detailed explanation

lib/staging/slac/fsm/evse/src/states/others.cpp
```diff
@@ -76,6 +83,57 @@ bool ResetState::handle_slac_message(slac::messages::HomeplugMessage& message) {
...
+        // RS_DEV.REQ is on protocol version 0
+        ctx.send_slac_message(cfg.plc_peer_mac, set_reset_req, 0);
+
+        chip_reset_has_been_sent = true;
+
+        return cfg.chip_reset_timeout_ms;
+    } else {
+        ctx.log_info("RS_DEV.REQ timeout, no response received - failed to reset the chip");
+        return {};
+    }
```
a-w50, line 121: Instead of having two bools, I would introduce an `enum class` for this small sub-state machine like
```
enum class ChipResetState {
  DELAY,
  SEND_RESET,
  DONE,
}
```
and have only a state variable of its type.  Checking the bools could also be consistent (either `not` or `== false`).

corneliusclaussen:
Added another commit to also wait for link to be established before entering matched state. Successfully tested on real cars that it retries if the link does not come up. Also this option is disabled by default as it may only work on Qualcomm chips.

lib/staging/slac/fsm/evse/include/slac/fsm/evse/context.hpp
```diff
@@ -68,8 +82,10 @@ struct Context {
     slac::messages::HomeplugMessage slac_message_payload;
 
     // FIXME (aw): message should be const, but libslac doesn't allow for const ptr - needs changes in libslac
-    template <typename SlacMessageType> void send_slac_message(const uint8_t* mac, SlacMessageType& message) {
+    template <typename SlacMessageType>
+    void send_slac_message(const uint8_t* mac, SlacMessageType& message, int protocol_version = 1) {
         slac::messages::HomeplugMessage hp_message;
+        hp_message.set_protocol_version(protocol_version);
```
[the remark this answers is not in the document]
corneliusclaussen: Yes, it depends on https://github.com/EVerest/libslac/pull/7 which needs to be merged first. Then dependency yaml can be updated

lib/staging/slac/fsm/evse/src/states/others.cpp
```diff
@@ -76,6 +83,57 @@ bool ResetState::handle_slac_message(slac::messages::HomeplugMessage& message) {
...
+}
+
+bool ResetChipState::handle_slac_message(slac::messages::HomeplugMessage& message) {
+    const auto mmtype = message.get_mmtype();
+    if (mmtype != (slac::defs::MMTYPE_CM_RESET_DEVICE | slac::defs::MMTYPE_MODE_CNF)) {
+        // unexpected message
+        // FIXME (aw): need to also deal with CM_VALIDATE.REQ
```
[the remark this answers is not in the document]
corneliusclaussen: removed

lib/staging/slac/io/src/io.cpp
```diff
@@ -33,6 +33,7 @@ void SlacIO::loop() {
 
     while (running) {
         if (slac_channel.read(incoming_msg, 10)) {
+            incoming_msg.set_protocol_version_from_rawmsg();
```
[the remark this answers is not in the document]
corneliusclaussen: see libslac modifications to support 1.0 and 1.1 protocols

modules/EvseSlac/main/slacImpl.cpp
```diff
@@ -84,6 +84,11 @@ void slacImpl::run() {
     fsm_ctx.slac_config.set_key_timeout_ms = config.set_key_timeout_ms;
     fsm_ctx.slac_config.ac_mode_five_percent = config.ac_mode_five_percent;
     fsm_ctx.slac_config.sounding_atten_adjustment = config.sounding_attenuation_adjustment;
+
+    fsm_ctx.slac_config.do_chip_reset = config.do_chip_reset;
+    fsm_ctx.slac_config.chip_reset_delay_ms = config.chip_reset_delay_ms;
+    fsm_ctx.slac_config.chip_reset_timeout_ms = config.chip_reset_timeout_ms;
```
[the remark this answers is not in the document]
corneliusclaussen: done, also for the link status of the next commit

lib/staging/slac/fsm/evse/src/states/others.cpp
corneliusclaussen: done

lib/staging/slac/io/src/io.cpp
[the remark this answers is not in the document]
corneliusclaussen: The custom vendor extensions use a different header format, so libslac was modified to be able to handle both protocols

modules/EvseSlac/manifest.yaml
```diff
@@ -46,6 +46,34 @@ provides:
           It is better to wait for the matching to be done.
         type: boolean
         default: false
+      do_chip_reset:
+        description: Perform a chip reset after setting NMK using the RS_DEV.REQ Vendor MME Extension (QCA7k)
+        type: boolean
+        default: false
+      chip_reset_delay_ms:
+        description: Delay between SET_KEY.CNF and RS_DEV.REQ
```
lategoodbye, line 54: Just a note: i've noticed that the QCA7K does not always send a SET_KEY.CNF (or the host has a chance to receive it). Unfortunately i never had the time to investigate this further. AFAIR this was the case for firmware before 3.0.0 and non-fast SLAC configuration which does a real reset after CM_SET_KEY, but i'm not totally sure.

This is just a warning, not a request to change.

corneliusclaussen: Interesting. The non-fast SLAC config, does that mean the QCA resets itself when setting a key? I think we always used a config where the QCA does not automatically reset when the key is changed, that is why we do a reset here manually afterwards. I have not seen the issue of missing CNF in our setup, but it would make sense that e.g. the QCA itself resets before the CNF packet is really transmitted to the host

corneliusclaussen: So maybe using fast config option is preffered as we have more control over when (or if at all) we need to reset

lategoodbye: > Interesting. The non-fast SLAC config, does that mean the QCA resets itself when setting a key? I think we always used a config where the QCA does not automatically reset when the key is changed, that is why we do a reset here manually afterwards. I have not seen the issue of missing CNF in our setup, but it would make sense that e.g. the QCA itself resets before the CNF packet is really transmitted to the host

From my understanding the behavior in this case is the following:
- EVSE host send CM_SET_KEY.REQ
- QCA7000 stores the the NMK in the internal flash
- QCA7000 send CM_SET_KEY.CNF
- QCA7000 triggers a reset, which clears the messages buffers

Yes, i agree this process is not recommend for production because of the flash wearout.

corneliusclaussen: The fast way also writes it to flash memory, after power cycle the key is still there. I think the chip was designed for home usage where the PLC modems are paired a few times per lifetime. It was not really designed for changing a key very often.

lategoodbye: Okay, the other advantage of fast way was avoiding the reset which takes ~ 7 second before SLAC is operational again.

lib/staging/slac/io/src/io.cpp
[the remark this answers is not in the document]
corneliusclaussen: This was handled in libslac update

corneliusclaussen:
I checked on VertexCom that it still works when the reset and link detect options are disabled. I'm not aware of VertexCom or Lumissil vendor specific messages that could allow for reset/link detection/version information etc. Anyone knows?

lategoodbye:
> I checked on VertexCom that it still works when the reset and link detect options are disabled. I'm not aware of VertexCom or Lumissil vendor specific messages that could allow for reset/link detection/version information etc. Anyone knows?

We implemented the link detection in a GreenPHY compatible way (not vendor specific) in our propietary stack, which has been successful tested on QCA7000 and VertexCom. Is the involved MME sufficient?

lib/staging/slac/fsm/evse/src/states/others.cpp
```diff
@@ -124,4 +178,125 @@ FSMSimpleState::HandleEventReturnType FailedState::handle_event(AllocatorType& s
...
+        std::string freq;
+        int line_freq = (msg.line_freq_zc) & 0x03;
+        if (line_freq == 0x01) {
+            freq = "50Hz";
+        } else if (line_freq == 0x02) {
+            freq = "60Hz";
+        } else {
+            freq = "Unknown (" + std::to_string(line_freq) + ")";
+        }
+        ctx.log_info("  Line frequency: " + freq);
```
a-w50, line 299: would move this into a separate function

corneliusclaussen:
> Is the involved MME sufficient?

Which one did you use?

lib/staging/slac/fsm/evse/src/states/others.cpp
corneliusclaussen: done

corneliusclaussen:
> > Is the involved MME sufficient?
> 
> Which one did you use?

CM_NW_INFO?

barsnick:
> CM_NW_INFO? NSCM_LINK_STATS?

CM_NW_STATS.CNF->NumSTAs

corneliusclaussen:
> > CM_NW_INFO? NSCM_LINK_STATS?
> 
> CM_NW_STATS.CNF->NumSTAs

ok. Funny to see how many ways there are to find out if there is a link. Lumissil has a bunch of different packages implemented as well. But CM_NW_STATS also works on Lumissil

corneliusclaussen:
Now link detection and version information also working for Lumissil. VertexCom is supported, but only without link detection/version information/reset as of now. Will add vendor specific support for VertexCom in a later pull request when we get access to the documentation.

hikinggrass (approved):

hikinggrass (approved):

Pietfried (approved):

a-w50 (approved):
Should be fine so far.  Please test with the latest changes.  If everything works as expected, feel free to merge.  Remember to squash.

PR #608, opened 2024-03-27
fix: parameter needs to be pass by reference rather than value

james-ctc:
updated unit test to ensure function modifies parameter as expected.

## Describe your changes

previous compile error led to an incorrect fix changing a parameter to pass by value rather than reference.
unit test updated to check that pass by reference updates the parameter as expected.


barsnick:
Good job adding tests! Perhaps the other `modify_error_()` could be tested as well.

Other LGTM by visual inspection.

I added the "include-in-release" label, because the incorrect change was also pulled into the release branch.

corneliusclaussen (approved):

hikinggrass (approved):

PR #610, opened 2024-03-28
fix: EnergyManager to select limit taking into account time

james-ctc:
## Describe your changes

EnergyManager calculates a list of optimised values and previously selected the 1st item to set the root side limit.

Instead EnergyManager now traverses the list to find the applicable limit based on the current time.
(the list can include schedules that have expired)

Unit tests added to ensure expected operation

### DIFF
--- a/modules/EnergyManager/EnergyManager.cpp
+++ b/modules/EnergyManager/EnergyManager.cpp
@@ -5,6 +5,7 @@
 #include "BrokerFastCharging.hpp"
 #include "Market.hpp"
 #include <fmt/core.h>
+#include <optional>
 
 using namespace std::literals::chrono_literals;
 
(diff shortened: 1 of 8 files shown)


modules/EnergyManager/EnergyManager.cpp
```diff
@@ -143,19 +144,37 @@ std::vector<types::energy::EnforcedLimits> EnergyManager::run_optimizer(types::e
...
+            for (const auto& s : sold_energy) {
+                const auto schedule_time = Everest::Date::from_rfc3339(s.timestamp);
+                if (globals.start_time < schedule_time) {
+                    // all further schedules will be further into the future
+                    break;
+                } else {
+                    // use this schedule as the starting point
+                    l.limits_root_side = s.limits_to_root;
+                }
+            }
```
corneliusclaussen, line 170: This looks correct to me. 
If we are in the middle of a schedule, it finds the entry that is closest to now, but still in the past.
If the complete schedule is in the past, it finds the last entry.
If the complete schedule is in the future, it finds the first entry.

corneliusclaussen (commented):
Looks good to me. Let me know when you are done with testing in the actual error case that we had.

modules/EnergyManager/EnergyManager.cpp
james-ctc: Tested on a charger and the previously observed delay to changing the charging rate is no longer observed.
Observations prior to the code change
[schedule-timings.pdf](https://github.com/EVerest/everest-core/files/14793267/schedule-timings.pdf)
Observations post to the code change
[schedule-timings-fix.pdf](https://github.com/EVerest/everest-core/files/14793268/schedule-timings-fix.pdf)

corneliusclaussen (approved):

PR #615, opened 2024-04-02
EvseV2G: SDP bind only on one interface

corneliusclaussen:
## Describe your changes

EvseV2G SDP listener was receiving SDP broadcasts from any interface. This causes issues when multiple ISO connections are done from a single EVerest instance.

### DIFF
--- a/modules/EvseV2G/sdp.cpp
+++ b/modules/EvseV2G/sdp.cpp
@@ -220,15 +220,6 @@ int sdp_init(struct v2g_context* v2g_ctx) {
         return -1;
     }
 
-    if (v2g_ctx->local_tcp_addr) {
-        memcpy(&sdp_addr.sin6_addr, &v2g_ctx->local_tcp_addr->sin6_addr, sizeof(sdp_addr.sin6_addr));
-    } else if (v2g_ctx->local_tls_addr) {
-        memcpy(&sdp_addr.sin6_addr, &v2g_ctx->local_tls_addr->sin6_addr, sizeof(sdp_addr.sin6_addr));
-    } else {
-        dlog(DLOG_LEVEL_ERROR, "Neither TLS nor TCP server on %s", v2g_ctx->if_name);
-        return -1;
-    }
-
     /* create receiving socket */
     v2g_ctx->sdp_socket = socket(AF_INET6, SOCK_DGRAM, IPPROTO_UDP);
     if (v2g_ctx->sdp_socket == -1) {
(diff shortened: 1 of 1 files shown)


hikinggrass (approved):

modules/EvseV2G/sdp.cpp
```diff
@@ -220,15 +220,6 @@ int sdp_init(struct v2g_context* v2g_ctx) {
...
 
-    if (v2g_ctx->local_tcp_addr) {
-        memcpy(&sdp_addr.sin6_addr, &v2g_ctx->local_tcp_addr->sin6_addr, sizeof(sdp_addr.sin6_addr));
-    } else if (v2g_ctx->local_tls_addr) {
-        memcpy(&sdp_addr.sin6_addr, &v2g_ctx->local_tls_addr->sin6_addr, sizeof(sdp_addr.sin6_addr));
-    } else {
-        dlog(DLOG_LEVEL_ERROR, "Neither TLS nor TCP server on %s", v2g_ctx->if_name);
-        return -1;
-    }
-
```
barsnick, line 231: I'm guessing that this (binding to a specific IP address) probably did not work, because the value of `sdp_addr.sin6_addr` was overwritten with `in6addr_any` 15 lines below:
https://github.com/EVerest/everest-core/blob/b5b315d3b56304030df45752f9b1ee9137fde662/modules/EvseV2G/sdp.cpp#L236

modules/EvseV2G/sdp.cpp
```diff
@@ -252,6 +243,16 @@ int sdp_init(struct v2g_context* v2g_ctx) {
 
     dlog(DLOG_LEVEL_INFO, "SDP socket setup succeeded");
 
+    /* bind only to specified device */
+    if (setsockopt(v2g_ctx->sdp_socket, SOL_SOCKET, SO_BINDTODEVICE, v2g_ctx->if_name, strlen(v2g_ctx->if_name)) ==
+        -1) {
+        dlog(DLOG_LEVEL_ERROR, "setsockopt(SO_BINDTODEVICE) failed: %s", strerror(errno));
+        close(v2g_ctx->sdp_socket);
+        return -1;
+    }
```
barsnick, line 252: My intuition says this should go before the `bind()` call, but I couldn't find any clear instructions on that, and didn't test myself.

modules/EvseV2G/sdp.cpp
```diff
@@ -252,6 +243,16 @@ int sdp_init(struct v2g_context* v2g_ctx) {
...
+
+    dlog(DLOG_LEVEL_TRACE, "bind only to specified device");
```
barsnick, line 254: I'm not sure this is useful, but if so, you should add `v2g_ctx->if_name` to the output.

modules/EvseV2G/sdp.cpp
corneliusclaussen: Not sure about that, but it works this way.

modules/EvseV2G/sdp.cpp
corneliusclaussen: will add that

modules/EvseV2G/sdp.cpp
barsnick: Thanks for the explanation. Sound reasonable. Probably the inaddr_any was added because the multicast didn't work in the first place...

I agree with the conclusions, so LGTM.

PR #630, opened 2024-04-15
Bazel: Allow users to consume Rust crates

dorezyuk:
Allows users to consume Rust crates in their code

Users can 

```starlack
git_repository(
    name = "everest-core",
    remote = <the remote>,
    commit = <the commit>,

git_repository(
    name = "everest-framework",
    remote = <the remote>,
    commit = <the commit>,
)

load("@everest-framework//third-party/bazel:repos.bzl", "everest_framework_repos")
everest_framework_repos()

load("@everest-framework//third-party/bazel:defs.bzl", "everest_framework_deps")
everest_framework_deps()

load("@everest-core//third-party/bazel:repos.bzl", "everest_core_repos")
everest_core_repos()

load("@everest-core//third-party/bazel:defs.bzl", "everest_core_defs")
everest_core_defs()
```

after which they can run e.x.

```sh
bazel build @everest-core//modules/RsIskraMeter
```
to build the desired module.

When adding new modules  we need to extend the third-party/bazel/BUILD.bazel manifest section

### DIFF
--- a/WORKSPACE.bazel
+++ b/WORKSPACE.bazel
@@ -16,34 +16,6 @@ rust_register_toolchains(
     versions = ["1.74.0"],
 )
 
-load("@rules_rust//crate_universe:repositories.bzl", "crate_universe_dependencies")
-
-crate_universe_dependencies()
-
-load("@rules_rust//crate_universe:defs.bzl", "crates_repository", "crate")
-
-crates_repository(
-    name = "crate_index",
-    cargo_lockfile = "//modules:Cargo.lock",
-    isolated = False,
-    manifests = [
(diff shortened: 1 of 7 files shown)


golovasteek (approved):
Looks good to me.
Can you please, update the description of PR: remove all mentions of vendoring. Since it is not there anymore.

SirVer (approved):
Yes, that solution looks much better to me. Isn't this also how we do it internally at Qwello?

dorezyuk:
> Yes, that solution looks much better to me. Isn't this also how we do it internally at Qwello?

I've merged the same already into our fork - yes

PR #640, opened 2024-04-18
Refactored OCMF powermeter transaction handling

Pietfried:
## Describe your changes
Refactored OCMF powermeter transaction handling:
* Powermeter interface now contains proper user identification arguments (IS, IT, IF, IT, ID) to start the transaction
* Refactored EvseManager and LemDCBM400600 modules to address the changed types
* Added module config params for LemDCBM400600 : cable_id, tariff_id
* Fixed test cases in LemDCBM400600 due to changes made

Why changes are required:
The `cable_id` is no OCMF parameter but it is specific to the LEM powermeter and shall therefore not be an argument within the TransactionReq type. A `tariff_id` is also no OCMF parameter (but Tariff Text (TT) is) and in the LEM communication protocol it is unclear how this parameter shall be set. This might require further changes in the LEM module, but it doesnt belong to the TransactionReq type.


modules/EvseManager/Charger.cpp
```diff
@@ -9,25 +9,18 @@
...
-
-#include <boost/uuid/random_generator.hpp>
-#include <boost/uuid/uuid.hpp>
-#include <boost/uuid/uuid_io.hpp>
 #include <generated/types/powermeter.hpp>
 #include <math.h>
 #include <string.h>
 #include <thread>
 #include <type_traits>
+#include <utils.hpp>
```
a-w50, line 17: `#include "utils.hpp"`, try to seperate includes by their scope ...

modules/EvseManager/Charger.cpp
```diff
@@ -1083,7 +1076,14 @@ bool Charger::start_transaction() {
...
     const types::powermeter::TransactionReq req{
-        evse_id, shared_context.session_uuid, shared_context.id_token.id_token.value, 0, 0, ""};
+        evse_id,
+        shared_context.session_uuid,
+        true,
+        {},
+        utils::convert_to_ocmf_identification_type(shared_context.id_token.id_token.type),
+        std::nullopt,
+        shared_context.id_token.id_token.value,
+        std::nullopt};
```
a-w50, line 1086: This is hard to read, might help to have comments or set each field by hand - also very error prone

a-w50 (changes requested):
Missing some documentation about why for ex. `cable_id` and `tariff_id` have been removed from the powermeter type and are now only local config options for the LEM powermeter

modules/EvseManager/utils.hpp
```diff
@@ -0,0 +1,39 @@
...
+        return types::powermeter::OCMFIdentificationType::ISO14443;
+    case types::authorization::IdTokenType::ISO15693:
+        return types::powermeter::OCMFIdentificationType::ISO15693;
+    case types::authorization::IdTokenType::KeyCode:
+        return types::powermeter::OCMFIdentificationType::KEY_CODE;
+    case types::authorization::IdTokenType::Local:
+        return types::powermeter::OCMFIdentificationType::LOCAL;
+    case types::authorization::IdTokenType::NoAuthorization:
+        return types::powermeter::OCMFIdentificationType::NONE;
+    default:
```
a-w50, line 33: in order for static analysis being able to see missing enum types in a switch, you shouldn't have a `default` case.  Therefor in most situations it is better practice to leave out the `default` case, just do a return of the default value after the switch

Pietfried:
> Missing some documentation about why for ex. `cable_id` and `tariff_id` have been removed from the powermeter type and are now only local config options for the LEM powermeter

The `cable_id` is no OCMF parameter but it is specific to the LEM powermeter and shall therefore not be an argument within the `TransactionReq` type. A `tariff_id` is also no OCMF parameter (but Tariff Text (TT) is) and in the LEM communication protocol it is unclear how this parameter shall be set. This might require further changes in the LEM module, but it doesnt belong to the `TransactionReq` type.

a-w50:
> > Missing some documentation about why for ex. `cable_id` and `tariff_id` have been removed from the powermeter type and are now only local config options for the LEM powermeter
> 
> The `cable_id` is no OCMF parameter but it is specific to the LEM powermeter and shall therefore not be an argument within the `TransactionReq` type. A `tariff_id` is also no OCMF parameter (but Tariff Text (TT) is) and in the LEM communication protocol it is unclear how this parameter shall be set. This might require further changes in the LEM module, but it doesnt belong to the `TransactionReq` type.

That is exactly what I would like to see in the commit or PR message, so that we're able to reason about why those changes happened.

a-w50 (approved):

modules/RsIskraMeter/src/main.rs
```diff
@@ -1277,12 +1276,11 @@ mod tests {
...
 
         let _unused = ready_state.start_transaction(TransactionReq {
-            cable_id: 1,
-            client_id: String::new(),
             evse_id: String::new(),
-            tariff_id: 1,
             transaction_id: String::new(),
-            user_data: String::new(),
+            identification_status: true,
+            identification_type: "ISO14443",
```
SirVer, line 1282: This should not compile. Did you try compiling with Rust support? there is at least a `.to_string()` missing here.

Pietfried:
> @Pietfried The MR description explains the "what" happened, but not the "why", please extend, i.e. I have a similar question than @a-w50.

Hi @SirVer , this response to @a-w50 request explains "why" the changes are required: https://github.com/EVerest/everest-core/pull/640#issuecomment-2067176131 .

Please have a look at the Rust build, which is currently failing. It looks like `minItems` and `maxItems` keywords are not supported in the rust part.

types/powermeter.yaml
```diff
@@ -7,29 +7,52 @@ types:
...
+          OCMF Identification Level (IL): Encoded overall status of the user assignment
+        type: string
+        $ref: /powermeter#/OCMFIdentificationLevel
+      identification_flags:
+        description: >-
+          OCMF Identification Flags (IF): Detailed statements about the user assignment, represented by one or more identifiers
+        type: array
+        items:
+          minItems: 0
+          maxItems: 4
```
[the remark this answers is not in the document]
SirVer: I think they should be, yes. If not it is a bug that should be fixed. @dorezyuk, can you make sure and fix this MR too?

SirVer:
> Hi @SirVer , this response to @a-w50 request explains "why" the changes are required: https://github.com/EVerest/everest-core/pull/640#issuecomment-2066376028 .

Yes, I saw that @Pietfried. My statement was more that this should go into the MR description, because it is important context for posterity and should not be buried in the discussion.

types/powermeter.yaml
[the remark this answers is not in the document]
dorezyuk: @Pietfried Your indentation level is one too deep - you basically say right now "an array of arbitary length with strings between 0 and 4  chars" `(std::vector<std::array<char, 4>>)`. If this is what you want then you need to rename the `minItems` and `maxItems` to `minLength` and `maxLength`.

If you want to day an array between 0 and 4 with any string, then you need to move the `maxItems` and `minItems` one up

```yaml
type:array
minItems: 0
maxItems: 1
items:
  type:string
```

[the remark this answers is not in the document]
Pietfried: good catch, thanks!

modules/RsIskraMeter/src/main.rs
```diff
@@ -1179,12 +1179,14 @@ mod tests {
             let ready_state = make_ready_state(mock);
 
             let _unused = ready_state.start_transaction(TransactionReq {
-                cable_id: 1,
-                client_id: String::new(),
                 evse_id: String::new(),
-                tariff_id: 1,
                 transaction_id: String::new(),
-                user_data: String::new(),
+                identification_status: generated::types::powermeter::OCMFUserIdentificationStatus::ASSIGNED,
```
dorezyuk, line 1184: nice great job with fixing. There is a generated::types::powermeter include block at the top of the file. Lets add OCMFIdentificationType to it so we can write here just OCMFIdentificationType::ISO14443 and so on

modules/RsIskraMeter/src/main.rs
```diff
@@ -1179,12 +1179,14 @@ mod tests {
...
+                identification_type: generated::types::powermeter::OCMFIdentificationType::ISO14443,
+                identification_flags: Vec::new(),
+                identification_data: Some(String::new()),
```
dorezyuk, line 1187: I think this (and the 2 fields below) could also just be `None`

dorezyuk (approved):

SirVer (approved):
Great, thanks for addressing the boolean. The resulting code is much better I think!

SebaLukas (approved):
:+1:

a-w50:
> @Pietfried The MR description explains the "what" happened, but not the "why", please extend, i.e. I have a similar question than @a-w50.

Exactly, that's what I was trying to say.

corneliusclaussen (approved):

PR #643, opened 2024-04-19
Adding EvManager module

MarzellT:
## Describe your changes
* rewrites the EvManager car simulator in C++
* adds unit tests for the added functions
* fixes a few small issues in the nodered config
* small fixes in the JsEvManager


modules/EvManager/main/car_simulatorImpl.cpp
```diff
@@ -0,0 +1,556 @@
...
+        }
+    }
+}
+
+void car_simulatorImpl::registerAllCommands() {
+    const auto sleepFuntion = [this](const std::vector<std::string>& arguments) {
+        static std::optional<size_t> sleepTimeLeftMs;
+        if (!sleepTimeLeftMs.has_value()) {
+            sleepTimeLeftMs =
+                std::stoll(arguments[0]) * static_cast<long>((1 / static_cast<float>(loopIntervalMs.value()))) + 1;
```
a-w50, line 114: possibly unchecked access to optional value

modules/EvManager/main/car_simulatorImpl.cpp
```diff
@@ -0,0 +1,556 @@
...
+            } else {
+                break;
+            }
+        }
+    }
+}
+
+void car_simulatorImpl::registerAllCommands() {
+    const auto sleepFuntion = [this](const std::vector<std::string>& arguments) {
+        static std::optional<size_t> sleepTimeLeftMs;
```
a-w50, line 111: I would consider using this static variable here a bit hacky

modules/EvManager/main/car_simulatorImpl.cpp
```diff
@@ -0,0 +1,556 @@
...
+        if (!sleepTimeLeftMs.has_value()) {
+            sleepTimeLeftMs =
```
a-w50, line 113: This is a bit of a misnomer here, it isn't `sleepTimerLeftMs`, but rather `sleepTimerLeftTicks`

modules/EvManager/main/RegisteredCommand.hpp
```diff
@@ -0,0 +1,67 @@
...
+        registeredCommands.try_emplace(commandName, std::make_unique<RegisteredCommand<FunctionT>>(
+                                                        commandName, argumentCount, std::forward<FunctionT>(function)));
+    }
+
+    static const RegisteredCommandBase& getRegisteredCommand(const std::string& commandName) {
+        return *registeredCommands.at(commandName).get();
+    }
+
+private:
+    inline static std::unordered_map<std::string, std::unique_ptr<RegisteredCommandBase>> registeredCommands;
```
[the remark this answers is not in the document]
a-w50: Looking at it, I would probably create only a Registry class, which contains a map of Commands with
```
struct Command {
  std::function<bool(std::vector<std::string>)> function;
  std::size_t arity;
}
```
Within this class you can register functions with their name and arity.  Furthermore this Registry class would have an `bool execute(const std::string& name, const std::vector<std::string>& args)` function which would search the command by name, check its arity and then just call it.

a-w50 (commented):
Nicely done.  Concerning the style - we don't have strict rules, but I think the majority of code is using only CamelCase for classes/structs and snake_case for variables and filenames.  Constants are usually written as SOME_CONSTANT (also for static constexpr vars).

modules/CMakeLists.txt
```diff
@@ -17,6 +18,7 @@ ev_add_module(PacketSniffer)
 ev_add_module(PersistentStore)
 ev_add_module(PN532TokenProvider)
 ev_add_module(PyEvJosev)
+ev_add_module(PySlacLoopback)
```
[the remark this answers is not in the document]
MarzellT: I'm just super confused how this ended up here. I think I made a mistake during a rebase

modules/EvManager/main/RegisteredCommand.hpp
```diff
@@ -0,0 +1,67 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Pionix GmbH and Contributors to EVerest
+
+#ifndef TMP_CLION_CLANG_TIDY_SIMCOMMAND_HPP
+#define TMP_CLION_CLANG_TIDY_SIMCOMMAND_HPP
```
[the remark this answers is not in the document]
MarzellT: True, I wanted to change that. It just somehow was the default. I thought about using `#pragma once` instead

modules/EvManager/main/RegisteredCommand.hpp
```diff
@@ -0,0 +1,67 @@
...
+#include <utility>
+#include <vector>
+namespace module::main {
+
+template <typename Func> class RegisteredCommand;
+
+class RegisteredCommandBase {
+public:
+    RegisteredCommandBase() = default;
+    virtual ~RegisteredCommandBase(){};
```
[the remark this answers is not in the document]
MarzellT: I think this was just a simple mistake. I changed it to `= default`.

modules/EvManager/main/RegisteredCommand.hpp
```diff
@@ -0,0 +1,67 @@
...
+    RegisteredCommandBase(const RegisteredCommandBase&) = default;
+    RegisteredCommandBase& operator=(const RegisteredCommandBase&) = default;
+    RegisteredCommandBase(RegisteredCommandBase&&) = default;
+    RegisteredCommandBase& operator=(RegisteredCommandBase&&) = default;
+
+    virtual bool operator()(const std::vector<std::string>& arguments) const = 0;
+
+    // TODO: maybe we could deduce the argument count from the function signature somehow
```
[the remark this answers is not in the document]
MarzellT: I'll look into that

modules/EvManager/main/car_simulatorImpl.cpp
```diff
@@ -0,0 +1,556 @@
...
+        return false;
+    }
+
+    return true;
+}
+
+void car_simulatorImpl::subscribeToVariablesOnInit() {
+    // subscribe bsp_event
+    using types::board_support_common::BspEvent;
+    mod->r_ev_board_support->subscribe_bsp_event([this](auto bsp_event) {
```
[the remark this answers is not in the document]
MarzellT: True, good catch!

modules/EvManager/main/car_simulatorImpl.cpp
MarzellT: True. I need to double check the usages of optional values everywhere

modules/EvManager/main/car_simulatorImpl.cpp
MarzellT: You are right. At the time I was thinking there will only be one instance of the `car_simulatorImpl` running so this would make sense but I guess there could be multiple. I will change that

modules/EvManager/main/car_simulatorImpl.cpp
MarzellT: Also true, thanks!

modules/EvManager/main/RegisteredCommand.hpp
```diff
@@ -0,0 +1,67 @@
...
+template <typename Func> class RegisteredCommand;
+
+class RegisteredCommandBase {
+public:
+    RegisteredCommandBase() = default;
+    virtual ~RegisteredCommandBase(){};
+    RegisteredCommandBase(const RegisteredCommandBase&) = default;
+    RegisteredCommandBase& operator=(const RegisteredCommandBase&) = default;
+    RegisteredCommandBase(RegisteredCommandBase&&) = default;
+    RegisteredCommandBase& operator=(RegisteredCommandBase&&) = default;
```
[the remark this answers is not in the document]
MarzellT: Honestly, I'm not sure what's best here. I was following https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-five. I'm not sure if this rule would apply here or not. What do you think?

modules/EvManager/main/RegisteredCommand.hpp
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+private:
+    inline static std::unordered_map<std::string, std::unique_ptr<RegisteredCommandBase>> registeredCommands;
+};
+
+template <typename FunctionT> class RegisteredCommand : public RegisteredCommandBase {
+public:
+    RegisteredCommand(std::string commandName, std::size_t argumentCount, FunctionT function) :
+        commandName{std::move(commandName)}, argumentCount{argumentCount}, function{std::move(function)} {
```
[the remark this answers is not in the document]
MarzellT: That makes sense. I will change that

modules/EvManager/main/RegisteredCommand.hpp
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    ~RegisteredCommand() override = default;
+
+    bool operator()(const std::vector<std::string>& arguments) const override {
+        if (!arguments.empty() && arguments.size() != argumentCount) {
```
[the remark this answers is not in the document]
MarzellT: Yes. That was just a mistake. I already removed that

modules/EvManager/main/SimCommand.hpp
```diff
@@ -0,0 +1,34 @@
...
+#include <mutex>
+#include <stdexcept>
+#include <string>
+#include <unordered_map>
+#include <utility>
+#include <vector>
+
+namespace module::main {
+
+class RegisteredCommandBase;
```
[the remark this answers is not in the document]
MarzellT: It didn't compile. I thought I pushed a compilable version but apparently not. I already changed that too

modules/EvManager/main/SimData.cpp
```diff
@@ -0,0 +1,77 @@
...
+    auto commandsVector{convertCommandsStringToVector(commands)};
+
+    auto commandsWithArguments{splitIntoCommandsWithArguments(commandsVector)};
+
+    return compileCommands(commandsWithArguments);
+}
+
+std::vector<std::string> convertCommandsStringToVector(const std::string_view& commandsView) {
+
+    auto commands = std::string{commandsView};
```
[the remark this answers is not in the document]
MarzellT: Yeah that makes no sense. I was thinking about using `string_view`s vs `strings` in different places. Probably, I forgot to change that here. Good catch

modules/EvManager/main/SimData.cpp
```diff
@@ -0,0 +1,77 @@
...
+
+    // convert to lower case inplace
+    std::transform(commands.begin(), commands.end(), commands.begin(),
+                   [](unsigned char character) { return std::tolower(character); });
```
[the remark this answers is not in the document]
MarzellT: Definitely, I think I just copy pasted this from somewhere

modules/EvManager/main/SimData.cpp
```diff
@@ -0,0 +1,77 @@
...
+
+    // replace newlines with semicolons
+    std::replace(commands.begin(), commands.end(), '\n', ';');
```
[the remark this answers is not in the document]
MarzellT: Honestly, I just "translated" the javascript version without thinking a lot about it. I will have a second look to see if it's necessary

modules/EvManager/main/RegisteredCommand.hpp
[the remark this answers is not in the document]
a-w50: You have an abstract base class because of `virtual bool operator()(...) const = 0`, therefor you'll never be able to create an instance of `RegisteredCommandBase`.  If you can't create an instance, you also can't move/copy it. What you have, are derived instances of `RegisteredCommandBase`, these you can move/copy, if you want.  Usually you're referring to these derived instance via a `RegisteredCommandBase` pointer or reference.  Now you could of course dereference them via their parent type and call the copy/move operator - but this will potentially not do what you want (like when you have a non-virtual base class destructor), as long as you don't make the move/copy constructors of the base class virtual.  But again, as you're usually using only the `RegisteredCommandBase` pointer or reference, you don't have to move at all because your just moving the pointer and not the instance itself.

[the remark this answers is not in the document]
MarzellT: That makes sense. Thanks!

modules/EvManager/main/SimData.cpp
```diff
@@ -0,0 +1,77 @@
...
+    std::stringstream commandsStream{commands};
+    auto command = std::string{};
+    auto commandsVector = std::vector<std::string>{};
+
+    while (std::getline(commandsStream, command, ';')) {
+        commandsVector.push_back(command);
+    }
+    return commandsVector;
+}
+std::vector<std::pair<std::string, std::vector<std::string>>>
```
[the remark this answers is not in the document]
MarzellT: True, I changed that

modules/EvManager/main/car_simulatorImpl.cpp
```diff
@@ -0,0 +1,556 @@
...
+#include "car_simulatorImpl.hpp"
+#include "main/SimCommand.hpp"
+#include <everest/logging.hpp>
+#include <queue>
+
+namespace module::main {
+
+void car_simulatorImpl::init() {
+    // TODO: simulation loop time from manifest.yaml?
+    simData = std::make_unique<main::SimData>();
```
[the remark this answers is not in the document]
MarzellT: The idea was to not initialize it when the constructor is called. I'm not sure if that's a good idea or not. It probably wouldn't matter much. I was just trying to be as close as possible to the jsevmanager

modules/EvManager/main/car_simulatorImpl.cpp
```diff
@@ -0,0 +1,556 @@
...
+        EVLOG_warning << "Already received data, but framework is not ready yet.";
+        return;
+    }
+
+    if (enabled == value) {
+        // ignore if value is the same
+        return;
+    }
+
+    simData = std::make_unique<main::SimData>();
```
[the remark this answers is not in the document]
MarzellT: That's a good point. I didn't properly think this through. I'm currently looking into finding a good solution.

modules/EvManager/main/SimData.hpp
```diff
@@ -0,0 +1,77 @@
...
+    size_t last_pwm_duty_cycle{0};
+
+    types::board_support_common::Event actualBspEvent{};
+
+
+
+    std::queue<SimCommand> commandQueue;
+};
+
+std::queue<SimCommand> parseSimCommands(const std::string& value);
```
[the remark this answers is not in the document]
MarzellT: No you are right. I changed all the functions to be static member functions with appropriate visibility

modules/EvManager/main/car_simulatorImpl.cpp
```diff
@@ -0,0 +1,556 @@
...
+    // Check enabled
+    if (!checkCanExecute()) {
+        return;
+    }
+
+    simData = std::make_unique<main::SimData>();
+
+    auto valueCopy = value;
+    auto& commandQueue = simData->commandQueue;
+    commandQueue = parseSimCommands(valueCopy);
```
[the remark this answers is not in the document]
MarzellT: Also true, I was thinking about what would be the best solution for passing around the commands and didn't adopt how I pass the value here

modules/EvManager/main/car_simulatorImpl.cpp
```diff
@@ -0,0 +1,556 @@
...
+}
+
+void car_simulatorImpl::handle_modifyChargingSession(std::string& value) {
+    if (!enabled) {
+        EVLOG_warning << "Simulation disabled, cannot execute charging simulation.";
+        return;
+    }
+
+    auto valueCopy = value;
+    simData->commandQueue = parseSimCommands(valueCopy);
```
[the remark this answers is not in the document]
MarzellT: That's something I need to clarify

modules/EvManager/tests/CMakeLists.txt
```diff
@@ -1,9 +1,16 @@
...
 add_dependencies(${TEST_TARGET_NAME} ${MODULE_NAME})
 
-target_sources(${TARGET_NAME}
+get_target_property(GENERATED_INCLUDE_DIR generate_cpp_files EVEREST_GENERATED_INCLUDE_DIR)
+target_include_directories(${TEST_TARGET_NAME} PRIVATE
+    . ..
+    ${GENERATED_INCLUDE_DIR}
+    ${CMAKE_BINARY_DIR}/generated/modules/${MODULE_NAME}
+)
+
```
a-w50, line 12:
```
target_include_directories(${TEST_TARGET_NAME}
    PRIVATE
        "$<TARGET_PROPERTY:generate_cpp_files,EVEREST_GENERATED_INCLUDE_DIR>"
)
```
might be shorter.  Also note, that you're depending here on files generated by the target `generate_cpp_files` - as cmake doesn't know, that you're depending on this target, it could happen, that the test gets build before the files are generated (although I haven't seen that happening yet).
Furthermore, including the `.` in `target_include_directories` doesn't make sense to me, if the target itself is located in that directory.

modules/EvManager/main/CommandRegistry.hpp
```diff
@@ -0,0 +1,86 @@
...
+
+    // TODO(MarzellT): maybe we could deduce the argument count from the function signature somehow
+    template <typename FunctionT>
+    void registerCommand(std::string commandName, size_t argumentCount, FunctionT&& function) {
+        registeredCommands.try_emplace(
+            commandName, std::make_unique<RegisteredCommand<FunctionT>>(simulator, commandName, argumentCount,
+                                                                        std::forward<FunctionT>(function)));
+    }
+
+    const RegisteredCommandBase* getRegisteredCommand(const std::string& commandName) const {
```
a-w50, line 35: If an exception is already thrown in this function, than I would return by reference and not by pointer.  If you wouldn't throw exceptions here, than a pointer could be alright because you can check if it null or not.  An optional might also be applicable but is a bit inconvenient to use, depending on if you want to return by reference or by value

modules/EvManager/tests/CMakeLists.txt
MarzellT: Yes this solution is nicer, thanks!

modules/EvManager/main/car_simulatorImpl.hpp
```diff
@@ -0,0 +1,84 @@
...
+    void callEVBoardSupportFunctions();
+    void subscribeToExternalMQTT();
+    void resetSimDataToDefaults();
+
+    std::unique_ptr<CommandRegistry> commandRegistry;
+    std::unique_ptr<SimData> simData;
+    std::optional<bool> enabled;
+    std::atomic<bool> executionActive{false};
+    const size_t defaultLoopIntervalMs{250};
+    std::optional<size_t> loopIntervalMs{};
```
SebaLukas, line 71: Why `optional`?

modules/EvManager/main/car_simulatorImpl.cpp
```diff
@@ -0,0 +1,586 @@
...
+// Copyright Pionix GmbH and Contributors to EVerest
+
+#include "car_simulatorImpl.hpp"
+#include "main/SimCommand.hpp"
+#include <everest/logging.hpp>
+
+namespace module::main {
+
+void car_simulatorImpl::init() {
+    loopIntervalMs = defaultLoopIntervalMs;
```
SebaLukas, line 11: Adding a `constexpr` here instead of `defaultLoopIntervalMs`. E.g. `constexpr auto DEFAULT_LOOP_INTERVAL_MS = 250;`

modules/EvManager/main/car_simulatorImpl.cpp
```diff
@@ -0,0 +1,586 @@
...
+}
+
+void car_simulatorImpl::handle_enable(bool& value) {
+    if (!enabled.has_value()) {
+        EVLOG_warning << "Already received data, but framework is not ready yet.";
+        return;
+    }
+
+    if (enabled == value) {
+        // ignore if value is the same
```
SebaLukas, line 40: Adding `EVLOG_warning << "Simulation is already enabled, ignoring enable!";`

modules/EvManager/main/car_simulatorImpl.cpp
```diff
@@ -0,0 +1,586 @@
...
+
+    if (!simData->commandQueue.empty()) {
+        executionActive = true;
+    }
+}
+
+void car_simulatorImpl::handleSimulationLoop() {
+    while (enabled) {
+        if (simData && executionActive) {
+            if (loopIntervalMs.has_value()) {
```
SebaLukas, line 97: Why `optional`?

modules/EvManager/main/car_simulatorImpl.cpp
```diff
@@ -0,0 +1,586 @@
...
+                                                    const std::vector<std::string>& arguments) {
+        simulator->simData->state = SimState::PLUGGED_IN;
+        return (simulator->simData->pwm_duty_cycle > 4.0f && simulator->simData->pwm_duty_cycle < 97.0f);
+    };
+    commandRegistry->registerCommand("iso_wait_pwm_is_running", 0, iso_wait_pwm_is_runningFunction);
+
+    const auto draw_power_regulatedFunction = [](car_simulatorImpl* const simulator,
+                                                 const std::vector<std::string>& arguments) {
+        simulator->mod->r_ev_board_support->call_set_ac_max_current(std::stod(arguments[0]));
+        if (arguments[1] == "3") {
```
SebaLukas, line 148: `"3"` define as const

modules/EvManager/main/car_simulatorImpl.cpp
```diff
@@ -0,0 +1,586 @@
...
+        }
+        simulator->simData->state = SimState::CHARGING_REGULATED;
+        return true;
+    };
+    commandRegistry->registerCommand("draw_power_regulated", 2, draw_power_regulatedFunction);
+
+    const auto draw_power_fixedFunction = [](car_simulatorImpl* const simulator,
+                                             const std::vector<std::string>& arguments) {
+        simulator->mod->r_ev_board_support->call_set_ac_max_current(std::stod(arguments[0]));
+        if (arguments[1] == "3") {
```
SebaLukas, line 161: see above

modules/EvManager/main/car_simulatorImpl.cpp
```diff
@@ -0,0 +1,586 @@
...
+                simulator->simData->slacState = "TRIGGERED";
+            }
+        }
+        if (simulator->simData->slacState == "MATCHED") {
+            EVLOG_debug << "Slac Matched";
+            return true;
+        }
+        return false;
+    };
+    commandRegistry->registerCommand("iso_wait_slac_matched", 0, iso_wait_slac_matchedFunction);
```
SebaLukas, line 223: Move `if (simulator->mod->r_slac.empty()) {` before `const auto iso_wait_slac_matchedFunction = ...`

modules/EvManager/main/car_simulatorImpl.cpp
```diff
@@ -0,0 +1,586 @@
...
+                return false;
+            }
+
+            simulator->mod->r_ev[0]->call_start_charging(simulator->simData->energymode);
+            return true;
+        };
+        commandRegistry->registerCommand("iso_start_v2g_session", 1, iso_start_v2g_sessionFunction);
+    }
+
+    const auto iso_wait_pwr_readyFunction = [](car_simulatorImpl* const simulator,
```
SebaLukas, line 251: Should be also under `if (mod->r_ev.empty())`

modules/EvManager/main/car_simulatorImpl.cpp
```diff
@@ -0,0 +1,586 @@
...
+                                               const std::vector<std::string>& arguments) {
+        if (simulator->simData->iso_pwr_ready) {
+            simulator->simData->state = SimState::ISO_POWER_READY;
+            return true;
+        }
+        return false;
+    };
+    commandRegistry->registerCommand("iso_wait_pwr_ready", 0, iso_wait_pwr_readyFunction);
+
+    const auto iso_dc_power_onFunction = [](car_simulatorImpl* const simulator,
```
SebaLukas, line 261: Should be also under `if (mod->r_ev.empty())`

modules/EvManager/main/car_simulatorImpl.cpp
```diff
@@ -0,0 +1,586 @@
...
+        if (simulator->simData->dc_power_on) {
+            simulator->simData->state = SimState::ISO_CHARGING_REGULATED;
+            simulator->mod->r_ev_board_support->call_allow_power_on(true);
+            return true;
+        }
+        return false;
+    };
+    commandRegistry->registerCommand("iso_dc_power_on", 0, iso_dc_power_onFunction);
+
+    const auto iso_draw_power_regulatedFunction = [](car_simulatorImpl* const simulator,
```
SebaLukas, line 273: Should be also under `if (mod->r_ev.empty())`

modules/EvManager/main/car_simulatorImpl.cpp
```diff
@@ -0,0 +1,586 @@
...
+        }
+        return false;
+    };
+    commandRegistry->registerCommand("iso_wait_slac_matched", 0, iso_wait_slac_matchedFunction);
+
+    if (!mod->r_ev.empty()) {
+        const auto iso_start_v2g_sessionFunction = [](car_simulatorImpl* const simulator,
+                                                      const std::vector<std::string>& arguments) {
+            const auto& argument = arguments[0];
+            if (argument == "ac_single_phase_core") {
```
SebaLukas, line 229: Define enum. E.g. `class enum EnergyMode : std::string`

modules/EvManager/main/car_simulatorImpl.cpp
```diff
@@ -0,0 +1,586 @@
...
+            return true;
+        }
+        return false;
+    };
+    commandRegistry->registerCommand("iso_dc_power_on", 0, iso_dc_power_onFunction);
+
+    const auto iso_draw_power_regulatedFunction = [](car_simulatorImpl* const simulator,
+                                                     const std::vector<std::string>& arguments) {
+        simulator->mod->r_ev_board_support->call_set_ac_max_current(std::stod(arguments[0]));
+        if (arguments[1] == "3") {
```
SebaLukas, line 276: see above

modules/EvManager/main/car_simulatorImpl.cpp
```diff
@@ -0,0 +1,586 @@
...
+            // Wait for physical plugin (ev BSP sees state A on CP and not Disconnected)
+
+            // If we have auto_exec configured, restart simulation when it was unplugged
+            EVLOG_info << "Unplug detected, restarting simulation.";
+            simData->slacState = "UNMATCHED";
+            mod->r_ev[0]->call_stop_charging();
+            if (mod->config.auto_exec) {
+                auto valueCopy = mod->config.auto_exec_commands;
+                handle_executeChargingSession(valueCopy);
+            }
```
SebaLukas, line 390: Please delete

modules/EvManager/main/car_simulatorImpl.cpp
```diff
@@ -0,0 +1,586 @@
...
+}
+
+bool car_simulatorImpl::checkCanExecute() {
+    if (!enabled.has_value()) {
+        EVLOG_warning << "Already received data, but framework is not ready yet.";
+        return false;
+    }
+
+    const auto enabledValue = enabled.value();
+    if (!enabledValue) {
```
SebaLukas, line 487: `if (!enabled.value())`

SebaLukas (changes requested):
A really well done from me too. But a few things should still be changed.

modules/EvManager/main/CommandRegistry.cpp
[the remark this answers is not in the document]
MarzellT: True, I will delete the file

config/config-sil-dc.yaml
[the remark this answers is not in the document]
MarzellT: Sure!

modules/EvManager/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
...
+# AUTO GENERATED - MARKED REGIONS WILL BE KEPT
+# template version 3
+#
+
+# module setup:
+#   - ${MODULE_NAME}: module name
+ev_setup_cpp_module()
+
+# ev@bcc62523-e22b-41d7-ba2f-825b493a3c97:v1
+# insert your custom targets and additional config variables here
```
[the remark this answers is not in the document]
MarzellT: Oh yes, good find!

modules/EvManager/main/SimCommand.cpp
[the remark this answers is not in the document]
MarzellT: I'm not sure if that would be the better approach. Wouldn't I need to `inline` the definitions in that case?

modules/EvManager/main/car_simulatorImpl.hpp
MarzellT: I basically wanted to mimic
```
function execute_charging_session(mod, args) {
  if (mod === undefined) {
    evlog.warning('Already received data, but framework is not ready yet');
    return;
  }
```
from the `JsEvManager` as I then would only set `enabled` to a value in the ready function but I'm not sure if that makes sense at all. What do you think?

modules/EvManager/main/car_simulatorImpl.cpp
MarzellT: Makes sense. I also removed the optional

modules/EvManager/main/car_simulatorImpl.cpp
MarzellT: Added the log

modules/EvManager/main/car_simulatorImpl.cpp
MarzellT: Sorry, I'm not sure what you mean. Could you give me an example?

modules/EvManager/main/car_simulatorImpl.cpp
MarzellT: Done

modules/EvManager/main/car_simulatorImpl.cpp
MarzellT: Done

modules/EvManager/main/car_simulatorImpl.cpp
MarzellT: Done

modules/EvManager/main/car_simulatorImpl.cpp
SebaLukas: E.g. `constexpr auto THREE_PHASES = "3"` -> `if (arguments[1] == THREE_PHASES) {` I find that easier to understand.

modules/EvManager/main/car_simulatorImpl.cpp
MarzellT: Makes sense. I had to create this myself though

modules/EvManager/main/car_simulatorImpl.cpp
MarzellT: Alright, I see! I changed that

modules/EvManager/main/car_simulatorImpl.cpp
MarzellT: Done

modules/EvManager/main/RegisteredCommand.hpp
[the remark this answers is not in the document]
MarzellT: I'm not sure if it's really necessary to do right now. I'll leave the TODO in the code. Maybe this would be something to look into later

SebaLukas:
Todos before merging:

- Adding documentation
- Testing
- Function names and variables in snake_case

SebaLukas (approved):
Very good :+1:

modules/EvManager/manifest.yaml
```diff
@@ -0,0 +1,104 @@
+description: >-
+  This module implements a Car simulator that can execute charging sessions using the yeti-simulation-control interface
```
hikinggrass, line 2: The yeti-simulation-control interface doesn't exist anymore

hikinggrass (approved):

modules/EvManager/manifest.yaml
SebaLukas: That's right, it should actually say `using the car_simulator interface`. Probably copied over from the JsEvManager module

MarzellT: Yes, it was copied from the `JsEvManager`. I will quickly change that

MarzellT: Fixed

PR #650, opened 2024-04-26
Fix simplified charging with waiting for energy

corneliusclaussen:
## Describe your changes

When waiting for Energy, PWM is switched off. A car should switch to state B when PWM is off. Some cars won't (e.g. if they do simplified charging). In this case, prepare charging after waiting for energy did not work as no B->C transition happens after PWM is enabled again. This PR fixes this issue.


SebaLukas (approved):
:+1: Nice, I didn't know `[[fallthrough]]` at all.

PR #651, opened 2024-04-26
CMakeLists: require C++17 support globally

barsnick:
## Describe your changes
Many modules use C++17 features already, either explicitly (requested support via CMake) or implicitly. Make this a global CMake requirement, and drop the local ones.

While the resulting flags also apply to the dependencies when building with EDM, the EVerest dependencies should nevertheless mark their requirements separately, as they can be built and used separately.

Perhaps the docs need a hint as well.
As [discussed on Zulip](https://lfenergy.zulipchat.com/#narrow/stream/417678-EVerest.3A-Framework-.26-Tools/topic/Marking.20code.20as.20requiring.20C.2B.2B17/near/435542432)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -28,6 +28,10 @@ endif()
 # make own cmake modules available
 list(INSERT CMAKE_MODULE_PATH 0 "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
 
+# specify minimum C++ standard
+set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_CXX_STANDARD_REQUIRED True)
+
 # test whether linking with libatomic is required on some platforms
 include(CheckAtomic)
 if(HAVE_CXX_ATOMICS_WITH_LIB OR HAVE_CXX_ATOMICS64_WITH_LIB)
--- a/lib/staging/can_dpm1000/tests/CMakeLists.txt
+++ b/lib/staging/can_dpm1000/tests/CMakeLists.txt
@@ -9,5 +9,3 @@ target_link_libraries(dpm1000_tester
         can_protocols::dpm1000
         Threads::Threads
 )
-
-target_compile_features(dpm1000_tester PRIVATE cxx_std_17)
--- a/lib/staging/slac/test/CMakeLists.txt
+++ b/lib/staging/slac/test/CMakeLists.txt
@@ -9,7 +9,6 @@ target_link_libraries(evse_slac_test
         slac_fsm_evse
         fmt::fmt
 )
-target_compile_features(evse_slac_test PRIVATE cxx_std_17)
 
 add_executable(evse_vs_ev)
 target_sources(evse_vs_ev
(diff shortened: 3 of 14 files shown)


a-w50 (commented):
Hi, as you've already written in the commit message - setting the c++ standard in a global fashion like this, has some drawbacks (like propagating into the dependencies and so on ...).  Furthermore, concerning "modern" cmake best practices, the `target_**` commands should be used instead of setting global flags, because they are more specific and you have more fine grained control.
Currently, every c++ module should be compiled with c++17 anyway as they depend on everest-framework, which publicly requires c++17.
If we want to force the c++ modules to have c++17 anyway, we could add the `target_compile_feature` command also to the `ev_add_cpp_module` cmake function.
For the other binaries, like small test utilities and so on, I'm not sure if we want to set this globally.
Has there been a specific need for this change?

corneliusclaussen:
So it looks like it is good to remove the per-module C++17 requirements as they should inherit this automatically from framework. But we should not require c++17 globally to avoid infecting other dependencies right? So merge without b9b763e5d5f686e21ec2eba1a00cb55825642eb8 ?

barsnick:
> Currently, every c++ module should be compiled with c++17 anyway as they depend on everest-framework, which publicly requires c++17.

Ah, okay, I didn't realize we inherit from the framework. (I saw no `--std=g++17` when compiling everest-core without the "global" requirement.)

> But we should not require c++17 globally to avoid infecting other dependencies right?

What is the effect of infecting the dependencies? I would guess the compiliation could trip over changes in what is allowed and not (like we remember from the days of early g++-4.x, when going to a new version), but the dependencies at least _built_ fine with the C++17 requirement, AFAICT.

a-w50:
Hi @barsnick

> Ah, okay, I didn't realize we inherit from the framework. (I saw no `--std=g++17` when compiling everest-core without the "global" requirement.)

Do you mean that it is not there or that you just don't saw it.  If I take the `Store` module for example, which has no `target_compile_definitions` for c++17 in its `CMakeLists.txt`, I can still see a `-std=gnu++17` in the `build/compile_commands.json`. That is because public `target_compile_definitions` are transitive in cmake.

> What is the effect of infecting the dependencies? I would guess the compiliation could trip over changes in what is allowed and not (like we remember from the days of early g++-4.x, when going to a new version), but the dependencies at least _built_ fine with the C++17 requirement, AFAICT.

Of course, in most cases it might not harm, but it is still considered bad practice, as a library should decide on its own, which compile flags to use.  In rare cases, where there are incompatibilities between different standard versions, this might lead to bugs, which would be hard to figure out.

corneliusclaussen:
Can we close this one?

barsnick:
Since there was no agreement here (Zulip: "let's define this globally"; GitHub: "this is bad practice"), we might as well close this.

It still leaves a bunch of modules failing to set the correct requirement (API, EnergyNode, ErrorHistory, EvseSecurity, GenericPowermeter, OCPP, OCPP201, PersistentStore, PowermeterBSM, Setup, System, YetiDriver).

PR #654, opened 2024-04-28
Bazel: choose tags or commit, when parsing dependencies.yaml

golovasteek:
## Describe your changes

Now bazel dep_tool, Is able to decide if the commit or tag is used as a reference and generate the *.bzl files accordingly.
There is a limitation: commit should we always be a full 40-character string, otherwise it's treated as a tag.

### DIFF
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -68,7 +68,7 @@ ext-openv2g:
 # mbedtls
 ext-mbedtls:
   git: https://github.com/EVerest/ext-mbedtls.git
-  git_tag: 8b3f26a
+  git_tag: 8b3f26a5ac38d4fdccbc5c5366229f3e01dafcc0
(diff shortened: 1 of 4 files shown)


golovasteek:
This should prevent issues like this one: https://github.com/EVerest/everest-core/pull/652#issuecomment-2079257584

SirVer (approved):
good solution I think.

dorezyuk (approved):

hikinggrass (approved):

dependencies.yaml
hikinggrass, line 1: One minor thing, could you add a comment to the top of this file that reminds people to use long git hashes from now on instead of the short ones used historically?

hikinggrass: thanks!

PR #661, opened 2024-04-30
Simpler power supply dc interface

corneliusclaussen:
## Describe your changes

The get_capabilities command of the power supply DC interface makes it hard to implement in some setups, as it requires the capabilities to be available at start up of the system immediately. 
The interface was changed to allow to publish the capabilities at any later point in time, when ever they become available.

It also uses the error framework now to report errors.


corneliusclaussen:
@a-w50 : As discussed earlier today, please have a look at the proposed power_supply_DC interface file. I will implement the actual code when the interface is agreed.

modules/EvseV2G/charger/ISO15118_chargerImpl.cpp
```diff
@@ -353,6 +294,32 @@ void ISO15118_chargerImpl::handle_stop_charging(bool& stop) {
     }
 }
 
+void ISO15118_chargerImpl::handle_set_physical_values(types::iso15118_charger::SetupPhysicalValues& physical_values) {
+    // FIXME IMPLEMENT ME CORRECTLY: can we just update those values any time? also during charging session?
```
corneliusclaussen, line 298: @SebaLukas : Please  check if something is missing here

SebaLukas: Yes, that works.
And there should be no problems with the values changing during a session.

corneliusclaussen: ok, then I'll remove the FIXME

modules/simulation/DCSupplySimulator/main/power_supply_DCImpl.hpp
```diff
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: Apache-2.0
-// Copyright chargebyte GmbH and Contributors to EVerest
+// Copyright Pionix GmbH and Contributors to EVerest
```
SebaLukas, line 2: The module was written by Chargebyte, right?

modules/simulation/JsDCSupplySimulator/index.js
```diff
@@ -156,8 +153,9 @@ boot_module(async ({
 
     if (connector_current > 0) energy_import_total += (connector_voltage * connector_current * 0.5) / 3600;
     if (connector_current < 0) energy_export_total += (connector_voltage * -connector_current * 0.5) / 3600;
-    
+
     mod.provides.powermeter.publish.powermeter(power_meter_external());
+    mod.provides.main.publish.capabilities(Capabilities);
```
SebaLukas, line 158: Here, the capabilities are published every 500ms? They do not change in the SIL. That's why I would call  `publish.capabilities(Capabilities);` before the `setInterval()`.

modules/EvseManager/energy_grid/energyImpl.cpp
```diff
@@ -125,10 +139,20 @@ void energyImpl::request_energy_from_energy_manager(bool priority_request) {
...
+        // copy complete external limit schedules for import
         if (mod->getLocalEnergyLimits().schedule_import.has_value() &&
             !mod->getLocalEnergyLimits().schedule_import.value().empty()) {
             energy_flow_request.schedule_import = mod->getLocalEnergyLimits().schedule_import;
+
+            if (mod->config.charge_mode == "DC") {
+                // For DC, apply our power supply capabilities as an additional limit on leaves side
+                const auto caps = mod->get_powersupply_capabilities();
+                for (auto& entry : energy_flow_request.schedule_import.value()) {
+                    if (entry.limits_to_leaves.total_power_W > caps.max_export_power_W) {
```
SebaLukas, line 151: Is export correct here?

modules/EvseManager/energy_grid/energyImpl.cpp
```diff
@@ -156,9 +180,20 @@ void energyImpl::request_energy_from_energy_manager(bool priority_request) {
...
+        // copy complete external limit schedules for export
         if (mod->getLocalEnergyLimits().schedule_export.has_value() &&
             !mod->getLocalEnergyLimits().schedule_export.value().empty()) {
             energy_flow_request.schedule_export = mod->getLocalEnergyLimits().schedule_export;
+
+            if (mod->config.charge_mode == "DC") {
+                // For DC, apply our power supply capabilities as an additional limit on leaves side
+                const auto caps = mod->get_powersupply_capabilities();
+                for (auto& entry : energy_flow_request.schedule_export.value()) {
+                    if (entry.limits_to_leaves.total_power_W > caps.max_import_power_W) {
```
SebaLukas, line 192: Is import correct here?

SebaLukas (approved):
Except for a few small things, it looks good :+1:

interfaces/ISO15118_charger.yaml
```diff
@@ -26,6 +26,11 @@ cmds:
       debug_mode:
         description: Enable/Disable debug mode
         type: boolean
+  set_physical_values:
```
a-w50, line 29: What are physical values?  Is this iso15118 naming?  Maybe there is a more descriptive name for that.

modules/EvseManager/EvseManager.cpp
```diff
@@ -167,52 +167,43 @@ void EvseManager::ready() {
...
-            setup_physical_values.dc_energy_to_be_delivered = 10000;
-
-            types::iso15118_charger::DC_EVSEMaximumLimits evseMaxLimits;
-            evseMaxLimits.EVSEMaximumCurrentLimit = powersupply_capabilities.max_export_current_A;
-            evseMaxLimits.EVSEMaximumPowerLimit = powersupply_capabilities.max_export_power_W;
-            evseMaxLimits.EVSEMaximumVoltageLimit = powersupply_capabilities.max_export_voltage_V;
-            setup_physical_values.dc_maximum_limits = evseMaxLimits;
-            charger->inform_new_evse_max_hlc_limits(evseMaxLimits);
+            // apply sane defaults capabilities settings once on boot
+            update_powersupply_capabilities(psu_caps);
```
a-w50, line 203: The initialization of `psu_caps` happens before setting up `present_values`.  Couldn't `update_powersupply_capabilities` be called directly after setting up `psu_caps` instead of having this `present_values` in between.  Might be more coherent.

modules/EvseManager/EvseManager.cpp
```diff
@@ -167,52 +167,43 @@ void EvseManager::ready() {
...
+            // Init power supply capabilities with safe defaults
+            types::power_supply_DC::Capabilities psu_caps;
+            psu_caps.bidirectional = false;
+            psu_caps.peak_current_ripple_A = 0.5;
+            psu_caps.current_regulation_tolerance_A = 0.5;
+            psu_caps.max_export_current_A = 0;
+            psu_caps.max_export_power_W = 0;
+            psu_caps.max_export_voltage_V = 60;
+            psu_caps.min_export_voltage_V = 0;
+            psu_caps.min_export_current_A = 0;
```
a-w50, line 194: if this would be wrapped into a helper function called `types::power_supply_DC::Capabilities get_sane_default_power_supply_capabilities()` or similar, the comment wouldn't be necessary and probably easier to read.

modules/EvseManager/EvseManager.cpp
```diff
@@ -167,52 +167,43 @@ void EvseManager::ready() {
...
-            charger->inform_new_evse_max_hlc_limits(evseMaxLimits);
+            // apply sane defaults capabilities settings once on boot
+            update_powersupply_capabilities(psu_caps);
 
-            types::iso15118_charger::DC_EVSEMinimumLimits evseMinLimits;
-            evseMinLimits.EVSEMinimumCurrentLimit = powersupply_capabilities.min_export_current_A;
-            evseMinLimits.EVSEMinimumVoltageLimit = powersupply_capabilities.min_export_voltage_V;
-            setup_physical_values.dc_minimum_limits = evseMinLimits;
+            // subscribe to run time updates for real initial values (and changes e.g. due to de-rating)
+            r_powersupply_DC[0]->subscribe_capabilities([this](auto caps) { update_powersupply_capabilities(caps); });
```
a-w50, line 206: Why doesn't the subscribe happen earlier?  Doing this during `ready` we might miss the capabilities if the dc power meter is quick.

a-w50: Furthermore, the lambda should have `(const auto& caps)`.

modules/EvseManager/EvseManager.hpp
```diff
@@ -196,13 +196,34 @@ class EvseManager : public Everest::ModuleBase {
...
-
-        powersupply_capabilities.max_import_current_A = caps.max_import_current_A;
-        powersupply_capabilities.max_import_power_W = caps.max_import_power_W;
+        powersupply_capabilities = caps;
+
+        // Inform HLC layer about update of physical values
+        types::iso15118_charger::SetupPhysicalValues setup_physical_values;
+        setup_physical_values.dc_current_regulation_tolerance = powersupply_capabilities.current_regulation_tolerance_A;
+        setup_physical_values.dc_peak_current_ripple = powersupply_capabilities.peak_current_ripple_A;
+        setup_physical_values.dc_energy_to_be_delivered = 10000;
```
a-w50, line 207: would put this also into a helper function, so the length of the main function `update_powersupply_capabilities` gets shorter and the logic more clear.

modules/EvseManager/EvseManager.hpp
```diff
@@ -196,13 +196,34 @@ class EvseManager : public Everest::ModuleBase {
...
+        r_hlc[0]->call_set_physical_values(setup_physical_values);
+
+        types::iso15118_charger::DC_EVSEMinimumLimits evseMinLimits;
+        evseMinLimits.EVSEMinimumCurrentLimit = powersupply_capabilities.min_export_current_A;
+        evseMinLimits.EVSEMinimumVoltageLimit = powersupply_capabilities.min_export_voltage_V;
```
a-w50, line 212: see above

modules/EvseManager/EvseManager.hpp
```diff
@@ -196,13 +196,34 @@ class EvseManager : public Everest::ModuleBase {
...
+        // energy management first. Those limits will be applied in energy_grid implementation when requesting energy,
+        // so it is enough to set the powersupply_capabilities here.
+        // FIXME: this is not implemented yet: enforce_limits uses the enforced limits to tell HLC, but capabilities
+        // limits are not yet included in request.
+
+        // Inform charger about new max limits
+        types::iso15118_charger::DC_EVSEMaximumLimits evseMaxLimits;
+        evseMaxLimits.EVSEMaximumCurrentLimit = powersupply_capabilities.max_export_current_A;
+        evseMaxLimits.EVSEMaximumPowerLimit = powersupply_capabilities.max_export_power_W;
+        evseMaxLimits.EVSEMaximumVoltageLimit = powersupply_capabilities.max_export_voltage_V;
```
a-w50, line 225: see above

interfaces/ISO15118_charger.yaml
SebaLukas: This is an ISO15118 thing. [V2G_CI_MsgDataTypes.xsd](https://standards.iso.org/iso/15118/-2/ed-2/en/V2G_CI_MsgDataTypes.xsd) search for `PhysicalValueType`

a-w50: Yes, this defines a type, but not the meaning of an aggregation of values which have this type.  I would rather call it something like `set_initial_charging_parameters`

modules/simulation/DCSupplySimulator/main/power_supply_DCImpl.hpp
corneliusclaussen: argh, we need to change the auto generator

modules/simulation/JsDCSupplySimulator/index.js
corneliusclaussen: ok

modules/EvseManager/energy_grid/energyImpl.cpp
corneliusclaussen: export on dc power supply is into the car, which is import from grid

modules/EvseManager/energy_grid/energyImpl.cpp
corneliusclaussen: confusing but correct

modules/simulation/JsDCSupplySimulator/index.js
corneliusclaussen: done

interfaces/ISO15118_charger.yaml
corneliusclaussen: renamed the new command, but leaving the types which were not changed by this PR untouched.

modules/EvseManager/EvseManager.hpp
corneliusclaussen: I think update_powersupply_capabilities() is not so long and splitting it up in 3-4 for functions would not really help with readability

Pietfried (approved):

PR #665, opened 2024-05-03
Bugfix/66 graceful crash handling

AssemblyJohn:
## Describe your changes

Updated interface usage for CSR generation

https://github.com/EVerest/libocpp/pull/604
https://github.com/EVerest/libevse-security/pull/70

### DIFF
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -56,7 +56,7 @@ libevse-security:
 # OCPP
 libocpp:
   git: https://github.com/EVerest/libocpp.git
-  git_tag: 3055563
+  git_tag: 3055563beadf47611033c2c8fe100ec6ddac925a
   cmake_condition: "EVEREST_DEPENDENCY_ENABLED_LIBOCPP"
 # Josev
 Josev:
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -50,13 +50,13 @@ libcurl:
 # and would otherwise be overwritten by the version used there
 libevse-security:
   git: https://github.com/EVerest/libevse-security.git
-  git_tag: 72e0b45fe41292227b6438e19e710b99976205b2
+  git_tag: 6e702ef5df568c2f9929a3c3b97a09c0cb4c5b21
   cmake_condition: "EVEREST_DEPENDENCY_ENABLED_LIBEVSE_SECURITY"
 
 # OCPP
 libocpp:
   git: https://github.com/EVerest/libocpp.git
-  git_tag: 37242de85aa536411c419da7c656123a228e4658
+  git_tag: 9deeb534c7f38981a422b53a6f0416bdb791b368
   cmake_condition: "EVEREST_DEPENDENCY_ENABLED_LIBOCPP"
 # Josev
 Josev:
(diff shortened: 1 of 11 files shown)


interfaces/evse_security.yaml
```diff
@@ -103,6 +103,16 @@ cmds:
...
         type: string
+  retrieve_ocsp_cache:
+    description: Command to retrieve the OCSP cache file path of the specified certificate hash
+    arguments:
+      certificate_hash_data:
+        description: Certificate hash data that identifies the certificate for which the cache should be retrieved
+        type: object
+        $ref: /evse_security#/CertificateHashData
+    result:
+      description: Path of data related to the certificate hash, if present
```
Pietfried, line 114: Like the changed response type for the CSR generation, I would also prefer a dedicated response type here that carries more information (status + ocsp result (if present)).

Pietfried (changes requested):

types/evse_security.yaml
```diff
@@ -183,13 +191,38 @@ types:
           minimum: 0
           type: object
           $ref: /evse_security#/OCSPRequestData
-  KeyPair:
+  CertificateOCSP:
+    description: OCSP data related to requested the certificates
+    type: object
+    required:
+      - hash
```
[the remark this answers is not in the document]
AssemblyJohn: Because if we have a chain of returned certificates, for example leaf<-subca2<-subca1, we agreed that a list of the same count of OCSP data will be returned, because matching the ocsp data to a certificate in everest-core will be very difficult, based on a certificate hash, because it is difficult to build that has, especially when intermediaries are involved. So, if a chain of 3 certificates is returned, then a vector of 3 ocsp data is returned, each matched to the certificate from the chain in order.

interfaces/evse_security.yaml
AssemblyJohn: I am not entirely sure what error could be present here, we can only have one state for the OCSP cache related to a certificate. Present or NotPresent. A std::optional seems to be taking into account those 2 cases. Is there anything else that I might be forgetting when handling this case?

AssemblyJohn:
Some comments resolved, awaiting further feedback.

dependencies.yaml
```diff
@@ -47,13 +47,13 @@ libcurl:
...
   git: https://github.com/EVerest/libevse-security.git
-  git_tag: v0.6.0
+  git_tag: b56760f
   cmake_condition: "EVEREST_DEPENDENCY_ENABLED_LIBEVSE_SECURITY"
 
 # OCPP
 libocpp:
   git: https://github.com/EVerest/libocpp.git
-  git_tag: v0.11.0
+  git_tag: 3055563
```
Pietfried, line 56:
```suggestion
  git_tag: "3055563"
```

SirVer (approved):
Reviewed for Bazel only. LGTM.

interfaces/evse_security.yaml
Pietfried: The issue is that the function returns an `std::string` and not and `std::optional<std::string>` . For this reason we need a more complex return type like for the CSR generation

AssemblyJohn: Understood, makes sense, since I've forgot that the API does not allow std::optional return types.

SebaLukas (approved):
:+1:

PR #669, opened 2024-05-06
EvseManager: Fix bug where HLC is not stopped on unexpected CP state

FaHaGit:
Before the HLC session was not stopped by the EvseManager module in case of an unexpected CP state transition while charging. Now the HLC session is stopped in case of a CP state transition C/D -> B/E/F.

## Describe your changes

Implemented CPEvent::BCDtoEF event. Stop HLC session in case of CP state transition C/D -> B/E/F.

https://github.com/EVerest/everest-core/issues/668

### DIFF
--- a/modules/EvseManager/Charger.cpp
+++ b/modules/EvseManager/Charger.cpp
@@ -793,6 +793,20 @@ void Charger::process_cp_events_state(CPEvent cp_event) {
         if (cp_event == CPEvent::CarRequestedStopPower) {
             shared_context.iec_allow_close_contactor = false;
             shared_context.current_state = EvseState::ChargingPausedEV;
+            // Tell HLC stack to stop the session. Normally the session should have already been stopped by the EV, but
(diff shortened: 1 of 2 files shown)


FaHaGit:
I would be happy if it could be tested again independently of me. I am not completly sure if the new event state CPEvent::BCDtoEF could lead to unexpected behavior.

corneliusclaussen:
> I would be happy if it could be tested again independently of me. I am not completly sure if the new event state CPEvent::BCDtoEF could lead to unexpected behavior.

Don't really have the time to test this at the moment, but it looks good to me. As far as I can see the additional BCDtoEF will not cause side effects. It will trigger a SLAC->leaveBCD, but this does not do anything as far as I can see in the slac implementation.

corneliusclaussen (approved):

SebaLukas (approved):
:+1:

PR #674, opened 2024-05-07
Enable DIN70121 by default

corneliusclaussen:
## Describe your changes

Up to now, DIN70121 was disabled by default. If no options are given, I think we should enable all protocols for maximum compatibility.


SebaLukas (approved):
:+1:

PR #677, opened 2024-05-08
Added OpenSSL server that supports TLS certificate status request

james-ctc:
## Describe your changes

feat: added optional OpenSSL server
feat: integration of OpenSSL into EvseV2G
feat: added cmake integration with optional OpenSSL
feat: added OpenSSL sha256 and ECDSA functions + ISO test vectors
feat: added and tested EXI signatures
feat: added base64 encode/decode
feat: OpenSSL or MbedTLS for EvseV2G

cmake -DEVEREST_CORE_BUILD_TESTING=ON -USING_MBED_TLS=OFF GNinja .. default is to use MBED TLS


james-ctc:
UDP with session key would be a good addition - the aim is to add this as a subsequent PR.
The aim at this point is to check that the OpenSSL refactor is a working option

james-ctc:
> Can or should some of this be in libevse-security?
Worthy of consideration - my take is to have something working in the first instance and then look to have common functionality in libevse-security. Aspects of libevse-security aren't abstracted to make unit testing easy especially since it is currently focused on being a TLS client rather than server.

modules/EvseV2G/CMakeLists.txt
```diff
@@ -31,17 +42,39 @@ target_link_libraries(${MODULE_NAME}
         mbedtls
         mbedx509
         everest::openv2g
+        OpenSSL::SSL
+        OpenSSL::Crypto
```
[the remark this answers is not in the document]
james-ctc: there were additional uses of mbedtls_base64_encode and mbedtls_base64_decode that have now been removed when building for OpenSSL.

modules/EvseManager/tests/CMakeLists.txt
```diff
@@ -6,7 +6,7 @@ add_dependencies(${TEST_TARGET_NAME} ${MODULE_NAME})
 get_target_property(GENERATED_INCLUDE_DIR generate_cpp_files EVEREST_GENERATED_INCLUDE_DIR)
 
 target_include_directories(${TEST_TARGET_NAME} PRIVATE
-    . ..
+    . .. ../../../tests/include
```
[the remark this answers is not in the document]
james-ctc: EvseManager and EvseV2G tests share a common stub file for unit tests. EvseManager needed to know the common location.

modules/EvseManager/tests/EvseManagerStub.hpp
```diff
@@ -6,6 +6,7 @@
 
 #include "ModuleAdapterStub.hpp"
 #include <ErrorHandling.hpp>
+#include <ModuleAdapterStub.hpp>
```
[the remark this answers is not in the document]
james-ctc: EvseManager and EvseV2G tests share a common stub file for unit tests. EvseManager needed to know the common location. <> preferred over "" when the file isn't in the same directory.

modules/EvseV2G/CMakeLists.txt
```diff
@@ -19,6 +29,7 @@ target_sources(${MODULE_NAME}
 # ev@c55432ab-152c-45a9-9d2e-7281d50c69c3:v1
 # Add pkg-config functionality
 find_package(PkgConfig REQUIRED)
+find_package(OpenSSL)
```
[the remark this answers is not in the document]
james-ctc: see below

modules/EvseV2G/tests/evse_secrutyIntfStub.hpp
```diff
@@ -0,0 +1,57 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Pionix GmbH and Contributors to EVerest
+
+#ifndef EVSE_SECURITYINTFSTUB_H_
+#define EVSE_SECURITYINTFSTUB_H_
```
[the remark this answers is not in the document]
james-ctc: file renamed

config/config-sil-dc.yaml
SebaLukas, line 1: Please change `tls_active: true` and `payment_enable_contract: false` to the original state.

james-ctc: done.

SebaLukas (commented):
First batch :sweat_smile:

modules/EvseV2G/CMakeLists.txt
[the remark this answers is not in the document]
james-ctc: done in lib/statging/tls/CMakeLists.txt and removed from modules/EvseV2G/CMakeLists.txt

modules/EvseV2G/EvseV2G.cpp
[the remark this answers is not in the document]
james-ctc: It was moved to ensure that EvseSecurity is ready to receive the MQTT request.

modules/EvseV2G/tests/tls_main.cpp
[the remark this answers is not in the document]
james-ctc: added blank line to quite a few files including that one.

modules/EvseV2G/connection/tls_connection.cpp
```diff
@@ -0,0 +1,256 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2020 - 2023 Pionix GmbH and Contributors to EVerest
+
+#include "tls_connection.hpp"
+#include "connection.hpp"
+#include "log.hpp"
+#include "tls.hpp"
```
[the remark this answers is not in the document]
james-ctc: fixed in a number of places.

modules/EvseV2G/EvseV2G.cpp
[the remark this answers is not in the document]
SebaLukas: Alright :+1:

modules/EvseV2G/tests/tls_main.cpp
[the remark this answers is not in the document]
SebaLukas: Is there any reason not to move this test to `lib/staging/tls`?

modules/EvseManager/tests/CMakeLists.txt
SebaLukas, line 1: Separate into an extra PR

modules/EvseManager/tests/EvseManagerStub.hpp
SebaLukas, line 1: Separate into an extra PR

tests/README.md
SebaLukas, line 1: Separate into an extra PR

tests/include/ModuleAdapterStub.hpp
SebaLukas, line 1: Separate into an extra PR

SebaLukas (commented):
These changes could be separated and moved to a separate PR.

modules/EvseV2G/tests/tls_main.cpp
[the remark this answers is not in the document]
james-ctc: none other than it is already moved on another branch which will be the next PR for this work (That PR includes the status_request_v2 functionality).

james-ctc:
`These changes could be separated and moved to a separate PR.`
They can be moved but this PR will then depend on that one being approved.

SebaLukas:
Yes, this is clear that there is a dependency. But the few changes in the extra PR can be merged more quickly than this large PR.

james-ctc:
I'll move them out, but there can be no further updates to this PR until it is approved and merged as this one will no longer compile.

james-ctc:
rebase/merged against main with the new libcbv2g.
addressed many merge issues.

lib/staging/tls/tests/README.md
```diff
@@ -0,0 +1,41 @@
...
+$ cd build
+$ cmake -GNinja -DEVEREST_CORE_BUILD_TESTING=ON ..
+$ ninja install
+```
+
+`touch release.json` may be needed if it hasn't been created (then re-run `ninja install`).
+
+## Unit tests
+
+- `./openssl_test`
```
SebaLukas, line 17: Is this up to date? Because I can not find this

modules/EvseV2G/charger/ISO15118_chargerImpl.hpp
```diff
@@ -10,7 +10,7 @@
 
 #include <generated/interfaces/ISO15118_charger/Implementation.hpp>
 
-#include "../EvseV2G.hpp"
```
SebaLukas, line 13: Why deleting `..`?

lib/staging/tls/tests/tls_main.cpp
```diff
@@ -0,0 +1,99 @@
...
+#else
+    // config.cipher_list = "ECDHE-ECDSA-AES128-SHA256:ECDH-ECDSA-AES128-SHA256";
+    config.cipher_list = "ECDHE-ECDSA-AES128-SHA256";
+    config.ciphersuites = "TLS_AES_128_GCM_SHA256:TLS_AES_256_GCM_SHA384";
+    // config.ciphersuites = "";
+#endif
+    config.certificate_chain_file = "server_chain.pem";
+    config.private_key_file = "server_priv.pem";
+    config.verify_locations_file = "client_root_cert.pem";
+    // config.ocsp_response_files = {"ocsp_response.der", nullptr};
```
SebaLukas, line 77: Delete comment

lib/staging/tls/tests/tls_main.cpp
```diff
@@ -0,0 +1,99 @@
...
+
+#if 0
+    config.cipher_list =
+        "ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES256-GCM-SHA384:AES128-GCM-SHA256:AES256-GCM-SHA384";
+    config.ciphersuites = "TLS_AES_256_GCM_SHA384:TLS_AES_128_GCM_SHA256";
+#else
+    // config.cipher_list = "ECDHE-ECDSA-AES128-SHA256:ECDH-ECDSA-AES128-SHA256";
+    config.cipher_list = "ECDHE-ECDSA-AES128-SHA256";
+    config.ciphersuites = "TLS_AES_128_GCM_SHA256:TLS_AES_256_GCM_SHA384";
+    // config.ciphersuites = "";
```
SebaLukas, line 72: Delete comment or specify if tls 1.3 should be used

SebaLukas (commented):
Some comments

modules/EvseV2G/charger/ISO15118_chargerImpl.hpp
james-ctc: relative paths in #include are generally not a good idea. The compiler flags already list the parent directory as place to look for include files, hence the .. isn't needed.

Main reason, I've used IDEs that try to be "helpful" and have automatically added unnecessary relative paths. It is habit to remove them. I don't know if there is any specific EVerest style guidance on whether .. in #include should be used or not.

Happy to revert that.

lib/staging/tls/tests/tls_main.cpp
james-ctc: The comment is there to allow the tls_server to be rebuild quickly with different configurations for testing.
same for the #if 0 lines.

lib/staging/tls/tests/tls_main.cpp
james-ctc: The comment is there to allow the tls_server to be rebuild quickly with different configurations for testing.
same for the #if 0 lines.

In this specific case to force TLS 1.3 for testing with TLS 1.2 clients (that may or may not support TLS 1.3)

modules/EvseV2G/charger/ISO15118_chargerImpl.hpp
SebaLukas: Thanks for the explanation. Then we'll keep it that way.

lib/staging/tls/tests/tls_main.cpp
SebaLukas: :+1:

lib/staging/tls/tests/tls_main.cpp
SebaLukas: :+1:

lib/staging/tls/openssl-patch.md
```diff
@@ -0,0 +1,66 @@
+# OpenSSL 3.0.8 patch
+
+The file `openssl-3.0.8-feat-updates-to-support-status_request_v2.patch` is a patch to OpenSSL 3.0.8 to support the `status_request_v2` TLS extension defined in [RFC 6961](https://datatracker.ietf.org/doc/html/rfc6961).
```
AssemblyJohn, line 3: I have a question here, how is this linked to the openssl library? Is it a local copy that this module is linked against? Does that mean that this module will use version 3.0.8 of openssl while other parts of the project will use another openssl version? Or do we have some mechanism in place in order to make sure that everyone uses this 3.0.8 version?

james-ctc: If someone wants to deploy EVerest with support for status_request_v2 then a patched OpenSSL will be required.

My expectation is that EVerest and other code will be built via Yocto to produce the image for deploying to a charging station. Yocto has support for applying patches as part of the build process and hence a single patched version of OpenSSL would be used.

Where Yocto isn't used then my recommendation would be to clone OpenSSL, apply the patch and install (e.g. to /usr/local) and use pkg config variables so that when EVerest is built the /usr/local version of OpenSSL is used.
I would not recommend having some parts of EVerest using different OpenSSL versions.

The main aim is that 15118 TLS connections will work on patched or unpatched versions of OpenSSL.
Unpatched will not support the TLS status_request_v2 handshake extension.

Note that status_request_v2 is only supported for TLS 1.2, TLS 1.3 sessions ignore the status_request_v2 extension.

In summary: the recommended mechanism to use a single patched OpenSSL version is to use Yocto - i.e. meta-everest

lib/staging/tls/openssl_util.cpp
```diff
@@ -0,0 +1,518 @@
...
+    auto* cert = d2i_X509(nullptr, &ptr, static_cast<std::int64_t>(len));
+    if (cert == nullptr) {
+        log_error("d2i_X509");
+    } else {
+        result = Certificate_ptr{cert, &X509_free};
+    }
+    return result;
+}
+
+verify_result_t verify_certificate(const x509_st* cert, const CertificateList& trust_anchors,
```
AssemblyJohn, line 389: My last point, I've seen that some utilities from the libevse-security were used for certificate handling. Can't code be unified and to use the verify certificate routines from that library instead of doing this again manually? Maybe it will make code easier to maintain in the future.

lib/staging/tls/openssl-patch.md
AssemblyJohn: Understood, but yes, this will make testing a bit difficult. Maybe some extra steps might be required for explanation in the readme in case someone requires this?

AssemblyJohn (approved):
For my part the openssl part looks ok, but with further possible improvements if we require to share more code (with the security lib), instead of duplicating it. In that way the code is also safer since we don't need to test for security vulnerabilities in two places, at least where's common functionality. The PR is very large, and I'd be against doing those modifications now.

lib/staging/tls/tests/README.md
james-ctc: updated.

SebaLukas (commented):
Last comment before testing PnC and debugging some tests with my setup

lib/staging/tls/tls.cpp
```diff
@@ -0,0 +1,1459 @@
...
+    // - SSL_TLSEXT_ERR_ALERT_FATAL abort connection
+    bool bSet{false};
+    int result = SSL_TLSEXT_ERR_NOACK;
+    openssl::sha_256_digest_t digest{};
+
+    if (ctx != nullptr) {
+        const auto* cert = SSL_get_certificate(ctx);
+        const auto* name = X509_get_subject_name(cert);
+        bSet = OcspCache::digest(digest, cert);
+        std::cout << "OCSP:  ";
```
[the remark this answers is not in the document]
james-ctc: std::cout should have been removed. (or been in a #ifdef block)
EVLOG isn't used because it has a memory leak (according to valgrind memcheck).
That makes it much harder to see if there are issues with the code under test.
openssl::log_error() has a #ifdef to prevent EVLOG being used for unit tests. EVLOG is used for the running system.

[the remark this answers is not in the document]
corneliusclaussen: The idea was that those libs are kind of everest independent. An alternative would be to use a logging callback from the lib and the module can then register a callback function that uses EVLOG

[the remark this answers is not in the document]
james-ctc: removed EVLOG. logging now determined by external handler

SebaLukas (approved):
Great job :+1:

corneliusclaussen (commented):
I think we should make openssl the default setting when compiling. We need some more real world EV testing, but I'm afraid we will not get enough testing if it is not enabled by default. Since it is merged after the June release we have some time until the next stable release to fix potential issues

corneliusclaussen (approved):

james-ctc:
Fixed integration test failure.

item|note
----|----
underlying cause|Evse Security could not find certificates and key for ISO 15118 TLS server causing the module to exit
why was it passing previously|The original code didn't attempt SSL configuration until there was an incoming connection (which there isn't in any of the tests)
simple fix|provide certificate chain and key as part of the CI pipeline
actual fix|allow module to start with missing configuration. Missing SSL configuration requested when there is an incoming connection
Codacy|mostly false positives - not sure how to correctly tag that the issue has been investigated

PR #685, opened 2024-05-14
EvseV2G: Sending the TLS session key over a UDP message

MarzellT:
## Describe your changes

* adds a udp socket to the EvseV2G module that sends the TLS secrets for debugging

### DIFF
--- a/modules/EvseV2G/connection.cpp
+++ b/modules/EvseV2G/connection.cpp
@@ -600,6 +600,19 @@ error_out:
 static void ssl_key_log_debug_callback(void* ACtx, int ALevel, const char* AFile, int ALine, const char* AStr) {
     keylogDebugCtx* ctx = static_cast<keylogDebugCtx*>(ACtx);
 
+    const auto add_to_udp_message = [&](const auto str) { ctx->udp_buffer.append(str); };
+
+    const auto add_character_to_udp_message = [&](const auto character) { ctx->udp_buffer.push_back(character); };
+
+    const auto send_udp_message = [&]() {
+        const auto message_size = ctx->udp_buffer.size();
+        const auto ret = send(ctx->udp_socket, ctx->udp_buffer.c_str(), message_size, 0);
+
+        if (ret != message_size) {
+            EVLOG_error << "send() failed: " << strerror(errno);
+        }
+    };
+
     ((void)ALevel);
     ((void)AFile);
     ((void)ALine);
@@ -607,12 +620,16 @@ static void ssl_key_log_debug_callback(void* ACtx, int ALevel, const char* AFile
     if (strstr(AStr, "dumping 'client hello, random bytes' (32 bytes)")) {
         ctx->inClientRandom = true;
         ctx->hexdumpLinesToProcess = 2;
-        fputs("CLIENT_RANDOM ", ctx->file);
+        const auto message = "CLIENT_RANDOM ";
+        fputs(message, ctx->file);
+        add_to_udp_message(message);
         return;
     } else if (strstr(AStr, "dumping 'master secret' (48 bytes)")) {
         ctx->inMasterSecret = true;
         ctx->hexdumpLinesToProcess = 3;
-        fputc(' ', ctx->file);
+        const auto message = ' ';
+        fputc(message, ctx->file);
+        add_character_to_udp_message(message);
         return;
     } else if (((ctx->inClientRandom == false) && (ctx->inMasterSecret == false)) ||
                (ctx->hexdumpLinesToProcess == 0)) {
@@ -632,7 +649,9 @@ static void ssl_key_log_debug_callback(void* ACtx, int ALevel, const char* AFile
         if ((('0' <= c1 && c1 <= '9') || ('a' <= c1 && c1 <= 'f')) &&
             (('0' <= c2 && c2 <= '9') || ('a' <= c2 && c2 <= 'f')) && c3 == ' ') {
             fputc(c1, ctx->file);
+            add_character_to_udp_message(c1);
             fputc(c2, ctx->file);
+            add_character_to_udp_message(c2);
         } else {
             goto reset; /* unexpected non-hex char */
         }
@@ -644,8 +663,13 @@ reset:
     ctx->hexdumpLinesToProcess = 0;
     ctx->inClientRandom = false;
     ctx->inMasterSecret = false;
-    fputc('\n', ctx->file); /* finish key log line */
+
+    const auto message = '\n';
+    fputc(message, ctx->file); /* finish key log line */
     fflush(ctx->file);
+
+    send_udp_message();
+    ctx->udp_buffer = {};
 }
 
 /**
@@ -735,6 +759,13 @@ static void* connection_handle_tls(void* data) {
 
         v2g_ctx->tls_log_ctx.file = fopen(tls_log_path.c_str(), "a");
 
+        const auto udp_socket = create_udp_socket(conn->ctx->udp_port, v2g_ctx->if_name);
+        if (udp_socket < 0) {
+            goto thread_exit;
+        }
+        EVLOG_info << "Sending TLS secrets to port: " << conn->ctx->udp_port;
+        v2g_ctx->udp_socket = udp_socket;
+
         if (v2g_ctx->tls_log_ctx.file == NULL) {
             dlog(DLOG_LEVEL_INFO, "%s", "Failed to open file path for TLS key logging: %s", strerror(errno));
             mbedtls_debug_set_threshold(MBEDTLS_DEBUG_LEVEL_NO_DEBUG);
@@ -744,6 +775,7 @@ static void* connection_handle_tls(void* data) {
             v2g_ctx->tls_log_ctx.inClientRandom = false;
             v2g_ctx->tls_log_ctx.inMasterSecret = false;
             v2g_ctx->tls_log_ctx.hexdumpLinesToProcess = 0;
+            v2g_ctx->tls_log_ctx.udp_socket = v2g_ctx->udp_socket;
             mbedtls_ssl_conf_dbg(ssl_config, ssl_key_log_debug_callback, &v2g_ctx->tls_log_ctx);
         }
     }
@@ -812,6 +844,7 @@ static void* connection_handle_tls(void* data) {
     if (conn->ctx->tls_log_ctx.file != NULL) {
         mbedtls_debug_set_threshold(MBEDTLS_DEBUG_LEVEL_NO_DEBUG);
         fclose(conn->ctx->tls_log_ctx.file);
+        close(conn->ctx->udp_socket);
         memset(&conn->ctx->tls_log_ctx, 0, sizeof(conn->ctx->tls_log_ctx));
     }
 
@@ -938,6 +971,9 @@ static void* connection_server(void* data) {
                  strerror(errno));
         }
 
+        // store the port to create a udp socket
+        conn->ctx->udp_port = ntohs(addr.sin6_port);
+
         if (pthread_create(&conn->thread_id, &attr,
                            conn->is_tls_connection ? connection_handle_tls : connection_handle_tcp, conn) != 0) {
             dlog(DLOG_LEVEL_ERROR, "pthread_create() failed: %s", strerror(errno));
@@ -984,4 +1020,64 @@ int connection_start_servers(struct v2g_context* ctx) {
     }
 
     return 0;
+}
+
+int create_udp_socket(const uint16_t udp_port, const char* interface_name) {
+    constexpr auto LINK_LOCAL_MULTICAST = "ff02::1";
+
+    int udp_socket = socket(AF_INET6, SOCK_DGRAM, 0);
+    if (udp_socket < 0) {
+        EVLOG_error << "Could not create socket: " << strerror(errno);
+        return udp_socket;
+    }
+
+    // source setup
+
+    // find port between 49152-65535
+    auto could_bind = false;
+    auto source_port = 49152;
+    for (; source_port < 65535; source_port++) {
+        sockaddr_in6 source_address = {AF_INET6, htons(source_port)};
+        if (bind(udp_socket, reinterpret_cast<sockaddr*>(&source_address), sizeof(sockaddr_in6)) == 0) {
+            could_bind = true;
+            break;
+        }
+    }
+
+    if (!could_bind) {
+        EVLOG_error << "Could not bind: " << strerror(errno);
+        return -1;
+    }
+
+    EVLOG_info << "UDP socket bound to source port: " << source_port;
+
+    const auto index = if_nametoindex(interface_name);
+    auto mreq = ipv6_mreq{};
+    mreq.ipv6mr_interface = index;
+    if (inet_pton(AF_INET6, LINK_LOCAL_MULTICAST, &mreq.ipv6mr_multiaddr) <= 0) {
+        EVLOG_error << "Failed to setup multicast address" << strerror(errno);
+        return -1;
+    }
+    if (setsockopt(udp_socket, IPPROTO_IPV6, IPV6_ADD_MEMBERSHIP, &mreq, sizeof(mreq)) < 0) {
+        EVLOG_error << "Could not add multicast group membership: " << strerror(errno);
+        return -1;
+    }
+
+    if (setsockopt(udp_socket, IPPROTO_IPV6, IPV6_MULTICAST_IF, &index, sizeof(index)) < 0) {
+        EVLOG_error << "Could not set interface name: " << interface_name << "with error: " << strerror(errno);
+    }
+
+    // destination setup
+    sockaddr_in6 destination_address = {AF_INET6, htons(udp_port)};
+    if (inet_pton(AF_INET6, LINK_LOCAL_MULTICAST, &destination_address.sin6_addr) <= 0) {
+        EVLOG_error << "Failed to setup server address" << strerror(errno);
+    }
+    const auto connected =
+        connect(udp_socket, reinterpret_cast<sockaddr*>(&destination_address), sizeof(sockaddr_in6)) == 0;
+    if (!connected) {
+        EVLOG_error << "Could not connect: " << strerror(errno);
+        return -1;
+    }
+
+    return udp_socket;
 }
--- a/modules/EvseV2G/connection.hpp
+++ b/modules/EvseV2G/connection.hpp
@@ -12,6 +12,7 @@
 
 int connection_init(struct v2g_context* ctx);
 int connection_start_servers(struct v2g_context* ctx);
+int create_udp_socket(const uint16_t udp_port, const char* interface_name);
 
 /*!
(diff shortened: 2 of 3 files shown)


modules/EvseV2G/connection.cpp
```diff
@@ -600,19 +601,37 @@ static bool connection_init_tls(struct v2g_context* ctx) {
 static void ssl_key_log_debug_callback(void* ACtx, int ALevel, const char* AFile, int ALine, const char* AStr) {
     keylogDebugCtx* ctx = static_cast<keylogDebugCtx*>(ACtx);
 
+    const auto addToUDPMessage = [&](const auto str) { ctx->udp_buffer.append(str); };
+
+    const auto addCharacterToUDPMessage = [&](const auto character) { ctx->udp_buffer.push_back(character); };
+
+    const auto sendUDPMessage = [&]() {
+        const auto message_size = ctx->udp_buffer.size();
+        auto ret = send(ctx->udp_socket, ctx->udp_buffer.c_str(), message_size, 0);
```
MarzellT, line 610: case

config/config-sil-dc.yaml
```diff
@@ -3,7 +3,8 @@ active_modules:
     module: EvseV2G
     config_module:
       device: auto
-      tls_security: allow
+      tls_security: force
+      tls_key_logging: true
     connections:
```
MarzellT, line 8: revert

modules/EvseV2G/connection.cpp
```diff
@@ -699,8 +724,14 @@ static void* connection_handle_tcp(void* data) {
 /**
  * This is the 'main' function of a thread, which handles a TLS connection.
  */
-static void* connection_handle_tls(void* data) {
-    struct v2g_connection* conn = static_cast<v2g_connection*>(data);
+template <typename v2g_connectionT, typename sdp_queryT> static void* connection_handle_tls(void* data) {
```
MarzellT, line 727: template or not?

modules/EvseV2G/connection.cpp
```diff
@@ -699,8 +724,14 @@ static void* connection_handle_tcp(void* data) {
...
+    auto* const conn_and_sdp_tuple = static_cast<std::tuple<v2g_connectionT, sdp_queryT>*>(data);
+    auto* const conn = std::get<0>(*conn_and_sdp_tuple);
+    auto* const sdp_query = std::get<1>(*conn_and_sdp_tuple);
+
+    // print port
```
MarzellT, line 732: unnecessary comment

modules/EvseV2G/connection.cpp
```diff
@@ -985,3 +1033,51 @@ int connection_start_servers(struct v2g_context* ctx) {
...
+    // find suitable addresses
+    addrinfo hints;
+    addrinfo *result, *result_pointer;
+    int udp_socket, s;
+
+    memset(&hints, 0, sizeof(addrinfo));
+    hints.ai_family = AF_UNSPEC;    // allow IPv4 or IPv6
+    hints.ai_socktype = SOCK_DGRAM; // Datagram socket
+    hints.ai_flags = AI_ALL;        // For IPv4 or IPv6
+    hints.ai_protocol = 0;          // Any protocol
```
MarzellT, line 1046: os selects protocol

modules/EvseV2G/connection.cpp
```diff
@@ -985,3 +1033,51 @@ int connection_start_servers(struct v2g_context* ctx) {
...
+
+    // convert port to string
+    char port_str[6];
+    snprintf(port_str, sizeof(port_str), "%d", ntohs(sdp_query_variable->remote_addr.sin6_port));
+
+    //                    alternatively:
+    //                    "localhost"
+    s = getaddrinfo(nullptr, port_str, &hints, &result);
+
```
MarzellT, line 1055: `nullptr` vs. `"localhost"`?

modules/EvseV2G/connection.cpp
```diff
@@ -985,3 +1033,51 @@ int connection_start_servers(struct v2g_context* ctx) {
...
+    if (s != 0) {
```
MarzellT, line 1056: better name than `s`

modules/EvseV2G/connection.cpp
```diff
@@ -985,3 +1021,51 @@ int connection_start_servers(struct v2g_context* ctx) {
...
+    int udp_socket;
+
+    memset(&hints, 0, sizeof(addrinfo));
+    hints.ai_family = AF_UNSPEC;    // allow IPv4 or IPv6
+    hints.ai_socktype = SOCK_DGRAM; // Datagram socket
+    hints.ai_flags = AI_ALL;        // For IPv4 or IPv6
+    hints.ai_protocol = 0;          // OS selects best protocol (best practise)
+
+    // convert port to string
+    char port_str[6];
```
a-w50, line 1037: I might be a bit picky here, but the function takes an int, so it could be also 100000, which won't fit into 5 chars.  So maybe do a validation or accept an `uint16_t`.  I would probably also prefer `std::to_string` instead of `snprintf`.

modules/EvseV2G/connection.cpp
```diff
@@ -985,3 +1021,51 @@ int connection_start_servers(struct v2g_context* ctx) {
...
+    snprintf(port_str, sizeof(port_str), "%d", udp_port);
+
+    //              alternatively:
+    //              "localhost"
+    auto get_addrinfo_result = getaddrinfo(nullptr, port_str, &hints, &result);
```
a-w50, line 1042: `const auto`

modules/EvseV2G/connection.cpp
```diff
@@ -985,3 +1021,51 @@ int connection_start_servers(struct v2g_context* ctx) {
...
+
+        close(udp_socket);
+    }
+
+    if (result_pointer == nullptr) {
+        dlog(DLOG_LEVEL_ERROR, "Could not connect");
+        return -1;
+    }
+
+    freeaddrinfo(result);
```
a-w50, line 1069: probably a memory leak, because you might return earlier

modules/EvseV2G/connection.cpp
```diff
@@ -985,3 +1021,51 @@ int connection_start_servers(struct v2g_context* ctx) {
...
+    auto get_addrinfo_result = getaddrinfo(nullptr, port_str, &hints, &result);
+
+    if (get_addrinfo_result != 0) {
+        dlog(DLOG_LEVEL_ERROR, "getaddrinfo: %s", gai_strerror(get_addrinfo_result));
+        return -1;
+    }
+
+    // create the socket
+    for (result_pointer = result; result_pointer != nullptr; result_pointer = result_pointer->ai_next) {
+        udp_socket = socket(result_pointer->ai_family, result_pointer->ai_socktype, result_pointer->ai_protocol);
```
a-w50, line 1051: what is the idea behind this?  In case creating an udp socket fails, just try the next available addresses?  This doesn't really look deterministic to me.

modules/EvseV2G/connection.cpp
```diff
@@ -985,3 +1021,51 @@ int connection_start_servers(struct v2g_context* ctx) {
...
+    //              "localhost"
+    auto get_addrinfo_result = getaddrinfo(nullptr, port_str, &hints, &result);
+
+    if (get_addrinfo_result != 0) {
+        dlog(DLOG_LEVEL_ERROR, "getaddrinfo: %s", gai_strerror(get_addrinfo_result));
+        return -1;
+    }
+
+    // create the socket
+    for (result_pointer = result; result_pointer != nullptr; result_pointer = result_pointer->ai_next) {
```
a-w50, line 1050: `result_pointer = result` sounds wrong, better use `initial_addr` and `current_addr` or something similar

modules/EvseV2G/v2g.hpp
```diff
@@ -166,6 +166,8 @@ typedef struct keylogDebugCtx {
     bool inClientRandom;
     bool inMasterSecret;
     uint8_t hexdumpLinesToProcess;
+    int udp_socket;
+    std::string udp_buffer{};
```
a-w50, line 170: the `{}` in case of string is not really necessary

modules/EvseV2G/connection.cpp
```diff
@@ -985,3 +1021,51 @@ int connection_start_servers(struct v2g_context* ctx) {
 
     return 0;
 }
+int create_udp_socket(const int udp_port) {
+    // find suitable addresses
+    addrinfo hints;
+    addrinfo *result, *result_pointer;
```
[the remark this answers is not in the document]
a-w50: Ok, most of the EvseV2G module was written in C90 style ...

modules/EvseV2G/connection.cpp
```diff
@@ -644,8 +664,12 @@ static void ssl_key_log_debug_callback(void* ACtx, int ALevel, const char* AFile
     ctx->hexdumpLinesToProcess = 0;
     ctx->inClientRandom = false;
     ctx->inMasterSecret = false;
-    fputc('\n', ctx->file); /* finish key log line */
+    const auto message = '\n';
+    fputc(message, ctx->file); /* finish key log line */
+    add_character_to_udp_message(message);
     fflush(ctx->file);
+    send_udp_message();
+    ctx->udp_buffer = {};
```
a-w50, line 672: I would not intertwine the udp and FILE handling like this, better do them one after another

modules/EvseV2G/connection.cpp
```diff
@@ -600,19 +600,37 @@ static bool connection_init_tls(struct v2g_context* ctx) {
 static void ssl_key_log_debug_callback(void* ACtx, int ALevel, const char* AFile, int ALine, const char* AStr) {
     keylogDebugCtx* ctx = static_cast<keylogDebugCtx*>(ACtx);
 
+    const auto add_to_udp_message = [&](const auto str) { ctx->udp_buffer.append(str); };
+
+    const auto add_character_to_udp_message = [&](const auto character) { ctx->udp_buffer.push_back(character); };
+
+    const auto send_udp_message = [&]() {
+        const auto message_size = ctx->udp_buffer.size();
+        auto ret = send(ctx->udp_socket, ctx->udp_buffer.c_str(), message_size, 0);
```
a-w50, line 609: `const auto`

modules/EvseV2G/connection.cpp
```diff
@@ -600,19 +600,37 @@ static bool connection_init_tls(struct v2g_context* ctx) {
...
+
+        // check ret
```
a-w50, line 611: Looking at the next `if`, I would consider this comment redundant.

modules/EvseV2G/connection.cpp
SebaLukas: I'm currently discussing this with @MarzellT, whether this is too much and could be done more simply.

modules/EvseV2G/connection.cpp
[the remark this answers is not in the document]
MarzellT: Yes, it's a weird situation, perhaps it would still be best to use c++ style as much as possible. Here was basically following the `getaddrinfo 3` manpage which I assumed would be best practicel

modules/EvseV2G/connection.cpp
MarzellT: True, that would be a bug. Maybe it would make sense to limit the function input to 16 bits instead as that's the ports max size

modules/EvseV2G/connection.cpp
MarzellT: I think it wouldn't be a memory leak as in any other return case the `result` should be a nullptr

modules/EvseV2G/connection.cpp
MarzellT: It might make sense to simplify this. I thought this was the "correct" way of creating sockets but I must admit I'm not sure. Basically, you would let the OS take care of creating an appropriate socket given the `hints`. It would be possible to request IPv6 explicitly changing the `ai_family` type in the hints but I'm not sure if the whole process is necessary. I don't really think it's necessary to check more than the first result of `getaddrinfo` but at the same time I'm not sure what could be the possible results of that function call

modules/EvseV2G/v2g.hpp
MarzellT: True

modules/EvseV2G/connection.cpp
```diff
@@ -644,8 +663,14 @@ static void ssl_key_log_debug_callback(void* ACtx, int ALevel, const char* AFile
     ctx->hexdumpLinesToProcess = 0;
     ctx->inClientRandom = false;
     ctx->inMasterSecret = false;
-    fputc('\n', ctx->file); /* finish key log line */
+
+    const auto message = '\n';
+    fputc(message, ctx->file); /* finish key log line */
     fflush(ctx->file);
+
+    add_character_to_udp_message(message);
```
SebaLukas, line 671: Newline is not necessary here

SebaLukas (changes requested):

modules/EvseV2G/connection.cpp
a-w50: Ah, well then it is even more wrong :).  `freeaddrinfo` needs to be called on the linked list, returned by `getaddrinfo`, so the initial one.  Calling `freeaddrinfo` on members of the linked list is probably undefined behaviour (UB).

modules/EvseV2G/connection.cpp
```diff
@@ -985,3 +1022,51 @@ int connection_start_servers(struct v2g_context* ctx) {
 
     return 0;
 }
+int create_udp_socket(const u_int16_t udp_port) {
```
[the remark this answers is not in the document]
a-w50: :+1: , still return values of `inet_pton` and `connect` should be checked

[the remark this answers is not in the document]
MarzellT: Why should this be a multicast address? Also as far as I'm aware if we use this address I need to specify a network interface. Is this what we need?

modules/EvseV2G/connection.cpp
```diff
@@ -985,3 +1022,44 @@ int connection_start_servers(struct v2g_context* ctx) {
...
+        dlog(DLOG_LEVEL_ERROR, "Could not get address info");
+        return -1;
+    }
+
+    const auto udp_socket = socket(result->ai_family, result->ai_socktype, result->ai_protocol);
+    if (udp_socket == -1) {
+        dlog(DLOG_LEVEL_ERROR, "Could not create socket");
+        return -1;
+    }
+    const auto connected = connect(udp_socket, result->ai_addr, result->ai_addrlen);
```
barsnick, line 1057: The TLS key extraction of the dSPACE Wireshark plugin requires a particular source port:

> The disclosure message is a UDP packet within the source port range 49152-65535

See also https://github.com/dspace-group/dsV2Gshark/blob/6da7c8f673c367a91a15d2da2ae9a9ea8dc68a9a/Wireshark/plugins/v2gtlssecret.lua#L207.

You need to `bind()` your socket to such a source port for this particular plugin to work. (The payload should work, at first glance.)

Untested, just from reading the plugin.

SebaLukas: That is a good point. I was experimenting with an udp socket client yesterday and unfortunately the OS sometimes selects ports that are not in the dynamic port radius (49152-65535). So we should bind a random port (between 49152 and 65535) for the UDP client as well.

modules/EvseV2G/connection.cpp
SebaLukas: Please delete here `add_character_to_udp_message(message);` A `\n` is not necessary here for the udp payload

modules/EvseV2G/connection.cpp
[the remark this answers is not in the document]
SebaLukas: We should use the same interface for the UDP message as for the ISO communication. Otherwise it can happen that a different interface is selected by the OS for sending this debug udp message. And when debugging the ISO messages via a pcap file, we will not be able to decrypt the TLS messages.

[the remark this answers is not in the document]
barsnick: Exactly, it must be on the same interface - also to not accidentally propagate. The secret is only supposed to be visible on this host, possibly the communication partner, and of course the captured trace.

I think a solution might be to also use `bind()`. When using it for the UDP source port, it can also bind to the source address, to the address of the TLS connection.

modules/EvseV2G/connection.cpp
```diff
@@ -985,3 +1021,63 @@ int connection_start_servers(struct v2g_context* ctx) {
 
     return 0;
 }
+
+int create_udp_socket(const u_int16_t udp_port, const char* interface_name) {
```
SebaLukas, line 1025: `uint_16t` instead of `u_int16_t`

modules/EvseV2G/connection.hpp
```diff
@@ -12,6 +12,7 @@
 
 int connection_init(struct v2g_context* ctx);
 int connection_start_servers(struct v2g_context* ctx);
+int create_udp_socket(const u_int16_t udp_port, const char* interface_name);
```
SebaLukas, line 15: See above `uint_16t`

modules/EvseV2G/connection.cpp
```diff
@@ -985,3 +1021,63 @@ int connection_start_servers(struct v2g_context* ctx) {
...
+    }
+
+    // source setup
+
+    // find port between 49152-65535
+    auto could_bind = false;
+    auto source_port = 49152;
+    for (; source_port < 65535; source_port++) {
+        sockaddr_in6 source_address = {AF_INET6, htons(source_port)};
+        if (bind(udp_socket, reinterpret_cast<sockaddr*>(&source_address), sizeof(sockaddr_in6)) == 0) {
```
SebaLukas, line 1041: `static_cast` instead of `reinterpret_cast`

modules/EvseV2G/connection.cpp
```diff
@@ -985,3 +1021,63 @@ int connection_start_servers(struct v2g_context* ctx) {
...
+        EVLOG_error << "Could not set interface name: " << interface_name << "with error: " << strerror(errno);
+    }
+
+    // destination setup
+    sockaddr_in6 destination_address = {AF_INET6, htons(udp_port)};
+    if (inet_pton(AF_INET6, LINK_LOCAL_MULTICAST, &destination_address.sin6_addr) <= 0) {
+        EVLOG_error << "Failed to setup server address" << strerror(errno);
+    }
+    const auto connected =
+        connect(udp_socket, reinterpret_cast<sockaddr*>(&destination_address), sizeof(sockaddr_in6)) == 0;
```
SebaLukas, line 1076: `static_cast` instead of `reinterpret_cast`

SebaLukas (approved):
Some small changes, but overall good job :+1:

modules/EvseV2G/connection.cpp
SebaLukas: reinterpret_cast is correct

modules/EvseV2G/connection.cpp
SebaLukas: reinterpret_cast is correct

PR #691, opened 2024-05-16
System: improve reset handling

mhei:
## Describe your changes

Updated description after addressing the review comments:

~~This PR bundles two smaller improvements:~~
- a new configuration key is introduced which allows to delay the actual reset execution a little bit (see commit message and added description for details)
- ~~the behavior of "hard" resets is now changed to perform a system reboot instead of just a restart of the EVerest processes~~

### DIFF
--- a/modules/System/System.hpp
+++ b/modules/System/System.hpp
@@ -22,6 +22,7 @@ namespace module {
 struct Conf {
     double DefaultRetries;
     double DefaultRetryInterval;
+    int ResetDelay;
 };
 
 class System : public Everest::ModuleBase {
--- a/modules/System/main/systemImpl.cpp
+++ b/modules/System/main/systemImpl.cpp
@@ -3,8 +3,10 @@
 
 #include "systemImpl.hpp"
 
+#include <chrono>
 #include <cstdlib>
 #include <fstream>
+#include <thread>
 #include <vector>
 
 #include <unistd.h>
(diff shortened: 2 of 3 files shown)


Pietfried:
Hi @mhei , thanks for adding the reset delay! 

I am a little concerned with adding the system reboot here since this module is used in a lot of SIL configs and it could cause your development machine to reboot which is probably undesired in most cases.

In most target platforms that run in production the `system` interface is implemented by a dedicated module because many of the commands of the interface needs special handling that depends on the target (e.g. `update_firmware`). 

I admit that this is not stated clearly in the documentation of this module, but we consider this module to be more of an example / SIL implementation rather than to be used in production. Therefore I'd propose to remove the system reboot instead keep killing the process.

barsnick:
> Therefore I'd propose to remove the system reboot instead keep killing the process.

How about we make this configurable? Like `hard_reset_is_soft_reset` 😉 or something like this, defaulting to soft (i.e. kill). Then we would have a good example.

We use this upstream `System` module as a template, and apply some system-specific patches for our platform.

mhei:
Thanks for clarification that you consider this module more as an example/HIL module. We discussed this internally and came to the conclusion that it would be better in our eyes that we do a fork for our purposes, which would then be expanded further to fit for out platforms. So in our eyes, this PR would be obsolete and I'd close it - unless you want e.g. the "delayed reset" feature. In this case, I'd drop the second commit only. (I'd keep the complexity of this example low and would not make the reset behavior configurable.)

Pietfried:
> Thanks for clarification that you consider this module more as an example/HIL module. We discussed this internally and came to the conclusion that it would be better in our eyes that we do a fork for our purposes, which would then be expanded further to fit for out platforms. So in our eyes, this PR would be obsolete and I'd close it - unless you want e.g. the "delayed reset" feature. In this case, I'd drop the second commit only. (I'd keep the complexity of this example low and would not make the reset behavior configurable.)

Ok, sounds good. Please keep the delay :+1:

Pietfried (approved):

PR #694, opened 2024-05-21
EvseV2G: Adding EVSE_Emergency_Shutdown handling

SebaLukas:
## Describe your changes
- Adding EVSE_Emergency_Shutdown to ISO2 StatusCode if a emergency shutdown is happening
- Adding additional umwc iso stop handlers

### DIFF
--- a/modules/EvseV2G/iso_server.cpp
+++ b/modules/EvseV2G/iso_server.cpp
@@ -538,6 +538,13 @@ static size_t getEmaidFromContractCert(const mbedtls_x509_name* ASubject, char* 
     return certEmaidLen;
 }
 
+static auto get_emergency_status_code(const struct v2g_context* ctx, uint8_t phase_type) {
+    if (ctx->intl_emergency_shutdown)
+        return iso1DC_EVSEStatusCodeType_EVSE_EmergencyShutdown;
+    else
+        return static_cast<iso1DC_EVSEStatusCodeType>(ctx->evse_v2g_data.evse_status_code[phase_type]);
+}
+
 //=============================================
 //             Publishing request msg
 //=============================================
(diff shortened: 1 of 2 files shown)


modules/EvseV2G/iso_server.cpp
```diff
@@ -2000,7 +2004,9 @@ static enum v2g_event handle_iso_cable_check(struct v2g_connection* conn) {
     res->DC_EVSEStatus.EVSEIsolationStatus_isUsed = conn->ctx->evse_v2g_data.evse_isolation_status_is_used;
     res->DC_EVSEStatus.EVSENotification = (iso1EVSENotificationType)conn->ctx->evse_v2g_data.evse_notification;
     res->DC_EVSEStatus.EVSEStatusCode =
-        (iso1DC_EVSEStatusCodeType)conn->ctx->evse_v2g_data.evse_status_code[PHASE_ISOLATION];
+        (true == conn->ctx->intl_emergency_shutdown)
+            ? iso1DC_EVSEStatusCodeType_EVSE_EmergencyShutdown
+            : (iso1DC_EVSEStatusCodeType)conn->ctx->evse_v2g_data.evse_status_code[PHASE_ISOLATION];
```
a-w50, line 2009: Two remarks here: 
- the syntax `(true == ...)` has the advantage that you can't use `=` instead of `==` by accident, but it might not read that natural.  In general, it might be better to have a constant ref to the ctx or similar, which would also prohibit the use of an accidential `=`.
- furthermore there is quite some code duplication, if you would have a helper function like
```
auto get_emergency_status_code(const CtxType& ctx, PhaseEnum phase_type) {
  if (ctx->intl_emergency_shutdown)
    return iso1DC_EVSEStatusCodeType_EVSE_EmergencyShutdown;
  else
    return static_cast<iso1DC_EVSEStatusCodeType>(ctx.evse_v2g_data.evse_status_code[phase_type]);
}
```
you would solve both problems in one go

SebaLukas: Good point, I think i copied only some code from the iso_server.cpp and it worked. I will add the helper function.

SebaLukas: Done

PR #698, opened 2024-05-24
Auth: Notify Evse also when authorization is rejected

Pietfried:
## Describe your changes
Extended Auth module so that it notifies the Evse also in case a token was rejected. This is especially required to allow the ISO15118 state machine to escape the authorize loop when a Plug&Charge authorization was rejected.


SebaLukas (approved):
Looks good!

corneliusclaussen (approved):

PR #700, opened 2024-05-27
Removed enabling of EVSE on startup

Pietfried:
## Describe your changes
Removed automatic enabling of evse on startup.

This change is required because otherwise a previously communicated Inoperative state could be overriden. The charger is enabled by default, so there is no reason the call enable on startup.

### DIFF
--- a/modules/EvseManager/EvseManager.cpp
+++ b/modules/EvseManager/EvseManager.cpp
@@ -888,7 +888,8 @@ void EvseManager::ready_to_start_charging() {
     timepoint_ready_for_charging = std::chrono::steady_clock::now();
     charger->run();
 
-    // this will publish a session event Enabled or Disabled that allows other modules the retrieve this state on startup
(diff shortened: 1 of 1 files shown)


corneliusclaussen:
We need to check if it confuses the API module as it may get stuck in state::unknown for example

Pietfried:
> We need to check if it confuses the API module as it may get stuck in state::unknown for example

Instead of removing the line we could publish either

`signal_simple_event(types::evse_manager::SessionEventEnum::Enabled);`

or 

`signal_simple_event(types::evse_manager::SessionEventEnum::Disabled);`

depending on the state. OCPP / OCPP201 have already called `call_enable` or `call_disable` based on the persisted state.

corneliusclaussen:
That would be cleaner as some modules may require an initial sync of state

Pietfried:
> That would be cleaner as some modules may require an initial sync of state

Added it, ready for review

corneliusclaussen (approved):

PR #703, opened 2024-05-31
Add timeout to token validation status

ydankner:
## Describe your changes
The Auth module previously never published an indication for a validated token timing out due to no connector being connected within the timeout. I added this status (TokenTimedOut), and published it in the appropriate place in the logic.

### DIFF
--- a/modules/Auth/tests/auth_tests.cpp
+++ b/modules/Auth/tests/auth_tests.cpp
@@ -1217,7 +1217,7 @@ TEST_F(AuthTest, test_token_timed_out) {
     ProvidedIdToken provided_token = get_provided_token(VALID_TOKEN_1, connectors);
 
     EXPECT_CALL(mock_publish_token_validation_status_callback,
-            Call(Field(&ProvidedIdToken::id_token, provided_token.id_token), TokenValidationStatus::Processing));
+                Call(Field(&ProvidedIdToken::id_token, provided_token.id_token), TokenValidationStatus::Processing));
     EXPECT_CALL(mock_publish_token_validation_status_callback,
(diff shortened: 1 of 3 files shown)


modules/Auth/tests/auth_tests.cpp
```diff
@@ -1200,4 +1206,31 @@ TEST_F(AuthTest, test_master_pass_group_id) {
     ASSERT_FALSE(this->auth_receiver->get_authorization(1));
 }
 
+/// \brief Test TokenTimedOut is published when authorization was provided but transaction has not yet been started.
```
golovasteek, line 1209:
```suggestion
/// \brief Test TokenTimedOut is published when authorization was provided but transaction has never been started.
```
I think "yet" is misleading here.

types/authorization.yaml
```diff
@@ -32,6 +32,7 @@ types:
       - Processing
       - Accepted
       - Rejected
+      - TokenTimedOut
```
golovasteek, line 35: Maybe add a comment, that it happens, when the validation was successful, but no charging session was initiated?

golovasteek (approved):
Looks good to me generally.
I would add some clarification in the comments.

types/authorization.yaml
a-w50, line 35: `TimedOut` might be enough here as its type is already called `TokenValidationStatus` and hence one can already deduce that the timeout is related to the token.

ydankner: Done :)

ydankner: Renamed :)

dorezyuk (approved):

Pietfried (approved):

PR #706, opened 2024-06-05
Fix EvseV2G: Prevent integer overflow 

SebaLukas:
## Describe your changes
Preventing an integer overflow after reading the v2gtp message length

### DIFF
--- a/modules/EvseV2G/v2g_server.cpp
+++ b/modules/EvseV2G/v2g_server.cpp
@@ -25,6 +25,7 @@
 #include "v2g_server.hpp"
 
 #define MAX_RES_TIME 98
+#define UINT32_MAX   0xFFFFFFFF
 
 static types::iso15118_charger::V2G_Message_ID get_V2G_Message_ID(enum V2gMsgTypeId v2g_msg,
                                                                   enum v2g_protocol selected_protocol, bool is_req) {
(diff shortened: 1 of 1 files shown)


modules/EvseV2G/v2g_server.cpp
```diff
@@ -162,6 +163,12 @@ static int v2g_incoming_v2gtp(struct v2g_connection* conn) {
         return -1;
     }
 
+    if (conn->payload_len >= UINT32_MAX - V2GTP_HEADER_LENGTH) {
+        dlog(DLOG_LEVEL_ERROR, "Prevent integer overflow - payload too long: have %d, would need %d",
+             DEFAULT_BUFFER_SIZE, conn->payload_len);
```
lategoodbye, line 168: Shouldn't the second parameter be "%u" because payload_len is of type uint32_t?

lategoodbye (changes requested):

modules/EvseV2G/v2g_server.cpp
SebaLukas: Good point. Then I have to change the next dlog() in line 173, too.

SebaLukas: Done

corneliusclaussen (approved):

lategoodbye (approved):

PR #708, opened 2024-06-05
Fixing empty string in EvseSecurity config

Pietfried:
## Describe your changes
Checking if private key password is empty and passing std::nullopt to EvseSecurity ctor in this case.

This change is required, because even if the config parameter `private_key_password` is not configured, the constructor will pass an empty string instead of std::nullopt. This will cause private keys to be encrypted every time.


AssemblyJohn (approved):

james-ctc:
Note there are other places in the codebase where nullopt is converted to an empty string for passwords:
e.g.
    EVP_PKEY_ptr evp_pkey(PEM_read_bio_PrivateKey(bio.get(), nullptr, nullptr, (void*)password.value_or("").c_str()));

Care is needed to ensure that existing keys that have been created with "" can still be used.

PR #710, opened 2024-06-06
Adapt error handling in JsYetiSimulator to recent error framework changes

hikinggrass:

### DIFF
--- a/modules/examples/error-framework/JsExampleErrorRaiser/index.js
+++ b/modules/examples/error-framework/JsExampleErrorRaiser/index.js
@@ -138,7 +138,7 @@ boot_module(async ({
   mod.provides.example_raiser.raise_error(error);
   check_conditions(mod);
   await sleep(1000);
-  mod.provides.example_raiser.clear_error(error.type, true);
+  mod.provides.example_raiser.clear_error(error.type, error.sub_type);
   check_conditions(mod);
   await sleep(1000);
   error = mod.provides.example_raiser.error_factory.create_error(
@@ -149,7 +149,7 @@ boot_module(async ({
   mod.provides.example_raiser.raise_error(error);
   check_conditions(mod);
   await sleep(1000);
-  mod.provides.example_raiser.clear_error(error.type, true);
+  mod.provides.example_raiser.clear_error(error.type, error.sub_type);
   check_conditions(mod);
   await sleep(1000);
   error = mod.provides.example_raiser.error_factory.create_error(
--- a/modules/examples/error-framework/JsExampleErrorRaiser/index.js
+++ b/modules/examples/error-framework/JsExampleErrorRaiser/index.js
@@ -138,7 +138,7 @@ boot_module(async ({
   mod.provides.example_raiser.raise_error(error);
   check_conditions(mod);
   await sleep(1000);
-  mod.provides.example_raiser.clear_error(error.type, error.sub_type);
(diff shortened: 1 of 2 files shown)


SebaLukas (approved):
Looks good :+1:

modules/examples/error-framework/JsExampleErrorRaiser/index.js
```diff
@@ -138,7 +138,7 @@ boot_module(async ({
   mod.provides.example_raiser.raise_error(error);
   check_conditions(mod);
   await sleep(1000);
-  mod.provides.example_raiser.clear_error(error.type, error.sub_type);
+  mod.provides.example_raiser.clear_error(error.type, true);
```
andistorm, line 141: FYI: This is not a bug

this will clear at maximum one error that matches type and sub_type:
```
mod.provides.example_raiser.clear_error(error.type, error.sub_type);
```

this will clear multiple errors (in this case 1) that match type
```
mod.provides.example_raiser.clear_error(error.type, true);
```

I would suggest to show both syntax in example module:
```suggestion
  // Clear exactly one error that matches type and sub_type of raised error object
  mod.provides.example_raiser.clear_error(error.type, error.sub_type);
```

modules/examples/error-framework/JsExampleErrorRaiser/index.js
```diff
@@ -149,7 +149,7 @@ boot_module(async ({
...
```
andistorm, line 152: See above:
```suggestion
  // Clear all errors that match type of raised error object (In this case: 1)
  mod.provides.example_raiser.clear_error(error.type, true);
```

modules/simulation/JsYetiSimulator/index.js
```diff
@@ -268,330 +268,528 @@ function dutyCycleToAmps(dc) {
...
+      'evse_board_support/BrownOut',
+      '',
+      'Simulated fault event',
+      'High'
+    );
+    mod.provides.board_support.raise_error(error);
     active_errors.BrownOut = true;
   } else if (!raise && active_errors.BrownOut) {
-    mod.provides.board_support.request_clear_all_of_type.evse_board_support_BrownOut();
+    mod.provides.board_support.clear_error('evse_board_support/BrownOut', true);
```
andistorm, line 280: Unfortunately there is no create_error function without sub_type parameter, but to clear an error it would also possible to do so:
```suggestion
    mod.provides.board_support.clear_error('evse_board_support/BrownOut');
```
This is same as
```
mod.provides.board_support.clear_error('evse_board_support/BrownOut', false);
````
and will clear at maximum one error.

See also https://everest.github.io/nightly/general/01_framework/error_framework.html#id13

It is your decision what to use here, but I would recommend to use my suggestion in case of not using sub_types here


... same for the rest

andistorm (changes requested):

modules/examples/error-framework/JsExampleErrorRaiser/index.js
hikinggrass: I must have overlooked that when looking at the everestjs code, but I've reverted the commit

modules/simulation/JsYetiSimulator/index.js
hikinggrass: Changed this

andistorm (approved):
LGTM

PR #711, opened 2024-06-07
Rename UINT32_MAX as it collides with definition in /usr/include/stdi…

corneliusclaussen:
…nt.h

UINT32_MAX is already defined in a system header

### DIFF
--- a/modules/EvseV2G/v2g_server.cpp
+++ b/modules/EvseV2G/v2g_server.cpp
@@ -25,7 +25,6 @@
 #include "v2g_server.hpp"
 
 #define MAX_RES_TIME 98
-#define UINT32_MAX   0xFFFFFFFF
 
 static types::iso15118_charger::V2G_Message_ID get_V2G_Message_ID(enum V2gMsgTypeId v2g_msg,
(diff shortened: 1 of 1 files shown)


modules/EvseV2G/v2g_server.cpp
```diff
@@ -163,7 +163,7 @@ static int v2g_incoming_v2gtp(struct v2g_connection* conn) {
         return -1;
     }
 
-    if (conn->payload_len >= UINT32_MAX - V2GTP_HEADER_LENGTH) {
+    if (conn->payload_len >= _UINT32_MAX - V2GTP_HEADER_LENGTH) {
```
SebaLukas, line 166: Good catch! We could also use UINT32_MAX from stdint instead of adding a extra define

corneliusclaussen: true, changed to the system one

SebaLukas (approved):
:+1:

hikinggrass (approved):
PR should probably be renamed before merging :D

a-w50:
Closed in favour of #715

PR #715, opened 2024-06-10
refactor: correct include order

a-w50:
- delete unnecessary redefinition of UINT32_MAX
- group and sort include files


a-w50:
Just saw, that #711 tries to fix the same issue.  I would still propose, that we should include `cstdint`, as `UINT32_MAX` is defined in there.  Furthermore, I rearranged/regrouped the `include`s

SebaLukas (approved):
👍🏻

PR #728, opened 2024-06-17
Enable compile warnings

barsnick:
## Describe your changes

Looking at `make VERBOSE=1`, I noticed that no compiler warnings were enabled at all. I'm not sure this was intentional. It is certainly not recommended, as warnings can uncover bugs, and also wrong intentions.

This change adds the "simplest" set of warnings with `-Wall`, but silences warnings for unused functions, as the generated interface implementations are full of them, as they serve as libraries.

No attempt was made yet at `-Wextra` or at defining a more useful set of warnings.

This CMake change also affects the building of dependencies, and thereby also uncovers warnings in internal dependencies.

The change set further tries to fix some obvious stuff, to make an initial effort of reducing the number of warnings. This is to be considered an example, as more such obvious stuff remains.

This was tested with g++11, and still manages to build the complete project from scratch, albeit with much more noise.


corneliusclaussen:
I like this one, but I should we move it after the June release and get rid of some more of the warnings?

corneliusclaussen:
Some of the warnings should even be treated as errors I think, e.g. the warning that an enum is not handled in a switch case. This caused quite some trouble before.

hikinggrass:
> I like this one, but I should we move it after the June release and get rid of some more of the warnings?

Sounds like a plan, we had a discussion about it in last week's ci working group meeting and I'm in the process of modifying this PR slightly so it only applies to the targets defined in everest-core. Should be done soon

barsnick:
I missed that you pushed your proposed change last week. I will rebase the branch and test soon.

hikinggrass:
> @hikinggrass, thanks for the proposed change. I missed that you had pushed it last week. I will rebase the branch and test soon.

I might add a "global compiler options" flag that's off by default as well, so we can have the same result as earlier

barsnick:
Work for me so far.

I added more fixes. We may want to separate them from this PR?

The libraries also need cmake options to enable compile warnings. These are not covered anymore with the recent change.

> Some of the warnings should even be treated as errors I think, e.g. the warning that an enum is not handled in a switch case. This caused quite some trouble before.

This obviously needs some work, since some enums are currently missed (possibly on purpose). Each code owner may need to check. (How about an optional (non-fatal) workflow pipeline which enables compile warnings (later warnings-as-errors?), so that they get flagged.)

hikinggrass (approved):

modules/Auth/lib/CMakeLists.txt
```diff
@@ -29,3 +29,8 @@ PRIVATE
 # needs c++ 14
 target_compile_features(auth_handler PRIVATE cxx_std_14)
 
+if(EVEREST_ENABLE_COMPILE_WARNINGS)
+    target_compile_options(auth_handler
+        PRIVATE ${EVEREST_COMPILE_OPTIONS}
+    )
+endif()
```
SebaLukas, line 36: Missing last empty line

SebaLukas (approved):
:+1:

Issue #523, opened 2024-02-01
Log is flooded by serialcommhub_x in case Modbus slave is not available


lategoodbye:
Hi,
we have tested the lastest everest-core 2024.01 and accidentially configured a meter on our Tarragon platform, which is not available via Modbus. After the start of EVerest the whole log is flooded (multiple times per second) by error messages from serialcommhub_x, which cause eMMC wearout:

````
Feb 01 13:53:51 tarragon manager[539]: 2024-02-01 13:53:51.074076 [ERRO] serialcommhub_x std::vector<short unsigned int> tiny_modbus::decode_reply(const uint8_t*, int, uint8_t, tiny_modbus::FunctionCode) :: Packet too small: 3 bytes.
Feb 01 13:53:51 tarragon manager[539]: 2024-02-01 13:53:51.293540 [ERRO] serialcommhub_x std::vector<short unsigned int> tiny_modbus::decode_reply(const uint8_t*, int, uint8_t, tiny_modbus::FunctionCode) :: Packet too small: 3 bytes.
Feb 01 13:53:51 tarragon manager[539]: 2024-02-01 13:53:51.423442 [ERRO] serialcommhub_x std::vector<short unsigned int> tiny_modbus::decode_reply(const uint8_t*, int, uint8_t, tiny_modbus::FunctionCode) :: Packet too small: 3 bytes.
Feb 01 13:53:51 tarragon manager[539]: 2024-02-01 13:53:51.553456 [ERRO] serialcommhub_x std::vector<short unsigned int> tiny_modbus::decode_reply(const uint8_t*, int, uint8_t, tiny_modbus::FunctionCode) :: Packet too small: 3 bytes.
Feb 01 13:53:52 tarragon manager[539]: 2024-02-01 13:53:52.733426 [ERRO] serialcommhub_x std::vector<short unsigned int> tiny_modbus::decode_reply(const uint8_t*, int, uint8_t, tiny_modbus::FunctionCode) :: Packet too small: 3 bytes.
Feb 01 13:53:52 tarragon manager[539]: 2024-02-01 13:53:52.863485 [ERRO] serialcommhub_x std::vector<short unsigned int> tiny_modbus::decode_reply(const uint8_t*, int, uint8_t, tiny_modbus::FunctionCode) :: Packet too small: 3 bytes.
[...]

PR #731, opened 2024-06-18
SerialCommHub: several improvements

mhei:
## Describe your changes

This PR accumulates several minor fixes and some improvements:
- the currently internal retry handling is made public and configurable
- the internal error handling is refactored so that logging can be improved and managed centrally
- some additional minor stuff

This PR is considered a fix for #523, in combination with #730.

- [x] chargebyte internal review approved this PR

### DIFF
--- a/modules/SerialCommHub/main/serial_communication_hubImpl.cpp
+++ b/modules/SerialCommHub/main/serial_communication_hubImpl.cpp
@@ -60,11 +60,12 @@ serial_communication_hubImpl::handle_modbus_read_holding_registers(int& target_d
     {
         std::scoped_lock lock(serial_mutex);
 
-        auto retry_counter = this->num_resends_on_error;
+        auto retry_counter = config.retries + 1;
         while (retry_counter > 0) {
+            auto current_trial = config.retries + 1 - retry_counter + 1;
 
-            EVLOG_debug << fmt::format("Try {} Call modbus_client->read_holding_register(id {} addr {} len {})",
-                                       (int)retry_counter, (uint8_t)target_device_id, (uint16_t)first_register_address,
+            EVLOG_debug << fmt::format("Trial {}/{}: call modbus_client->read_holding_register(id {} addr {} len {})",
+                                       current_trial , config.retries + 1, (uint8_t)target_device_id, (uint16_t)first_register_address,
                                        (uint16_t)num_registers_to_read);
 
             response = modbus.txrx(target_device_id, tiny_modbus::FunctionCode::READ_MULTIPLE_HOLDING_REGISTERS,
@@ -96,11 +97,12 @@ serial_communication_hubImpl::handle_modbus_read_input_registers(int& target_dev
     {
         std::scoped_lock lock(serial_mutex);
 
-        uint8_t retry_counter{this->num_resends_on_error};
-        while (retry_counter-- > 0) {
+        auto retry_counter = config.retries + 1;
+        while (retry_counter > 0) {
+            auto current_trial = config.retries + 1 - retry_counter + 1;
 
-            EVLOG_debug << fmt::format("Try {} Call modbus_client->read_input_register(id {} addr {} len {})",
-                                       (int)retry_counter, (uint8_t)target_device_id, (uint16_t)first_register_address,
+            EVLOG_debug << fmt::format("Trial {}/{}: call modbus_client->read_input_register(id {} addr {} len {})",
+                                       current_trial , config.retries + 1, (uint8_t)target_device_id, (uint16_t)first_register_address,
                                        (uint16_t)num_registers_to_read);
 
             response = modbus.txrx(target_device_id, tiny_modbus::FunctionCode::READ_INPUT_REGISTERS,
@@ -108,6 +110,7 @@ serial_communication_hubImpl::handle_modbus_read_input_registers(int& target_dev
             if (response.size() > 0) {
                 break;
             }
+            retry_counter--;
         }
     }
 
@@ -134,11 +137,12 @@ types::serial_comm_hub_requests::StatusCodeEnum serial_communication_hubImpl::ha
     {
         std::scoped_lock lock(serial_mutex);
 
-        uint8_t retry_counter{this->num_resends_on_error};
-        while (retry_counter-- > 0) {
+        auto retry_counter = config.retries + 1;
+        while (retry_counter > 0) {
+            auto current_trial = config.retries + 1 - retry_counter + 1;
 
-            EVLOG_debug << fmt::format("Try {} Call modbus_client->write_multiple_registers(id {} addr {} len {})",
-                                       (int)retry_counter, (uint8_t)target_device_id, (uint16_t)first_register_address,
+            EVLOG_debug << fmt::format("Trial {}/{}: call modbus_client->write_multiple_registers(id {} addr {} len {})",
+                                       current_trial , config.retries + 1, (uint8_t)target_device_id, (uint16_t)first_register_address,
                                        (uint16_t)data.size());
 
             response = modbus.txrx(target_device_id, tiny_modbus::FunctionCode::WRITE_MULTIPLE_HOLDING_REGISTERS,
@@ -146,6 +150,7 @@ types::serial_comm_hub_requests::StatusCodeEnum serial_communication_hubImpl::ha
             if (response.size() > 0) {
                 break;
             }
+            retry_counter--;
         }
     }
 
@@ -167,11 +172,12 @@ serial_communication_hubImpl::handle_modbus_write_single_register(int& target_de
     {
         std::scoped_lock lock(serial_mutex);
 
-        uint8_t retry_counter{this->num_resends_on_error};
-        while (retry_counter-- > 0) {
+        auto retry_counter = config.retries + 1;
+        while (retry_counter > 0) {
+            auto current_trial = config.retries + 1 - retry_counter + 1;
 
-            EVLOG_debug << fmt::format("Try {} Call modbus_client->write_single_register(id {} addr {} data 0x{:04x})",
-                                       (int)retry_counter, (uint8_t)target_device_id, (uint16_t)register_address,
+            EVLOG_debug << fmt::format("Trial {}/{}: call modbus_client->write_single_register(id {} addr {} data 0x{:04x})",
+                                       current_trial , config.retries + 1, (uint8_t)target_device_id, (uint16_t)register_address,
                                        (uint16_t)data);
 
             response = modbus.txrx(target_device_id, tiny_modbus::FunctionCode::WRITE_SINGLE_HOLDING_REGISTER,
@@ -179,6 +185,7 @@ serial_communication_hubImpl::handle_modbus_write_single_register(int& target_de
             if (response.size() > 0) {
                 break;
             }
+            retry_counter--;
         }
     }
     EVLOG_debug << fmt::format("Done writing (size {})", response.size());
--- a/modules/SerialCommHub/main/serial_communication_hubImpl.cpp
+++ b/modules/SerialCommHub/main/serial_communication_hubImpl.cpp
@@ -40,6 +40,8 @@ void serial_communication_hubImpl::init() {
     rxtx_gpio_settings.line_number = config.rxtx_gpio_line;
     rxtx_gpio_settings.inverted = config.rxtx_gpio_tx_high;
 
+    system_error_logged = false;
+
     if (!modbus.open_device(config.serial_port, config.baudrate, config.ignore_echo, rxtx_gpio_settings,
                             static_cast<tiny_modbus::Parity>(config.parity), config.rtscts,
                             milliseconds(config.initial_timeout_ms), milliseconds(config.within_message_timeout_ms))) {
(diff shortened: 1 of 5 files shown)


barsnick:
@hikinggrass, @corneliusclaussen, @Pietfried
Ping for review (pre-release).

modules/SerialCommHub/tiny_modbus_rtu.cpp
```diff
@@ -7,28 +7,63 @@
...
+    case FunctionCode::WRITE_MULTIPLE_HOLDING_REGISTERS:
+        return "WRITE_MULTIPLE_HOLDING_REGISTERS";
+    default:
+        return "unknown";
+    }
+}
 
-namespace tiny_modbus {
+std::string FunctionCode_to_string_with_hex(FunctionCode fc) {
+    std::stringstream ss;
```
corneliusclaussen, line 57: Could probably be written more readable with fmt::format, but just my taste ;-)

modules/SerialCommHub/main/serial_communication_hubImpl.cpp
```diff
@@ -48,37 +50,44 @@ void serial_communication_hubImpl::init() {
...
+        auto current_trial = config.retries + 1 - retry_counter + 1;
+
+        EVLOG_debug << fmt::format("Trial {}/{}: calling {}(id {} addr {}(0x{:04x}) len {})", current_trial,
+                                   config.retries + 1, tiny_modbus::FunctionCode_to_string_with_hex(function),
+                                   device_address, first_register_address, first_register_address, register_quantity);
+
+        try {
+            response = modbus.txrx(device_address, function, first_register_address, register_quantity,
+                                   config.max_packet_size, wait_for_reply, request);
+        } catch (tiny_modbus::TinyModbusException& e) {
```
corneliusclaussen, line 72: should be const

modules/SerialCommHub/tiny_modbus_rtu.cpp
```diff
@@ -103,92 +134,112 @@ static std::vector<uint16_t> decode_reply(const uint8_t* buf, int len, uint8_t e
...
+        [[fallthrough]];
+    case FunctionCode::READ_COILS:
+    case FunctionCode::READ_DISCRETE_INPUTS:
+        // yes
+        // adapt byte count and starting pos
+        byte_cnt = buf[RES_RX_LEN_POS];
+        start_of_result = RES_RX_START_OF_PAYLOAD;
+        break;
+    default:
+        throw std::logic_error("Missing implementation for function code " + FunctionCode_to_string_with_hex(function));
```
corneliusclaussen, line 214: This exception will not be caught as far as I can see. We are trying to get rid of all uncaught exception throughout the code base (as well as all exit() from modules in case of error). Even if it is an error in the code if the implementation is missing here, it should not trigger a complete restart of the system including ocpp disconnection. We should just indicate that the modbus transfer failed, as we do with other transport errors and not exit the process.

corneliusclaussen (changes requested):
See uncaught exception, the rest is just my taste.

modules/SerialCommHub/main/serial_communication_hubImpl.cpp
barsnick: Will fix.

modules/SerialCommHub/tiny_modbus_rtu.cpp
barsnick: Will fix.

So we should also avoid the other 
```throw std::system_error(errno, std::generic_category(), "Could not send Modbus request");```
when `write()` fails, right?

**Comment by barsnick** [comment]
I have added the fixes. I used four separate commits, for easier review, and so that they can be squashed with their respective commits (and the whole bunch will be squashed anyway later).

**Comment by mhei** [comment]
Regarding the uncaught exception: usually this should never trigger, since the if-constructs a few lines above already cover it and in case the exceptions is raised, then the code review process failed. That said, adding `UnimplementedFunction` as child of `TinyModbusException` is the wrong way to go in my eyes because it is a complete different error domain and has nothing in common with the other TinyModbusException exceptions. I see two options: either just drop raising an exception here, or catch this specific exception (`std::logic_error`), too - in this case, spamming the log with messages would be fine with me 😇.

Regarding failed write: since this indicates major system trouble, it should be caught on the highest level possible and in turn force a shutdown (as clean as possible in this case) and reboot of the whole system. But according to my understanding, this still requires more infrastructure around and parts are already in progress.

Just my 2 cts, back in holiday mode again 🌞.

**Comment by corneliusclaussen** [comment]
If it never triggers, we don't need to throw it ;-) The problem that we need to change our minds a bit how to handle errors in modules in general (not just here). If a module cannot perform its work anymore (e.g. because the modbus device disappeared as someone unplugged the USB dongle), it should go into an error mode but it should never kill the process. I think the clean solution here would be what we start using in other modules: Define an error (e.g. CommunicationFault to align with other modules) to indicate to the modules that are using modbus that the connection is not available. All calls will then return with a failure.
Other modules using the modbus will also need to go into a fault mode etc.
This way, we have proper error reporting through ErrorHistory Module, OCPP etc and we do not kill everything.

There are many things that are still functional even if one device cannot be reached on the modbus anymore: E.g. the charger may be configured to still charge for free etc and cloud communication still works and reports that error properly instead of getting stuck in a reboot loop.

So I disagree that this always needs to result in a system reboot.

File: modules/SerialCommHub/tiny_modbus_rtu.cpp (hunk 3 above)
Comment (corneliusclaussen) [reply modules/SerialCommHub/tiny_modbus_rtu.cpp]: yes, see my comment further down

**Comment by corneliusclaussen** [comment]
A better way of handling full system reboots could for example be a separate module that observes the errors of all modules and if they do not self recover (which all modules should try as best as they can) e.g. within 24h or so it reboots. This way it could also do this when e.g. no charging session is active and the behaviour could be configured (e.g. to select which errors/modules are uncritical and should be ignored in the actual setup

File: modules/SerialCommHub/tiny_modbus_rtu.cpp [hunk 4]
```diff
@@ -103,92 +131,113 @@ static std::vector<uint16_t> decode_reply(const uint8_t* buf, int len, uint8_t e
...
+        throw OddByteCountException("For " + FunctionCode_to_string_with_hex(function) +
+                                    " an even byte count is expected in the response.");
+    }
+
+    // ready to copy actual result data to output, so pre-allocate enough memory for the output
+    result.reserve((byte_cnt + 1) / 2);
+
+    for (int i = start_of_result; i < start_of_result + byte_cnt; i += 2) {
+        uint16_t t;
+        memcpy(&t, buf + i, 2);
```
dorezyuk, line 231: wont this read invalid memory in byte_cnt is odd?

mhei: Some lines above, it is ensured that `byte_cnt` is even.

barsnick:
> I think the clean solution here would be what we start using in other modules: Define an error (e.g. CommunicationFault to align with other modules) to indicate to the modules that are using modbus that the connection is not available. All calls will then return with a failure.

This requires somewhat of an infrastructure though, including interfaces like in your PR https://github.com/EVerest/everest-core/pull/724 (why was that closed), and handling in EvseManager, right? I guess this is something which goes beyond this PR right now, even though that is some of the thought behind it.

corneliusclaussen:
> > I think the clean solution here would be what we start using in other modules: Define an error (e.g. CommunicationFault to align with other modules) to indicate to the modules that are using modbus that the connection is not available. All calls will then return with a failure.
> 
> This requires somewhat of an infrastructure though, including interfaces like in your PR #724 (why was that closed), and handling in EvseManager, right? I guess this is something which goes beyond this PR right now, even though that is some of the thought behind it.

Yes, this is not for this PR

modules/SerialCommHub/tiny_modbus_rtu.cpp
```diff
@@ -103,92 +131,113 @@ static std::vector<uint16_t> decode_reply(const uint8_t* buf, int len, uint8_t e
...
+        throw UnimplementedFunction("Missing implementation for function code " +
+                                    FunctionCode_to_string_with_hex(function));
+    }
+
+    // check if result is completely in received data
+    if (start_of_result + byte_cnt > len) {
+        throw IncompletePacketException("Result data not completely in received message.");
+    }
+
+    // check if result is completely in received data
```
corneliusclaussen, line 220: This comment is a copy paste error

modules/SerialCommHub/tiny_modbus_rtu.cpp
corneliusclaussen: The last item in the buffer it writes to potentially is buf[start_of_result + byte_cnt] for odd byte count (and no expected even byte count).

As far as I can see the

if (start_of_result + byte_cnt > len)

does not catch it. So I also think it writes one byte past the buffer.

dorezyuk: you also don't need to copy element-wise -  just doing memcpy on the whole buffer or constructing the result directly from the buffer should work (given that you handle the odd case)

modules/SerialCommHub/tiny_modbus_rtu.cpp
```diff
@@ -103,92 +131,113 @@ static std::vector<uint16_t> decode_reply(const uint8_t* buf, int len, uint8_t e
...
+        t = be16toh(t);
+        result.push_back(t);
+    }
+
+    return result;
 }
 
 TinyModbusRTU::~TinyModbusRTU() {
-    if (fd)
+    if (fd != -1)
```
dorezyuk, line 240: there are bunch of cases in this file where we still check if `fd <= 0` . can you make them consistent?

barsnick:
I resolved most comments, squashed the fix-ups with their respective commits, and rebased to current main HEAD.

I will look at the byte count tomorrow.

modules/SerialCommHub/tiny_modbus_rtu.cpp
barsnick: > The last item in the buffer it writes to potentially is buf[start_of_result + byte_cnt] for odd byte count (and no expected even byte count).

I re-reviewed the code. (This is part of the original implementation, BTW, not of the changes in this PR.)
Indeed, when receiving an odd number of bytes, the `memcpy()` will read beyond the end of `buf`.

E.g. read coils, five bytes:
```C++
    for (int i = start_of_result; i < start_of_result + byte_cnt; i += 2) {
        uint16_t t = 0;
        memcpy(&t, buf + i, 2);
        t = be16toh(t);
        result.push_back(t);
    }
```
len = 8
start_of_result = 3
byte_cnt = 5
Then `i` will take the values 3, 5, 7. The last iteration will `memcopy()` two bytes from `buf[7]`, the second of which is beyond its end. This is either undefined behavior or illegal access (likely not due to alignment), and potentially returns data not belonging to the response.

I don't know if there is even a point of re-interpreting the return value of "read coils" and "read inputs" as big-endian (I understand these are bit fields). I am open for comments. I will just keep the be16 conversion in place, and make sure `memcpy()` places the single, final byte from `buf` into the correct position.

Is that okay for you?

barsnick: > you also don't need to copy element-wise - just doing memcpy on the whole buffer or constructing the result directly from the buffer should work (given that you handle the odd case)

You cannot just memcopy from the byte array to the word array, as the bytes need to (potentially) be converted from Modbus's big-endian order. (You could do some casting and union tricks for conversion, but for me, `memcpy()` is good to read.)

barsnick:
I believe all issues are resolved with my latest commits.
I fixed the reading beyond the end of the buffer, and rebased on top of current main HEAD.

corneliusclaussen (approved):
Looks good now!

dorezyuk (approved):

modules/SerialCommHub/tiny_modbus_rtu.cpp
hikinggrass: Is this still relevant? As far as I can see this wasn't addressed

modules/SerialCommHub/tiny_modbus_rtu.cpp
```diff
@@ -438,7 +487,17 @@ std::vector<uint16_t> TinyModbusRTU::txrx_impl(uint8_t device_address, FunctionC
...
         rxtx_gpio.set(false);
-        write(fd, req.data(), req.size());
+
+        uint8_t* buffer = req.data();
+        ssize_t written = 0;
+
+        while (written < req.size()) {
+            ssize_t c = write(fd, &buffer[written], req.size() - written);
+            if (c == -1)
+                throw std::system_error(errno, std::generic_category(), "Could not send Modbus request");
```
hikinggrass, line 497: That's the system_error that was mentioned earlier that should be avoided

barsnick: Yes, but now it's being caught. I that not okay?

hikinggrass: ah somehow I missed that :sweat_smile: all good

modules/SerialCommHub/tiny_modbus_rtu.cpp
hikinggrass: same here

PR #732, opened 2024-06-20
Make change to the libocpp interface for the device model set variabl…

maaikez:
## Describe your changes

Change interface so it is possible to set the source of the config value changes in the device model.

[libocpp issue 656](https://github.com/EVerest/libocpp/issues/656)


modules/OCPP/ocpp_generic/ocppImpl.cpp
```diff
@@ -185,7 +185,8 @@ ocppImpl::handle_get_variables(std::vector<types::ocpp::GetVariableRequest>& req
 }
 
 std::vector<types::ocpp::SetVariableResult>
-ocppImpl::handle_set_variables(std::vector<types::ocpp::SetVariableRequest>& requests) {
+ocppImpl::handle_set_variables(std::vector<types::ocpp::SetVariableRequest>& requests,
+                               std::string& source) { // TODO do something with source?
```
Pietfried, line 189: Nothing to do with `source` here. You can remove the todo

dependencies.yaml
```diff
@@ -60,7 +60,7 @@ libevse-security:
 # OCPP
 libocpp:
   git: https://github.com/EVerest/libocpp.git
-  git_tag: v0.13.0
+  git_tag: feature/656-device-model-initialization-in-cpp
```
Pietfried, line 63: We usually delete the branches, so the PR in libocpp will be merged first and then we need to update the reference to the commit hash here.

Pietfried (approved):
:+1: 
Please update the commit git_tag in the dependency yaml once the PR in libocpp has been merged.

dependencies.yaml
maaikez: Yes so this needs to change before merging indeed. But now you can just checkout this commit and it will build (which is nice to know for the buildserver)

PR #750, opened 2024-06-26
Added in patch which supports PNC

louisg1337:
## Describe your changes

Hello everyone, I am apart of the team over at NREL who has been working on the SIL demos in `everest/everest-demo`. Recently, @shankari went to the CharIN Testival and showcased EVerest by running a custom demo that was tailored to have PnC functioning, found [here](https://github.com/US-JOET/everest-demo/blob/charin-e2e-demo/demo-iso15118-2-ac-plus-ocpp.sh). To get PnC to work, she had to apply a handful of patches to different parts of EVerest, all which can be found [here](https://github.com/US-JOET/everest-demo/blob/879548aafdf9ff2366de45f1b77cd25d5667ebb2/demo-iso15118-2-ac-plus-ocpp.sh#L354). We now would like to slowly integrate these patches into EVerest so that it can fully support PnC.

This patch in particular involves adding in support for a `Contract` payment method to `PyEvJosev`.

**Testing Done**

As mentioned above, this patch is one of many, so it has only been tested in conjunction with the other patches in the SIL demo. The particular demo which has PnC working can be found linked above. To run this demo, I would recommend...

1. Clone repo and check out branch.
   ```
   git clone https://github.com/US-JOET/everest-demo.git
   git checkout -b test-demo origin/charin-e2e-demo  
   ```
2. Run the software simulation demo.
   ```
   bash demo-iso15118-2-ac-plus-ocpp.sh -r $(pwd) -b test-demo -3   
   ```
3. Wait until the EVerest software boots up and you see the following.
   ```
   2024-06-18 18:23:22.836788 [INFO] ocpp:OCPP201     :: Received BootNotificationResponse: {
       "currentTime": "2024-06-18T18:23:22.000Z",
       "interval": 300,
       "status": "Accepted"
   }
   ```
4. Navigate to http://localhost:1880/ui/, choose from the `Car Simulation` dropdown menu `AC ISO15118-2 Plug&Charge`, then plug the car in and observe it authenticate on its own.

The issue can be found [here](https://github.com/EVerest/everest-demo/issues/61) in `everest-demo`. 

### DIFF
--- a/modules/PyEvJosev/module.py
+++ b/modules/PyEvJosev/module.py
@@ -95,6 +95,7 @@ class PyEVJosevModule():
 
     def _handler_start_charging(self, args) -> bool:
 
+        self._es.PaymentOption =args['PaymentOption']
         self._es.EnergyTransferMode = args['EnergyTransferMode']
 
         self._ready_event.set()
--- a/modules/simulation/JsEvManager/index.js
+++ b/modules/simulation/JsEvManager/index.js
@@ -362,17 +362,27 @@ function registerAllCmds(mod) {
   });
 
   if (mod.uses_list.ev.length > 0) {
-    registerCmd(mod, 'iso_start_v2g_session', 1, (mod, c) => {
-      switch (c.args[0]) {
+    registerCmd(mod, 'iso_start_v2g_session', 2, (mod, c) => {
+      if (c.args[0] === 'externalpayment') mod.payment = 'ExternalPayment';
+        else if (c.args[0] === 'contract') mod.payment = 'Contract';
+      else {
+        evlog.debug('Found invalid payment method' + c.args[0]);
+	return false;
+      }
+
+      switch (c.args[1]) {
         case 'ac':
           if (mod.config.module.three_phases !== true) mod.energymode = 'AC_single_phase_core';
           else mod.energymode = 'AC_three_phase_core';
           break;
         case 'dc': mod.energymode = 'DC_extended'; break;
-        default: return false;
+        default:
+          evlog.debug('Found invalid payment method' + c.args[1]);
+	        return false;
       }
 
-      mod.uses_list.ev[0].call.start_charging({ EnergyTransferMode: mod.energymode });
+      args = { PaymentOption: mod.payment, EnergyTransferMode: mod.energymode };
+      mod.uses_list.ev[0].call.start_charging(args);
 
       return true;
     });


SebaLukas:
Hello :)

thank you very much for your contribution. However, after looking through it, I have decided not to accept this PR and close it in a few days.

Because, the changes this PR describes, I deleted from everest-core a few months ago. The payment option in the `iso_start_v2g_session` cmd had no effect on Josev at that time.
I have seen that there is another PR in the Josev fork and that the option selected here is actually also used by Josev.

However, I find this approach less realistic and less effective than the current situation.

At the moment, the Josev EV side itself decides whether to select PnC. Certain conditions must be met: TLS must be active and the EVSE must also offer PnC as a payment option. PnC only works in the SIL if these conditions are met. The config also makes it possible to test certain scenarios, such as TLS + EIM.

This PR can lead to strange behavior on the EVCC side if, for example, PnC is selected and the charger only supports EIM. This cannot happen with the current status. For me, this means that another potential source of error is being added here when executing the SIL. 

I'm sorry, but I see no reason why the payment option should be added again.

SebaLukas:
After a brief discussion with @corneliusclaussen, I can understand why you would want to select `PnC` or `EIM` via node-red when EVerest is running. And don't always want to restart everest with a different config. 

I still don't like the possibility of setting the payment option directly for the car sim.

As an alternative, however, I could imagine that you can set a payment option priority using the `iso_start_v2g_session` cmd.
A priority list could be defined in Josev. E.g. `EIM: 1`, `PnC: 2` (small number -> higher priority) The same principle applies to the SupportAppProtocol message.
I.e. if Josev can select both options (`EIM` and `PnC`), Josev selects the relevant payment option depending on the priority. The cmd can be used to change the priority. E.g. `iso_start_ DC, PnC` -> `EIM: 2`, `PnC: 1` 
This has the advantage that if the priority is set to `PnC` but the Charger can only do `EIM`, `PnC` is ignored by Josev and `EIM` is selected. And you can "change" the payment options during running EVerest via node-red.

The options for not sending a payment option via the cmd do not work. The interface system does not allow optional arguments. Therefore the `iso15118_ev` interface should also be adapted here. All node-red yaml files in the config folder would also have to be adapted.

What do you think of the suggestion?

SebaLukas:
@louisg1337 Any thoughts on my suggestion?

corneliusclaussen:
Is this still being worked on or should we close it?

SebaLukas:
I'll close this next friday if nobody answers

SebaLukas:
So I close this one. If there is a need, then please reopen a comment on my proposal.

Abby-Wheelis:
These message pairs are already in the logs:

```
Decoded message (ns=Namespace.ISO_V2_MSG_DEF): 
{"V2G_Message":{"Header":{"SessionID":"BB1CFF3F1F6BE78E"},"Body":{"ServiceDiscoveryRes":{"ResponseCode":"OK",
**"PaymentOptionList":{"PaymentOption":["ExternalPayment"]},**
"ChargeService":{"ServiceID":1,"ServiceCategory":"EVCharging","FreeService":false,"SupportedEnergyTransferMode":{"EnergyTransferMode":["AC_single_phase_core"]}},
"ServiceList":{"Service":[{"ServiceID":2,"ServiceName":"Certificate","ServiceCategory":"ContractCertificate","FreeService":true}]}}}}}

...

Message to encode (ns=Namespace.ISO_V2_MSG_DEF): {"V2G_Message": {"Header": {"SessionID": "BB1CFF3F1F6BE78E"}, 
"Body": {"PaymentServiceSelectionReq": {**"SelectedPaymentOption": "ExternalPayment",** "SelectedServiceList": {"SelectedService": [{"ServiceID": 1}]}}}}}
```

Based on the list and the selection, I think the suggestion of a ranked list set via nodered makes sense. If multiple payment options are supported, then the car can select which of those it prefers.

SebaLukas:
My problem with this is still that it can lead to more issues or problems, as the user may set the Node-RED script incorrectly.
If you can now blindly set the payment option for the car, e.g., PnC, but the charger or EVerest for the SIL only offers ExternalPayment, then the user is initially confused and has to check why the session is being canceled. Unlike with the charger, there is no option to add an enforce in NodeRed. That would make it a little clearer.

That's why I still think my alternative is best. The SIL transmit a priority list. Of course, this can only be filled with PnC or EIM, too. But that makes it a little clearer for the "user experience".

Abby-Wheelis:
We really want to set it in the user interface. This will allow users of the UI to control and understand the simulation they are running. 

I would propose:
- allow the user to select the payment type in the UI
- list is sent to the car simulator
- if supported by car simulator and found in `PaymentOptionsList`, charge with that payment type
- if not supported in current configuration of car simulator or charger - throw a clear error of what happened

This will allow the user to control the simulator they are running, and with a clear error they can understand what went wrong and how to fix it, without being too stuck or confused. 

Alternate list approach:
- user selects a ranked list of payment types -- require a "complete ballot" where all choices are ranked
- list is sent to the car simulator
- car simulator reconciles user list with what it is configured to support and `PaymentOptionsList`, selects most preferred which is supported

This approach is more confusing, in my opinion, because it is hiding more from the user of what is happening. We could log clearly what is selected and why - but if a user is trying to test `PnC` they would probably prefer to know why that's not happening, rather than a silent default to `ExternalPayment`. It could lead to more failed charging sessions, but the goal of this testing environment is to test different scenarios, not for charging to always succeed.

shankari:
At a higher level, my understanding is that the goal of the car simulator is **not** to have a successful charge every time - it is not a real car, and nobody is relying on it to get home. Instead, it is to allow users to simulate various charging scenarios, both successful and unsuccessful. That is why most of the node-red flows support generating errors from the UI (e.g. the MREC sliders or "AC RCD Error".

This is just another error scenario. We should improve error reporting so that the user is not confused, regardless of the source of the error.

SebaLukas:
To cut this short and avoid pointless discussions about what is more understandable for the user, you could also open a PR with your idea. I can review this PR and approve and merge it if it is good enough. An "improved" error reporting system could also be part of this PR.

Yes, the SIL is really powerful and can be used to test not only the right way, but also a wide variety of scenarios. But not all of them make sense.
E.g. what is the point of selecting PnC in NodeRed if I don't have any certificates available? Josev should actually abort here if no TLS connection has been established and only PnC is available. Of course, this scenario can be tested, but does it really benefit anyone?
For me, it's only important to have a backup case in the car simulation and not to blindly enforce things.

Furthermore, NodeRed is not the only way to control the SIL. Please be aware of this.

I would be interested to know in which scenarios you need to set the PaymentOption for the car side.

PR #752, opened 2024-07-01
Perform ev-cli setup in ev_add_project()

hikinggrass:
This eases the use of out-of-source repositories

### DIFF
--- a/cmake/everest-generate.cmake
+++ b/cmake/everest-generate.cmake
@@ -26,7 +26,7 @@ set_target_properties(generate_cpp_files
 #
 # out-of-tree interfaces/types/modules support
 #
-function(ev_add_project)
+function(_ev_add_project)
     # FIXME (aw): resort to proper argument handling!
     if (ARGC EQUAL 2)
         set (EVEREST_PROJECT_DIR ${ARGV0})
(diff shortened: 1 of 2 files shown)


cmake/everest-generate.cmake
```diff
@@ -26,13 +26,7 @@ set_target_properties(generate_cpp_files
 #
 # out-of-tree interfaces/types/modules support
 #
-function(ev_add_project)
-    # FIXME (aw): resort to proper argument handling!
-    if (ARGC EQUAL 2)
-        set (EVEREST_PROJECT_DIR ${ARGV0})
-        set (EVEREST_PROJECT_NAME ${ARGV1})
```
andistorm, line 33: Why can those args be omitted?

cmake/everest-generate.cmake
```diff
@@ -125,6 +119,20 @@ function(ev_add_project)
...
+    ev_setup_cmake_variables_python_wheel()
+    option(${PROJECT_NAME}_INSTALL_EV_CLI_IN_PYTHON_VENV "Install ev-cli in python venv instead of using system" ON)
+    set(${PROJECT_NAME}_PYTHON_VENV_PATH "${CMAKE_BINARY_DIR}/venv" CACHE PATH "Path to python venv")
+
+    ev_setup_python_executable(
+        USE_PYTHON_VENV ${${PROJECT_NAME}_USE_PYTHON_VENV}
+        PYTHON_VENV_PATH ${${PROJECT_NAME}_PYTHON_VENV_PATH}
+    )
+    setup_ev_cli()
+
```
andistorm, line 132: In case of actual defining those vars for example `some-out-of-tree-project_PYTHON_VENV_PATH=${CMAKE_BINARY_DIR}/venv1` and `everest-core_PYTHON_VENV_PATH=${CMAKE_BINARY_DIR}/venv2`, there would be two python virtualvenvs be installed which could lead to undefined behaviour during the build process. Same for `${PROJECT_NAME}_INSTALL_EV_CLI_IN_PYTHON_VENV`.

Maybe we should change to non-project-prefixed vars with global validity as for example:
`EVEREST_INSTALL_EV_CLI_IN_PYTHON_VENV`

Those vars could be setup in a dedicated function which can be triggered from multiple places. This function could be guarded to be only executed once.

cmake/everest-generate.cmake
hikinggrass: I can re-add them but I've not seen them used anywhere

cmake/everest-generate.cmake
hikinggrass: wasn't this the original intention to be able to specify multiple venvs? Otherwise the ${PROJECT_NAME}_ prefixes would be a bit redundant. For now I can at least report it works well with the in tree and out-of-tree repos, as well as yocto. And doesn't need any more modifications on out-of-tree repos :partying_face:

cmake/everest-generate.cmake
andistorm: oh nvmd, I have overseen that there are set default values.

So the following can be shortened a bit, since with removing the args they are definetly not defined:
```
    if (NOT EVEREST_PROJECT_DIR)
        # if we don't get a directory, we're assuming project directory
        set (EVEREST_PROJECT_DIR ${PROJECT_SOURCE_DIR})
        set (CALLED_FROM_WITHIN_PROJECT TRUE)
    elseif (NOT EXISTS ${EVEREST_PROJECT_DIR})
        message(FATAL_ERROR "${CMAKE_CURRENT_FUNCTION} got non-existing project path: ${EVEREST_PROJECT_DIR}")
    endif ()
    if (NOT EVEREST_PROJECT_NAME)
        set (EVEREST_PROJECT_NAME ${PROJECT_NAME})
    endif ()
```

hikinggrass: I've readded them, they were needed 😅

corneliusclaussen (approved):

andistorm (approved):
:+1: 🐧

PR #755, opened 2024-07-02
Added temperature sensors values in Powermeter

florinmihut:
…mperature2)

    - added in interface a variable containing the public OCMF key (publicKeyOcmf)
    - added in config the following options: meter_timezone, meter_dst, SC, UV, UD
    - added support for setting the following parameters: meter_dst and meter_timezone
    - added support for V2 interface


modules/RsIskraMeter/src/main.rs
```diff
@@ -629,6 +629,8 @@ impl ReadyState {
             signed_meter_value: None,
             var_signed: None,
             voltage_v_signed: None,
+            temperature_1: Option::None,
```
SirVer, line 632: no need to specficy the full path (i.e. `Option::None`), `None` is enough.

SirVer (approved):
only one nit for rust code.

modules/LemDCBM400600/main/lem_dcbm_400600_controller.hpp
```diff
@@ -78,17 +88,22 @@ class LemDCBM400600Controller {
 
 private:
     const std::unique_ptr<HttpClientInterface> http_client;
-    std::string meter_id;
+    std::string meter_id = "";
+    std::string public_key = "";
+    std::string public_key_ocmf = "";
+    std::string version = "";
```
Pietfried, line 94: std::string initializes to empty string by default

modules/LemDCBM400600/main/lem_dcbm_400600_controller.hpp
```diff
@@ -78,17 +88,22 @@ class LemDCBM400600Controller {
...
+    bool trasaction_is_ongoing = false;
     Conf config;
     std::unique_ptr<LemDCBMTimeSyncHelper> time_sync_helper;
 
+    std::vector<std::string> split(const std::string& str, char delimiter);
     void fetch_meter_id_from_device();
     void request_device_to_start_transaction(const types::powermeter::TransactionReq& value);
     void request_device_to_stop_transaction(const std::string& transaction_id);
     std::string fetch_ocmf_result(const std::string& transaction_id);
     void convert_livemeasure_to_powermeter(const std::string& livemeasure, types::powermeter::Powermeter& powermeter);
```
Pietfried, line 105: not part of the PR but could also be defined as a free function

types/powermeter.yaml
```diff
@@ -299,3 +299,10 @@ types:
           This is intended for meters that only support signing a collection of meter values.
         type: object
         $ref: /units_signed#/SignedMeterValue
+      temperature1:
```
Pietfried, line 302: This looks to LEM specific to me. Not sure which temperatures it actually measures, but would it make sense to you to add more generic properties like:
* ambient_temperature
* internal_temperature
* conductor_temperature
* ...

Pietfried (changes requested):

types/powermeter.yaml
florinmihut: I agree but since this is not defined by all the power meters, this has to cover everyone. 
What if I have 2 cable temps? Or a high and low temp? 
Starting to give them "actual functional" names might pollute the interface with a lot of unused variables.
I spoke with Cornelius as well and he proposed to use temp 1 and 2 and their meaning is Powermeter dependent. Since anyway it is - some have it, some have it not, some measure the ambient some measure the internal some measure the high and the low. 
You get the picture. 
So I would be in the favor of keeping them temp 1 and 2 and their meaning is Powermeter actual implementation dependent.

hikinggrass: If it's really that opaque we could also just add an array of temperature values here, what if there's a third sensor for example :D

Pietfried: > I agree but since this is not defined by all the power meters, this has to cover everyone.
> Starting to give them "actual functional" names might pollute the interface with a lot of unused variables

For this we have optional properties. There are a lot of properties in the powermeter already defined this way.

> What if I have 2 cable temps? Or a high and low temp?

As @hikinggrass pointed out, what if we have 3? That is indeed specific to the powermeter and should therefore not be covered in a generic interface like powermeter like this. An interface is meant to be required by other modules. How should other modules know what temp1 or temp2 or temp3 are, especially if they mean different things for different implementations. Defining it this way doesn't seem useful to me.

If we add it like this, Id agree with @hikinggrass and add an array of temperature values, but I would rather add more meaningful properties as described above, maybe defining a type for a temperature and adding optional low / high values to it as properties.

corneliusclaussen: I think an array is a good idea

modules/LemDCBM400600/main/lem_dcbm_400600_controller.cpp
```diff
@@ -197,8 +199,18 @@ void LemDCBM400600Controller::convert_livemeasure_to_powermeter(const std::strin
...
+                                         types::temperature::Temperature{data.at("temperatureL"), "temperatureL"}});
+    } else {
+        for (auto& sensor : *powermeter.temperatures) {
+            if (sensor.identification == "temperatureL") {
+                sensor.temperature = data.at("temperatureL");
+            } else if (sensor.identification == "temperatureH") {
+                sensor.temperature = data.at("temperatureH");
+            }
+        }
+    }
```
Pietfried, line 213: the `powermeter` argument passed to this function is default initialized and the properties are then set in this function. Therefore this `else` block is never executed and can be dropped. Its also not required to check for `has_value()` in L202.

Apart from that I'd prefer a function signature like:
`types::powermeter::Powermeter LemDCBM400600Controller::convert_livemeasure_to_powermeter(const std::string& livemeasure)` that returns the powermeter object instead of passing it as an argument, but since this is legacy code im ok with keeping the signature like that

Pietfried (changes requested):

modules/LemDCBM400600/main/lem_dcbm_400600_controller.cpp
hikinggrass: But isn't powermeter.temperatures an optional vector so that check is still needed?

Pietfried: maybe I miss something but afaics the `temperatures` property is always not set so the else block is never executed

Pietfried: because the passed `powermeter` is default initialized and passed to this function

hikinggrass: ah you're right, then it's a good idea to drop the else block indeed!

florinmihut: You're right, I didn't check where the powermeter value comes from. 
Since it is created by the calling function the else branch would never execute.
Thank you for catching that.
I made the changes, including the signature of the function since I am changing it again and the change is minor.

Pietfried (approved):

hikinggrass (approved):

PR #756, opened 2024-07-02
Install edm and deps on the fly

golovasteek:
## Describe your changes
Current version of everest requires edm to be installed in the system to run bazel-build. Which is fine for the everest developers, but may be annoying for the downstream repositories (to ensure every developer has edm on the laptop).

Now the edm is fetched on the fly into virtual environment.

### DIFF
--- a/third-party/bazel/edm-wrapper.sh
+++ b/third-party/bazel/edm-wrapper.sh
@@ -1,0 +1,11 @@
+#!/usr/bin/env bash
+set -eu -o pipefail
+
+EDM_LINK=git+https://github.com/Everest/everest-dev-environment.git#subdirectory=dependency_manager
+mkdir -p venv > /dev/null
+TARGET_PATH=$(realpath venv)
+pip3 install --upgrade -t $TARGET_PATH $EDM_LINK > /dev/null
+export PYTHONPATH=$TARGET_PATH
+export PATH=$TARGET_PATH/bin:$PATH
+
+edm "$@"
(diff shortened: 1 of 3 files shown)


third-party/bazel/edm-wrapper.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+set -eu -o pipefail
+
+EDM_LINK=git+https://github.com/Everest/everest-dev-environment.git#subdirectory=dependency_manager
+mkdir -p venv > /dev/null
+TARGET_PATH=$(realpath venv)
+pip3 install --upgrade -t $TARGET_PATH $EDM_LINK > /dev/null
+export PYTHONPATH=$TARGET_PATH
```
dorezyuk, line 8: should we restore the old PYTHONPATH after the command?

third-party/bazel/edm-wrapper.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+set -eu -o pipefail
+
+EDM_LINK=git+https://github.com/Everest/everest-dev-environment.git#subdirectory=dependency_manager
+mkdir -p venv > /dev/null
```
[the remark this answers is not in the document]
golovasteek: I wouldn't do it. I would keep things hermetic.

third-party/bazel/edm-wrapper.sh
golovasteek: We exit after the command, why bother restoring anything?

dorezyuk: ah brainfart - we never source this but just run

third-party/bazel/edm-wrapper.sh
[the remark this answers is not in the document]
dorezyuk: but this will run everytime we change anything in the sources right?

PR #757, opened 2024-07-02
Charger: Transition to PausedByEV if EV does not start charging after…

corneliusclaussen:
… wakeup

## Describe your changes

In PrepareCharging state PWM is enabled to signal to the EV that we are ready to deliver power. If the car does not switch to state C to start charging, a legacy wakeup was performed after 30s. 
If it still does not charge after that, it probably just does not want to charge (e.g. because it is full).
In the current implementation the state stays in PrepareCharging. This can lead to an issue that OCPP does not reflect the state correctly (and also local display apps stay in PrepareCharging).
This PR switches to PausedByEV which is in that case the correct state.


SebaLukas (approved):
:+1:

hikinggrass (approved):

Pietfried (approved):

PR #762, opened 2024-07-05
Charger: Fix soft over current detection when PWM changes often

corneliusclaussen:
## Describe your changes

Soft over current detection triggers when the EV draws more current then the PWM limit for several seconds. The current implementation checked the EV's current against the internally set current limit, which may be different from the PWM signalled current limit for up to 5 seconds (as IEC allows PWM updates only every 5 seconds).
If PWM is changed frequently (e.g. due to solar charging) this may lead to unintended over current triggers (it takes 5 seconds for the PWM to change and another 6s or so for some EVs to actually reduce current).
With this PR, the OC detection is based on the current actually signalled on the PWM for basic AC charging.


hikinggrass (approved):

SebaLukas (approved):
:+1:

PR #765, opened 2024-07-05
EvseManager: Wait for initial meter value

Pietfried:
## Describe your changes

If EVerest starts up with an EV plugged in (and authorization is present immediately), there is a race condition between the TransactionStarted event and the first powermeter value received. Without this change, transactions may start and use an initial meter value of 0.

This PR adds handling to wait for an initial powermeter value before the EvseManager becomes operational (is ready to start charging). The waiting time is configurable ( `initial_meter_value_timeout_ms` ) or can be disabled (by configuring 0)

### DIFF
--- a/modules/EvseManager/EvseManager.cpp
+++ b/modules/EvseManager/EvseManager.cpp
@@ -678,7 +678,7 @@ void EvseManager::ready() {
             }
 
             {
-                std::lock_guard<std::mutex> lg(powermeter_mutex);
+                std::scoped_lock<std::mutex> lk(powermeter_mutex);
                 initial_powermeter_value_received = true;
             }
             powermeter_cv.notify_one();
--- a/modules/EvseManager/EvseManager.cpp
+++ b/modules/EvseManager/EvseManager.cpp
@@ -899,10 +899,12 @@ void EvseManager::ready() {
         }
     });
 
-    // wait for first powermeter value
-    std::unique_lock<std::mutex> lk(powermeter_mutex);
-    this->powermeter_cv.wait_for(lk, std::chrono::milliseconds(this->config.initial_meter_value_timeout_ms),
-                                 [this] { return initial_powermeter_value_received; });
+    {
+        // wait for first powermeter value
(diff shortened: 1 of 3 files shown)


lategoodbye:
@Pietfried Is there a possible relation to this [issue](https://github.com/EVerest/libocpp/issues/678)?

modules/EvseManager/EvseManager.cpp
```diff
@@ -677,6 +677,12 @@ void EvseManager::ready() {
                 latest_powermeter_data_billing = p;
             }
 
+            {
+                std::lock_guard<std::mutex> lg(powermeter_mutex);
```
corneliusclaussen, line 681: I would prefer std::scoped_lock here as used in the rest of the module

modules/EvseManager/EvseManager.cpp
```diff
@@ -893,6 +899,11 @@ void EvseManager::ready() {
         }
     });
 
+    // wait for first powermeter value
+    std::unique_lock<std::mutex> lk(powermeter_mutex);
```
corneliusclaussen, line 903: Doesn't that deadlock if the first power meter value arrives after that lock? Maybe we should make initial_powermeter_value_received an std::atomic_bool instead and remove the locking in the power meter handler?

Pietfried:
> @Pietfried Is there a possible relation to this [issue](https://github.com/EVerest/libocpp/issues/678)?

Good catch, looks like it. If the powermeter is not updated, the timestamp string defaults to `""`

modules/EvseManager/EvseManager.cpp
Pietfried: The `this->powermeter_cv.wait_for` releases the mutex so there should not be a deadlock

dorezyuk: but you don't need the lock after the wait_for returns. Maybe its better to place both in an own scope

corneliusclaussen (approved):

PR #768, opened 2024-07-08
Feature/376 ocpp 201 california pricing requirements tariff and cost

maaikez:
## Describe your changes

Interface changes needed for California Pricing Requirements: TariffAndCost and DisplayMessage.

libocpp / 376


types/session_cost.yaml
```diff
@@ -17,13 +50,25 @@ types:
...
       cost:
         description: >-
           Cost of the energy (or other things like time, base price, ...) 
           consumed during this chunk in the currency specified in the session
           price
         type: object
         $ref: /money#/MoneyAmount
+      costAmount:
+        description: >-
+          Cost in the currency specified in the configuration.
```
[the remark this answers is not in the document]
maaikez: Removed costAmount, but in the configuration there will be a currency which is used on the charging station.

interfaces/display_message.yaml
```diff
@@ -0,0 +1,83 @@
...
+          status:
+            description: Whether the charging station is able to display the message
+            $ref: /display_message#/DisplayMessageStatusEnum
+            type: object
+          status_info:
+            description: Detailed status information
+            $ref: /ocpp#/StatusInfoType
+            type: object
+        required:
+          - status
```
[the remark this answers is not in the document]
maaikez: I will, but it does just build.

types/session_cost.yaml
```diff
@@ -1,5 +1,38 @@
...
+        description: >-
+          Grace period in minutes before idle time is charged.
+          Grace minutes start counting from the timestamp of the message in which state
+          changed from "Charging" to "Idle".
+        type: integer
+      hourPrice:
+        description: >-
+          Price per hour while idle
+        type: number
+  ChargingPriceComponent:
```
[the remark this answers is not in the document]
maaikez: I kept it but changed the enum type to 'category'. Because we need an array with this information in SessionCost (otherwise I have to define it there, which is also fine by me).

interfaces/display_message.yaml
```diff
@@ -0,0 +1,43 @@
+description: >-
+  A module that implements this interface should be able to:
+  - store (add, remove, change) and retrieve predefined messages
+  - show messages on a display
+
+  When a display message contains a session id, the display message must be removed once the session has ended.
+cmds:
+  set_display_message:
```
dorezyuk, line 8: is `set` here like `display` a screen? If yes, then what is the meaning/usecase of get?

interfaces/display_message.yaml
```diff
@@ -0,0 +1,43 @@
...
+    arguments:
+      request:
+        description: The request for display messages
+        type: object
+        $ref: /display_message#/GetDisplayMessageRequest
+    result:
+      description: The display messages or an empty array if there are none
+      type: object
+      $ref: /display_message#/GetDisplayMessageResponse
+  clear_display_message:
```
dorezyuk, line 33: then the display should stop showing the current screen?

interfaces/display_message.yaml
dorezyuk: ah its from the ocpp spec - maybe we can add this to the description of the module

types/session_cost.yaml
```diff
@@ -1,5 +1,62 @@
 description: Types to provide the session price
 types:
+  IdlePrice:
```
dorezyuk, line 3: i'm a little confused - from what i understand we refer to idle to the state when the station has no customer. What does idle mean in this context?

types/session_cost.yaml
```diff
@@ -28,17 +93,15 @@ types:
...
           Cost category. Typically this is energy or time.
         type: string
-        enum:
-          - Energy
-          - Time
-          - Other
+        $ref: /session_cost#/CostCategory
   SessionStatus:
     description: >-
       Session status enum. Session can be running or finished. Costs of the
```
dorezyuk, line 99: the doc is then outdated

types/session_cost.yaml
```diff
@@ -17,6 +74,14 @@ types:
           Absolute timestamp for the end of this chunk in RFC3339 UTC format
         type: string
         format: date-time
+      metervalue_from:
+        description: >-
+          Metervalue (energie Wh import) for the start of this chunk
+        type: integer
```
dorezyuk, line 80: why not using powermeter types here?

types/session_cost.yaml
```diff
@@ -1,5 +1,62 @@
...
+      hour_price:
+        description: >-
+          Price per hour while idle
+        type: number
+  CostCategory:
+    description: Cost category
+    type: string
+    enum:
+      - Energy
```
dorezyuk, line 23: can't we model the idle price not as a category in here?

types/session_cost.yaml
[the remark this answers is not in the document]
maaikez: A SessionCostChunk is the 'real cost' of the session (the total cost of the charging session until now for example), ChargingPriceComponent is the price per kWh, hour or flat fee. So if you want to calculate the total cost (for example if the charging station goes offline), you need the charging price components to be able to do that.

interfaces/display_message.yaml
maaikez: correct

we did not add that to the description because we also want to use it for internal messages.

types/session_cost.yaml
maaikez: For more info, you can read the California Pricing Requirements.
https://openchargealliance.org/whitepapers/ <- to be downloaded there. 

Idle is when a car is attached but currently not charging. For example because the battery is already full. There are charging stations where you have to pay something when you are not charging (I mainly found them in the field with DC chargers, but also AC chargers sometimes have this).

types/session_cost.yaml
```diff
@@ -1,5 +1,62 @@
...
+        description: >-
+          Price for this component
+        type: object
+        $ref: /money#/MoneyAmount
+  NextPeriodPrice:
+    description: Pricing for the next period
+    type: object
+    properties:
+      timestamp_from:
+        description: Time when these prices become active.
```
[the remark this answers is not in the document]
maaikez: Well it refers to the charging prices and idle prices.

types/session_cost.yaml
maaikez: Because this is as specified in California Pricing Requirements. We just get a metervalue from the CSMS. But if you think we should change this: what is your proposal?

types/session_cost.yaml
maaikez: So you would like to add a IdlePriceTime, ChargingTime, IdleGraceMinutes and remove Time? I thought about it but I don't know if that makes things more clear.

types/session_cost.yaml
dorezyuk: cool understood - maybe you can link the req in the description

types/session_cost.yaml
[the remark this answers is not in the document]
dorezyuk: understood - but can it be modeled similarly to the SessionCostChunk and also named similarly, e.x. `PriceChunk` and allow the user to pass an array of chunks

types/session_cost.yaml
```diff
@@ -1,5 +1,64 @@
...
+      category:
+        description: >-
+          Component type for this charging price
+        type: string
+        $ref: /session_cost#/CostCategory
+      price:
+        description: >-
+          Price for this component
+        type: object
+        $ref: /money#/Price
```
dorezyuk, line 41: money is probably not the right thing here. The price.yaml defines the amount as integer in cents but it can be in this case that the kWh price is even a fraction of that. You sometimes see this for example at gas stations.

modules/OCPP/display_message/display_messageImpl.cpp
```diff
@@ -0,0 +1,47 @@
...
+void display_messageImpl::init() {
+}
+
+void display_messageImpl::ready() {
+}
+
+types::display_message::SetDisplayMessageResponse
+display_messageImpl::handle_set_display_message(std::vector<types::display_message::DisplayMessage>& request) {
+    return this->mod->r_display_message.at(0)->call_set_display_message(request);
+    // types::display_message::SetDisplayMessageResponse response;
```
dorezyuk, line 18: remove?

types/session_cost.yaml
dorezyuk: i don't really understand why we need to define actually idle as a separate category. Your linked whitepater says
> It is up to the CSMS and not the charging station, to decide when a transaction is considered to be in idle time.

So at some point when the station sends a `chargingState` or `atPowerkW` and the CSMS thinks that its now idle it will send a price containing just the time component. The CSMS can internally also account for grace times or other conditions

types/session_cost.yaml
maaikez: What should it be then?
Price has a currency, which has a 'number of decimals' and currency. So with the 'number of decimals' you can define how many decimals the number has.

types/session_cost.yaml
maaikez: Yes but what if the CSMS is not online? 
I just reflect the fields from the California Pricing Requirements, which sends the idle price as well.

types/session_cost.yaml
maaikez: > i would use the powermeter type since it would fit better in the ecosystem

I am fine with that, @Pietfried  is that ok for you?

types/session_cost.yaml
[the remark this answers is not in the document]
maaikez: Well that is only renaming ChargingPriceComponent to PriceChunk isn't it? As it already is an array.

types/session_cost.yaml
maaikez: Ah yes the link is already in the issue:
https://github.com/EVerest/libocpp/issues/376

Pietfried (changes requested):
Looks good overall, see in line comments :+1:

types/display_message.yaml
```diff
@@ -0,0 +1,162 @@
...
+  DisplayMessage:
+    description: Message to show on a display
+    type: object
+    additionalProperties: false
+    properties:
+      id:
+        description: The message id
+        type: integer
+      priority:
+        description: Priority of the message
```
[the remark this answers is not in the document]
maaikez: Added a description (I will push later). I also added a better description to MessagePriorityEnum. Priority is mandatory in OCPP 2.0.1 but as we want to use this internally as well, and OCPP 1.6 California Pricing does not use the priority, I did not make it required.

modules/OCPP/conversions.cpp
```diff
@@ -454,5 +454,345 @@ to_everest_registration_status(const ocpp::v16::RegistrationStatus& registration
...
+types::session_cost::SessionCostChunk create_session_cost_chunk(const double& price, const uint32_t& number_of_decimals,
+                                                                const std::optional<ocpp::DateTime>& timestamp,
+                                                                const std::optional<uint32_t>& meter_value) {
+    types::session_cost::SessionCostChunk chunk;
+    chunk.cost = types::money::MoneyAmount();
+    chunk.cost->value = static_cast<int>(price * (pow(10, number_of_decimals)));
+    if (timestamp.has_value()) {
+        chunk.timestamp_to = timestamp.value().to_rfc3339();
+    }
+    chunk.metervalue_to = meter_value;
```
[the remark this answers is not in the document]
maaikez: Well in ocpp there is no metervalue_from. But there is a metervalue_from and metervalue_to in the interface (they were already there).

modules/DisplayMessage/DisplayMessage.cpp
```diff
@@ -0,0 +1,68 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Pionix GmbH and Contributors to EVerest
+#include "DisplayMessage.hpp"
+
+namespace module {
+
+void DisplayMessage::init() {
+    invoke_init(*p_display_message);
+    this->r_session_cost->subscribe_session_cost([](const types::session_cost::SessionCost& session_cost) {
```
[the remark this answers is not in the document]
maaikez: Hmm yes, why I also added this is because there is a message in the session cost that should be shown on the display, just like the qr code. So maybe I should not show all the charging price info etc, but the message should be shown. 

Or I send two messages of mqtt, one session cost with only the price information and one display message with the message only. But I think you might want to show stuff from the session cost in the screen as well (convert that to a message).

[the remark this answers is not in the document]
maaikez: ok changed the name of the module.

dependencies.yaml
```diff
@@ -60,7 +60,7 @@ libevse-security:
 # OCPP
 libocpp:
   git: https://github.com/EVerest/libocpp.git
-  git_tag: aab1d5785bde141203b1a89b03d66fa91edf8093
+  git_tag: 1009ca1550a54824b9e2c1eacdad69f9373fd542
```
maaikez, line 63: TODO set to correct git tag before merging to main!!!

SirVer (approved):

modules/OCPP/display_message/display_messageImpl.cpp
```diff
@@ -0,0 +1,31 @@
+// SPDX-License-Identifier: Apache-2.0
```
Pietfried, line 1: the whole display_message directory shall be removed from OCPP. Its not something that the module provides, but optionally requires (see the manifest)

modules/OCPP/OCPP.cpp
```diff
@@ -752,6 +752,42 @@ void OCPP::ready() {
             this->p_ocpp_generic->publish_boot_notification_response(everest_boot_notification_response);
         });
 
+    this->charge_point->register_session_cost_callback(
+        [this](const ocpp::RunningCost& session_cost,
+               const uint32_t number_of_decimals) -> ocpp::v16::DataTransferResponse {
+            const types::session_cost::SessionCost cost =
+                conversions::create_session_cost(session_cost, number_of_decimals, {});
+            ocpp::v16::DataTransferResponse response;
+            if (this->p_session_cost == nullptr) {
```
Pietfried, line 761: nit: `p_session_cost` cannot really be null, since providing an interface implementation is not optional for a module.

Pietfried (changes requested):

modules/OCPP/OCPP.cpp
maaikez: ok i'll remove it. on your risk ;)

Pietfried (approved):

PR #769, opened 2024-07-09
Remove get_hardware_capablities() and 1ph3ph switching support

corneliusclaussen:
Implementation of evse_board_support was simplified a bit. The sync command to get hw capablities on startup is no longer needed as the async publish of capablities is enough. Each BSP implementation module is required to publish its capablities at least once in the ready function. It may publish it later on to update the values e.g. for thermal derating.

1ph/3ph switching support during charging with a hyteresis if the 1ph/3ph power ranges overlap (e.g. on 22kW).
Hysteresis only really works if EV and EVSE have the same max current rating, e.g. both 16A See doc.rst for details.


modules/YetiDriver/yetiR1_2.2_firmware.bin
Pietfried, line 1: This requires a change here as well
https://github.com/EVerest/everest-core/blob/main/modules/YetiDriver/CMakeLists.txt#L39

corneliusclaussen: fixed

modules/EvseManager/EvseManager.hpp
```diff
@@ -231,6 +235,7 @@ class EvseManager : public Everest::ModuleBase {
             charger->inform_new_evse_max_hlc_limits(evseMaxLimits);
         }
     }
+    std::atomic_int ac_nr_ph_phases_active{0};
```
Pietfried, line 238: `ac_nr_phases_active` ?

modules/PhyVersoBSP/connector_1/evse_board_supportImpl.cpp
```diff
@@ -59,15 +59,6 @@ void evse_board_supportImpl::init() {
 void evse_board_supportImpl::ready() {
 }
 
-void evse_board_supportImpl::handle_setup(bool& three_phases, bool& has_ventilation, std::string& country_code) {
```
Pietfried, line 62: Looks like you need to publish the caps for connector 1 as well like its done for connector 2

Pietfried (changes requested):
See comments + Lint and Tests are currently failing

Pietfried (approved):

PR #771, opened 2024-07-09
PhyVersoBSP: Keep alive, config rework, RCD handling

rckstrh:
## Describe your changes
1Hz keep alive messages from EVerest to MCU have been implemented so that when EVerest crashes and keep alive messages stop the MCU will timeout and go into a fault mode.

Rework has been done on the PhyVersoBSP module regarding configuration options that have been managed in a separate json config file until now. JSON config file support remains for phyverso_cli testing tool but the config handling has been changed so that the PhyVersoBSP module is configurable regarding MCU config data using normal module config parameters in EVerest config like any other module.

MCU can now be configured to have different motor lock styles on each charging port and GPIO bank and pin can be configured for accessing the nRST signal in case one wants to hard reset the MCU from EVerest/phyverso_cli.

RCD error reporting and handling has been added as well as the ability to set and reset the TEST signal for the RCD by MCU and reset the RCD from EVerest after it has been triggered and MCU/system is in emergency-off mode. 


modules/PhyVersoBSP/phyverso_mcu_comms/evConfig.cpp
```diff
@@ -31,16 +32,21 @@ bool evConfig::open_file(std::string path) {
 
 // todo: needs to be refactored a lot
 bool evConfig::check_validity() {
-    std::cout << config_file.dump() << "\n\n";
+    std::cout << config_file.dump() << "\n\n"; // DEBUG
```
corneliusclaussen, line 35: Maybe left over from debugging?

corneliusclaussen (approved):

modules/PhyVersoBSP/phyverso_mcu_comms/evConfig.cpp
rckstrh: Yes, this was left over from debugging, the comment above the function was from pre config rework so I removed that as well in [44c90c1](https://github.com/EVerest/everest-core/pull/771/commits/44c90c1cac008ed939ff12ddb9f067ed3cb2612c).

modules/PhyVersoBSP/rcd_1/ac_rcdImpl.cpp
```diff
@@ -7,17 +7,23 @@ namespace module {
 namespace rcd_1 {
 
 void ac_rcdImpl::init() {
+    mod->serial.signal_error_flags.connect([this](int connector, ErrorFlags error_flags) {
+        if ((connector == 1) && error_flags.rcd_triggered) {
+            EVLOG_info << "[1] RCD triggered: " << error_flags.rcd_triggered;
+            // TODO: publish somewhere
```
hikinggrass, line 13: This could probably raise a ac_rcd/AC error: https://github.com/EVerest/everest-core/blob/main/errors/ac_rcd.yaml

modules/PhyVersoBSP/rcd_2/ac_rcdImpl.cpp
```diff
@@ -7,17 +7,23 @@ namespace module {
 namespace rcd_2 {
 
 void ac_rcdImpl::init() {
+    mod->serial.signal_error_flags.connect([this](int connector, ErrorFlags error_flags) {
+        if ((connector == 2) && error_flags.rcd_triggered) {
+            EVLOG_info << "[2] RCD triggered: " << error_flags.rcd_triggered;
```
hikinggrass, line 12: same as rcd_1

hikinggrass (approved):

modules/PhyVersoBSP/rcd_1/ac_rcdImpl.cpp
rckstrh: I have changed the code to raise a ac_rcd/DC error because the phyVERSO is specified to use a residual direct current monitoring device, so normally this would be blind to AC fault currents. I could also change it to MREC2GroundFailure error because this error seems to not discriminate between AC or DC faults in case one does not only want to attach, like specified, a RDC-MD but some kind of monitoring device that is not blind to AC fault currents.

PR #776, opened 2024-07-12
Multiplexer module for ISO communication

corneliusclaussen:
This module starts and SDP server and waits
for incoming V2G connections (TCP, and optionally TLS). Upon connection, it reads the SupportedAppProtocol.req message to determine if the EV supports ISO-20 or not.

If ISO-20 is supported, it transparently bridges the connection to the ISO-20 module (using TLS->TCP).

If the EV does not support ISO-20, it bridges to the ISO-2 module instead.

Both the required iso-2 and iso-20 modules only need TCP enabled, but they require to accept e.g. PnC on TCP sessions (as the actual communication to the EV may be TLS).
Also the need to listen on localhost (::1) instead of the actual PLC device.


corneliusclaussen:
What's missing: D20 module needs support for ::1 TCP connections via option, both modules need to allow PnC via option on TCP connection. Config files to be updated to actually use D20 module once it is merged.

a-w50 (commented):
Stopping reviewing here as most of the code base is copied from the EvseV2G module (and therefore this would be more like a review for that code base).  Nice to see that the proxying can be implemented that easy via the `connection_proxy` functions.
I would recommend for the next time, when copying source files, that they are checked in with their initial/original content in a first commit - so it would be possible to track changes.

corneliusclaussen:
> Stopping reviewing here as most of the code base is copied from the EvseV2G module (and therefore this would be more like a review for that code base). Nice to see that the proxying can be implemented that easy via the `connection_proxy` functions. I would recommend for the next time, when copying source files, that they are checked in with their initial/original content in a first commit - so it would be possible to track changes.

Thanks for your review, I'm not sure if it makes sense to fix the issues that are copy past from the EvseV2G module as it will make it harder to merge fixes done to EvseV2G into this module. This module is also just an intermediate solution until the D20 module supports DIN and ISO-2 as well.

SebaLukas (approved):
:+1:

modules/IsoMux/v2g_server.cpp
```diff
@@ -0,0 +1,179 @@
...
+            dlog(DLOG_LEVEL_ERROR, "out-of-memory condition");
+            return V2G_EVENT_TERMINATE_CONNECTION;
+        }
+
+        dlog(DLOG_LEVEL_INFO,
+             "handshake_req: Namespace: %s, Version: %" PRIu32 ".%" PRIu32 ", SchemaID: %" PRIu8 ", Priority: %" PRIu8,
+             proto_ns, app_proto->VersionNumberMajor, app_proto->VersionNumberMinor, app_proto->SchemaID,
+             app_proto->Priority);
+
+        free(proto_ns);
```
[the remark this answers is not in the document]
SebaLukas: fixed

lib/staging/tls/tls.hpp
```diff
@@ -221,6 +222,19 @@ class Connection {
...
+    /**
+     * \brief obtain the underlying SSL context
+     * \returns the underlying SSL context pointer
+     */
+    [[nodiscard]] SSL* ssl_context() const;
+
+    /**
+     * \brief set the read timeout in ms
+     */
+    void set_read_timeout(int ms) {
```
james-ctc, line 235: Not sure this is needed.  
`[[nodiscard]] result_t read(std::byte* buf, std::size_t num, std::size_t& readbytes, int timeout_ms);`
supports a different timeout ...
Full non-blocking IO is supported, so may not be needed at all.

lib/staging/tls/tls.hpp
```diff
@@ -221,6 +222,19 @@ class Connection {
      * \note the certificate must not be freed
      */
     [[nodiscard]] const Certificate* peer_certificate() const;
+
+    /**
+     * \brief obtain the underlying SSL context
+     * \returns the underlying SSL context pointer
+     */
+    [[nodiscard]] SSL* ssl_context() const;
```
james-ctc, line 230: The idea was to avoid making OpenSSL internals available.
shouldn't be needed since both read() and write() return
        want_read,  //!< non-blocking - operation waiting for read available on socket
        want_write, //!< non-blocking - operation waiting for write available on socket
(see above)

modules/IsoMux/connection/tls_connection.cpp
```diff
@@ -0,0 +1,413 @@
...
+    struct pollfd poll_list[2];
+    poll_list[0].fd = proxy_fd;
+    poll_list[1].fd = ev_fd;
+    poll_list[0].events = POLLIN;
+    poll_list[1].events = POLLIN;
+
+    unsigned char buf[2048];
+
+    // Set reading to (more or less) non-blocking
+    conn->tls_connection->set_read_timeout(10);
```
james-ctc, line 343: TLS code fully supports non-blocking IO. Used when timeout == 0.
both read() and write() have methods that take the timeout parameter.

modules/IsoMux/connection/tls_connection.cpp
```diff
@@ -0,0 +1,413 @@
...
+        if (r < 0) {
+            // something is wrong with the connection, exiting...
+            break;
+        } else if (r > 0) {
+            // successfully read bytes, forward to proxy module
+            write(proxy_fd, buf, r);
+        }
+
+        // check if SSL was actually waiting on write
+        int e = SSL_get_error(conn->tls_connection->ssl_context(), r);
```
james-ctc, line 363: TLS read() and write() already return this.
```
        want_read,  //!< non-blocking - operation waiting for read available on socket
        want_write, //!< non-blocking - operation waiting for write available on socket
```

lib/staging/tls/tls.hpp
corneliusclaussen: Agreed that this should not be exposed. Given the temporary nature of IsoMux module (until the -2/DIN spec implementation in libiso15118 is ready) I marked this (and the set_timeout) function as deprecated so that they are not used by any other module. They will be removed together with IsoMux module then.

hikinggrass (approved):

PR #778, opened 2024-07-12
feat: add optional support for TLS trusted_ca_keys extension

james-ctc:
feat: add documentation
fix: consistent type naming convention applied
fix: failed tests
feat: added signal to stop stop server
feat: SA_RESTART flag + notes on client certificates fix: memory leak
feat: testing non-blocking read
feat: patched test only includes the test that would fail
      on an unpatched OpenSSL
fix: logging to omit filename and line information fix: partial config support improved where trust anchors are missing

## Describe your changes

Added support for TLS trusted_ca_keys extension.
Note its use is disabled since changes to EVSE Security are needed to obtain the trust anchor certificates needed by the extension


modules/EvseV2G/EvseV2G.cpp
```diff
@@ -16,11 +17,13 @@ void log_handler(openssl::log_level_t level, const std::string& str) {
         // ignore debug logs
         break;
     case openssl::log_level_t::warning:
-        EVLOG_warning << str;
+        // avoid logging function and line number
+        BOOST_LOG_SEV(::global_logger::get(), ::Everest::Logging::warning) << str;
```
corneliusclaussen, line 21: Is this really neccessary? We may want to replace boost::log at some point, so the EVLOG_nn abstraction should be used I think. Maybe we can remove the funtion and line number from all warning logs instead as it makes logs quite unreadable

modules/EvseV2G/EvseV2G.cpp
```diff
@@ -16,11 +17,13 @@ void log_handler(openssl::log_level_t level, const std::string& str) {
...
         break;
     case openssl::log_level_t::error:
     default:
-        EVLOG_error << str;
+        // avoid logging function and line number
+        BOOST_LOG_SEV(::global_logger::get(), ::Everest::Logging::error) << str;
```
corneliusclaussen, line 26: see other comment about direct access to boost log

modules/EvseV2G/EvseV2G.cpp
james-ctc: I don't have a good solution.
We should use a consistent logging approach.
I was asked to make the TLS server logging configurable rather than use the standard approach.

EVLOG_warning etc. all have a memory leak which makes use of valgrind/memcheck problematic (especially for unit tests).

With EVLOG_warning ... they will always show EvseV2G.cpp and that line number despite the error occurring elsewhere (staging/tls ...). 

The file & line number have been very useful in logs in the past to help track down an issue. But showing the wrong file and line isn't useful.

I will revert to EVLOG_warning << str; and EVLOG_error << str; since that is the least worst option.

modules/EvseV2G/EvseV2G.cpp
james-ctc: see other comment's response

modules/EvseV2G/EvseV2G.cpp
corneliusclaussen: Maybe we can change it such that info and warning don't output line numbers. Critical and error maybe still do. It would be nice to have a warning level that is not as verbose as it is now

james-ctc: added a note to liblog so that removing file & line can be considered.
https://github.com/EVerest/liblog/issues/22

modules/EvseV2G/connection/tls_connection.cpp
```diff
@@ -132,17 +154,18 @@ bool build_config(tls::Server::config_t& config, struct v2g_context* ctx) {
             // workaround (see above libevse-security comment)
             const auto key_password = info.password.value_or("");
 
-            config.certificate_chain_file = cert_path.c_str();
-            config.private_key_file = key_path.c_str();
-            config.private_key_password = key_password.c_str();
+            config.chains.emplace_back();
+            config.chains[0].certificate_chain_file = cert_path.c_str();
+            config.chains[0].private_key_file = key_path.c_str();
+            config.chains[0].private_key_password = key_password.c_str();
```
SebaLukas, line 160: You can use an reference `auto&` instead of `config.chains[0]`.
For example:
```
auto& config_chain_0 = config.chains.emplace_back();
config_chain_0.certificate_chain_file = cert_path.c_str();
[...]
```

SebaLukas (approved):
The changes in the EvseV2G module look good :+1: I just have one small comment.

Pietfried:
Please wait for a review from @AssemblyJohn before merging

modules/EvseV2G/connection/tls_connection.cpp
james-ctc: done

AssemblyJohn (approved):

james-ctc:
> Please wait for a review from @AssemblyJohn before merging

@Pietfried review complete, awaiting code owner review/approval.

lib/staging/tls/extensions/helpers.cpp
```diff
@@ -0,0 +1,51 @@
...
+    }
+    return out;
+}
+
+std::ostream& operator<<(std::ostream& out, const openssl::DER& obj) {
+    const auto sav = out.flags();
+    const auto* ptr = obj.get();
+    for (std::size_t i = 0; i < obj.size(); i++) {
+        out << std::setw(2) << std::setfill('0') << std::hex << static_cast<std::uint32_t>(*ptr++);
+    }
```
barsnick, line 48: Nit: I think you should set the stream properties only once, not for every element. (I'm not sure whether the compiler is capable of optimizing this, godbolt says it isn't):
```suggestion
    out << std::setw(2) << std::setfill('0') << std::hex;
    for (std::size_t i = 0; i < obj.size(); i++) {
        out << static_cast<std::uint32_t>(*ptr++);
    }
```
Same for other `operator<<()`.

james-ctc: I tried that change and it doesn't work.
(the code is only used in unit tests anyway)

Before
```
[ RUN      ] DER.equal
010203040506070809
[       OK ] DER.equal (0 ms)
```
After
```
[ RUN      ] DER.equal
0123456789
[       OK ] DER.equal (0 ms)
```

Pietfried (approved):

PR #780, opened 2024-07-15
Add start_precharge variable to iso15118_charger interface

Yuanzjls:
feat: add start_precharge to iso15118_charger, so other modules such as dc power supply can know what stage it is at.

### DIFF
--- a/interfaces/ISO15118_charger.yaml
+++ b/interfaces/ISO15118_charger.yaml
@@ -193,6 +193,9 @@ vars:
   Start_CableCheck:
     description: The charger should now start a cable check
     type: "null"
+  Start_PreCharge:
+    description: The charger should now start a pre charge
+    type: "null"
   DC_Open_Contactor:
     description: The contactor should be opened
     type: "null"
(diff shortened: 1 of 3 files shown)


SebaLukas:
Hi :)

Thank you very much for your contribution.

Is there an exact reason why you want to introduce the `Start_PreCharge` var?

With your contribution the var is published by the EvseV2G module but the EvseManager does not subscribe to it and generally does not react to `Start_PreCharge`.

The Power Supply already knows when the PreCharge starts, because the `DC_EVTargetVoltageCurrent` var is only published for the first time in PreCharge and forwarded to the PowerSupply module by the EvseManager.

Yuanzjls:
Hi Sebastian, 

Thank you for the reply.

We are using a dc power supply that is designed by ourself, and on different stages (cable check, precharge, charging), it has to set itself to different mode, so that's why we want to dc power supply to know the stages explicitly.

However, current everest doesn't have an interface to tell power supply what stage it is at. So I am wondering, can we add a cmd to dc power supply interface to tell what stage it is, and our dc power supply can change its mode based on the command.

corneliusclaussen:
I think we should add it to the set mode in DC power supply interface. I can look into it on Monday. In general, there are potentially several power supplies with that requirement.

corneliusclaussen:
@Yuanzjls please check out this PR:

https://github.com/EVerest/everest-core/pull/803

It adds the pre charge information in a similar way to this PR, but transfers this information all the way to the power_supply_DC interface so that it is available when the setMode command is received by the driver. If you are ok, we can close this PR and merge 803 instead.

Yuanzjls:
It looks good to me 👍 @corneliusclaussen

PR #784, opened 2024-07-16
feat: use hex digits for SSID

james-ctc:
feat: added SSID string to hex and back functions

## Describe your changes
Over the everest-api you can request a list of visible wifi networks and then connect.
SSIDs can contain non-printable characters (as well as ASCII/UTF-8) these are escaped (\xnn) in the SSID list.
The problem was that when attempting to connect the escaped characters were not being accepted.
The solution is to pass the SSID to wpa_supplicant as a string of hex digits.


modules/Setup/WiFiSetup.cpp
```diff
@@ -217,7 +261,11 @@ bool WpaCliSetup::set_network(const std::string& interface, int network_id, cons
     }
 
     auto network_id_string = std::to_string(network_id);
-    auto ssid_parameter = "\"" + ssid + "\"";
+    // auto ssid_parameter = "\"" + ssid + "\"";
```
hikinggrass, line 264: remove

modules/Setup/WiFiSetup.cpp
```diff
@@ -379,4 +427,151 @@ bool WpaCliSetup::is_wifi_interface(const std::string& interface) {
...
+        output = static_cast<int>(c);
+    }
+    return output;
+}
+
+int Ssid::backslash(std::uint8_t c) {
+    int output{static_cast<int>(c)};
+    handler = &Ssid::standard;
+    switch (c) {
+    case '"':
```
hikinggrass, line 444: use `[[fallthrough]];` if that's intended

hikinggrass (changes requested):

modules/Setup/WiFiSetup.cpp
james-ctc: That doesn't appear to be a standard approach.
e.g. https://github.com/EVerest/everest-core/blob/e4b30fac7defcce81b3de9b944de7b9be535a77d/modules/EvseManager/Charger.cpp#L775
Doesn't do this:
```
    switch (cp_event) {
    case CPEvent::CarPluggedIn:
        [[fallthrough]];
    case CPEvent::CarRequestedPower:
        [[fallthrough]];
    case CPEvent::CarRequestedStopPower:
        [[fallthrough]];
    case CPEvent::CarUnplugged:
        [[fallthrough]];
    case CPEvent::BCDtoEF:
        [[fallthrough]];
```
The example in 
https://en.cppreference.com/w/cpp/language/attributes/fallthrough
uses [[fallthrough]]; between case 2 and 3 but not between case 1 and 2.

So I don't think adding [[fallthrough]]; is needed here: 
```
    case '"':
    case '\\':
        break;
```

modules/Setup/WiFiSetup.cpp
james-ctc: removed and rebased to latest main

modules/Setup/WiFiSetup.cpp
hikinggrass: You're right :+1:

hikinggrass (approved):

PR #790, opened 2024-07-19
Raise a CommunicationFault if we can't communicate with the LEM modul…

florinmihut:
…e and clear it once the connection is re-established

### DIFF
--- a/modules/LemDCBM400600/main/powermeterImpl.cpp
+++ b/modules/LemDCBM400600/main/powermeterImpl.cpp
@@ -7,6 +7,12 @@
 #include <chrono>
 #include <string>
 #include <thread>
+#include <utils/error/error_factory.hpp>
+#include <utils/error/error_state_monitor.hpp>
+
+using Error = Everest::error::Error;
+using Condition = Everest::error::ErrorStateMonitor::StateCondition;
+
 
 namespace module::main {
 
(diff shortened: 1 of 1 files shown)


modules/LemDCBM400600/main/powermeterImpl.cpp
```diff
@@ -35,15 +40,26 @@ void powermeterImpl::ready() {
     // Start the live_measure_publisher thread, which periodically publishes the live measurements of the device
     this->live_measure_publisher_thread = std::thread([this] {
         this->publish_public_key_ocmf(this->controller->get_public_key_ocmf());
+        const Condition condition = Condition("powermeter/CommunicationFault", "Communication timed out", true);
```
corneliusclaussen, line 43: This looks a bit over complicated, maybe all you need is error_state_monitor->is_error_active() below?

florinmihut: You're right, is_condition_satisfied calls is_error_active when used with one condition.
Fixed!

modules/LemDCBM400600/main/powermeterImpl.cpp
```diff
@@ -7,6 +7,11 @@
 #include <chrono>
 #include <string>
 #include <thread>
+#include <utils/error/error_factory.hpp>
+#include <utils/error/error_state_monitor.hpp>
+
+using Error = Everest::error::Error;
+using Condition = Everest::error::ErrorStateMonitor::StateCondition;
```
corneliusclaussen, line 14: those two usings can be removed

modules/LemDCBM400600/main/powermeterImpl.cpp
```diff
@@ -7,6 +7,11 @@
 #include <chrono>
 #include <string>
 #include <thread>
+#include <utils/error/error_factory.hpp>
+#include <utils/error/error_state_monitor.hpp>
```
corneliusclaussen, line 11: aren't they automatically included?

modules/LemDCBM400600/main/powermeterImpl.cpp
```diff
@@ -39,11 +44,23 @@ void powermeterImpl::ready() {
...
+                }
             } catch (LemDCBM400600Controller::DCBMUnexpectedResponseException& dcbm_exception) {
                 EVLOG_error << "Failed to publish powermeter value due to an invalid device response: "
                             << dcbm_exception.what();
             } catch (HttpClientError& client_error) {
                 EVLOG_error << "Failed to publish powermeter value due to an http error: " << client_error.what();
+                if (!this->error_state_monitor->is_error_active("powermeter/CommunicationFault",
+                                                                "Communication timed out")) {
+                    Error error =
+                        this->error_factory->create_error("powermeter/CommunicationFault", "Communication timed out",
```
corneliusclaussen, line 60: you can put it in one line raise(factory->create) or use auto instead of the Error type

corneliusclaussen (changes requested):

modules/LemDCBM400600/main/powermeterImpl.cpp
florinmihut: right again, fixed

modules/LemDCBM400600/main/powermeterImpl.cpp
florinmihut: right again, fixed

corneliusclaussen (approved):

PR #793, opened 2024-07-24
added set charging profile callback in OCPP201

Pietfried:

### DIFF
--- a/modules/OCPP201/OCPP201.cpp
+++ b/modules/OCPP201/OCPP201.cpp
@@ -525,6 +525,7 @@ void OCPP201::ready() {
     };
 
     callbacks.set_charging_profiles_callback = [this]() {
+        // TODO: implement once charging profiles are available in libocpp
         return;
     };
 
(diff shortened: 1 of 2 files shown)


modules/OCPP201/OCPP201.cpp
```diff
@@ -524,6 +524,10 @@ void OCPP201::ready() {
         this->p_ocpp_generic->publish_security_event(event);
     };
 
+    callbacks.set_charging_profiles_callback = [this]() {
+        return;
```
maaikez, line 528: Euh we don't have to do anything with it? Or is it a todo which will be done in another ticket? If so, add ticket number?

maaikez (approved):

modules/OCPP201/OCPP201.cpp
Pietfried: The charging profile(s) cannot yet be retrieved from libocpp, but this callback is now required since https://github.com/EVerest/libocpp/commit/01f064f45f6d0ff44605db07373e4b9ea02b63b8#diff-dfc83f0a09179c3413c08a198f22698f37ee784eb8755743d6cac96192edc982

hikinggrass (approved):

PR #799, opened 2024-07-26
Adding Ocpp Configuration Module

MarzellT:


modules/Cargo.lock
a-w50, line 1: This file shouldn't change due to this commit

modules/OCPPConfiguration/main/util.hpp
```diff
@@ -0,0 +1,16 @@
...
+
+#include <c4/yml/std/map.hpp>
+#include <ryml.hpp>
+
+namespace module {
+class Util {
+public:
+    static ryml::Tree load_yaml_file(const std::string& file_path);
+    static void save_yaml_file(const std::string& file_path, ryml::Tree& tree);
+};
```
a-w50, line 14: classes shouldn't be used for namespacing,  just do
```
namespace module::util {
  ryml::Tree load_yaml_file(..);
  void save_yaml_file(..);
}
```
I probably also wouldn't expose all the ryml stuff.  Rather have a container or opaque data type for that.

modules/OCPPConfiguration/main/mapping_reader.cpp
```diff
@@ -0,0 +1,31 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Pionix GmbH and Contributors to EVerest
+
+#include "mapping_reader.hpp"
+#include "util.hpp"
+#include <iostream>
```
a-w50, line 6: put includes into sections

modules/OCPPConfiguration/main/event_handler.hpp
```diff
@@ -0,0 +1,27 @@
...
+class EventHandler {
+public:
+    EventHandler(const std::string& config_mapping_path, const std::string& user_config_path);
+    void handleEvent(const types::ocpp::EventData& event_data);
+
+private:
+    using EventMap = MappingReader::ConfigMap;
+    using EverestConfigPair = MappingReader::EverestConfigPair;
+
+    const EventHandler::EverestConfigPair find_event_in_map(const std::string& event_name) const;
```
a-w50, line 21: what if the event can't be found?  What will the pair look like?

modules/Cargo.lock
MarzellT: That's true. I'm honestly not sure why it would have changed. I'll have a look

modules/OCPPConfiguration/main/util.hpp
MarzellT: True. I changed that

modules/OCPPConfiguration/main/event_handler.hpp
MarzellT: I think it should either throw an exception or just ignore it. It depends on the configuration of the module whether a variable should be in that map or not

modules/OCPPConfiguration/OCPPConfiguration.cpp
```diff
@@ -0,0 +1,49 @@
...
+    // Split the monitor variables string by comma
+    auto component_variables = std::vector<types::ocpp::ComponentVariable>();
+    auto ss = std::stringstream{monitor_variables_string};
+    auto variable_name = std::string{};
+
+    while (std::getline(ss, variable_name, ',')) {
+        auto component_variable = types::ocpp::ComponentVariable{};
+        component_variable.variable.name = variable_name;
+        component_variables.push_back(component_variable);
+    }
```
Pietfried, line 44: This doenst work for OCPP2.0.1, because the component is not considered, only the variable.

modules/CMakeLists.txt
```diff
@@ -37,8 +38,8 @@ ev_add_module(PhyVersoBSP)
 add_subdirectory(examples)
 add_subdirectory(simulation)
 
-if(${EVEREST_ENABLE_RS_SUPPORT})
+if (${EVEREST_ENABLE_RS_SUPPORT})
```
Pietfried: remove space

modules/CMakeLists.txt
```diff
@@ -37,8 +38,8 @@ ev_add_module(PhyVersoBSP)
...
     ev_add_module(RsIskraMeter)
     ev_add_module(RsPaymentTerminal)
     add_subdirectory(rust_examples)
-endif()
+endif ()
```
Pietfried: remove space

modules/OCPPConfiguration/manifest.yaml
```diff
@@ -0,0 +1,33 @@
...
+    default: >
+      AllowOfflineTxForUnknownId,AuthorizationCacheEnabled,AuthorizeRemoteTxRequests,
+      BlinkRepeat,ClockAlignedDataInterval,ConnectionTimeOut,ConnectorPhaseRotation,
+      HeartbeatInterval,LightIntensity,LocalAuthorizeOffline,LocalPreAuthorize,
+      MaxEnergyOnInvalidId,MeterValuesAlignedData,MeterValuesSampledData,
+      MeterValueSampleInterval,MinimumStatusDuration,ResetRetries,
+      StopTransactionOnEVSideDisconnect,StopTransactionOnInvalidId,StopTxnAlignedData,
+      StopTxnSampledData,TransactionMessageAttempts,TransactionMessageRetryInterval,
+      UnlockConnectorOnEVSideDisconnect,WebsocketPingInterval
+  user_config_path:
```
Pietfried, line 15: @hikinggrass : Can the user-config path be accessed from within a module?

modules/OCPPConfiguration/config-mapping.yaml
```diff
@@ -0,0 +1,312 @@
+- component:
```
Pietfried, line 1: This file should serve as an example. 
* Reduce entries in this file and rename it to `example_mapping.yaml`
* add reasonable example for OCPP1.6 (e.g. ConnectionTimeOut maps to module_id: auth, config_param: connection_timeout)
* add reasonable example for OCPP2.0.1 (e.g. Component=TxCtrlr, Variable=EVConnectionTimeOut maps to module_id: auth, config_param: connection_timeout

modules/OCPPConfiguration/OCPPConfiguration.hpp
```diff
@@ -0,0 +1,71 @@
...
+    // ev@4714b2ab-a24f-4b95-ab81-36439e1478de:v1
+
+private:
+    friend class LdEverest;
+    void init();
+    void ready();
+
+    // ev@211cfdbe-f69a-4cd6-a4ec-f8aaa3d1b6c8:v1
+    // insert your private definitions here
+    std::vector<types::ocpp::ComponentVariable> parse_config_monitor_variables();
```
Pietfried, line 59: as mentioned in the manifest you don't need this function but you can read the variables to monitor from the mapping

modules/OCPPConfiguration/OCPPConfiguration.cpp
```diff
@@ -0,0 +1,49 @@
...
+    invoke_init(*p_main);
+}
+
+void OCPPConfiguration::ready() {
+    invoke_ready(*p_main);
+
+    const auto mapping_file_path = std::filesystem::path{config.mapping_file_path};
+    event_handler = std::make_unique<EventHandler>(mapping_file_path);
+
+    r_ocpp_module->call_monitor_variables(parse_config_monitor_variables());
```
Pietfried, line 19: use the variables from the mapping

modules/OCPPConfiguration/main/event_handler.hpp
```diff
@@ -0,0 +1,27 @@
...
+// Copyright Pionix GmbH and Contributors to EVerest
+
+#pragma once
+
+#include "generated/types/ocpp.hpp"
+#include "mapping_reader.hpp"
+
+namespace module {
+
+class EventHandler {
```
Pietfried, line 11: Please add a brief description what this class does

modules/OCPPConfiguration/main/event_handler.hpp
```diff
@@ -0,0 +1,27 @@
...
+public:
+    EventHandler(const std::filesystem::path& config_mapping_path);
+    void handle_event(const types::ocpp::EventData& event_data, const std::string& user_config_path_string);
+
+private:
+    const std::optional<EverestConfigMapping>
+    find_mapping_by_component_variable_or_log_error(const types::ocpp::ComponentVariable& component_variable) const;
```
Pietfried, line 18: It should be considered normal to receive event data for variables that are not part of the mapping, so I wouldnt throw a `runtime_error` exception here but simply do nothing in case nothing can be found in the mapping. I think its enough to only have this function:

`const std::optional<EverestConfigMapping>
    find_mapping_by_component_variable(const types::ocpp::ComponentVariable& component_variable) const;`

modules/OCPPConfiguration/OCPPConfiguration.cpp
```diff
@@ -0,0 +1,49 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Pionix GmbH and Contributors to EVerest
+#include "OCPPConfiguration.hpp"
+#include "main/event_handler.hpp"
+#include <everest/logging.hpp>
+
+namespace module {
+
+void OCPPConfiguration::init() {
+    invoke_init(*p_main);
```
Pietfried, line 10: You should verify the configuration parameters in the init:
* Check if mapping file exists and is valid and warn if it doesn't
* Check if user config exists and warn if it doesn't

In case the configuration could not be verified this module is not operational, but it should not crash, since it doesn't provide any interface and shold not cause a crash of everest ever.

modules/OCPPConfiguration/config-mapping.yaml
Pietfried, line 1: Please add a yaml schema file to validate the format of the mapping file

Pietfried (changes requested):

modules/OCPPConfiguration/main/mapping_reader.cpp
```diff
@@ -0,0 +1,111 @@
...
+            : std::nullopt;
+    return {std::move(variable_name), std::move(instance)};
+}
+
+std::optional<types::ocpp::EVSE> MappingReader::parse_evse_node(const c4::yml::NodeRef node) {
+    if (!node.has_child("id")) {
+        throw std::runtime_error("EVSE node must have an id");
+    }
+    const auto id_val = node["id"].val();
+    auto id = std::stoi(std::string{id_val.str, id_val.len});
```
Pietfried, line 84: This should catch `std::invalid_argument`

modules/OCPPConfiguration/main/mapping_reader.cpp
```diff
@@ -0,0 +1,111 @@
...
+
+    const auto connector_id_optional_val = node.has_child("id") ? std::optional{node["id"].val()} : std::nullopt;
+    auto connector_id = connector_id_optional_val.has_value()
+                            ? std::optional{std::stoi(std::string{connector_id_optional_val.value().str,
```
Pietfried, line 88: This should catch `std::invalid_argument`

modules/OCPPConfiguration/main/mapping_reader.hpp
```diff
@@ -0,0 +1,37 @@
...
+
+namespace module {
+
+using types::ocpp::Component;
+using types::ocpp::ComponentVariable;
+using types::ocpp::Variable;
+
+using OcppToEverestConfigMapping = std::unordered_map<ComponentVariable, EverestConfigMapping>;
+
+class MappingReader {
```
Pietfried, line 23: Please add a brief description for this class

modules/OCPPConfiguration/main/mapping_reader.hpp
```diff
@@ -0,0 +1,37 @@
...
+public:
+    static OcppToEverestConfigMapping read_mapping(const std::filesystem::path& file_path);
```
Pietfried, line 25: Please add a brief description for public functions

modules/OCPPConfiguration/main/event_handler.hpp
```diff
@@ -0,0 +1,27 @@
...
+
+#include "generated/types/ocpp.hpp"
+#include "mapping_reader.hpp"
+
+namespace module {
+
+class EventHandler {
+public:
+    EventHandler(const std::filesystem::path& config_mapping_path);
+    void handle_event(const types::ocpp::EventData& event_data, const std::string& user_config_path_string);
```
Pietfried, line 14: Please add a brief description for public functions

modules/OCPPConfiguration/main/util.hpp
```diff
@@ -0,0 +1,26 @@
...
+#include "generated/types/ocpp.hpp"
+#include "mapping_reader.hpp"
+#include "utils/yaml_loader.hpp"
+#include <c4/yml/std/map.hpp>
+#include <filesystem>
+#include <ryml.hpp>
+#include <sstream>
+
+namespace module {
+namespace Util {
```
Pietfried, line 15: Please use lower case for namespace e.g. `util`

modules/OCPPConfiguration/manifest.yaml
hikinggrass: No, you also don't get the path to the original config. Since this is just a temporary solution until the config service in everest-framework is ready it's fine to do it like this

modules/OCPPConfiguration/config-mapping.yaml
```diff
@@ -0,0 +1,312 @@
...
+    instance: 1
+    evse:
+      id: 1
+      connector: 1
+  variable:
+    name: AuthorizationCacheEnabled
+    instance: 1
+  maps_to:
+    module_id: authorization_cache_enabled
+    config_key: authorization_cache_enabled_config
```
Pietfried, line 36: Please separate the definition of one mapping into:
* one level for ocpp definition
* one level for everest definition

modules/OCPPConfiguration/manifest.yaml
```diff
@@ -0,0 +1,33 @@
...
+    description: Path to the user configuration file
+    type: string
+    default: "config/user-config/ocpp-remote-config.yaml"
```
Pietfried, line 18: remove the default

modules/OCPPConfiguration/OCPPConfiguration.cpp
MarzellT: Done

modules/OCPPConfiguration/OCPPConfiguration.cpp
MarzellT: Done by taking the monitor_variables from the mapping

modules/OCPPConfiguration/OCPPConfiguration.hpp
MarzellT: Done

modules/OCPPConfiguration/config-mapping.yaml
MarzellT: Done

modules/OCPPConfiguration/config-mapping.yaml
MarzellT: Done

modules/OCPPConfiguration/main/event_handler.hpp
MarzellT: I decided to keep both functions but change the high level function to not log any error.    
My reasoning was that the high level function checks if the low level function returned a value and if not returns a nullopt and the other function makes the actual lookup.    
This way both functions have just one responsibility and I can still use the raii approach in the caller function of the high level function

modules/OCPPConfiguration/main/util.hpp
MarzellT: Done

modules/OCPPConfiguration/manifest.yaml
```diff
@@ -0,0 +1,33 @@
+description: This module reacts to OCPP configuration changes.
+config:
+  monitor_variables:
```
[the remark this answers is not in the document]
MarzellT: Done

config/config-sil-ocpp.yaml
```diff
@@ -183,7 +189,7 @@ active_modules:
       fuse_limit_A: 40.0
       phase_count: 3
     connections:
-      price_information: []
+      price_information: [ ]
```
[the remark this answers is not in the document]
MarzellT: Done

config/config-sil-ocpp.yaml
```diff
@@ -1,4 +1,10 @@
 active_modules:
+  ocpp-configuration-instance:
+    module: OCPPConfiguration
+    connections:
+      ocpp_module:
+        - module_id: ocpp
+          implementation_id: ocpp_generic
```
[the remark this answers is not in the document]
MarzellT: Done

modules/OCPPConfiguration/manifest.yaml
MarzellT: Done

modules/OCPPConfiguration/main/event_handler.hpp
MarzellT: Done

modules/OCPPConfiguration/main/mapping_reader.hpp
MarzellT: Done

modules/OCPPConfiguration/main/mapping_reader.hpp
MarzellT: Done

modules/OCPPConfiguration/main/event_handler.hpp
MarzellT: Done

SirVer (dismissed):
does not touch Rust or Bazel, so lgtm.

modules/OCPPConfiguration/main/mapping_reader.cpp
MarzellT: Done but I think this should be done automatically by some tool instead of having to do it manually. Isn't this set up somehow already?

modules/OCPPConfiguration/OCPPConfiguration.cpp
MarzellT: The first case is handled implicitly now but I think the second case shouldn't generate a warning and just silently create the file. Do you disagree?

modules/OCPPConfiguration/main/mapping_reader.cpp
```diff
@@ -0,0 +1,111 @@
...
+
+#include "mapping_reader.hpp"
+#include "everest/logging.hpp"
+#include "util.hpp"
+#include <iostream>
+
+namespace module {
+OcppToEverestConfigMapping MappingReader::read_mapping(const std::filesystem::path& file_path) {
+    const auto hacked_file_path =
+        std::filesystem::path{"modules"} / "OCPPConfiguration" / file_path; // TODO: this is very hacky
```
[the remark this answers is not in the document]
MarzellT: I completely changed this and I think the solution is good, so this comment would be irrelevant then

modules/OCPPConfiguration/config-mapping.yaml
MarzellT: Done

modules/OCPPConfiguration/docs/index.rst
```diff
@@ -0,0 +1,23 @@
...
+
+..  The documentation can be written in reStructuredText,
+and will be converted to HTML and PDF by Sphinx.
+This index.rst file is the entry point for the module documentation.
+
+*******************************************
+OCPPConfiguration
+*******************************************
+
+:ref:`Link <everest_modules_OCPPConfiguration>` to the module's reference.
```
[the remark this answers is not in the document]
MarzellT: Done

suda:
Hi @MarzellT, do you need a hand taking it to the finish line? Looks like it's almost done and could be quite useful for the `everest-admin-panel`

Pietfried:
The runtime configuration of EVerest modules will be integrated in a different way as part of the everest-framework in the future.

PR #800, opened 2024-07-29
Command "iso_wait_for_stop" now works as intended in EvManager

SebaLukas:
## Describe your changes
The "iso_wait_for_stop" implementation was not correct. 

`r_ev[0]->call_stop_charging();` was not called

### DIFF
--- a/modules/EvManager/main/car_simulation.cpp
+++ b/modules/EvManager/main/car_simulation.cpp
@@ -94,8 +94,8 @@ bool CarSimulation::sleep(const CmdArguments& arguments, size_t loop_interval_ms
         const auto sleep_time_ms = sleep_time * 1000;
         sim_data.sleep_ticks_left = static_cast<long long>(sleep_time_ms / loop_interval_ms) + 1;
     }
-    sim_data.sleep_ticks_left = sim_data.sleep_ticks_left.value() - 1;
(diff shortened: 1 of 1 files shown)


modules/EvManager/main/car_simulation.cpp
```diff
@@ -234,21 +234,23 @@ bool CarSimulation::iso_stop_charging(const CmdArguments& arguments) {
 }
 
 bool CarSimulation::iso_wait_for_stop(const CmdArguments& arguments, size_t loop_interval_ms) {
-    if (!sim_data.sleep_ticks_left.has_value()) {
-        sim_data.sleep_ticks_left =
-            std::stoll(arguments[0]) * static_cast<long>(1 / static_cast<float>(loop_interval_ms)) + 1;
+    if (not sim_data.sleep_ticks_left.has_value()) {
```
corneliusclaussen, line 237: Not introduced in this PR, but std::stoll et al are used without catching exceptions here and in other places in EVManager. We should add try catch everywhere probably in a separate PR.

corneliusclaussen (approved):

MarzellT (approved):
Looks good to me

modules/EvManager/main/car_simulation.cpp
SebaLukas: Good point! I'll plan that.

PR #803, opened 2024-07-30
DC Power supply: Add charging phase information

corneliusclaussen:
## Describe your changes

In some special setups, the DC power supply needs to know if it is switched on for cable check, pre charge or actual charging phase as it may handle that differently internally. Most PSUs should not need to know as they should always operate in CCCV mode independent of the phase.

This PR adds the charging phase information to the setMode command.

It also drops the JsDCSupplySimulator module as it is not used anymore.


SebaLukas (approved):
:+1:

PR #804, opened 2024-08-01
Add plc chip reset to ev_slac

SebaLukas:
## Describe your changes
The reset function from the Evse ported to the Ev.

The Ev Slac module lacks the function to reset the PLC chip after each session

### DIFF
--- a/lib/staging/slac/fsm/ev/include/slac/fsm/ev/context.hpp
+++ b/lib/staging/slac/fsm/ev/include/slac/fsm/ev/context.hpp
@@ -37,6 +37,23 @@ template <> struct MMTYPE<slac::messages::cm_set_key_req> {
     static const uint16_t value = slac::defs::MMTYPE_CM_SET_KEY | slac::defs::MMTYPE_MODE_REQ;
 };
 
+template <> struct MMTYPE<slac::messages::qualcomm::cm_reset_device_req> {
+    static const uint16_t value = slac::defs::qualcomm::MMTYPE_CM_RESET_DEVICE | slac::defs::MMTYPE_MODE_REQ;
+};
+
+template <> struct MMTYPE<slac::messages::qualcomm::cm_reset_device_cnf> {
+    static const uint16_t value = slac::defs::qualcomm::MMTYPE_CM_RESET_DEVICE | slac::defs::MMTYPE_MODE_CNF;
+};
+
+// This message has no CNF counterpart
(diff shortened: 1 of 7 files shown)


SebaLukas:
The test with real HW is missing

corneliusclaussen:
check out this draft PR: 

https://github.com/EVerest/everest-core/pull/809

It is not ready for merging yet because of a hardcoded 10s delay after reset, but it changes the sections in EvseSlac that you copied over to EvSlac in this PR. Maybe you can change it to be more similar to #809, then it will work with both QCA and Lumissil.

SebaLukas:
Thx for the suggestion. I will have a look :+1:

SebaLukas:
I took a closer look at #809. As I understand it, I'm adding a 10s delay after the Lumissil reset. Until we have found a better method for the Lumissil.
The new NMK state does not need the EV, as the network is only joined in the matched state.

lategoodbye:
Hi, could you please give a short motivation of this change?

There are several methods to trigger a reset of the PLC. The chose one here has the best independence from hardware & kernel, but it's also the most unreliable way to do because it needs a working firmware.

SebaLukas:
Hi :)

As the QCA sometimes has problems joining the network, it always needs a soft reset after each charging session. This is solved using this method. As far as I know, we cannot reset the QCA via the HW pin on Linux.

On the Evse side, the reset is implemented in the same way and has worked very well since it was added. That's why I simply ported it to the EV side.

What would the other methods look like?

corneliusclaussen:
I think we should add reset via GPIO as well (in a separate PR) as an option for boards that have this pin connected to the Linux board (also for the EVSE side)

SebaLukas:
This is on hold because the feature getting the info which chips is installed is missing

Pietfried:
@SebaLukas do you plan to pick this up again?

SebaLukas:
> @SebaLukas do you plan to pick this up again?

Yeah. I'll continue here if it's really needed. At the moment it has a low priority. I can also close the PR and open it later. The branch will stay.

Pietfried:
> > @SebaLukas do you plan to pick this up again?
> 
> Yeah. I'll continue here if it's really needed. At the moment it has a low priority. I can also close the PR and open it later. The branch will stay.

Ok, lets close and reopen then :+1:

SebaLukas:
If work continues here, I will open this PR again

PR #814, opened 2024-08-05
Registered time_sync_callback in OCPP201 module

Pietfried:


modules/OCPP201/OCPP201.cpp
```diff
@@ -529,6 +529,10 @@ void OCPP201::ready() {
         return;
     };
 
+    callbacks.time_sync_callback = [this](const ocpp::DateTime& current_time) {
```
a-w50, line 532: Instead of capturing *everything* (i.e. `this`), it is advised to capture only the things that are needed.  So you could probably also write:
```
callbacks.time_sync_callback = [&system = this->r_system](const auto& current_time) {
  system->call_set_system_time(current_time.to_rfc3339());
}
```
But note, that this would not be the same in case `this->r_system` changes during the lifetime of the assigned function.

Pietfried: Right, since this covers almost every other callback definition, I have raised an issue for that: https://github.com/EVerest/everest-core/issues/828 . It therefore won't be covered as part of this PR.

hikinggrass (approved):

hikinggrass (approved):

PR #823, opened 2024-08-08
bug(EvseManager): legacy wakeup

a-w50:
- a previous commit #757 probably introduced a bug
- it should bring the EVSE state machine into state B1 back from B2 after the legacy wakeup routine using state F didn't succeed
- but the code (which checks for the timeout in order start the wakeup) unconditionally runs into the code which transitions back to B1, effectively disabling the legacy wakeup


corneliusclaussen:
Cannot test it right now but the correct sequence would be: if it prepares charging by enabling pwm (B1 to B2) it expects the ev to proceed to C2 to actually start charging. If the ev does not within a timeout, it does b2-F-B2 to try one wakeup. If after another timeout it still does not go to C, it goes to state paused by ev as we can assume the ev just does not want to charge. It is still in b2 as the ev may still start charging anytime. Going to b1 would actually be a pausedbyevse instead, which is not what we want here.

With your change, does it work that way?

Pietfried (approved):

a-w50:
> With your change, does it work that way?

yes

PR #843, opened 2024-09-02
EvseSlac: Fix regression that NMK was not regenerated on reset

corneliusclaussen:
## Describe your changes

A regression had been introduced between 2024.2.0 and 2024.3.0 that removed the re-generation of the NMK in between sessions. This PR re-adds this functionality.

### DIFF
--- a/lib/staging/slac/fsm/evse/src/states/others.cpp
+++ b/lib/staging/slac/fsm/evse/src/states/others.cpp
@@ -30,6 +30,7 @@ static auto create_cm_set_key_req(uint8_t const* session_nmk) {
 
 void ResetState::enter() {
     ctx.log_info("Entered Reset state");
+    ctx.slac_config.generate_nmk();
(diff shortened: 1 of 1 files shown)


corneliusclaussen:
Verfied on real HW that it now generates new NMK keys for each session. Also verfied it only sends one set_key.req at startup and not two.

SebaLukas:
Despite no code change, CM_SET_KEY.REQ is only sent once?

SebaLukas (approved):
:+1:

a-w50 (commented):
Just a note (I know I initially wrote this),  but the `generate_nmk` function should be a free function and the code should look like
`ctx.slac_config.session_nmk = generate_nmk()` this is more explicit.  Unfortunately standard c arrays are not copyable like this, but this can be mitigated by using `std::array` which would also be more *safe*

PR #845, opened 2024-09-03
Fix legacy wakeup for PWM charging

corneliusclaussen:
## Describe your changes

Resuming charging after legacy wakeup sequence did not work since no allow_power_on(true) was sent to the BSP in this case. This PR fixes this issue.


modules/EvseManager/IECStateMachine.cpp
```diff
@@ -372,6 +375,7 @@ void IECStateMachine::set_pwm_F() {
 
 // The higher level state machine in Charger.cpp calls this to indicate it allows contactors to be switched on
 void IECStateMachine::allow_power_on(bool value, types::evse_board_support::Reason reason) {
+    EVLOG_info << "Allow power on: " << value << " Reason: " << reason;
```
hikinggrass, line 378: Is that log message desired here or left over from debugging?

corneliusclaussen: It is desired

SebaLukas (approved):
:+1:

PR #856, opened 2024-09-05
fix: API module; delay sending commands to EvseManager until it is ready

james-ctc:
## Describe your changes

The API module can send commands to EvseManager before the manager has completed initialisation.
When this happens there is a potential for the manager to dereference a null pointer and terminate.

This change holds requests to EvseManagers until they have all responded with `ready`.

### DIFF
--- a/modules/API/API.cpp
+++ b/modules/API/API.cpp
@@ -269,11 +269,15 @@ void API::init() {
     std::vector<std::string> connectors;
     std::string var_connectors = this->api_base + "connectors";
 
+    evse_manager_ready = this->r_evse_manager.size();
+
     for (auto& evse : this->r_evse_manager) {
         auto& session_info = this->info.emplace_back(std::make_unique<SessionInfo>());
         auto& hw_caps = this->hw_capabilities_str.emplace_back("");
         std::string evse_base = this->api_base + evse->module_id;
         connectors.push_back(evse->module_id);
+
+        evse->subscribe_ready([this](bool ready) { this->notify_evse_manager_ready(ready); });
 
         // API variables
         std::string var_base = evse_base + "/var/";
@@ -395,7 +399,7 @@ void API::init() {
         std::string cmd_base = evse_base + "/cmd/";
 
         std::string cmd_enable_disable = cmd_base + "enable_disable";
-        this->mqtt.subscribe(cmd_enable_disable, [&evse](const std::string& data) {
+        this->mqtt.subscribe(cmd_enable_disable, [this, &evse](const std::string& data) {
             auto connector_id = 0;
             types::evse_manager::EnableDisableSource enable_source{types::evse_manager::Enable_source::LocalAPI,
                                                                    types::evse_manager::Enable_state::Enable, 100};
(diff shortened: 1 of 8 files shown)


modules/API/API.hpp
```diff
@@ -192,6 +193,12 @@ class API : public Everest::ModuleBase {
     std::vector<std::thread> api_threads;
     bool running = true;
 
+    std::condition_variable evse_manager_cv;
+    std::mutex evse_manager_mux;
+    std::int8_t evse_manager_ready{0};
```
corneliusclaussen, line 198: According to the manifest, 128 connections can be configured to evse managers. In this case, the evse_manager_ready variable will overflow to -128 and it will be ready before they actually are.

corneliusclaussen (changes requested):

modules/API/API.hpp
james-ctc: good spot - will use std::int16_t instead.

modules/API/API.cpp
```diff
@@ -634,4 +646,23 @@ void API::ready() {
...
+    std::unique_lock lock(evse_manager_mux);
+    evse_manager_cv.wait(lock, [this] { return this->evse_manager_ready <= 0; });
+}
+
+void API::notify_evse_manager_ready(bool ready) {
+    if (ready) {
+        bool notify{false};
+        {
+            std::lock_guard lock(evse_manager_mux);
+            evse_manager_ready--;
```
Pietfried, line 659: These changes rely on the fact that each EvseManager publishes `ready` only once (which is currently given). Instead of using a counter that counts down you could actually check for each EVSE individually (e.g. by using a map).

modules/API/API.hpp
```diff
@@ -192,6 +193,12 @@ class API : public Everest::ModuleBase {
     std::vector<std::thread> api_threads;
     bool running = true;
 
+    std::condition_variable evse_manager_cv;
+    std::mutex evse_manager_mux;
+    std::int16_t evse_manager_ready{0};
```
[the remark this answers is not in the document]
james-ctc: `r_evse_manager.size()` will never equal 0.
```
  evse_manager:
    interface: evse_manager
    min_connections: 1
    max_connections: 128
```
If there are 10 EVSE managers then `r_evse_manager.size() == 10`
`evse_manager_ready`  is set to 10 and decremented when a manager becomes ready.
`evse_manager_ready` being 0 or less means that at least 10 ready messages have been received.

modules/API/API.cpp
```diff
@@ -269,12 +269,16 @@ void API::init() {
     std::vector<std::string> connectors;
     std::string var_connectors = this->api_base + "connectors";
 
+    evse_manager_ready = this->r_evse_manager.size();
```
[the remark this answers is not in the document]
james-ctc: `this->` is commonly needed for lambdas. Elsewhere `this->` should not be added unless is is actually required.

modules/API/API.hpp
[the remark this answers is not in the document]
MarzellT: Alright :) then I must have misunderstood how this variable works

modules/API/API.cpp
james-ctc: @Pietfried updated the approach so that an EVSE manager can be ready multiple times and not cause an issue.

modules/API/API.hpp
james-ctc: @corneliusclaussen approach changed removing the evse_manager_ready counter.

corneliusclaussen (approved):

hikinggrass (approved):

PR #859, opened 2024-09-06
Adding Slac Simulator

MarzellT:
## Describe your changes

This module implements a C++ rewrite for the JsSlacSimulator


modules/SlacSimulator/SlacSimulator.cpp
```diff
@@ -0,0 +1,37 @@
...
+void SlacSimulator::init() {
+    invoke_init(*p_evse);
+    invoke_init(*p_ev);
+}
+
+void SlacSimulator::ready() {
+    invoke_ready(*p_evse);
+    invoke_ready(*p_ev);
+
+    loop_interval_ms = 250;
```
SebaLukas, line 20: Maybe define loop_interval_ms as constexpr

SebaLukas (approved):
Looking good 👍🏻

modules/SlacSimulator/util/state.hpp
```diff
@@ -0,0 +1,20 @@
...
+enum class State : uint8_t {
+    UNMATCHED = 0,
+    MATCHING = 1,
+    MATCHED = 2,
+};
+
+std::string state_to_string(State state);
+
+} // namespace util
+} // namespace module
```
hikinggrass, line 20: missing newline

modules/SlacSimulator/doc.rst
hikinggrass, line 1: might be a good idea to write a few words about what this module does, or to delete these doc files if not

hikinggrass (changes requested):

modules/SlacSimulator/doc.rst
MarzellT: deleted them

modules/SlacSimulator/util/state.hpp
MarzellT: corrected

modules/SlacSimulator/util/state.hpp
```diff
@@ -0,0 +1,18 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Pionix GmbH and Contributors to EVerest
+
+#pragma once
+
+#include <string>
+
```
SebaLukas, line 7: Missing `#include <cstdint>`

SebaLukas (changes requested):
- Without including `cstdint` gcc 13+ does not build the SlacSimulator.
- `config-sil` does not work

modules/SlacSimulator/util/state.hpp
MarzellT: Fixed

SebaLukas (approved):

modules/EvManager/main/car_simulatorImpl.cpp
```diff
@@ -13,14 +13,14 @@ void car_simulatorImpl::init() {
     register_all_commands();
     subscribe_to_variables_on_init();
 
+    car_simulation = std::make_unique<CarSimulation>(mod->r_ev_board_support, mod->r_ev, mod->r_slac);
+
```
SebaLukas, line 17: @hikinggrass I am not sure why you moved this to the init()? Publishes from other modules are handled in the `subscribe_to_variables_on_init()` in line 14.

hikinggrass: Looks like my commit message wasn't as clear as I intended. Yes, the subscriptions all happen in init(), using the car_simulation in the lambdas that is still a nullptr at that time (which isn't a problem during setup of the callbacks), however it can then happen that the EvManager and SlacSimulator receiver their ready() and immediately after a publish on one of these variables happens, before the car_simulation is initialized. I think this wasn't a problem before because the JsSlacSimulator was a bit slower with its first publish than the c++ version.
Moving this to init() ensures that car_simulation cannot be a nullptr at any time in ready(), which lead to a segfault in one of the ci runs (that's how I spotted it)

SebaLukas: Ah I see :+1: Yes, then this should definitely be moved to init()

hikinggrass (approved):

modules/SlacSimulator/manifest.yaml
```diff
@@ -0,0 +1,32 @@
+description: SIL Implementation of SLAC data link negotiation according to ISO15118-3.
+provides:
+  evse:
+    interface: slac
+    description: SLAC interface implementation for EVSE side
+    config:
+      evse_id:
```
corneliusclaussen, line 7: The config options for evse_id, nid should be removed, they are not used and not present in the real slac module. The number_of_sounds exists in the real slac module, but should be removed here as well I think as it is not used in this module.

modules/SlacSimulator/manifest.yaml
```diff
@@ -0,0 +1,32 @@
...
+        type: string
+        default: pionix!
+      number_of_sounds:
+        description: SLAC number of sounds.
+        type: integer
+        default: 10
+  ev:
+    interface: ev_slac
+    description: SLAC interface implementation for EV side
+    config:
```
corneliusclaussen, line 22: config option should be removed

corneliusclaussen (changes requested):

corneliusclaussen (approved):

PR #860, opened 2024-09-09
Bump cmake version to 3.16

corneliusclaussen:
## Describe your changes

cmake 3.14.7's find python defaults to find the highest version instead of the first found on the path. In cross compile environments this means it may use the hosts python instead of the toolchain's sysroot version.


a-w50 (approved):
Probably we should create an issue within everest-cmake, that this should also be updated for other repositories

hikinggrass (approved):

PR #862, opened 2024-09-10
Adding yeti simulator

MarzellT:


modules/simulation/JsYetiSimulator/index.js
```diff
@@ -1243,7 +1243,7 @@ boot_module(async ({
 
   setup.provides.board_support.register.enable((mod, args) => {
     if (args.value) {
-      if (mod.current_state === STATE_DISABLED) mod.state = STATE_A;
+      if (mod.current_state === STATE_DISABLED) mod.state = STATE_A;  // TODO: is this a bug?
```
corneliusclaussen, line 1246: state F would be better here

corneliusclaussen, line 1246: Thats ok

corneliusclaussen (approved):

SebaLukas (changes requested):
I am divided on the PR. I think that the code basically works, but I find the architecture confusing.

I think that the complete code from the `simulation.hpp` can also be moved and executed in the `YetiSimulator.cpp`.
This saves the many get functions and the module structure would also be like other EVerest modules. And I think that the clarity will not suffer.

I also don't like the large `ModuleState` struct. This could be reduced by moving the code to YetiSimulator.cpp and define the individual variables and structs, such as `PowermeterData` and `SimulationData`directly in the `YetiSimulator` class.

modules/YetiSimulator/board_support/evse_board_supportImpl.cpp
```diff
@@ -0,0 +1,121 @@
...
+        .max_current_A_import = 32.0,
+        .min_current_A_import = 6.0,
+        .max_phase_count_import = 3,
+        .min_phase_count_import = 1,
+        .max_current_A_export = 16.0,
+        .min_current_A_export = 0.0,
+        .max_phase_count_export = 3,
+        .min_phase_count_export = 1,
+        .supports_changing_phases_during_charging = true,
+        .connector_type = types::evse_board_support::Connector_type::IEC62196Type2Cable};
```
barsnick, line 23: Designated initializers are only supported in C++20 and above.

(Also below with `return {.ampacity = Ampacity::None};`.)

a-w50 (dismissed):
See my comments in https://github.com/EVerest/everest-core/pull/887#discussion_r1797430249. Not sure why this is another PR.
I created a working example in this branch: https://github.com/EVerest/everest-core/tree/refactor/cleanup_yeti_sim_error

barsnick:
Shouldn't this be in subdir `modules/simulation` like the previous JS module?

SebaLukas (approved):
Looking good :+1:

PR #869, opened 2024-09-18
Fix 0A in EV pause handling

corneliusclaussen:
## Describe your changes

This fixes the handling of a 0 amp current limit in EV pause state. It now proceeds to Waiting for energy (effectivly paused by the evse and not the ev).
Also removes a warning that is shown in at least one case where the behaviour was actually valid.


SebaLukas (approved):
:+1:

PR #880, opened 2024-09-25
OCPP&OCPP201: explicitly handle all cases in switch statements

hikinggrass:
Remove default: statements, turn on compiler error on enum conversion switch-case statements. This ensures that all cases are explicitly handled. This should also ensure that throw statements in the conversions functions are never be reached, they remain to prevent "control reaches end of non-void function" compiler warnings


barsnick:
> Good idea! Looks good.

Same. We should do the same for all code, in the long run.

BTW, it also protects against changes in the enum declarations and missing them in some handling functions.

PR #891, opened 2024-10-06
Add simple EvAPI and extend EvManager with simplistic SoC calculation

hikinggrass:


SebaLukas:
Maybe we can combine the new `ev_manager` interface with the existing `car_simulator` interface. I see no problem with  moving the cmds and var from the `car_simulator` interface to the `ev_manager` interface.

SebaLukas (commented):
In general, things are looking good. See my two comments.

modules/EvManager/main/car_simulation.hpp
```diff
@@ -98,8 +110,17 @@ class CarSimulation {
 
 private:
     SimulationData sim_data;
+    const module::Conf& config;
+    std::chrono::time_point<std::chrono::steady_clock> timepoint_last_update;
+    double charge_current_a{0};
+    double charge_voltage_v{0};
```
MarzellT, line 116: Is there a reason this is a class member? As far as I can see this is only used in one function where it's just copying the value of the config so it could live only there

hikinggrass: changed to use the configured voltage for AC and target voltage for DC

modules/EvManager/main/car_simulation.cpp
```diff
@@ -92,8 +94,59 @@ void CarSimulation::state_machine() {
         sim_data.state = SimState::UNPLUGGED;
         break;
     }
+
+    if (not state_has_changed and
+        (sim_data.state == SimState::CHARGING_REGULATED or sim_data.state == SimState::CHARGING_FIXED or
+         sim_data.state == SimState::ISO_CHARGING_REGULATED)) {
+        simulate_soc();
```
hikinggrass, line 101: Maybe this should already be triggered at startup, right now the ev_info is only published when a charging session is ongoing, but the SoC information is available all the time

SirVer:
I am not sure why I was added into this code review. @hikinggrass This does not seem to touch Rust or Bazel?

hikinggrass:
> I am not sure why I was added into this code review. @hikinggrass This does not seem to touch Rust or Bazel?

Doesn't touch rust or bazel, review request were triggered by accident from a botched merge of main into this branch. I've removed the accidentally added reviewers again

modules/EvManager/main/car_simulation.cpp
SebaLukas: Yes, I think you could do this and only `sim_data.battery_charge_wh` will be increased during charging.

SebaLukas (dismissed):
:+1:

config/config-sil-two-evse-dc.yaml
SebaLukas, line 1: Adding EVAPi for every config-sil yaml file?

SebaLukas (dismissed):
Looking good :+1:

SebaLukas (dismissed):

SebaLukas (approved):

PR #897, opened 2024-10-08
Remove deprecated BSM powermeter module

corneliusclaussen:
## Describe your changes

The PowermeterBSM module does not use the SerialCommHub, so it cannot be used with other modules on the same modbus. It also adds an extra dependency for EVerest to the libmodbus, which causes trouble from time to time because another library with the same name exists.
Support for Eichrecht is broken in the module limiting its use quite a bit. If we still need support for this powermeter it should be re-written, using the the SerialCommHub and a proper Eichrecht implementation.


golovasteek (approved):
Makes sense to me. It was confusing for a while already.

mhei (approved):
Kind of house-keeping, really appreciated.

hikinggrass (approved):

dorezyuk (approved):

SirVer (approved):
I am not sure if this is blocking on me, since my team has already approved, but since this has been sitting for a while just to be sure: lgtm.

PR #898, opened 2024-10-08
Remove external libtimer dependency

corneliusclaussen:
Copy the libtimer header file to lib/staging. Maintaining an external repository for a single header file only used in everest-core seems overkill.

### DIFF
--- a/lib/staging/async_timer/async_timer.hpp
+++ b/lib/staging/async_timer/async_timer.hpp
@@ -1,0 +1,184 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2020 - 2021 Pionix GmbH and Contributors to EVerest
+#ifndef EVEREST_TIMER_HPP
+#define EVEREST_TIMER_HPP
+
+#include <boost/asio.hpp>
+#include <chrono>
+#include <condition_variable>
+#include <date/date.h>
+#include <date/tz.h>
+#include <functional>
+#include <mutex>
+#include <thread>
+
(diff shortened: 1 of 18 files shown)


lib/staging/async_timer/async_timer.hpp
```diff
@@ -0,0 +1,184 @@
...
+#include <chrono>
+#include <condition_variable>
+#include <date/date.h>
+#include <date/tz.h>
+#include <functional>
+#include <mutex>
+#include <thread>
+
+namespace Everest {
+// template <typename TimerClock = date::steady_clock> class Timer {
```
maaikez, line 16: Remove out commented code?

hikinggrass:
I'm not sure about this, it's also used in libocpp and libevse-security, so in total 3 repositories that would have to copy this header now. And when we want to get rid of the threaded implementation as discussed earlier we would have to do this in multiple places now

corneliusclaussen:
> I'm not sure about this, it's also used in libocpp and libevse-security, so in total 3 repositories that would have to copy this header now. And when we want to get rid of the threaded implementation as discussed earlier we would have to do this in multiple places now

Good point, I was not aware that libevse-security/libocpp also uses this. I just think in generel we have too many separate repositories for really small things. We can leave it as is, I'm trying to remove some other dependencies instead ;-)
E.g. libslac should not be separate. It is spread across libslac in a separate repository, /lib/staging/slac in core and the everest modules for Ev and EvseSlac for example.

PR #902, opened 2024-10-09
fix: TLS server certificate_sha_1 calculated incorrect hash

james-ctc:
## Describe your changes

certificate_sha_1() was originally used to calculate a unique value representing a certificate. For trusted_ca_keys the certificate hash needs to be over the whole certificate.

certificate_sha_1() updated to calculate the correct hash. unit tests added to check the calculated hash
python test script to print trusted_ca_keys and certificate hashes

### DIFF
--- a/lib/staging/tls/openssl_util.cpp
+++ b/lib/staging/tls/openssl_util.cpp
@@ -592,6 +592,20 @@ std::string certificate_to_pem(const X509* cert) {
     return result;
 }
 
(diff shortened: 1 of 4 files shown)


AssemblyJohn (approved):
Looks good, I have only one question. Is the hash calculated from the DER or PEM? All certs have to be converted to DER first, right?

james-ctc:
> Is the hash calculated from the DER or PEM?
The hash is calculated of the full DER encoding of the certificate.

PR #904, opened 2024-10-11
Fixed issue when recursively creating Market objects

Pietfried:
## Issue
The Market class previously used a `std::vector<Market> _children` to store child `Market` objects, which were initialized recursively within the `Market` constructor:

```cpp
for (auto& flow_child : _energy_flow_request.children) {
    _children.emplace_back(flow_child, _nominal_ac_voltage, this); // third argument is _parent of the object
}
```

The problem with that is that `std::vector` may need to resize and move its elements to a new memory location when additional elements are added. When this happens, all existing `Market` elements are moved, which invalidates any pointers or references to those elements. In this case, the `_parent` pointer in could end up pointing to an invalid memory location, leading to segmentation faults when accessed.

## Changes
* Switched `_children` from `std::vector` to `std::list`, as it guarantees that its elements remain at stable memory addresses even when new elements are inserted. This ensures that `_parent` pointers remain valid and point to the correct objects
* Removed the `children()` function because it was dead code and no longer needed.

### DIFF
--- a/modules/EnergyManager/EnergyManager.cpp
+++ b/modules/EnergyManager/EnergyManager.cpp
@@ -185,7 +185,7 @@ std::vector<types::energy::EnforcedLimits> EnergyManager::run_optimizer(types::e
     std::vector<types::energy::EnforcedLimits> optimized_values;
     optimized_values.reserve(brokers.size());
 
-    for (auto broker : brokers) {
+    for (auto& broker : brokers) {
         auto& local_market = broker->get_local_market();
         const auto sold_energy = local_market.get_sold_energy();
 
--- a/modules/EnergyManager/EnergyManager.cpp
+++ b/modules/EnergyManager/EnergyManager.cpp
@@ -186,7 +186,7 @@ std::vector<types::energy::EnforcedLimits> EnergyManager::run_optimizer(types::e
     optimized_values.reserve(brokers.size());
 
     for (auto broker : brokers) {
-        auto local_market = broker->get_local_market();
+        auto& local_market = broker->get_local_market();
         const auto sold_energy = local_market.get_sold_energy();
 
         if (sold_energy.size() > 0) {
--- a/modules/EnergyManager/Market.cpp
+++ b/modules/EnergyManager/Market.cpp
@@ -245,12 +245,8 @@ Market::Market(types::energy::EnergyFlowRequest& _energy_flow_request, const flo
 
     // Recursion: create one Market for each child
     for (auto& flow_child : _energy_flow_request.children) {
-        _children.emplace_back(flow_child, _nominal_ac_voltage, this);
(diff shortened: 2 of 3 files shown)


modules/EnergyManager/EnergyManager.cpp
```diff
@@ -186,7 +186,7 @@ std::vector<types::energy::EnforcedLimits> EnergyManager::run_optimizer(types::e
     optimized_values.reserve(brokers.size());
 
     for (auto broker : brokers) {
-        auto local_market = broker->get_local_market();
+        auto& local_market = broker->get_local_market();
```
a-w50, line 189: :+1:

modules/EnergyManager/EnergyManager.cpp
```diff
@@ -186,7 +186,7 @@ std::vector<types::energy::EnforcedLimits> EnergyManager::run_optimizer(types::e
     optimized_values.reserve(brokers.size());
 
     for (auto broker : brokers) {
```
a-w50, line 188: This should also be `auto&`, otherwise you needlessly copy a `shared_ptr` (note that this copy will also take a lock/mutex in order to internally update the ref count)

modules/EnergyManager/Market.cpp
```diff
@@ -245,14 +245,10 @@ Market::Market(types::energy::EnergyFlowRequest& _energy_flow_request, const flo
...
     for (auto& flow_child : _energy_flow_request.children) {
-        _children.emplace_back(flow_child, _nominal_ac_voltage, this);
+        _children.emplace_back(std::make_unique<Market>(flow_child, _nominal_ac_voltage, this));
     }
 }
 
-const std::vector<Market>& Market::children() {
-    return _children;
-}
-
```
a-w50, line 255: Is this dead code?  If so, I would appreciate if this either is mentioned in the commit message or goes to a separate commit

modules/EnergyManager/Market.hpp
```diff
@@ -77,7 +76,7 @@ class Market {
 
 private:
     Market* _parent;
-    std::vector<Market> _children;
+    std::vector<std::unique_ptr<Market>> _children;
```
a-w50, line 79: I'm not sure why you added `std::unique_ptr` here - I would appreciate if you mention this in the commit message (what was the failure and how it got solved).
I guess, the `Market` objects don't behave well, when the vector gets enhanced or references are taken to these elements and become dangling when the vector gets resized.  If this is the case, a simpler solution would be to use `std::list` instead of `std::vector` in combination with `std::unique_ptr`.  `std::list` also guarantees, that it's elements *stay* at the same address, when other elements get added or removed.

Pietfried: Thanks, changed it to `std::list` and added the reasoning in the PR description

a-w50 (approved):
:100:

corneliusclaussen (approved):

Issue #857, opened 2024-09-05
OpenSSL server for ISO 15118 to emit TLS key for connection debugging


james-ctc:
### Describe the problem

There is an option when using MBedTLS to emit the TLS key for use with Wireshark and other tools to analyse TLS TCP logs allowing decryption of the packets.

This feature request is to add the same capability when using OpenSSL (which is now the default provider)

### EVerest Domain

ISO15118

### Affected EVerest Module

EvseV2G

### Describe your solution

provide the TLS key in a similar method as done with MBedTLS

james-ctc:
PR #910 has been approved and merged.

SebaLukas:
#910 is merged

PR #910, opened 2024-10-15
Adding tls key logger for EvseV2G (openssl part)

SebaLukas:
## Describe your changes
Right now only the EvseV2G mbedtls side had the functionality to store the tls session key. Now if you build EvseV2G with openssl, you can store the tls session key, too. For debugging pcap files with a tls connection this key is absolutely necessary.

#857 

### DIFF
--- a/lib/staging/tls/tls.cpp
+++ b/lib/staging/tls/tls.cpp
@@ -14,7 +14,6 @@
 #include <cstdint>
 #include <cstring>
 #include <fcntl.h>
-#include <filesystem>
 #include <fstream>
 #include <memory>
 #include <mutex>
@@ -699,36 +698,41 @@ std::condition_variable ServerConnection::m_cv;
 
 namespace {
 
-std::unique_ptr<std::filesystem::path> keylog_file;
-std::unique_ptr<TlsKeyLoggingServer> keylog_server;
+int ssl_keylog_file_index{-1};
+int ssl_keylog_server_index{-1};
 
 void keylog_callback(const SSL* ssl, const char* line) {
-    std::string key_log_msg = "TLS Handshake keys: " + std::string(line);
+
+    auto keylog_server = static_cast<TlsKeyLoggingServer*>(SSL_get_ex_data(ssl, ssl_keylog_server_index));
+
+    std::string key_log_msg = "TLS Handshake keys on port ";
+    key_log_msg += std::to_string(keylog_server->get_port()) + ": ";
+    key_log_msg += std::string(line);
+
     log_info(key_log_msg);
 
-    if (keylog_server and keylog_server->get_fd() != -1) {
+    if (keylog_server->get_fd() != -1) {
         const auto result = keylog_server->send(line);
         if (result not_eq strlen(line)) {
             log_error("key_logging_server send() failed!");
         }
-
-        keylog_server.reset();
     }
 
-    if (keylog_file) {
+    auto keylog_file_path =
+        static_cast<std::filesystem::path*>(SSL_CTX_get_ex_data(SSL_get_SSL_CTX(ssl), ssl_keylog_file_index));
+
+    if (not keylog_file_path->empty()) {
         std::ofstream ofs;
-        ofs.open(keylog_file->string(), std::ofstream::out | std::ofstream::app);
+        ofs.open(keylog_file_path->string(), std::ofstream::out | std::ofstream::app);
         ofs << line << std::endl;
         ofs.close();
-
-        keylog_file.reset();
     }
 }
 
 } // namespace
 
 ServerConnection::ServerConnection(SslContext* ctx, int soc, const char* ip_in, const char* service_in,
-                                   std::int32_t timeout_ms) :
+                                   std::int32_t timeout_ms, const ConfigItem& tls_key_interface) :
     Connection(ctx, soc, ip_in, service_in, timeout_ms), m_tck_data{m_trusted_ca_keys, m_flags} {
     {
         std::lock_guard lock(m_cv_mutex);
@@ -738,6 +742,12 @@ ServerConnection::ServerConnection(SslContext* ctx, int soc, const char* ip_in, 
         SSL_set_accept_state(m_context->ctx.get());
         ServerStatusRequestV2::set_data(m_context->ctx.get(), &m_flags);
         ServerTrustedCaKeys::set_data(m_context->ctx.get(), &m_tck_data);
+
+        if (tls_key_interface != nullptr) {
+            const auto port = std::stoul(service_in);
+            m_keylog_server = std::make_unique<TlsKeyLoggingServer>(std::string(tls_key_interface), port);
+            SSL_set_ex_data(m_context->ctx.get(), ssl_keylog_server_index, m_keylog_server.get());
+        }
     }
 }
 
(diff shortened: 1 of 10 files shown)


lib/staging/tls/tls.cpp
```diff
@@ -692,6 +697,36 @@ std::uint32_t ServerConnection::m_count{0};
 std::mutex ServerConnection::m_cv_mutex;
 std::condition_variable ServerConnection::m_cv;
 
+namespace {
+
+std::unique_ptr<std::filesystem::path> keylog_file;
+std::unique_ptr<TlsKeyLoggingServer> keylog_server;
+
```
james-ctc, line 704: I think SSL session variables should be used rather than file scoped variables.
There is a chance that keylog_server/keylog_file change before or during the execution of keylog_callback().

CRYPTO_get_ex_new_index(CRYPTO_EX_INDEX_SSL, 0, nullptr, nullptr, nullptr, nullptr)
SSL_set_ex_data(ctx, s_index, ptr)
SSL_get_ex_data

These are used in extensions/trusted_ca_keys.cpp

lib/staging/tls/tls.cpp
```diff
@@ -692,6 +697,36 @@ std::uint32_t ServerConnection::m_count{0};
...
+void keylog_callback(const SSL* ssl, const char* line) {
+    std::string key_log_msg = "TLS Handshake keys: " + std::string(line);
+    log_info(key_log_msg);
+
+    if (keylog_server and keylog_server->get_fd() != -1) {
```
james-ctc, line 709: SSL_get_ex_data could be used here to get the correct keylog_server and keylog_file
The callback could also then handle the case where keylog_server etc. are not available.

lib/staging/tls/tls.cpp
```diff
@@ -1058,6 +1101,12 @@ void Server::wait_for_connection(const ConnectionHandler& handler) {
                 // new connection, pass to handler
                 auto* ip = BIO_ADDR_hostname_string(peer.get(), 1);
                 auto* service = BIO_ADDR_service_string(peer.get(), 1);
+
+                if (m_tls_key_interface) {
+                    const auto port = std::stoul(service);
+                    keylog_server = std::make_unique<TlsKeyLoggingServer>(std::string(m_tls_key_interface), port);
```
james-ctc, line 1107: SSL_set_ex_data could be used here to set the server for this connection.
logging path could also be set
Better to set it in the connection object with the actual SSL object.
(line 740 in ServerConnection constructor)

lib/staging/tls/tls.cpp
```diff
@@ -1341,4 +1390,72 @@ Client::override_t Client::default_overrides() {
     };
 }
 
+TlsKeyLoggingServer::TlsKeyLoggingServer(const std::string& interface_name, uint16_t port) {
```
james-ctc, line 1393: The general approach in this code is to follow MISRA guidelines and only have a single return from a function/method

lib/staging/tls/tls.cpp
james-ctc: Looking at the code the log file is the same for all connections and can be set in the common SSL context.
The log server seems to be created for every connection and could be created as part of the ServerConnection object and a pointer made available to the callback as a SSL session variable. Thus the keylog_server would be consistent for the whole lifetime of the connection.
This would support sending out the new information when a TLS connection updates its keys.

james-ctc (changes requested):

lib/staging/tls/tls.cpp
SebaLukas: With multiple returns, the code is much easier to read. That's why I don't like this MISRA rule and wouldn't like to rewrite the constructor.

james-ctc: I'm used to MISRA code I find it easier to read when there are not multiple returns. It all depends on what you are used to. I don't know what standards we are supposed to work to, however I think mixing approaches in the same file should be avoided. Happy to propose updated code if you would like.

SebaLukas: If you like you can propose some code :+1:  I am working on the remaining points.

lib/staging/tls/tls.cpp
```diff
@@ -885,6 +920,14 @@ bool Server::init_ssl(const config_t& cfg) {
         // use the first server chain
         result = configure_ssl_ctx(ctx, cfg.ciphersuites, cfg.cipher_list, cfg.chains[0], true);
         if (result) {
+
+            if (cfg.tls_key_logging) {
+                const auto file_path = std::filesystem::path(cfg.tls_key_logging_path) /= "tls_session_keys.log";
+                keylog_file = std::make_unique<std::filesystem::path>(file_path);
+                SSL_CTX_set_keylog_callback(ctx, keylog_callback);
+                this->m_tls_key_interface = cfg.host;
```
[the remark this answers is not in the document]
SebaLukas: No, I removed `this`.

lib/staging/tls/tls.cpp
SebaLukas: Switched to `SSL_set_ex_data`, `SSL_get_ex_data`

lib/staging/tls/tls.cpp
SebaLukas: Done

lib/staging/tls/tls.cpp
```diff
@@ -885,6 +920,14 @@ bool Server::init_ssl(const config_t& cfg) {
         // use the first server chain
         result = configure_ssl_ctx(ctx, cfg.ciphersuites, cfg.cipher_list, cfg.chains[0], true);
         if (result) {
+
+            if (cfg.tls_key_logging) {
+                const auto file_path = std::filesystem::path(cfg.tls_key_logging_path) /= "tls_session_keys.log";
```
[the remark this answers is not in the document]
SebaLukas: Done

lib/staging/tls/tls.cpp
```diff
@@ -692,6 +697,36 @@ std::uint32_t ServerConnection::m_count{0};
 std::mutex ServerConnection::m_cv_mutex;
 std::condition_variable ServerConnection::m_cv;
 
+namespace {
+
+std::unique_ptr<std::filesystem::path> keylog_file;
+std::unique_ptr<TlsKeyLoggingServer> keylog_server;
+
+void keylog_callback(const SSL* ssl, const char* line) {
+    std::string key_log_msg = "TLS Handshake keys: " + std::string(line);
```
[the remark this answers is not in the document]
SebaLukas: Good idea :+1: I added that

lib/staging/tls/tls.cpp
SebaLukas: Yeah, your changes work for me :+1:

lib/staging/tls/tls.hpp
```diff
@@ -57,6 +59,28 @@ class ConfigItem {
...
+        return fd;
+    }
+
+    auto get_port() const {
+        return port;
+    }
+
+private:
+    int fd{-1};
+    std::array<uint8_t, 2048> buffer{};
```
james-ctc, line 79: I'm not sure buffer is used any more.

james-ctc (approved):
It looks like TlsKeyLoggingServer::buffer isn't used.
It is a good improvement.
Many thanks.

lib/staging/tls/tls.hpp
SebaLukas: You are right, buffer is not used. I will delete this one

hikinggrass (approved):

PR #919, opened 2024-10-18
feat: add support for new EVSE security API

james-ctc:
## Describe your changes

Now uses call_get_all_valid_certificates_info() to obtain multiple server certificate chains in order to support trusted_ca_keys.

### DIFF
--- a/modules/EvseV2G/connection/tls_connection.cpp
+++ b/modules/EvseV2G/connection/tls_connection.cpp
@@ -3,7 +3,6 @@
 
 #include "tls_connection.hpp"
 #include "connection.hpp"
-#include "everest/logging.hpp"
 #include "log.hpp"
 #include "v2g.hpp"
 #include "v2g_server.hpp"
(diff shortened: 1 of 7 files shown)


modules/EvseV2G/connection/tls_connection.cpp
```diff
@@ -142,30 +141,34 @@ bool build_config(tls::Server::config_t& config, struct v2g_context* ctx) {
...
+            for (const auto& chain : cert_info.info) {
+                const auto cert_path = chain.certificate.value_or("");
+                const auto key_path = chain.key;
+                const auto root_pem = chain.certificate_root.value_or("");
+
+                // workaround (see above libevse-security comment)
+                const auto key_password = chain.password.value_or("");
+
+                auto& ref = config.chains.emplace_back();
+                ref.certificate_chain_file = cert_path.c_str();
```
dorezyuk, line 159: actually i think this will eventually lead to a seg fault

the c_str is the same as data https://en.cppreference.com/w/cpp/string/basic_string/c_str
and data returns the pointer to raw memory of the stirng

if i read it correctly the cert_info and with it the cert_path are going out of scope once this function returns so the pointer you're accessing here will be invalid on function return

james-ctc: certificate_chain_file is a ConfigItem which is a std::optional<std::string>.
`ref.certificate_chain_file = cert_path.c_str();` creates a new string object that is a copy of `cert_path.c_str()`.
ref.certificate_chain_file remains in scope after the function return as the config object is passed in by reference as a parameter.

AssemblyJohn:
The integration looks fine to me, all good!

AssemblyJohn (approved):

SebaLukas (approved):
:+1:

PR #920, opened 2024-10-18
fix: removed duplicate file

james-ctc:
## Describe your changes
EnumFlags.hpp was in two places. Removed duplicate file


SebaLukas (approved):
I have compared the two EnumFlags.hpp. There are a few small differences. But they are so small that it is okay.

SebaLukas:
The bazel build fails because EnumFlags.hpp is missing. So you should change the bazel stuff, too.

dorezyuk:
@james-ctc we can probably help fix bazel but removing it to "fix" cannot a serious idea

modules/EvseManager/CMakeLists.txt
```diff
@@ -21,6 +21,10 @@ target_sources(${MODULE_NAME}
         PersistentStore.cpp
 )
 
+target_include_directories(${MODULE_NAME} PRIVATE
+    ../../lib/staging/util
```
dorezyuk, line 25: why don't we just link against the lib/stating/util?

james-ctc: > why don't we just link against the lib/stating/util?

because there are no object files to link against. EnumFlags.hpp is a header file only.

james-ctc:
> @james-ctc we can probably help fix bazel but removing it to "fix" cannot a serious idea

please feel free to add a fix - I don't know bazel at all.

modules/EvseManager/CMakeLists.txt
james-ctc: @a-w50 thank you for the useful information. I've not heard the term header-only library. I'm still relatively new to cmake (used make for 30+ years) so your help is much appreciated.

a-w50: I will add a commit here

a-w50: @james-ctc I added a commit as a draft on how to include additional libraries (in this case `everest::staging::util`) including the test via cmake.
It is working now, but I'm not really happy with the namespacing/naming so far.  The namespace is quite long.  Maybe we can reduce this a bit by using `ev` instead of `everest` and `staging` could become `draft`.  And once it becomes stable it might go to `ev::std::util` or something similar.  Accordingly the include directory structure would need to be adopted.
For now the commit could be taken as is, but soon we should come up with some naming conventions.

james-ctc: thank you for your help.

lib/staging/util/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
...
+add_library(everest::staging::util ALIAS everest_staging_util)
+
+target_include_directories(everest_staging_util
+    INTERFACE
+        $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
+)
+
+if (BUILD_TESTING)
+    add_subdirectory(tests)
+endif()
```
hikinggrass, line 11: missing newline

lib/staging/util/tests/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
...
+)
+
+target_link_libraries(EnumFlagsTest
+    PRIVATE
+        GTest::gtest_main
+        everest::staging::util
+)
+
+include(GoogleTest)
+gtest_discover_tests(EnumFlagsTest)
```
hikinggrass, line 12: missing newline

modules/EvseManager/tests/CMakeLists.txt
```diff
@@ -6,7 +6,7 @@ add_dependencies(${TEST_TARGET_NAME} ${MODULE_NAME})
 get_target_property(GENERATED_INCLUDE_DIR generate_cpp_files EVEREST_GENERATED_INCLUDE_DIR)
 
 target_include_directories(${TEST_TARGET_NAME} PRIVATE
-    . .. ../../../tests/include
+    . .. ../../../tests/include ../../../lib/util
```
hikinggrass, line 9: is this still needed now that there's the utils lib?

modules/EvseV2G/tests/CMakeLists.txt
```diff
@@ -59,7 +59,7 @@ add_executable(${V2G_MAIN_NAME})
 add_dependencies(${V2G_MAIN_NAME} generate_cpp_files)
 
 target_include_directories(${V2G_MAIN_NAME} PRIVATE
-    . .. ../connection ../../../tests/include ../../../lib/staging/util
+    . .. ../connection ../../../tests/include
```
hikinggrass, line 62: same as above

hikinggrass (approved):

lib/staging/util/include/everest/staging/EnumFlags.hpp
```diff
@@ -7,7 +7,7 @@
 #include <atomic>
 #include <type_traits>
 
-namespace util {
+namespace everest::staging::util {
```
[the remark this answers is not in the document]
a-w50: You're right, for now this would be more consistent.  I'll change that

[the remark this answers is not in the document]
a-w50: @dorezyuk I removed the one line from the EvseManager/bazel file, as this should not be needed there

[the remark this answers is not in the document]
dorezyuk: I see the build works but where do we include it then?

dorezyuk (approved):

PR #923, opened 2024-10-21
Feature/multi root trusted ca

AssemblyJohn:
## Describe your changes

Introduces one interface API call that allows the retrieval of multiple leaf certificate chains, that are linked to different roots.

https://github.com/EVerest/libevse-security/pull/91

### DIFF
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -56,11 +56,11 @@ libcurl:
   cmake_condition: "EVEREST_DEPENDENCY_ENABLED_LIBCURL"
 
 # EvseSecurity
-# This has to appear before libocpp in this file since it is also a direct dependency of libocpp
-# and would otherwise be overwritten by the version used there
+# This has to appear before libocpp in this file since it is also a direct dependency 
+# of libocpp and would otherwise be overwritten by the version used there
 libevse-security:
   git: https://github.com/EVerest/libevse-security.git
-  git_tag: v0.9.0
+  git_tag: v0.9.1
   cmake_condition: "EVEREST_DEPENDENCY_ENABLED_LIBEVSE_SECURITY"
 
 # OCPP
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -60,7 +60,7 @@ libcurl:
 # and would otherwise be overwritten by the version used there
 libevse-security:
   git: https://github.com/EVerest/libevse-security.git
-  git_tag: v0.8.0
+  git_tag: 5c0b6655a49167e8140f08e155fa43c823134897
   cmake_condition: "EVEREST_DEPENDENCY_ENABLED_LIBEVSE_SECURITY"
 
 # OCPP
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -60,7 +60,7 @@ libcurl:
 # and would otherwise be overwritten by the version used there
 libevse-security:
   git: https://github.com/EVerest/libevse-security.git
-  git_tag: 5c0b6655a49167e8140f08e155fa43c823134897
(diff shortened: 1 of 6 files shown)


james-ctc (approved):
looks good to me.
Once libevese-security has the updates then dependencies.yaml will need to be updated.

AssemblyJohn:
Will update the security part based on the comments, make the issues with the remaining fixes (cross, non-self-root), merge and will update the deps here & merge.

dependencies.yaml
```diff
@@ -60,7 +60,7 @@ libcurl:
 # and would otherwise be overwritten by the version used there
 libevse-security:
   git: https://github.com/EVerest/libevse-security.git
-  git_tag: v0.8.0
+  git_tag: 5c0b6655a49167e8140f08e155fa43c823134897
```
corneliusclaussen, line 63: will need to be changed before merge

modules/EvseSecurity/main/evse_securityImpl.cpp
```diff
@@ -121,6 +121,26 @@ evse_securityImpl::handle_get_leaf_certificate_info(types::evse_security::LeafCe
     return response;
 }
 
+types::evse_security::GetCertificateFullInfoResult
+evse_securityImpl::handle_get_all_valid_certificates_info(types::evse_security::LeafCertificateType& certificate_type,
```
corneliusclaussen, line 125: arguments could be const I think

types/evse_security.yaml
```diff
@@ -260,4 +263,22 @@ types:
...
+  GetCertificateFullInfoResult:
+    description: Response to the command get_all_valid_certificates_info
+    type: object
+    required:
+      - status
+      - info
+    properties:
+      status:
+        description: The status of the requested command
+        type: string
```
corneliusclaussen, line 275: Is it really a string?

corneliusclaussen (approved):
Approve with some small comments, please have a look before merging

dependencies.yaml
AssemblyJohn: I have a generic question for this. Do we really need to make a version for each small version release that we have? Libocpp is also a hash: https://github.com/EVerest/everest-core/blob/b11c690bc2158b713a16c5e81be73341a94cfd74/dependencies.yaml#L67

modules/EvseSecurity/main/evse_securityImpl.cpp
AssemblyJohn: By the way, that is auto-generated code. It is the case of all the headers, where they are not passed as const. Any method I can tell the code generator to make those 'const'?

types/evse_security.yaml
AssemblyJohn: Yes, string: https://github.com/EVerest/everest-core/blob/b11c690bc2158b713a16c5e81be73341a94cfd74/types/evse_security.yaml#L87

dependencies.yaml
hikinggrass: Since libocpp updates relatively frequently we only tag close to releases. But I think libevse-security doesn't update quite as often, so I would probably tag it. But this probably would need a version bump in the libevse-security CMakeLists.txt anyway in an additional commit

modules/EvseSecurity/main/evse_securityImpl.cpp
AssemblyJohn: As this doesn't seem clear with the code-gen at the moment, will leave as-is.

AssemblyJohn:
Implemented all suggestions.

Pietfried (approved):

PR #925, opened 2024-10-21
EvseV2G/IsoMux: check if if_name is nullptr

hikinggrass:
This can happen if there are no interfaces with IPv6 addresses and would lead to a segfault later on

### DIFF
--- a/modules/EvseV2G/connection/connection.cpp
+++ b/modules/EvseV2G/connection/connection.cpp
@@ -192,6 +192,10 @@ int check_interface(struct v2g_context* v2g_ctx) {
         v2g_ctx->if_name = choose_first_ipv6_interface();
     }
 
+    if (v2g_ctx->if_name == nullptr) {
+        return -1;
+    }
+
     mreq.ipv6mr_interface = if_nametoindex(v2g_ctx->if_name);
     if (!mreq.ipv6mr_interface) {
         dlog(DLOG_LEVEL_ERROR, "No such interface: %s", v2g_ctx->if_name);
--- a/modules/IsoMux/connection/connection.cpp
+++ b/modules/IsoMux/connection/connection.cpp
@@ -112,6 +112,10 @@ int check_interface(struct v2g_context* v2g_ctx) {
         v2g_ctx->if_name = choose_first_ipv6_interface();
     }
 
+    if (v2g_ctx->if_name == nullptr) {
+        return -1;
+    }
+
     mreq.ipv6mr_interface = if_nametoindex(v2g_ctx->if_name);
     if (!mreq.ipv6mr_interface) {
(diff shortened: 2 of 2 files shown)


james-ctc (approved):
looks good to me

modules/EvseV2G/connection/connection.cpp
```diff
@@ -192,6 +192,10 @@ int check_interface(struct v2g_context* v2g_ctx) {
         v2g_ctx->if_name = choose_first_ipv6_interface();
     }
 
+    if (v2g_ctx->if_name == nullptr) {
```
SebaLukas, line 195: In line 205 there is already a check if `if_name == nulltpr`. So the segfault is happening in line 199 (`mreq.ipv6mr_interface = if_nametoindex(v2g_ctx->if_name);`) ?

modules/IsoMux/connection/connection.cpp
```diff
@@ -112,6 +112,10 @@ int check_interface(struct v2g_context* v2g_ctx) {
         v2g_ctx->if_name = choose_first_ipv6_interface();
     }
 
+    if (v2g_ctx->if_name == nullptr) {
```
SebaLukas, line 115: In line 125 there is already a check if `if_name == nulltpr`. So the segfault is happening in line 119 (`mreq.ipv6mr_interface = if_nametoindex(v2g_ctx->if_name);`) ?

SebaLukas (commented):
See comments. I dont know if these checks are necessary

modules/EvseV2G/connection/connection.cpp
james-ctc: Just tried a simple test program and strcmp doesn't like nullptr as an argument.
There could be a segfault at line 191

james-ctc (changes requested):
strcmp call needs to be protected

modules/EvseV2G/connection/connection.cpp
hikinggrass: Yes the segfault happened in `if_nametoindex()` after `choose_first_ipv6_interface();` returned nullptr. I agree that we should already return early if `v2g_ctx->if_name` is already nullptr in the `strcmp` in line 191

SebaLukas: `v2g_ctx->if_name` can only be a `nullptr` if `choose_first_ipv6_interface()` returns nullptr. Before that it is a string from `config.device`. So strcmp should not be a problem.

hikinggrass: I've added some additional checks at the beginning of the function

james-ctc (approved):
looks good to me

modules/IsoMux/connection/connection.cpp
hikinggrass: exactly

SebaLukas (approved):

PR #927, opened 2024-10-22
Feature/add pn7160 token provider module

cburandt:
Add module PN7160TokenProvider and the required library linux_libnfc-nci.


lib/staging/linux_libnfc-nci/.gitignore
hikinggrass, line 1: I think this library should probably be forked into the EVerest org from https://github.com/NXPNFCLinux/linux_libnfc-nci (with appropriate patches applied, like the added CMakeLists.txt)

cburandt: Sounds like a sensible thing to do.

cburandt: But since linux_libnfc-nci needs to be modified again by the user (choice of SPI vs I²C, selection of GPIO pins in a header) it may not be that straight-forward. Is there an example of similar kind?

hikinggrass: I guess you could either use a header to set this information and include it before including the lib header(s) or work directly with cmake compile definitions. In this project no example comes to mind, for openssl we have a patch file that is stored in everest-core and then applied to the openssl recipe in to to, but that isn't quite the same I suppose

PR #928, opened 2024-10-23
- Fix TC_SECC_CMN_VTB_CmSlacParm_004/5/6

corneliusclaussen:
- Fix TC_SECC_CMN_VTB_CmSlacMatch_003
- Add reset event handling while waiting for Link
- Add link detection to matching state and reset if link got lost

### DIFF
--- a/lib/staging/slac/fsm/evse/include/slac/fsm/evse/context.hpp
+++ b/lib/staging/slac/fsm/evse/include/slac/fsm/evse/context.hpp
@@ -173,6 +173,7 @@ struct EvseSlacConfig {
     struct link_status_struct {
         bool do_detect = false;
         int retry_ms = 100;
+        int poll_in_matched_state_ms = 1000;
         int timeout_ms = 5000;
         bool debug_simulate_failed_matching = false;
     } link_status;
--- a/lib/staging/slac/fsm/evse/src/states/matching.cpp
+++ b/lib/staging/slac/fsm/evse/src/states/matching.cpp
@@ -72,7 +72,8 @@ FSMSimpleState::CallbackReturnType MatchingState::callback() {
 
     if (!seen_slac_parm_req) {
         if (now_tp >= timeout_slac_parm_req) {
-            return Event::RETRY_MATCHING;
+            ctx.log_info("CM_SLAC_PARM_REQ timed out -> FAILED");
+            return Event::FAILED;
         }
 
         call_back_ms = remaining_milliseconds(timeout_slac_parm_req, now_tp);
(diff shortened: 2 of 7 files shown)


SebaLukas (approved):
Looks good :+1:

lib/staging/slac/fsm/evse/src/states/others.cpp
```diff
@@ -163,17 +164,75 @@ FSMSimpleState::HandleEventReturnType IdleState::handle_event(AllocatorType& sa,
...
+    } else if (ev == Event::RESET) {
         return sa.create_simple<ResetState>(ctx);
+    }
+    return sa.PASS_ON;
+}
+
+FSMSimpleState::CallbackReturnType MatchedState::callback() {
+    const auto& cfg = ctx.slac_config;
+
+    if (not ctx.slac_config.link_status.do_detect) {
```
a-w50, line 225: Why not using `cfg` here, already? Could even use `const auto& link_status = ctx.slac_config.link_status`

lib/staging/slac/fsm/evse/src/states/others.cpp
```diff
@@ -163,17 +164,75 @@ FSMSimpleState::HandleEventReturnType IdleState::handle_event(AllocatorType& sa,
...
+        slac::messages::qualcomm::link_status_req link_status_req;
+        ctx.send_slac_message(ctx.slac_config.plc_peer_mac, link_status_req);
+        return true;
+    } else if (ctx.modem_vendor == ModemVendor::Lumissil) {
+        slac::messages::lumissil::nscm_get_d_link_status_req link_status_req;
+        ctx.send_slac_message(ctx.slac_config.plc_peer_mac, link_status_req);
+        return true;
+    } else {
+        return false;
+    }
```
a-w50, line 195: would drop the `else` here.  Just `return false;` at the end of the function

lib/staging/slac/fsm/evse/include/slac/fsm/evse/context.hpp
```diff
@@ -212,6 +213,7 @@ struct Context {
     void log_info(const std::string& text);
 
     ModemVendor modem_vendor{ModemVendor::Unknown};
+    slac::messages::cm_slac_match_cnf match_cnf_message;
```
a-w50, line 216: This doesn't really belong here, as it is shared information only between Matching and WaitForLink state.  Matching state should keep it for itself and pass it to WaitForLink state with `create_simple<WaitForLink>(ctx, std::move(match_cnf_message))`.  You would need to modify the constructor of WaitForLink state.

a-w50 (commented):
Will add these changes by myself

a-w50 (commented):
Style-wise I'm fine.  But it looks like the only necessary change for the 4/5/6 tests is the checking for a valid cm_slac_match_req message.
Everything else seems to be unrelated and I would appreciate if we could add this in a separate commit with corresponding documentation (i.e. what that the WaitForLinkState is for and why it can resent the slac_match_cnf message).

lib/staging/slac/fsm/evse/src/states/matching.cpp
```diff
@@ -72,7 +72,8 @@ FSMSimpleState::CallbackReturnType MatchingState::callback() {
 
     if (!seen_slac_parm_req) {
         if (now_tp >= timeout_slac_parm_req) {
-            return Event::RETRY_MATCHING;
+            ctx.log_info("CM_SLAC_PARM_REQ timed out -> FAILED");
+            return Event::FAILED;
```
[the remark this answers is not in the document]
corneliusclaussen: This is required for the tests to pass

PR #942, opened 2024-10-31
EvseManager: disable the bsp when disabling charger

dorezyuk:


dorezyuk:
I'm not sure if it was on purpose or just forgotten but we should disable the bsp so the locks don't lock when the charger is disabled

Pietfried (approved):

PR #943, opened 2024-10-31
Reservation changes for 2.0.1

maaikez:
## Describe your changes

Implement reservation changes for ocpp 2.0.1 (connector type, reservation for not specific evse)

OCTT tests: 

OCPP 2.0.1

Passed: 
TC_H_01_CS
TC_H_02_CS
TC_H_03_CS
TC_H_04_CS
TC_H_06_CS
TC_H_07_CS
TC_H_09_CS
TC_H_10_CS
TC_H_12_CS
TC_H_13_CS
TC_H_16_CS
TC_H_17_CS
TC_H_18_CS
TC_H_19_CS
TC_H_22_CS
TC_H_23_CS

Failed: 
TC_B_24_CS: After reset, BootNotification should have reason 'RemoteReset', but is 'PowerUp'.
TC_H_08_CS: Did not receive expected status change to 'Available'.
TC_H_14_CS: Two 'global' reservations and 2 evse's, should set them both to 'reserved'.
TC_H_15_CS: Did not receive expected status change to 'Available'. ??? -> Reservation is 'Cancelled', there is no reservation status update sent.
TC_H_21_CS: When evse is set to Unavailable and a reservation is cancelled, evse should go to 'Unavailable' directly, but it first goes to 'Available'.
TC_H_24_CS: Received unexpected request message: ReservationStatusUpdateRequest (Waiting for StatusNotificationRequest; NotifyEventRequest) ????

OCPP 1.6

Passed: 
TC_046_1_CS
TC_046_2_CS
TC_047_CS
TC_048_2_CS
TC_048_3_CS
TC_048_4_CS
TC_050_2_CS
TC_050_3_CS
TC_051_CS
TC_052_CS
TC_053_CS

Failed:
TC_049_CS: Test makes a reservation for connector id 0. Expects a StatusNotification.req with 'Reserved'. Is the test wrong or should we send a StatusNotification.req with connector id 0 and status reserved????


Pietfried (changes requested):

modules/EvseManager/EvseManager.cpp
```diff
@@ -1196,7 +1196,9 @@ bool EvseManager::update_max_current_limit(types::energy::ExternalLimits& limits
     return true;
 }
 
-bool EvseManager::reserve(int32_t id) {
+bool EvseManager::reserve(int32_t id, bool signal_others) {
```
Pietfried, line 1199: probably a name like `signal_reservation_event` is a better name here

modules/Auth/include/ReservationHandler.hpp
```diff
@@ -1,94 +1,355 @@
...
+    ///
+    /// If there are multiple global reservations, while a charging station might look available, it is possible that
+    /// charging is not possible because then cars that made reservations can not charge anymore.
+    ///
+    /// Only use this function to check if a car can charge without having a reservation id.
+    ///
+    /// \param evse_id  The evse on which a car wants to charge.
+    /// \return True if charging is possible.
+    ///
+    bool is_charging_possible(const uint32_t evse_id);
```
Pietfried, line 135: It looks like this function is never used except in tests

modules/Auth/include/ReservationHandler.hpp
```diff
@@ -1,94 +1,355 @@
...
+    ///
+    /// @brief Function checks if the given \p id_token or \p parent_id_token matches the reserved token of the given \p
+    /// evse_id
+    ///
+    /// @param evse_id           Evse id
+    /// @param id_token          Id token
+    /// @param parent_id_token   Parent id token
+    /// @return The reservation id when there is a matching identifier, otherwise std::nullopt.
+    ///
+    std::optional<int32_t> matches_reserved_identifier(const std::optional<uint32_t> evse_id,
```
Pietfried, line 184: Same here with order of args

modules/Auth/Auth.cpp
```diff
@@ -70,11 +76,35 @@ void Auth::ready() {
...
-    this->auth_handler->register_reserved_callback([this](const int32_t evse_index, const int32_t& reservation_id) {
-        this->r_evse_manager.at(evse_index)->call_reserve(reservation_id);
-    });
+    this->auth_handler->register_reserved_callback(
+        [this](const std::optional<int32_t> evse_id, const int32_t& reservation_id) {
+            // Only call the evse manager to store the reservation if it is done for a specific evse.
+            if (evse_id.has_value()) {
+                EVLOG_info << "Call reserved callback for evse id " << evse_id.value();
+
+                this->r_evse_manager.at(evse_id.value() - 1)->call_reserve(reservation_id);
```
Pietfried, line 85: `call_reserve` returns a bool, we should check for the result. Im not sure, but maybe this also allows to move a little bit of the business logic to check whether a reservation can be placed (e.g. in case of faulted, inoperative) to the EvseManager. If I look at `handle_reserve` / `reserve` a few checks are done there

modules/EvseManager/manifest.yaml
```diff
@@ -6,6 +6,9 @@ config:
   connector_id:
     description: Connector id of this evse manager
     type: integer
+  connector_type:
+    description: The connector type of this evse manager
```
Pietfried, line 10: In this description you should point to the enum of available values for this parameter

Pietfried (changes requested):

modules/Auth/include/ReservationHandler.hpp
maaikez: Good one and interesting, because I think this needs implementation then. So if there are two reservations and three charging stations and a car comes in and plugs in, it is possible that it can not charge on a specific evse because then the reservations can not be fulfilled anymore. But I have to think about where to implement this then. Thanks for pointing this out!

modules/Auth/Auth.cpp
maaikez: Hmm yes that's a good one. Since the ReservationHandler has a copy of this states, it should just always be possible. But ok, if the evse manager returns false here, should we cancel the reservation in that case then? Because I think at this point the reservation has already been made.

modules/Auth/include/ReservationHandler.hpp
maaikez: I added two calls to this function in AuthHandler and created two tests for that (auth_tests) as well. Can you take a look if that's how it should work?

modules/Auth/include/ReservationHandler.hpp
```diff
@@ -1,94 +1,355 @@
...
+        std::vector<EvseConnectorType> connectors; ///< @brief The connectors of this EVSE
+    };
 
-    std::mutex timer_mutex;
-    std::mutex reservation_mutex;
-    std::map<int, std::unique_ptr<Everest::SteadyTimer>> connector_to_reservation_timeout_timer_map;
+    const std::string kvs_store_key_id;
+    kvsIntf* store;
+    /// \brief Map of EVSE's, with EVSE id as key and the EVSE struct as value.
+    std::map<uint32_t, Evse> evses;
```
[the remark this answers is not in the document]
maaikez: Ok I changed it now in ReservationHandler so it is a reference to this structure. And added a callback to notify the ReservationHandler if a status has changed.

types/reservation.yaml
```diff
@@ -63,3 +67,55 @@ types:
...
+      connector_id:
+        description: >-
+          The id of the connector to check. A value of 0 means that the reservation is not for a specific connector.
+        type: integer
+      id_token:
+        description: Check if there is a reservation for the given id token.
+        type: string
+      group_id_token:
+        description: Check if there is a reservation for the given group / parent id token.
+        type: string
```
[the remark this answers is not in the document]
maaikez: Added it to the description.

modules/Auth/lib/AuthHandler.cpp
```diff
@@ -38,19 +40,40 @@ std::string token_handling_result_to_string(const TokenHandlingResult& result) {
...
+        // evse already exists.
+        Connector c(connector_id, connector_type);
+        evses[evse_id]->connectors.push_back(c);
+    } else {
+        this->evses[evse_id] = std::make_unique<EVSEContext>(connector_id, evse_index, connector_type);
+    }
+}
+
+void AuthHandler::initialized() {
 }
```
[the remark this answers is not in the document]
maaikez: Oops. Added functionality.

modules/Auth/include/AuthHandler.hpp
```diff
@@ -172,15 +196,16 @@ class AuthHandler {
      *
      * @param callback
      */
-    void
-    register_reserved_callback(const std::function<void(const int& evse_index, const int& reservation_id)>& callback);
+    void register_reserved_callback(const std::function<void(const std::optional<int>&, const int&)>& callback);
```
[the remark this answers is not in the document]
maaikez: Agreed, sometimes my ide removes it or something and I do not always see that. Thanks for mentioning.

modules/Auth/include/AuthHandler.hpp
```diff
@@ -57,8 +58,16 @@ class AuthHandler {
      *
      * @param connector_id
      * @param evse_index
+     * @param connector_type    The connector type. Defaults to Unknown
      */
-    void init_connector(const int connector_id, const int evse_index);
+    void init_connector(
```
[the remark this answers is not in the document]
maaikez: Is evse index always evse id - 1?
Or can evse id also be 5 while evse index is 1?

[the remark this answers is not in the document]
maaikez: Ok changed, added evse_id and changed the function name and the parameters.

types/reservation.yaml
```diff
@@ -63,3 +67,55 @@ types:
...
+      - reservation_status
+    properties:
+      reservation_id:
+        description: The reservation id
+        type: integer
+      reservation_status:
+        description: The reservation status
+        type: string
+        enum:
+          - Expired
```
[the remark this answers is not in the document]
maaikez: I added this and also 'Used', but 'Used' still has to be implemented, I made a ticket for that.

modules/Auth/Auth.cpp
maaikez: I just left a todo there for now.

modules/Auth/include/ReservationHandler.hpp
```diff
@@ -1,94 +1,335 @@
...
+    ///
+    /// \brief Change the EVSE state.
+    ///
+    /// This is important for the reservation handler, to know which EVSE is in which state, to know if a reservation
+    /// can be made or not.
+    ///
+    /// \param available    True if evse is now available.
+    /// \param evse_id      The EVSE id.
+    ///
+    void on_evse_available_changed(const bool available, const uint32_t evse_id);
```
Pietfried, line 89: looks like this is never called

modules/Auth/include/ReservationHandler.hpp
```diff
@@ -1,94 +1,335 @@
...
+    /// \brief Change a specific connector state.
+    ///
+    /// This is important for the reservation handler, to know which connector is in which state, to know if a
+    /// reservation can be made or not.
+    ///
+    /// \param connector_state  The state of the connector.
+    /// \param evse_id          The EVSE id the connector belongs to.
+    /// \param connector_id     The connector id.
+    ///
+    void on_connector_state_changed(const ConnectorState connector_state, const uint32_t evse_id,
```
Pietfried, line 101: You don't need to pass the `connector_state` here since the `ReservationManager` has a reference the the `evse` map and can check the status internally

modules/Auth/Auth.cpp
Pietfried: The EvseManager shall stay autonoumus with the decision if a reservation can be made, so we should be able to handle the response at some point

modules/Auth/include/ReservationHandler.hpp
```diff
@@ -1,94 +1,335 @@
...
+                   const std::map<uint32_t, types::reservation::Reservation>& evse_specific_reservations);
+
+    ///
+    /// \brief Helper function to check if the connector of a specific EVSE is available.
+    /// \param evse_id          The evse id the connector belongs to.
+    /// \param connector_type   The connector type to check.
+    /// \return The `ReservationResult` to return fot his specific connector.
+    ///
+    types::reservation::ReservationResult
+    is_connector_available(const uint32_t evse_id, const types::evse_manager::ConnectorTypeEnum connector_type);
```
Pietfried, line 210: These function signatures look confusing. `get_evse_state` and `is_connector_available` return a `ReservationResult`. I see that you are using the responses to return the value in `make_reservation`, but I think something more readable is possible here

Pietfried (changes requested):

modules/Auth/Auth.cpp
maaikez: So should the flow be:
- check in the reservation handler if the reservation can be made, but do not make it yet
- if it can be made, call 'reserve' on the evse manager
- if evse manager returns 'true', we make the reservation in the reservation handler

?

Is that all going well when a new car just arrives in the meantime?

modules/Auth/include/AuthHandler.hpp
```diff
@@ -48,17 +48,24 @@ class AuthHandler {
...
+     * @brief Initializes the evse with the given \p connectors and the given \p evse_id . It instantiates new
      * connector objects and fills data sturctures of the class.
      *
-     * @param connector_id
+     * @param evse_id
      * @param evse_index
+     * @param connectors    The connectors.
      */
-    void init_connector(const int connector_id, const int evse_index);
+    void init_evse(const int evse_id, const int evse_index, const std::vector<Connector>& connectors);
```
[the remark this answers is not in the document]
maaikez: Is evse index always evse id - 1?
Or can evse id also be 5 while evse index is 1?

SirVer (approved):
Looks good for Bazel.

modules/OCPP201/manifest.yaml
```diff
@@ -98,6 +98,10 @@ requires:
     interface: display_message
     min_connections: 0
     max_connections: 1
+  reservation:
+    interface: reservation
+    min_connections: 1
```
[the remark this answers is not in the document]
Pietfried: I have discussed this with Cornelius and Kai , lets make it optional to allow compatability of older configs

modules/Auth/lib/ReservationHandler.cpp
```diff
@@ -1,110 +1,805 @@
...
+    }
+
+    if (evse_reservations.count(evse_id) != 0 && evse_reservations[evse_id].connector_type.has_value() &&
+        (connector_it->type == evse_reservations[evse_id].connector_type.value() ||
+         connector_it->type == types::evse_manager::ConnectorTypeEnum::Unknown ||
+         evse_reservations[evse_id].connector_type.value() == types::evse_manager::ConnectorTypeEnum::Unknown)) {
+        cancel_reservation(evse_reservations[evse_id].reservation_id, true,
+                           types::reservation::ReservationEndReason::Cancelled);
+        return;
+    }
```
Pietfried, line 212: nit: I find that quite hard to read. Id prefer something like:

```cpp
bool reservation_exists = evse_reservations.count(evse_id) != 0;
bool reserved_connector_specified = evse_reservations[evse_id].connector_type.has_value();
bool connector_type_matches = 
    connector_it->type == evse_reservations[evse_id].connector_type.value();
bool connector_type_unknown =
    connector_it->type == types::evse_manager::ConnectorTypeEnum::Unknown ||
    evse_reservations[evse_id].connector_type.value() == types::evse_manager::ConnectorTypeEnum::Unknown;

if (reservation_exists && reserved_connector_specified && (connector_type_matches || connector_type_unknown)) {
    cancel_reservation(evse_reservations[evse_id].reservation_id, 
                       true, 
                       types::reservation::ReservationEndReason::Cancelled);
    return;
}
```

modules/Auth/include/AuthHandler.hpp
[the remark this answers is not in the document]
Pietfried: yes you can make this assumption and add this requirement to the documentation in the `index.rst` with something like

`The module connections of the evse_manager requirement must be connected in the correct order in the EVerest config file, i.e. the module representing the EVSE with evse id 1 must listed first, EVSE with evse id 2 second and so on.`

modules/Auth/include/ReservationHandler.hpp
Pietfried: it looks better now :+1:

Pietfried:
> Failed:
> TC_B_24_CS: After reset, BootNotification should have reason 'RemoteReset', but is 'PowerUp'.

That is just a SIL issue, the system module doesnt remember the reset reason.

> TC_H_08_CS: Did not receive expected status change to 'Available'.

Probably unrelated to this PR

> TC_H_14_CS: Two 'global' reservations and 2 evse's, should set them both to 'reserved'.

> TC_H_15_CS: Did not receive expected status change to 'Available'. ??? -> Reservation is 'Cancelled', there is no reservation status update sent.

We need to investigate.

> TC_H_21_CS: When evse is set to Unavailable and a reservation is cancelled, evse should go to 'Unavailable' directly, but it first goes to 'Available'.

Probably unrelated to this PR

> TC_H_24_CS: Received unexpected request message: ReservationStatusUpdateRequest (Waiting for StatusNotificationRequest; NotifyEventRequest) ????

> Failed:
TC_049_CS: Test makes a reservation for connector id 0. Expects a StatusNotification.req with 'Reserved'. Is the test wrong or should we send a StatusNotification.req with connector id 0 and status reserved????

Great that all OCPP1.6 PASS ! TC_049_CS can fail in my view since we can configure `ReserveConnectorZeroSupported` to false, right?

modules/Auth/include/AuthHandler.hpp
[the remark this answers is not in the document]
maaikez: Added this to the Auth index.rst

maaikez:
> > Failed:
> > TC_B_24_CS: After reset, BootNotification should have reason 'RemoteReset', but is 'PowerUp'.
> 
> That is just a SIL issue, the system module doesnt remember the reset reason.

Yes all the identified issues were not for this pull request, only some we indeed need to investigate.

> 
> > TC_H_08_CS: Did not receive expected status change to 'Available'.
> 
> Probably unrelated to this PR

I wonder if it is our bug of a bug in OCTT.

> > TC_H_15_CS: Did not receive expected status change to 'Available'. ??? 
> 
> We need to investigate.

Yep I also wonder here if it isn't an OCTT bug. Or some feature I don't understand (yet).

> 
> > TC_H_21_CS: When evse is set to Unavailable and a reservation is cancelled, evse should go to 'Unavailable' directly, but it first goes to 'Available'.
> 
> Probably unrelated to this PR

Yes you'd already taken a look.

> > TC_H_24_CS: Received unexpected request message: ReservationStatusUpdateRequest (Waiting for StatusNotificationRequest; NotifyEventRequest) ????

We need to investigate this one as well.

> 
> > Failed:
> > TC_049_CS: Test makes a reservation for connector id 0. Expects a StatusNotification.req with 'Reserved'. Is the test wrong or should we send a StatusNotification.req with connector id 0 and status reserved????
> 
> Great that all OCPP1.6 PASS ! TC_049_CS can fail in my view since we can configure `ReserveConnectorZeroSupported` to false, right?

All other tests pass except this one indeed.
Also for this one: I am not sure if this is indeed expected behaviour to have a status notification for connector 0. 
We can indeed configure that to pass the test.

modules/Auth/lib/ReservationHandler.cpp
maaikez: Hmm I did this, but the order is important because of the optionals. I tried what happens if I keep those bools but it's a lot extra checks and I don't know if it will be more readable.

modules/Auth/reservation/reservationImpl.cpp
```diff
@@ -12,24 +12,37 @@ void reservationImpl::init() {
...
     // your code for cmd reserve_now goes here
 
-    const auto reservation_result = this->mod->auth_handler->handle_reservation(connector_id, reservation);
+    EVLOG_debug << "Handle reservation for evse id " << (request.evse_id.has_value() ? request.evse_id.value() : -1);
+
+    const auto reservation_result = this->mod->auth_handler->handle_reservation(request);
     if (reservation_result == ReservationResult::Accepted) {
-        this->mod->auth_handler->call_reserved(connector_id, reservation.reservation_id);
+        if (!this->mod->auth_handler->call_reserved(request.reservation_id, request.evse_id)) {
+            return ReservationResult::Rejected;
```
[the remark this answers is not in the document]
maaikez: Well that was because you said the evse manager always has the last saying about the reservation. So if the reservation is made but the evse manager sais it is not possible, it is cancelled again.

I don't like that behaviour as well.

modules/Auth/CMakeLists.txt
```diff
@@ -21,6 +21,7 @@ target_link_libraries(${MODULE_NAME}
         date::date
         date::date-tz
         everest::timer
+        nlohmann_json::nlohmann_json
```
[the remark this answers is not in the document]
maaikez: I removed it (and will check if it still builds). But it is auto generated so it will be put back in again later probably?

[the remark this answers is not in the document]
Pietfried: This part is not autogenerated

modules/Auth/reservation/reservationImpl.cpp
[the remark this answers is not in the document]
Pietfried: Yes but I can not see where the reservation is canceled if the evse manager rejects it

Pietfried (approved):

PR #947, opened 2024-11-06
refactor(EvseManager): more generic error checking

a-w50:
- added utility function `util::find_first_match`, which behaves similar to `std::visit`:
  - it is given a matcher class as the first argument and a variadic amount of additional arguments, that possibly can have different types
  - the matcher class needs to overload its call operator for each of these types and implement a corresponding matching function
  - `find_first_match` will then execute the corresponding matching function for each argument in order and return the first value, which evaluates to true
- this utility function is used to write the check for active fatal errors within the EvseManager in a more generic fashion

- [] I have made corresponding changes to the documentation


SirVer (approved):
looks good for BAZEL. Thanks!

Pietfried:
@corneliusclaussen any plans how to proceed with this PR?

corneliusclaussen:
Not sure if we need it still. I don't see that it actually simplifies the code, in fact it is much harder to understand imho

corneliusclaussen:
Closing as we do not really need it anymore

PR #951, opened 2024-11-07
Add module for NXP PN7160 NFC chip as token provider

cburandt:
Add module for NXP PN7160 NFC chip as token provider.

Adds [EVerest/linux_libnfc-nci](https://github.com/EVerest/linux_libnfc-nci) (forked from [NXPNFCLinux](https://github.com/NXPNFCLinux/linux_libnfc-nci) and patched to work well with EVerest).


dependencies.yaml
```diff
@@ -91,6 +91,10 @@ ext-mbedtls:
 everest-utils:
   git: https://github.com/EVerest/everest-utils.git
   git_tag: v0.4.2
+# linux_libnfc-nci for RFID
+libnfc-nci:
+  git: https://github.com/EVerest/linux_libnfc-nci.git
+  git_tag: 202411.1.0
```
hikinggrass, line 97: Just a minor observation, for consistency with the other repos that use a date based version this could be 2024.11.0 (since it's date based it's not a semantic version anyway)

hikinggrass: Can also have a
`cmake_condition: "EVEREST_DEPENDENCY_ENABLED_LIBNFC_NCI"`

hikinggrass: A find_package somewhere around here needs to be added as well: https://github.com/EVerest/everest-core/blob/5cc8b6d765e267f6181047891e786dea7b9a90e3/CMakeLists.txt#L111

modules/PN7160TokenProvider/libnfc-nci_config/libnfc-nci.conf
a-w50, line 1: I'm not sure here, but these files are copied from somewhere, and hence we should check and indicate their licensing.

modules/PN7160TokenProvider/main/libnfc_handler.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+bool NfcHandler::start(Callback callback_) {
+    if (this->callback) {
+        return false;
+    }
+
+    this->callback = std::move(callback_);
+
+    registerTagCallback(&nfc_callbacks);
+    doEnableDiscovery(DEFAULT_NFA_TECH_MASK, 0x01, 0x0, 0x1);
```
a-w50, line 48: This line would benefit from some more explicit documentation about the flags

modules/PN7160TokenProvider/libnfc-nci_config/libnfc-nci.conf
cburandt: They are Apache-2.0-licensed. I will add according headers to them.

a-w50 (approved):

dependencies.yaml
cburandt: Since I don't know in which situation it would be used: Is a BUILD.libnfc_nci.bazel required?

hikinggrass: With this entry: https://github.com/EVerest/everest-core/blob/ab1bb79c0404a4ad6c8ee291bcd6d751680054df/module-dependencies.cmake#L65 it will only be built when you build the new module and excluded if you exclude the module

hikinggrass (approved):

PR #952, opened 2024-11-11
DC Power supply: perform mode switch only when it is on

corneliusclaussen:
## Describe your changes

In some cases for bidi DC, output voltage stayed on after the session was finished. The relays were always open in that case.


SebaLukas (approved):
:+1:

PR #955, opened 2024-11-11
Slac: avoid second chip reset if it was already performed before unplug

corneliusclaussen:
## Describe your changes

On unplug, slac was reset every time. Slac also needs to be reset whenever the iso session ends, leading to an unneccessary reset if the unplug happens directly after the session end (with no resumed charging etc)


a-w50 (approved):
Couldn't `r_slac` reset itself on `leave_bcd`?  Wonder if the EvseManager should be aware of that.

SebaLukas (approved):
:+1:

PR #956, opened 2024-11-11
Add MREC9AuthorizationTimeout Error

corneliusclaussen:
## Describe your changes

Add support for MREC9 authorization timeout error code reporting (car was plugged but never authorized)

### DIFF
--- a/errors/evse_manager.yaml
+++ b/errors/evse_manager.yaml
@@ -9,6 +9,8 @@ errors:
     description: Internal error of the state machine
   - name: MREC4OverCurrentFailure
     description: Over current event
+  - name: MREC9AuthorizationTimeout
+    description: No authorization was provided within timeout after plugin
   - name: PowermeterTransactionStartFailed
     description: Transaction could not be started at the powermeter
   - name: Inoperative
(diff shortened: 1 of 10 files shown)


Pietfried:
This currently causes some issues with OCPP1.6 since the error raising and SessionEvents do overlap in a way here, potentially causing something like:

image

Is it intended that this error prevents charging (also causing yet another `StatusNotification.req` in OCPP)?

corneliusclaussen:
> Is it intended that this error prevents charging (also causing yet another `StatusNotification.req` in OCPP)?

This is debatable. In principle yes, as no charging shall be possible without replugging. I'm not sure if the Auth manager already takes care of this, in which case we would not need a second inoperative here. ChargeX is unfortunately absolutely unclear on how they thought this should be used.

Pietfried:
> > Is it intended that this error prevents charging (also causing yet another `StatusNotification.req` in OCPP)?
> 
> This is debatable. In principle yes, as no charging shall be possible without replugging. I'm not sure if the Auth manager already takes care of this, in which case we would not need a second inoperative here. ChargeX is unfortunately absolutely unclear on how they thought this should be used.

Ok. The `EvseManager` already publishes a `PluginTimeout` session event so the information is already transmitted (`StatusNotification(status=Finishing, info=ConnectionTimeout`) without the MREC error. I still think its ok to send another informational StatusNotification for it

corneliusclaussen:
@Pietfried I like the config option change with default off. I'm ok with merging now as is.

Pietfried:
> @Pietfried I like the config option change with default off. I'm ok with merging now as is.

Ok, Im fine with merging this as well

PR #961, opened 2024-11-12
Fix empty exiResponse

Pietfried:
## Describe your changes
Fixed issue in EvseV2G and OCPP modules. The exiResponse is a required value in the OCPP type but its optional in the EVerest type. This could lead to the situation that the EvseV2G module receives an empty exiResponse that could lead to a segmentation fault. This commit only sets the exiResponse in OCPP when not empty and adds an additional check to EvseV2G to check if the exiResponse string is not empty


SebaLukas (approved):
:+1:

PR #962, opened 2024-11-13
Always try to unlock the Connector, as long as the relays are not on

ydankner:
## Describe your changes
This is a followup to a [previous attempt](https://github.com/EVerest/everest-core/pull/939). The idea is the `is_locked` variable and the physical lock might be out of sync. In this MR I'm trying to unconditionally unlock the lock, as long as the relays are open.

Additionally, I'm ensuring that we always lock in `check_connector_lock` if the relays are on.


modules/EvseManager/IECStateMachine.cpp
```diff
@@ -505,17 +505,26 @@ void IECStateMachine::connector_force_unlock() {
...
+    }
+
     if (cp == RawCPState::B or cp == RawCPState::C) {
         force_unlocked = true;
         check_connector_lock();
     }
 }
 
 void IECStateMachine::check_connector_lock() {
-    if (should_be_locked and not force_unlocked and not is_locked) {
```
ydankner, line 515: Previously, relais_on condition was not here. This meant that if force_unlock was on in cp state B, and then the state changed to C, the relais will not lock.

lategoodbye:
@ydankner The word "Contactor" in the pull request subject is confusing. Is it possible you meant "connector" or "plug lock"?

ydankner:
> @ydankner The word "Contactor" in the pull request subject is confusing. Is it possible you meant "connector" or "plug lock"?

Yes, thanks :)

modules/EvseManager/IECStateMachine.cpp
```diff
@@ -505,17 +505,26 @@ void IECStateMachine::connector_force_unlock() {
...
+    bool should_be_locked_considering_relais_and_force = relais_on or (should_be_locked and not force_unlocked);
+
+    if (not is_locked and should_be_locked_considering_relais_and_force) {
         signal_lock();
         is_locked = true;
-    } else if ((not should_be_locked or force_unlocked) and is_locked and not relais_on) {
+    } else if (is_locked and not should_be_locked_considering_relais_and_force) {
```
corneliusclaussen, line 527: This will always unlock if relais are off, i.e. always in state B. This is however not what we want, as in public spaces it should really only be unlocked in state A if not forced.

corneliusclaussen (changes requested):

corneliusclaussen (approved):
The logic seems correct to me

hikinggrass (approved):

Issue #958, opened 2024-11-11
Reservation: set all connectors to 'reserved' when there are as many 'global' reservations as there are free evse's


maaikez:
### Describe the bug

If there are three free evse's and a third reservation for a non-specific evse is made, all free evse's should go to reserved. This currently does not happen.

If there are as many free evse's as reservations for non-specific evse's and all connectors are set to reserved, as soon as one evse comes available, all connectors should go back to 'available'.

### EVerest Domain

Authorization, Other

### Affected EVerest Module

Auth
EVSE Manager

### Anything else?

The ReservationHandler can keep track of this and send 'reserved' to the EVSEManager. However, currently that interface requires a reservation id, so there must be made some changes in that interface as well. Because we can not provide the reservation id, because one reservation is not tight to a specific evse or connector.

PR #972, opened 2024-11-27
Reservation all connectors reserved (#958)

maaikez:
## Describe your changes

When there are as many reservations as evse's available, all available evse's must go to 'reserved'. And back to 'available' when there are more evse's than reservations.

Also, added unittests.

See other related pull requests:
https://github.com/EVerest/libocpp/pull/878
https://github.com/EVerest/everest-utils/pull/171

#958 

### DIFF
--- a/interfaces/evse_manager.yaml
+++ b/interfaces/evse_manager.yaml
@@ -49,8 +49,8 @@ cmds:
     arguments:
       reservation_id:
         description: >-
-          The reservation id (should be added to the TransactionStarted
-          event)
+          The reservation id (should be added to the TransactionStarted event). Negative value if there was
+          no specific reservation id for this evse (if there are 'global' reservations for 'any' evse).
         type: integer
     result:
       description: Returns true if the EVSE accepted the reservation, else false.
--- a/interfaces/evse_manager.yaml
+++ b/interfaces/evse_manager.yaml
@@ -49,8 +49,9 @@ cmds:
     arguments:
       reservation_id:
         description: >-
-          The reservation id (should be added to the TransactionStarted event). Negative value if there was
-          no specific reservation id for this evse (if there are 'global' reservations for 'any' evse).
+          The reservation id (should be added to the TransactionStarted event). Set this to a negative value if there is
+          no specific reservation id for this evse but the evse should still move to a Reserved state because of total
+          global reservations.
         type: integer
     result:
       description: Returns true if the EVSE accepted the reservation, else false.
--- a/modules/Auth/lib/ReservationHandler.cpp
+++ b/modules/Auth/lib/ReservationHandler.cpp
@@ -1,7 +1,6 @@
 #include <ReservationHandler.hpp>
 
 #include <algorithm>
-#include <iostream>
 
 #include <Connector.hpp>
 
@@ -414,7 +413,6 @@ bool ReservationHandler::has_reservation_parent_id(const std::optional<uint32_t>
 }
 
 ReservationEvseStatus ReservationHandler::check_number_global_reservations_match_number_available_evses() {
-    ReservationEvseStatus evse_status;
     std::set<int32_t> available_evses;
     std::unique_lock<std::recursive_mutex> lock(this->evse_mutex);
     // Get all evse's that are not reserved or used.
--- a/modules/Auth/lib/ReservationHandler.cpp
+++ b/modules/Auth/lib/ReservationHandler.cpp
@@ -403,7 +403,7 @@ bool ReservationHandler::has_reservation_parent_id(const std::optional<uint32_t>
 ReservationEvseStatus ReservationHandler::check_number_global_reservations_match_number_available_evses() {
     ReservationEvseStatus evse_status;
     std::set<int32_t> available_evses;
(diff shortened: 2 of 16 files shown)


interfaces/evse_manager.yaml
```diff
@@ -49,8 +49,8 @@ cmds:
     arguments:
       reservation_id:
         description: >-
-          The reservation id (should be added to the TransactionStarted
-          event)
+          The reservation id (should be added to the TransactionStarted event). Negative value if there was
+          no specific reservation id for this evse (if there are 'global' reservations for 'any' evse).
```
Pietfried, line 53:
```suggestion
          The reservation id (should be added to the TransactionStarted event). Set tis to a negative value if there is 
          no specific reservation id for this evse but the evse should still move to a Reserved state because of total 
          global reservations.
```

types/reservation.yaml
```diff
@@ -54,11 +54,15 @@ types:
         Expired: When the reservation expired before the reserved token was used for a session
         Cancelled: When the reservation was cancelled manually
         UsedToStartCharging: When the reservation ended because the reserved token was used for a session
+        GlobalReservationConnectorFree: When the reservation ended because there is a connector free and there are less
+                                        reservations than available evse's (reservation is still there but it is not
+                                        occupying the EVSE anymore).
```
Pietfried, line 59:
```suggestion
                                        occupying this EVSE anymore).
```

types/reservation.yaml
```diff
@@ -54,11 +54,15 @@ types:
...
     type: string
     enum:
       - Expired
       - Cancelled
       - UsedToStartCharging
+      - GlobalReservationConnectorFree
```
Pietfried, line 65: Since I was a little confused by the name.. maybe `GlobalReservationRequirementDropped` ?

tests/ocpp_tests/test_sets/ocpp16/ocpp_compliance_tests.py
```diff
@@ -4719,6 +4798,10 @@ async def test_reservation_faulted_state(
         call_result.ReserveNowPayload(ReservationStatus.faulted),
     )
 
+    test_controller.clear_error(1)
+
+    await asyncio.sleep(1)
+
```
Pietfried, line 4804: This doesnt look like its required at the end of the test case

tests/ocpp_tests/test_sets/ocpp16/ocpp_compliance_tests.py
```diff
@@ -4693,15 +4770,17 @@ async def test_reservation_faulted_state(
 ):
     logging.info("######### test_reservation_faulted_state #########")
 
-    test_controller.diode_fail()
+    test_controller.raise_error("MREC6UnderVoltage", 1)
+
+    await asyncio.sleep(1)
```
Pietfried, line 4775: I dont think that we have to sleep here since `wait_for_and_validate` will wait for the error event already

tests/ocpp_tests/test_sets/ocpp201/reservations.py
```diff
@@ -0,0 +1,1209 @@
...
+        evse_id=1
+    )
+
+    # which should return 'faulted'
+    assert await wait_for_and_validate(
+        test_utility,
+        charge_point_v201,
+        "ReserveNow",
+        call_result201.ReserveNowPayload(ReserveNowStatusType.faulted),
+    )
```
Pietfried, line 360: also here you can do drop the sleep if you do

1. `raise_error`
2. wait and validate for StatusNotification.req(Faulted)
3. ReserveNow
4. wait and validate ReserveNow

tests/ocpp_tests/test_sets/ocpp201/reservations.py
```diff
@@ -0,0 +1,1209 @@
...
+    assert await wait_for_and_validate(
+        test_utility,
+        charge_point_v201,
+        "StatusNotification",
+        call_201.StatusNotificationPayload(
+            ANY, 'Faulted', 1, 1
+        ),
+    )
+
+    await asyncio.sleep(1)
```
Pietfried, line 861: I think this can also be dropped.

modules/Auth/lib/ReservationHandler.cpp
```diff
@@ -400,6 +413,52 @@ bool ReservationHandler::has_reservation_parent_id(const std::optional<uint32_t>
     return false;
 }
 
+ReservationEvseStatus ReservationHandler::check_number_global_reservations_match_number_available_evses() {
+    ReservationEvseStatus evse_status;
```
Pietfried, line 417:
```suggestion
```
This looks like it is not used

Pietfried (approved):
lgtm, added a few suggestions

tests/ocpp_tests/test_sets/ocpp201/reservations.py
maaikez: Same here:
We have a race condition here. The statusnotification is sent earlier than the status is received by the auth handler so the reservation handler does not know of the state of the connector yet. :(

PR #991, opened 2024-12-12
Reduce mutexes in Auth module

Pietfried:
## Describe your changes
Reduce mutexes in Auth module:
* Only using event_mutex for public functions that can be called from different threads
* Unlocking before select_evse and lock only referenced_evse mutexes in it

This change leads to less complexity and reduced the number of mutexes

### DIFF
--- a/modules/Auth/lib/ReservationHandler.cpp
+++ b/modules/Auth/lib/ReservationHandler.cpp
@@ -407,8 +407,8 @@ bool ReservationHandler::has_reservation_parent_id(const std::optional<uint32_t>
 }
 
 ReservationEvseStatus ReservationHandler::check_number_global_reservations_match_number_available_evses() {
+    std::unique_lock<std::recursive_mutex> reservation_lock(this->event_mutex);
     std::set<int32_t> available_evses;
-    std::unique_lock<std::recursive_mutex> lock(this->evse_mutex);
     // Get all evse's that are not reserved or used.
     for (const auto& evse : this->evses) {
         if (get_evse_connector_state_reservation_result(static_cast<uint32_t>(evse.first), this->evse_reservations) ==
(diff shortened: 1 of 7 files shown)


modules/Auth/lib/ReservationHandler.cpp
```diff
@@ -253,19 +248,14 @@ ReservationHandler::cancel_reservation(const int reservation_id, const bool exec
 
     std::pair<bool, std::optional<uint32_t>> result;
 
-    std::unique_lock<std::recursive_mutex> lock(reservation_mutex);
```
maaikez, line 256: This function is also called from the timer within the reservation handler (see line 581, in 'set_reservation_timer'). So I think the mutex should not be removed here.

Pietfried: I added that one again

maaikez (approved):

modules/Auth/tests/auth_tests.cpp
```diff
@@ -198,6 +199,147 @@ TEST_F(AuthTest, test_two_referenced_connectors) {
...
+
+    ASSERT_TRUE(result1 == TokenHandlingResult::TIMEOUT);
+    ASSERT_TRUE(result2 == TokenHandlingResult::TIMEOUT);
+    ASSERT_FALSE(this->auth_receiver->get_authorization(0));
+    ASSERT_FALSE(this->auth_receiver->get_authorization(1));
+    ASSERT_FALSE(this->auth_receiver->get_authorization(2));
+    ASSERT_FALSE(this->auth_receiver->get_authorization(3));
+    ASSERT_FALSE(this->auth_receiver->get_authorization(4));
+
+    EVLOG_critical << "HQH";
```
hikinggrass, line 340: hqh? :smile:

hikinggrass (approved):

modules/Auth/tests/auth_tests.cpp
Pietfried, line 340:
```suggestion
```

PR #992, opened 2024-12-13
Add withdraw authorization command in auth interface

Pietfried:
## Describe your changes
This PR adds a `withdraw_authorization` command in the `auth` interface and the Auth module.

https://github.com/EVerest/everest-core/issues/978

### DIFF
--- a/modules/Auth/Auth.cpp
+++ b/modules/Auth/Auth.cpp
@@ -19,8 +19,10 @@ void Auth::init() {
         this->info.id, (!this->r_kvs.empty() ? this->r_kvs.at(0).get() : nullptr));
 
     for (const auto& token_provider : this->r_token_provider) {
-        token_provider->subscribe_provided_token(
-            [this](ProvidedIdToken provided_token) { this->auth_handler->on_token(provided_token); });
+        token_provider->subscribe_provided_token([this](ProvidedIdToken provided_token) {
+            std::thread t([this, provided_token]() { this->auth_handler->on_token(provided_token); });
+            t.detach();
+        });
     }
 }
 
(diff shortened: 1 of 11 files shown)


modules/Auth/include/AuthHandler.hpp
```diff
@@ -227,9 +263,12 @@ class AuthHandler {
     std::map<int, std::unique_ptr<EVSEContext>> evses;
 
     std::list<int> plug_in_queue;
-    std::set<std::string> tokens_in_process;
+    std::set<ProvidedIdToken> tokens_in_process;
     std::condition_variable cv;
+    std::condition_variable processing_finished_cv;
     std::mutex event_mutex;
+    std::mutex withdraw_mutex;
```
hikinggrass, line 270: this could probably also use the event_mutex now, or am I mistaken?

hikinggrass (approved):

modules/Auth/include/AuthHandler.hpp
Pietfried: Initially I thought so too. If we only use the `event_mutex` multiple withdraw requests could run in parallel and they could override `this->withdraw_request` in that case, because the `event_mutex` is released [here](https://github.com/EVerest/everest-core/pull/992/files#diff-8db9b5c55c5494b78d2e0be7e5aa44d64c9361a2862b47ae6f13a4cd86a63baaR894) while waiting for the targeted processing threads to finish. It ensures that only one withdraw_request can run at a time. Since a withdraw request should not block for a significant amount of time this seems to be the best solution to me

modules/Auth/Auth.cpp
```diff
@@ -19,10 +19,8 @@ void Auth::init() {
         this->info.id, (!this->r_kvs.empty() ? this->r_kvs.at(0).get() : nullptr));
 
     for (const auto& token_provider : this->r_token_provider) {
-        token_provider->subscribe_provided_token([this](ProvidedIdToken provided_token) {
-            std::thread t([this, provided_token]() { this->auth_handler->on_token(provided_token); });
```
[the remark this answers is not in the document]
Pietfried: The handlers are running in their own thread as long as the tokens are published from different token_providers. If tokens are published from the same provider you are right though, the handlers will only be executed once the previous one has returned. I will therefore change it back.

PR #993, opened 2024-12-16
Introduce EVSE Manager configuration option for failing charging if t…

florinmihut:
…he powermeter has errors

### DIFF
--- a/modules/EvseManager/manifest.yaml
+++ b/modules/EvseManager/manifest.yaml
@@ -192,7 +192,7 @@ config:
     default: 0.5
   hack_fix_hlc_integer_current_requests:
     description: >-
-      Some cars request only integer ampere values during DC charging. For low power DC charging that 
+      Some cars request only integer ampere values during DC charging. For low power DC charging that
       means that they charge a few hundred watts slower then needed. If enabled, this will charge at full power if the difference
       between EV requested current (integer) and HLC current limit is less then 1.0
     type: boolean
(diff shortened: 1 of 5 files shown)


hikinggrass (approved):

Pietfried (changes requested):
Please add the added errors which now cause the evse manager to become Inoperative to the documentation in the `doc.rst`

modules/EvseManager/manifest.yaml
```diff
@@ -283,14 +283,19 @@ config:
...
       in this case EVerest cannot recover the charging session if the fault is cleared.
       In this case you can set this parameter to 0, which will avoid to use state F in case of a fault
       and only disables PWM (C2->C1) while switching off power. This will violate IEC 61851-1:2017 however.
       The default is 300ms as the minimum suggested by IEC 61851-1:2017 Table A.5 (description) to be compliant.
       This setting is only active in BASIC charging mode.
     type: integer
     default: 300
+  fail_on_powermeter_errors:
+    description: >-
+      Set the EVSE Manager as inoperable if the powermeter is configured and has errors
```
Pietfried, line 296:
```suggestion
      Set the EVSE Manager to an inoperative state if the powermeter requirement is configured and has reported errors
```

Pietfried (approved):

Pietfried:
Currently, a transaction can never be started if `call_start_transaction` at the powermeter returns  `UNEXPECTED_ERROR`. My suggestion would be to use the introduced configuration option to control this flow.

I think we should check `config.fail_on_powermeter_errors` in this line too: https://github.com/EVerest/everest-core/blob/add88f3d72dd571d90560c865892007893ec4b49/modules/EvseManager/Charger.cpp#L1206
and we should not return early in case `fail_on_powermeter_errors` is `false` so that we can still start the transaction.

This would mean the `fail_on_powermeter_errors` option controls:
* If transactions can be started if `call_start_transaction` at the powermeter returns `UNEXPECTED_ERROR`.
* If charging is stopped during a transaction if a powermeter raises an error

What do you think @corneliusclaussen @florinmihut ?

I would suggest that we move the changes of [the powermeter documentation PR](https://github.com/EVerest/everest-core/pull/834) into this PR.

florinmihut:
Made the requested changes and add the documentation as mermaid files.

corneliusclaussen (approved):

PR #997, opened 2024-12-18
Accept bank cards only for specific connectors

wku12:
## Describe your changes
Add to the config a comma separated list (since config can't have arrays) of connectors to which the bank cards should be accepted 

### DIFF
--- a/modules/RsPaymentTerminal/manifest.yaml
+++ b/modules/RsPaymentTerminal/manifest.yaml
@@ -41,7 +41,7 @@ config:
     description: Password for the Feig terminal.
     type: integer
   accept_credit_cards:
-    description: Indicates whether credit cards should be accepted for all connectors
+    description: Indicates whether credit cards should be accepted for the connectors given with `credit_card_connectors`
     type: boolean
     default: true
   credit_card_connectors:
--- a/modules/RsPaymentTerminal/manifest.yaml
+++ b/modules/RsPaymentTerminal/manifest.yaml
@@ -40,6 +40,18 @@ config:
   password:
     description: Password for the Feig terminal.
     type: integer
+  accept_credit_cards:
+    description: Indicates whether credit cards should be accepted for all connectors
+    type: boolean
+    default: true
+  credit_card_connectors:
+    description: >-
+      If the `accept_credit_cards` is set to true credit cards
+      will only provide authentication to the given connectors.
+      If list is empty, authentication will happen to all the connectors.
+    type: string
+    default: ""
+
 requires:
     session:
       interface: session_cost
--- a/modules/RsPaymentTerminal/src/main.rs
+++ b/modules/RsPaymentTerminal/src/main.rs
@@ -127,18 +127,28 @@ mod sync_feig {
 #[mockall_double::double]
 use sync_feig::SyncFeig;
 
+fn string_to_vec(s: &String) -> Vec<i64> {
+    s.split(',')
+        .filter_map(|s| s.trim().parse::<i64>().ok())
+        .collect()
+}
+
 impl ProvidedIdToken {
-    fn new(id_token: String, authorization_type: AuthorizationType) -> Self {
+    fn new(
+        id_token: String,
+        authorization_type: AuthorizationType,
+        connectors: Option<Vec<i64>>,
+    ) -> Self {
         Self {
             parent_id_token: None,
             id_token: IdToken {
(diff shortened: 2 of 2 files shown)


modules/RsPaymentTerminal/manifest.yaml
```diff
@@ -40,6 +40,18 @@ config:
   password:
     description: Password for the Feig terminal.
     type: integer
+  accept_credit_cards_for_all_connectors:
```
dorezyuk, line 43: i think accept_credit_cards is actually enough

modules/RsPaymentTerminal/manifest.yaml
```diff
@@ -40,6 +40,18 @@ config:
...
+    description: Indicates wheter credit card should be accepted for all the connectors
```
dorezyuk, line 44: whether

dorezyuk: either a credit card or credit cards

dorezyuk: i don't think you need the `the`

wku12: English classes 1-1 :D

modules/RsPaymentTerminal/manifest.yaml
```diff
@@ -40,6 +40,18 @@ config:
...
+    type: boolean
+    default: true
+  credit_card_connectors:
+    description: >-
+      If the `accept_credit_cards_for_all_connectors` is set to false credit cards
+      will only provide authentication to the given connectors.
+      If list is empty, authentication won't happen.
+    type: string
+    default: "1,2"
```
dorezyuk, line 53: i thought we're doing it the reverse way - default should be empty. in this case we don't specify the connectors at all

modules/RsPaymentTerminal/src/main.rs
```diff
@@ -188,26 +205,43 @@ impl PaymentTerminalModule {
...
-                ProvidedIdToken::new(token, AuthorizationType::BankCard)
-            }
-            CardInfo::MembershipCard(id_token) => {
-                ProvidedIdToken::new(id_token, AuthorizationType::RFID)
+                if !self.accept_credit_cards {
+                    None
+                } else {
+                    self.feig.begin_transaction(&token)?;
+                    let credit_cards_connectors = if self.credit_cards_connectors.is_empty() {
+                        None
```
dorezyuk, line 215: this is not what we've documented. in this case we want to publish the token open for all connectors

modules/RsPaymentTerminal/src/main.rs
```diff
@@ -154,6 +164,13 @@ pub struct PaymentTerminalModule {
 
     /// The Feig interface.
     feig: SyncFeig,
+
+    /// For which connectors credit cards should be accepted
+    /// If None, authentication happens to all connectors
+    credit_cards_connectors: Vec<i64>,
```
dorezyuk, line 170: None can be only true for Option. Maybe lets make it an option?

wku12: It doesn't have to. It's a config - it should always be set to something. Even empty

dorezyuk: I mean in the comment above

dorezyuk (approved):

corneliusclaussen (approved):

modules/RsPaymentTerminal/manifest.yaml
```diff
@@ -40,6 +40,18 @@ config:
   password:
     description: Password for the Feig terminal.
     type: integer
+  accept_credit_cards:
+    description: Indicates whether credit cards should be accepted for all connectors
```
barsnick, line 44: The description "for all connectors" is in contradiction to the description for `credit_card_connectors`, where it says "only [...] to the given connectors".
```suggestion
    description: Indicates whether credit cards should be accepted for the connectors given with `credit_card_connectors`
```

modules/RsPaymentTerminal/manifest.yaml
```diff
@@ -40,6 +40,18 @@ config:
   password:
     description: Password for the Feig terminal.
     type: integer
+  accept_credit_cards:
```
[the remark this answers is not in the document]
dorezyuk: everest does not support enums in config parameter

PR #1006, opened 2025-01-16
bugfix(tls_test): use timed_mutex

a-w50:
- on some machines, the `tls_test` might not be able to run, because it can't open a port
- instead of hanging, it now will skip the test
- NOTE: some other tls test cases still fail and don't skip, this should be fixed


james-ctc (approved):
Looks good to me. Many thanks

PR #1017, opened 2025-01-21
Departure time patches

Abby-Wheelis:
## Describe your changes
updates in EvseV2G/iso_server.cpp which accounts for the departure time in the calculation of current, voltage, and power
updates in PyEvJosev/module.py which adds DepartureTime and EAmount initialization

[Issue #220 ](https://github.com/EVerest/EVerest/issues/220)

Opening PR as a draft to continue conversation about the changes.

### DIFF
--- a/modules/EvseV2G/iso_server.cpp
+++ b/modules/EvseV2G/iso_server.cpp
@@ -1125,32 +1125,55 @@ static enum v2g_event handle_iso_charge_parameter_discovery(struct v2g_connectio
     res->EVSEChargeParameter_isUsed = 0;
     res->EVSEProcessing = (iso2_EVSEProcessingType)conn->ctx->evse_v2g_data.evse_processing[PHASE_PARAMETER];
 
+    struct linked_ac_params {
+	float max_current;
+        int64_t voltage;
+        int64_t pmax;
+    };
+
+    linked_ac_params sel_params;
+
     /* Configure SA-schedules*/
     if (res->EVSEProcessing == iso2_EVSEProcessingType_Finished) {
         /* If processing is finished, configure SASchedule list */
         if (conn->ctx->evse_v2g_data.evse_sa_schedule_list_is_used == false) {
+	    int64_t departure_time_duration = req->AC_EVChargeParameter.DepartureTime;
+
             /* If not configured, configure SA-schedule automatically for AC charging */
             if (conn->ctx->is_dc_charger == false) {
                 /* Determin max current and nominal voltage */
-                float max_current = conn->ctx->basic_config.evse_ac_current_limit;
-                int64_t nom_voltage =
+		linked_ac_params default_params;
+                /* Setup default params (before the departure time overrides) */
+                default_params.max_current = conn->ctx->basic_config.evse_ac_current_limit;
+                default_params.voltage =
                     conn->ctx->evse_v2g_data.evse_nominal_voltage.Value *
                     pow(10, conn->ctx->evse_v2g_data.evse_nominal_voltage.Multiplier); /* nominal voltage */
-
-                /* Calculate pmax based on max current, nominal voltage and phase count (which the car has selected
-                 * above) */
-                int64_t pmax =
-                    max_current * nom_voltage *
+		default_params.pmax =
+                    default_params.max_current * default_params.voltage *
                     ((req->RequestedEnergyTransferMode == iso2_EnergyTransferModeType_AC_single_phase_core) ? 1 : 3);
+
+		dlog(DLOG_LEVEL_WARNING, "before adjusting for departure time, max_current %f, nom_voltage %d, pmax %d, departure_duration %d",
+			default_params.max_current, default_params.voltage,
+			default_params.pmax, departure_time_duration);
+                double req_eamount = calc_physical_value(req->AC_EVChargeParameter.EAmount.Value,
+                        req->AC_EVChargeParameter.EAmount.Multiplier);
+                dlog(DLOG_LEVEL_WARNING, "Requested departure time %u, requested energy %f",
+			departure_time_duration, req_eamount);
+                sel_params = default_params;
+
                 populate_physical_value(&conn->ctx->evse_v2g_data.evse_sa_schedule_list.SAScheduleTuple.array[0]
                                              .PMaxSchedule.PMaxScheduleEntry.array[0]
                                              .PMax,
-                                        pmax, iso2_unitSymbolType_W);
+					     sel_params.pmax, iso2_unitSymbolType_W);
             } else {
                 conn->ctx->evse_v2g_data.evse_sa_schedule_list.SAScheduleTuple.array[0]
                     .PMaxSchedule.PMaxScheduleEntry.array[0]
                     .PMax = conn->ctx->evse_v2g_data.evse_maximum_power_limit;
             }
+	    if (departure_time_duration == 0) {
+                departure_time_duration = SA_SCHEDULE_DURATION; // one day, per spec
+            }
+
             conn->ctx->evse_v2g_data.evse_sa_schedule_list.SAScheduleTuple.array[0]
                 .PMaxSchedule.PMaxScheduleEntry.array[0]
                 .RelativeTimeInterval.start = 0;
@@ -1159,7 +1182,7 @@ static enum v2g_event handle_iso_charge_parameter_discovery(struct v2g_connectio
                 .RelativeTimeInterval.duration_isUsed = 1;
             conn->ctx->evse_v2g_data.evse_sa_schedule_list.SAScheduleTuple.array[0]
                 .PMaxSchedule.PMaxScheduleEntry.array[0]
-                .RelativeTimeInterval.duration = SA_SCHEDULE_DURATION;
+                .RelativeTimeInterval.duration = departure_time_duration;
             conn->ctx->evse_v2g_data.evse_sa_schedule_list.SAScheduleTuple.array[0]
                 .PMaxSchedule.PMaxScheduleEntry.arrayLen = 1;
             conn->ctx->evse_v2g_data.evse_sa_schedule_list.SAScheduleTuple.arrayLen = 1;
@@ -1207,14 +1230,8 @@ static enum v2g_event handle_iso_charge_parameter_discovery(struct v2g_connectio
         populate_physical_value_float(&res->AC_EVSEChargeParameter.EVSEMaxCurrent, max_current, 1,
                                       iso2_unitSymbolType_A);
 
-        /* Nominal voltage */
-        res->AC_EVSEChargeParameter.EVSENominalVoltage = conn->ctx->evse_v2g_data.evse_nominal_voltage;
-        int64_t nom_voltage = conn->ctx->evse_v2g_data.evse_nominal_voltage.Value *
-                              pow(10, conn->ctx->evse_v2g_data.evse_nominal_voltage.Multiplier);
-
         /* Calculate pmax based on max current, nominal voltage and phase count (which the car has selected above) */
-        int64_t pmax = max_current * nom_voltage *
-                       ((iso2_EnergyTransferModeType_AC_single_phase_core == req->RequestedEnergyTransferMode) ? 1 : 3);
+        int64_t pmax = sel_params.pmax;
 
         /* Check the SASchedule */
         if (res->SAScheduleList_isUsed == (unsigned int)1) {
--- a/modules/PyEvJosev/module.py
+++ b/modules/PyEvJosev/module.py
@@ -94,7 +94,10 @@ class PyEVJosevModule():
     # implementation handlers
 
     def _handler_start_charging(self, args) -> bool:
+        log.info('_handler_start_charging() args: %s' % str(args))
 
+        self._es.DepartureTime = args['DepartureTime']
+        self._es.EAmount = args['EAmount']
         self._es.EnergyTransferMode = args['EnergyTransferMode']
 
         self._ready_event.set()


SebaLukas:
I will have a look next week

shankari:
@SebaLukas @Abby-Wheelis we made these changes to pyjosev before it was deprecated; unfortunately there was a large time lag between our implementation and the PR review due to external constraints. However, time and master wait for no person so the codebase has moved on.

We need to refactor this PR to adjust for the current circumstances. In my mind, there are two main components to this change.
1. **Supporting specifying EAmount and Departure Time from the UI:** To my knowledge all communication between the UI and EVerest happens via MQTT messages. We should have the `EVManager` (the new C++ implementation) accept these values and put them into the ISO message that we send to the station. There is an `void subscribe_to_external_mqtt();` in `car_simulatorImpl.hpp` so I would assume that we can listen to the MQTT messages directly instead of polluting the API with a bunch of messages that are non-EVSE related.
2. **Computing the curve on the EV:** The current implementation has the charging curve computed on the EV using a python optimization library. It makes sense for the EV to compute the charging curve, since in the real world, the BMS will/should control the rate of charge depending on battery characteristics. However, migrating the optimization algorithms to C++ will likely be complicated, since they use the python optimization libraries such as [numpy](https://github.com/EVerest/everest-demo/blob/a426034050a10f4f15fe0a9bc41b9d62a9b15f3b/manager/demo-patches/power_curve.py#L5). So we need to figure out a C++ friendly initial implementation so that we can get this change checked in.

We use these curves in two places:
   - to display potential curves in the nodered UI: I think we can drop this functionality for now since it is not clear that shape of the curve is meaningful to the user
   - to compute the values that are sent in the PowerDeliveryReq
https://github.com/EVerest/everest-demo/blob/a426034050a10f4f15fe0a9bc41b9d62a9b15f3b/manager/demo-patches/simulator_enabledt.patch#L24 We do need to keep that in place, but we need to think about how we can compute it. Maybe https://docs.python.org/3/extending/index.html#extending-index is an option to invoke python code from C++?

modules/EvseV2G/iso_server.cpp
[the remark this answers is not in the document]
shankari: `default_params` is the default parameters (from the config.yml); after adjusting based on the parameters that are passed in as part of the `ChargeParameterDiscoveryReq`, we get the `sel_params` which we return in `ChargeParameterDiscoveryRes`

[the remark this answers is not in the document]
SebaLukas: I'm sorry but I still don't understand it. From my point of view, `sel_params` are not needed at all. `sel_params` is not changed based on the req message.

SebaLukas:
I think you have misunderstood something.
`PyJosev` was the **EVSE** ISO module before `EvseV2G`, which has been deprecated for almost 2 years. `PyEvJosev`, **EV** ISO module, on the other hand, is not deprecated and will continue to be used in the future.

Lately we have been rewriting the Js modules one by one into cpp. But these cpp modules are one-to-one conversions with the same interfaces. So I would like to know how you came up with deprecated. Since this PR also has nothing to do with `PyJosev`.

Regarding point 1.) Another possibility would be to subscribe to external Mqtt topics via the `EvManager` module and forward these topics to `PyEvJosev` via the internal interfaces. As you have already written, we are already using the external mqtt in the `EvManager`. The only important thing for me is that the `car_simulation` interface for communication between `EvManager` and `PyEvJosev` and the two modules are extended accordingly.

Regarding point 2.) Theoretically, you could also calculate the curve directly in the `PyEvJosev` module or in `Josev`. `PyEvJosev` and `Josev` are written in Python and you could also use numpy etc. here. However, I would prefer `Josev`, as it is relatively easy to install the Python module using the requirements.txt file. It would also be possible via `PyEvJosev`.

Abby-Wheelis:
>PyJosev was the EVSE ISO module before EvseV2G, which has been deprecated for almost 2 years. PyEvJosev, EV ISO module, on the other hand, is not deprecated and will continue to be used in the future.

We must have been confused between `PyEvJosev` and `PyJosev`, since `PyEvJosev` is not deprecated, I will continue with cleaning up the changes in place and ensure that changes are kept in line with the interface. 

>Another possibility would be to subscribe to external Mqtt topics via the EvManager module and forward these topics to PyEvJosev via the internal interfaces. 

This seems like it might be the most straightforward option, since we do already use the external MQTT in the `EvManger`, and then the simulation of the BMS (curve calculations) could continue to take place in the car simulator.

shankari:
@SebaLukas

> So I would like to know how you came up with deprecated. 

Based on this https://github.com/EVerest/everest-core/pull/527 and the patch being in `3rd_party/josev/iso15118/evcc/controller/simulator.py`
Thank you for the clarification around the difference between `PyJosev` and `PyEvJosev` We have been making small, incremental changes directly to the installed codebase for demos, so are not intimately familiar with the structure of the source code.

@catarial just a note that the `josev` library is apparently still used on the EV side. The meeting yesterday must have referred to the station side. So you **do** still need to fix https://github.com/EVerest/everest-demo/issues/63

Abby-Wheelis:
>>Another possibility would be to subscribe to external Mqtt topics via the EvManager module and forward these topics to PyEvJosev via the internal interfaces.

>This seems like it might be the most straightforward option, since we do already use the external MQTT in the EvManger, and then the simulation of the BMS (curve calculations) could continue to take place in the car simulator.

This actually might not solve things the way that I thought, the confusion over how to use MQTT is not in the `EvManger` or in `PyEvJosev`, but rather in `ext-switchev-iso15118/evcc/controller`. Maybe there is a gap in my understanding of the role of the EV communication controller. Could the messages be received by the `EvManager` then sent to the `iso15118/evcc/controller`? Or can this controller get messages directly from the UI?

modules/EvseV2G/iso_server.cpp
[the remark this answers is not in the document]
shankari: @SebaLukas that's a great question; the answer is that this is an artifact from a previous change for the departure time/pmax calculations that we reverted, but might merit additional discussion with the community.

First, note that we do currently modify the response based on the request; we just don't do it using `sel_params`

```
int64_t departure_time_duration = req->AC_EVChargeParameter.DepartureTime;
...
.RelativeTimeInterval.duration = departure_time_duration;
```

When I first implemented departure time in EVerest, I computed the pmax in the `ChargeParameterDiscoveryRes` based on the departure time and EAmount. So the code handled three conditions (see full set of examples: https://github.com/EVerest/everest-demo/issues/64#issue-2409427125)
- no DT; current behavior
- DT specified and was short, so we needed pmax for the entire departure time
- DT specified and was long, so we can get away with lower power

But then, the feedback from other JOET experts was that the station should not modify the pmax based on the departure time, and instead, the charging curve should be controlled by the EV. So I reverted those changes for the demo, although I didn't get around to cleaning up the variables 😄 

But I think that this is the kind of topic that would be good to discuss in the energy management working group. 

How **should** the station handle the EAmount and Departure time?
- Should the SASchedule be dependent only on the available power (e.g. should it be controlled only by the OCPP charging profiles through the CSMS)
- OR
- should it also take the EV needs into account in the SASchedule?

I think that the expectation from the JOET expert was that the SASchedule should reflect network/grid constraints and the PowerDeliveryReq should reflect EV constraints. The departure time should be used only to determine the time horizon of the grid constraints. But I am not sure why the EAmount needs to be provided if that is the case.

Concrete example (for @Abby-Wheelis):
- grid -> CSMS sends a charging profile to the station with 20A or 13,800 W for the next 12 hours (43200 secs, overnight, no sun is shining)
- Car plugs in and requests EAmount of 85kWh over a departure time of 57600 (16 hours)

The two options are:
- Only dependent on available power: SASchedule with pmax = 13.8kW for 16 hours (current behavior)
   - Note that ideally, we would send two SASchedules but `iso_server.cpp` only currently only supports sending one. Accordingly to @corneliusclaussen this is because sending 2 SASchedules can crash some EVCCs
       1. pmax = 13.8kW for 12 hours
       2. pmax = default pmax for 4 hours (12 to 16 hours after session start)
- Takes EV needs into account: SASchedule with pmax $\approx$ 5312 W (85000/16) for 16 hours. Since the car is going to be there for 16 hours, it only needs 5312 W per hour to meet its needs.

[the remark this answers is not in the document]
shankari: Here's the code that takes EV needs into account in case we want to go back to it, or a variant of it
[iso_server_with_orig_dt_time_modifications.cpp.gz](https://github.com/user-attachments/files/19477746/iso_server_with_orig_dt_time_modifications.cpp.gz)

SebaLukas:
> This actually might not solve things the way that I thought, the confusion over how to use MQTT is not in the `EvManger` or in `PyEvJosev`, but rather in `ext-switchev-iso15118/evcc/controller`. Maybe there is a gap in my understanding of the role of the EV communication controller. Could the messages be received by the `EvManager` then sent to the `iso15118/evcc/controller`? Or can this controller get messages directly from the UI?

You should use the external mqtt in the `EvManager` module to subscribe some topic and forward this topic payload with the `iso15118_ev interface` to the `PyEvJosev` module. Then in the `PyEvJosev` module you are setting some variable with this payload and this variable can then be used from the josev state machine. Have a look in the `iso15118_ev` interface, the `PyEvJosev` and Josev how to forward a var from the `EvManager` to `Josev`.

modules/PyEvJosev/module.py
```diff
@@ -94,7 +94,10 @@ def _ready(self):
     # implementation handlers
 
     def _handler_start_charging(self, args) -> bool:
+        log.info('_handler_start_charging() args: %s' % str(args))
 
+        self._es.DepartureTime = args['DepartureTime']
+        self._es.EAmount = args['EAmount']
```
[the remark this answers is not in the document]
Abby-Wheelis: I have now updated the interface and the `EvManager`, if the flow looks right in the `EvManager`, I will update in the `JsEvManager` as well.

Abby-Wheelis:
I am no longer subscribing to the MQTT messages in such a direct manner, and I think the flow is improved. Right now the departure time and eamount are passed from the UI in the call to `iso_set_departure_time` that happens during the `execute_start_charging` flow. If this implementation looks good, I will copy the format in the `JsEVManager` unless that is being deprecated.

Once we have it working with the departure time and energy amount on charging start, we can think about what should happen if they are updated during a charging session, and can use the existing MQTT subscription process in the `EvManger` to recieve and handle those messages (I have already had success logging those messages locally, but don't know what proper update handling would look like). 

This is ready for another high level review.

modules/EvManager/main/car_simulation.cpp
```diff
@@ -302,3 +305,24 @@ bool CarSimulation::wait_for_real_plugin(const CmdArguments& arguments) {
     }
     return false;
 }
+
+
+bool CarSimulation::iso_set_departure_time(const CmdArguments& arguments) {
```
SebaLukas, line 310: You could solve the default setting of the two values differently.
In the EvManager manifest.yaml file, you could add `departure_time` and `e_amount` as config options and set default values when starting the session.
If you want to update the two values, you need an external mqtt subscribe function in the EvManager and e.g. an `update_departure_time_e_amount` cmd in the iso15118_ev interface. The cmd can be used to update both values in the current session.

Is there a specific reason for CarSimulation::iso_set_departure_time?

SebaLukas (commented):
Looks better already. See my comment about `iso_set_departure_time`

modules/EvManager/main/car_simulation.cpp
Abby-Wheelis: We're trying to simulate a car plugging in and a user providing this information. This way the information is "coming from the car". If it were in the config, it would be like the station were setting these values, which doesn't make sense. We can have the car send a default if the user doesn't set anything, but the user should really be the one setting these values, not the station.

SebaLukas: EvManager and PyEvJosev is independet of the charger. Except the connection via the YetiSimulator. For example we (Pionix Developer) are running both modules with a YetiEvDriver on a kids quad to simulate a real EV. So these informations are coming from the car and config. So this is **not** like the station were setting these values.

Abby-Wheelis: It should still really come from the UI, since it is coming from the user. We could put a default in the EvManager manifest.yaml for situations where there is no UI, but we also want the ability to set it from the UI at the beginning of the charging session. We can use the mqtt subscriptions to update the time/amount during the charging session once we decide how to handle renegotiation on those updates. 

I think we have very similar ideas of how to update the values, but just need to align our ideas of what the flow should be for startup, I think I'm currently using the `set` ... function similarly to how you're describing the `update` ... function. I would also be happy to set up a call to discuss.

Abby-Wheelis: I haven't committed it, because it currently does nothing (ie does not trigger any renegotiation), but this is what I have sketched in for handling updates

<details>
<summary> in car_simulatorImpl.cpp  </summary>

```
void car_simulatorImpl::handle_update_iso_departure_time(std::string& dt, std::string& eamount) {
    const CmdArguments arguments = {dt, eamount};
    EVLOG_debug << "DepartureTime: " << dt << " eamount: " << eamount;
    this->car_simulation->iso_set_departure_time(arguments)
}

...

void car_simulatorImpl::subscribe_to_external_mqtt() {
    const auto& mqtt = mod->mqtt;
   
    ...
    
    //subscribe to EAmount
    mqtt.subscribe("everest_external/nodered/evcc/confirm_eamount",
                    [this](auto eamount) {
                       std::string& dt = "";
                       handle_update_iso_departure_time(dt, eamount);
                   });
    //subscribe to the DepartureTime
    mqtt.subscribe("everest_external/nodered/evcc/confirm_departure_time",
                    [this](auto dt) {
                        std::string eamount = ""; // Remove the 'const' qualifier
                        handle_update_iso_departure_time(dt, eamount);
                   });
...
```

</details>

Hopefully that clears things up a bit as far as aligning our visions of how this should work? As the nodered UI flow stands, it calls `iso_set_departure_time` during start charging with the values in the UI. The MQTT messages are sent out when those values are updated during the session. If calling `iso_set_departure_time` from nodered as we do now is the concern, I could work on nodered to send the mqtt messages instead of making the call, but I initially worried about the manager not being fully started up by the time the messages were sent, such that the subscriptions were not established and the message would be missed.

SebaLukas: With the config it also comes from the user. But like you wrote we should combine this -> Adding both parameters to the EvManager manifest.yaml and leave the new cmd in. Then the user can decide how to set both parameters

SebaLukas: Your example code works only if you want to update both parameters before an EV session begins. If you want to update both parameters during a session, then you should add a cmd to the iso15118_ev interface to update both paramaters dynamically. Then Josev can use the updated values for the running session.

Abby-Wheelis: Correct, I wasn't sure how renegotiation during a charging session should work. 

>I haven't committed it, because it currently does nothing (ie does not trigger any renegotiation)

I will look into renegotiation, and can implement mid-charging update in this PR if it is fairly simple, but if not it may be best to begin with just the ability to set them before starting the session.

modules/EvManager/manifest.yaml
```diff
@@ -76,6 +76,14 @@ config:
     description: Support three phase
     type: boolean
     default: true
+  departure_time:
```
SebaLukas, line 79: It is not enough to add the two parameters in the manifest.yaml. You should update the `EvManager` module with `ev-cli`. In this case, however, it is enough to add both parameters to the config struct: https://github.com/EVerest/everest-core/blob/main/modules/EvManager/EvManager.hpp#L28
Then the `EvManager` module has access to both static parameters.

Abby-Wheelis:
The commit history got messy, but I was able to resolve the sign-off issues.

SebaLukas:
I'm sorry, but I think the best thing to do in this case would be to open a new PR with just your changes. I have no idea what would happen if we squash and merge all the commits.

PR #1021, opened 2025-01-22
feat: add TPM2 support for EvseV2G TLS server private key

james-ctc:
## Describe your changes

feat: add TPM2 support for EvseV2G TLS server private key
Note arbitrary sign/verify is not currently supported in the OpenSSL utility layer (openssl_util.cpp)

TPM2 support requires -DUSING_TPM2 to cmake

### DIFF
--- a/lib/staging/tls/openssl_util.cpp
+++ b/lib/staging/tls/openssl_util.cpp
@@ -407,6 +407,7 @@ pkey_ptr load_private_key(const char* filename, const char* password) {
     OpenSSLProvider provider;
 
     if (custom_key) {
+        // set global provider to custom settings
         provider.set_global_mode(OpenSSLProvider::mode_t::custom_provider);
     } else {
         provider.set_global_mode(OpenSSLProvider::mode_t::default_provider);
@@ -425,6 +426,7 @@ pkey_ptr load_private_key(const char* filename, const char* password) {
     }
 
     if (custom_key) {
+        // reset global provider back to default settings
         provider.set_global_mode(OpenSSLProvider::mode_t::default_provider);
     }
 
--- a/lib/staging/tls/openssl_util.cpp
+++ b/lib/staging/tls/openssl_util.cpp
@@ -11,6 +11,8 @@
 
 #include "openssl_util.hpp"
 
+#include <evse_security/crypto/openssl/openssl_provider.hpp>
+
 #include <openssl/bio.h>
 #include <openssl/crypto.h>
 #include <openssl/ecdsa.h>
@@ -130,6 +132,7 @@ template <> bool sha(const void* data, std::size_t len, openssl::sha_512_digest_
 } // namespace
 
 namespace openssl {
+using evse_security::OpenSSLProvider;
 
 DER::DER(std::size_t size) : DER(nullptr, size) {
 }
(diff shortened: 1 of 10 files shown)


lib/staging/tls/openssl_util.cpp
```diff
@@ -399,6 +423,11 @@ pkey_ptr load_private_key(const char* filename, const char* password) {
         }
         BIO_free(bio);
     }
+
+    if (custom_key) {
+        provider.set_global_mode(OpenSSLProvider::mode_t::default_provider);
+    }
```
SebaLukas, line 429: Is this correct? In line 409-410 the provider is set to a custom_provider if `custom_key` is `true`. Here to a default_provider. Or is this here a provider reset?

lib/staging/tls/tls.cpp
```diff
@@ -413,11 +417,38 @@ ssl_result_t ssl_shutdown(SSL* ctx) {
  * \return true when successful
  * \note required will be true for a TLS server and can be false for a TLS client
  */
-bool configure_ssl_ctx(SSL_CTX* ctx, const char* ciphersuites, const char* cipher_list,
+bool configure_ssl_ctx(bool is_server, SSL_CTX*& ctx, const char* ciphersuites, const char* cipher_list,
```
SebaLukas, line 420: `SSL_CTX*&` <- `*&` ? Reference to a pointer?

SebaLukas (approved):
Looks good :+1:
Do we have some documentation or example where `-DUSING_TPM2` is already used? If not, then we should make a note of how to activate TPM support.

james-ctc:
> Looks good 👍 Do we have some documentation or example where `-DUSING_TPM2` is already used? If not, then we should make a note of how to activate TPM support.

 `-DUSING_TPM2` is a livevse-security flag. It should be already documented there.
 It is only used here to include or exclude the TMP2 unit tests. (same as libocpp).

lib/staging/tls/openssl_util.cpp
james-ctc: This is a provider reset - there are probably better ways of doing this but they would require changes to libevse-security ...
another alternative is to use PEM_read_bio_PrivateKey_ex() where the context and propquery can be specified.
When the approach was used in TLS to the CSMS there were occasional problems with creating and removing contexts. The approach has been to minimise creating short lived contexts.

lib/staging/tls/tls.cpp
james-ctc: previously the SSL context was created by the caller of configure_ssl_ctx(). Context creation depends on whether TPM2 is being used or not so configure_ssl_ctx() was changed to return the new context.
An alternative would be SSL_CTX ** and then the code in configure_ssl_ctx() would need to change every use of the variable and an additional nullptr check required.

ultimately it is converted to a unique_ptr but only on success.
see Server::init_ssl()

lib/staging/tls/openssl_util.cpp
james-ctc: done!

AssemblyJohn: As a note, there were some issues with the previous approach of re-creating the TLS context, and the context switch that we are using here is much more stable than the initial approach that caused instability in the TPM.

lib/staging/tls/tests/tls_connection_test.hpp
```diff
@@ -285,5 +285,40 @@ class TlsTest : public testing::Test {
...
+        auto& ref0 = server_config.chains.emplace_back();
+        ref0.certificate_chain_file = "tpm_pki/server_chain.pem";
+        ref0.private_key_file = "tpm_pki/server_priv.pem";
+        ref0.trust_anchor_file = "tpm_pki/server_root_cert.pem";
+        ref0.ocsp_response_files = {"ocsp_response.der", "ocsp_response.der"};
+        // auto& ref1 = server_config.chains.emplace_back();
+        // ref1.certificate_chain_file = "alt_server_chain.pem";
+        // ref1.private_key_file = "alt_server_priv.pem";
+        // ref1.trust_anchor_file = "alt_server_root_cert.pem";
+        // ref1.ocsp_response_files = {"ocsp_response.der", "ocsp_response.der"};
```
lategoodbye, line 303: Is it intended to submit all these commented out stuff?

james-ctc: yes

lategoodbye: Since this is not self-explaining, could you please add a comment about the intention of this?
Personally i don't think such dead code, because there is no CI coverage in case of refactoring.

james-ctc: This is test code, not part of the deployed code.
Those lines will be uncommented when the tests that use them are added.
It saves having to remember what the changes are.

PR #1054, opened 2025-02-17
Fixing DcPowerSupplySimulator race condition

SebaLukas:
## Describe your changes
Adding missing initialization. Now publishing the `voltage_current` var the voltage and current should be 0.0 and not null. 

In rare cases the DcPowerSupplySimulator publishes voltage and current not as double but as null. In this case EVerest will crash because nlohmann can not convert null to a number

### DIFF
--- a/modules/simulation/DCSupplySimulator/main/power_supply_DCImpl.hpp
+++ b/modules/simulation/DCSupplySimulator/main/power_supply_DCImpl.hpp
@@ -65,7 +65,7 @@ private:
     double settings_connector_import_voltage{0.0};
     double settings_connector_max_export_current{0.0};
     double settings_connector_max_import_current{0.0};
-    types::power_supply_DC::Mode mode;
+    types::power_supply_DC::Mode mode{types::power_supply_DC::Mode::Off};
     double connector_voltage{0.0};
     double connector_current{0.0};
     double energy_import_total{0.0};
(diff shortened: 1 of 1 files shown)


modules/simulation/DCSupplySimulator/main/power_supply_DCImpl.hpp
```diff
@@ -61,15 +61,15 @@ class power_supply_DCImpl : public power_supply_DCImplBase {
...
     // insert your private definitions here
-    double settings_connector_export_voltage;
-    double settings_connector_import_voltage;
-    double settings_connector_max_export_current;
-    double settings_connector_max_import_current;
+    double settings_connector_export_voltage{0.0};
+    double settings_connector_import_voltage{0.0};
+    double settings_connector_max_export_current{0.0};
+    double settings_connector_max_import_current{0.0};
     types::power_supply_DC::Mode mode;
```
hikinggrass, line 68: This should probably also be initialized with
`types::power_supply_DC::Mode::Off`

hikinggrass (approved):

modules/simulation/DCSupplySimulator/main/power_supply_DCImpl.hpp
SebaLukas: Nice catch! I will add this, too :)

PR #1059, opened 2025-02-20
Adding IEM-DCR module

jherbert-isa:
## Describe your changes
This pull request is about adding a new module for communicating with Isabellenhuette IEM-DCR powermeter.

- [x] I read the [contribution documentation](https://github.com/EVerest/EVerest/blob/main/CONTRIBUTING.md) and made sure that my changes meet its requirements. Remark: Tests were performed with embedding the module in existing run-sil-dc.sh environment. If a fully automated test is required, please let me know.

### DIFF
--- a/modules/IsabellenhuetteIemDcr/main/powermeterImpl.cpp
+++ b/modules/IsabellenhuetteIemDcr/main/powermeterImpl.cpp
@@ -1,0 +1,272 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Pionix GmbH and Contributors to EVerest
+
+#include "powermeterImpl.hpp"
+#include "httpClient.hpp"
+#include <chrono>
+#include <string>
+#include <thread>
+
+namespace module {
+namespace main {
+
+void powermeterImpl::init() {
+    // Check Config values (essential plausibility checks)
+    check_config();
+
+    // Dependency injection pattern: Create the HTTP client first,
+    // then move it into the controller as a constructor argument
+    auto http_client = std::make_unique<HttpClient>(mod->config.ip_address, mod->config.port_http);
+
+    // Create controller object
+    this->controller = std::make_unique<IsaIemDcrController>(
+        std::move(http_client),
+        IsaIemDcrController::SnapshotConfig{
+            mod->config.timezone, mod->config.timezone_handle_DST, mod->config.datetime_resync_interval,
+            mod->config.resilience_initial_connection_retry_delay, mod->config.resilience_transaction_request_retries,
+            mod->config.resilience_transaction_request_retry_delay, mod->config.CT, mod->config.CI,
+            mod->config.TT_initial, mod->config.US});
+}
+
+void powermeterImpl::ready() {
+    // Start the live_measure_publisher thread, which periodically publishes the live measurements of the device
+    this->live_measure_publisher_thread = std::thread([this] {
+        while (true) {
+            try {
+                // Wait for at least one second (more if handle_start_transaction() or handle_stop_transaction() active)
+                do {
+                    std::this_thread::sleep_for(std::chrono::milliseconds(1000));
+                } while (start_transaction_running.load() == true || stop_transaction_running.load() == true);
+                // Init if needed
+                if (is_initialized.load() == false) {
+                    is_initialized.store(this->controller->init());
+                    if (is_initialized.load() == true) {
+                        // Publish public key once after init
+                        std::this_thread::sleep_for(std::chrono::milliseconds(1000));
+                        this->publish_public_key_ocmf(this->controller->get_publickey(false));
+                    } else {
+                        if (!this->error_state_monitor->is_error_active("powermeter/CommunicationFault",
+                                                                        "Communication timed out")) {
+                            auto error = this->error_factory->create_error(
+                                "powermeter/CommunicationFault", "Communication timed out",
+                                "This error is raised due to communication timeout");
+                            raise_error(error);
+                        }
+                        EVLOG_warning << "Connecting to IEM-DCR failed. Retry in "
+                                      << mod->config.resilience_initial_connection_retry_delay << " milliseconds";
+                        std::this_thread::sleep_for(
+                            std::chrono::milliseconds(mod->config.resilience_initial_connection_retry_delay));
+                    }
+                } else {
+                    // Publish metervalue node (named powermeter in EVerest) and update status information
+                    auto meter_value_response = this->controller->get_metervalue();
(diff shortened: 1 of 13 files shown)


florinmihut (changes requested):
We are not accepting different formatting rules or lint rules. 
Tabs are not allowed, etc.
Please fix this

modules/IsabellenhuetteIemDcr/main/isabellenhuette_IemDcr_controller.cpp
```diff
@@ -0,0 +1,408 @@
...
+	int offsetHours = std::stoi(snapshotConfig.timezone.substr(1, 2));
+    int offsetMinutes = std::stoi(snapshotConfig.timezone.substr(3, 2));
+    auto timeOffset = std::chrono::hours(offsetHours) + std::chrono::minutes(offsetMinutes);
+    std::time_t nowWithOffset;
+	if(signChar == '+') {
+		nowWithOffset = std::chrono::system_clock::to_time_t(now + timeOffset);
+	} else if(signChar == '-') {
+		nowWithOffset = std::chrono::system_clock::to_time_t(now - timeOffset);
+	} else {
+		throw std::runtime_error("manifest.yaml: Format of timezone not supported. Expected: something like \"+0100\".");
```
barsnick, line 373: `std::stoi()` can also throw, e.g. if it finds only letters, so the `catch` block should be extended.

And the parsing of the config perhaps shouldn't be done on every call of the helper. I suggest that parsing to a member variable `timeOffset` (and failing/exiting on unexpected format) should happen in `init()`.

BTW, EVerest uses snake_case for variable names, CamelCase for classes and structs.

And you can use more `const` for intermediate calculations such as `now`, `signChar`, `offsetHours`, `offsetMinutes`, to help understand that the value will not be modified again (and assist the compiler in optimizing).

jherbert-isa: Thanks for reviewing. Is CamelCase also to be used for method names inside classes?

If it is OK, I will omit adapting class names to CamelCase in order to avoid changing file names in this PR.

barsnick: Yeah, I wonder whether there's a style guide somewhere. Do check the other modules in everest-core (though some do still make use of deprecated style).

Variables and member functions use snake_case, as do namespaces. Classes use CamelCase (sometimes camelCase, for some reason), as do types.

Don't worry about renaming files, GitHub covers that nicely, i.e. in terms of reviewing.

As @florinmihut mentioned, do remove all those extra files, and please reformat the code using [the top-level clang-format config file](https://github.com/EVerest/everest-core/blob/main/.clang-format) for readability. Some of your indentation is "wild".

Remember to sign off _all_ commits in the PR, in order to pass the DCO check.

jherbert-isa:
I've added some commits with the requested changes. Static code analysis now only shows "notice" findings which all seems to be false-positives. One question regarding DCO: As some commits of the past were not performed with signoff, I've added a `past_commits.txt` file as described [here](https://github.com/lf-energy/tac/blob/main/process/contribution_guidelines.md#signoff-for-commits-where-the-dco-signoff-was-missed). Shall I instead do a git rebase (e.g. `git rebase HEAD~15 --signoff`)? Although I'm the only one implementing the driver, I'm not sure if this will fit into your already started review process.

modules/IsabellenhuetteIemDcr/main/powermeterImpl.cpp
```diff
@@ -0,0 +1,281 @@
...
+
+    return return_value;
+}
+
+void powermeterImpl::check_config() {
+    if (mod->config.ip_address.length() <= 0) {
+        EVLOG_error << "Incorrect module config: parameter ip_address is empty." << std::endl;
+        throw std::runtime_error("ip_address invalid. Please check configuration.");
+    }
+    if (mod->config.port_http < 0) {
```
mhei, line 228: You can here use the comfort of the manifests to specify minimum and maximum values. Then you don't need to check this stuff here.

modules/IsabellenhuetteIemDcr/manifest.yaml
```diff
@@ -0,0 +1,58 @@
+description: Module implements Isabellenhuette IEM-DCR power meter driver, connecting via HTTP/REST
+config:
+  ip_address:
+    description: IPv4 Address of the power meter API.
+    type: string
+    default: "192.168.60.12"
+  port_http:
+    description: HTTP-Port of the power meter API.
+    type: integer
```
mhei, line 9: As mentioned above, just add
```
    minimum: 1
    maximum: 65535
```
so that EVerest bails out if user configures stuff not in this range.

modules/IsabellenhuetteIemDcr/manifest.yaml
```diff
@@ -0,0 +1,58 @@
...
+    default: 80
+  timezone:
+    description: The timezone offset information according to ISO8601 (version without colon).
+    type: string
+    default: "+0100"
```
mhei, line 14: This looks conceptual wrong to me. While the timezone as-that is bound to the location of the charger (and these timezones are usually referred to via strings, e.g. "Europe/Berlin"), the string you used here must  actually change between daylight saving periods. You don't want users to reconfigure the station 2 times per year, do you?

I did not yet understand from the remaining code, whether the timezone can be changed for each charging session or whether this needs a reboot of the meter?

But in any case, I strongly recommend to determine this value on-the-fly from the underlying Linux system at runtime. You can/should expect that users running EVerest should provide a proper configured environment.

modules/IsabellenhuetteIemDcr/manifest.yaml
```diff
@@ -0,0 +1,58 @@
...
+    type: boolean
+    default: false
+provides:
+  main:
+    description: This is the main unit of the module
+    interface: powermeter
+metadata:
+  license: https://opensource.org/licenses/Apache-2.0
+  authors:
+    - Josef Herbert, josef.herbert@isabellenhuette.com
```
mhei, line 58: I'd recommend the typical e-mail address format: Your Name <your.name@...>

modules/IsabellenhuetteIemDcr/manifest.yaml
jherbert-isa: Thanks for bringing this up. I've added a feature for automatic handling of DST in the new commit.

florinmihut (dismissed):

hikinggrass (approved):

PR #1067, opened 2025-02-26
Reservation Handling in EvseManager

Pietfried:
## Describe your changes
Changed reservation handling in EvseManager:
* Signaling reservation end after Signaling Disabled event. This allows OCPPs state machine to move from
**Reserved->Unavailable** instead of **Reserved->Available->Unavailable**
* Fixed issue that reservation was cancelled on every SessionFinished event

## Conflicting requirements of OCPP2.0.1 and OCPP1.6

OCPP2.0.1 and OCPP1.6 use a shared implementation of the Reservation handling (implemented by the Auth module). We identified conflicting requirements between OCPP2.0.1 and OCPP1.6:
OCPP1.6 requires: 
> A reservation SHALL be terminated on the Charge Point when either (1) a transaction is started

OCPP2.0.1 requires:
> H03.FR.09 When an idToken or groupIdToken is presented that matches a reservation, Charging Station SHALL consider the reservation to be used (consumed)

This means, for OCPP1.6 a reservation is consumed/terminated when a transaction starts (Authorization present + EV plugged in) while in OCPP2.0.1 valid authorization is sufficient (e.g. swipe reserved RFID without EV plugin) to consume/terminate the reservation. 

Since the handling of the reservations is mainly implemented in the EvseManager and the Auth module (which are unaware of the OCPP version) and not in libocpp, this can't be solved properly for both versions at the same time.

For now and with these changes, the requirement of OCPP1.6 is applied and therefore  H03.FR.09 of OCPP2.0.1 is violated.

[Companion PR in libocpp](https://github.com/EVerest/libocpp/pull/998)

### DIFF
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -67,7 +67,7 @@ libevse-security:
 # OCPP
 libocpp:
   git: https://github.com/EVerest/libocpp.git
-  git_tag: bug/ocpp16-reservation-inoperative-multiple-state-change
+  git_tag: 6533694e8a25925985079bda33d5610f65c6d58f
   cmake_condition: "EVEREST_DEPENDENCY_ENABLED_LIBOCPP"
 # Josev
 Josev:
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -67,7 +67,7 @@ libevse-security:
 # OCPP
 libocpp:
   git: https://github.com/EVerest/libocpp.git
-  git_tag: 04059e2cb6ea34029b661b6803dce49727ea65c3
+  git_tag: bug/ocpp16-reservation-inoperative-multiple-state-change
   cmake_condition: "EVEREST_DEPENDENCY_ENABLED_LIBOCPP"
 # Josev
 Josev:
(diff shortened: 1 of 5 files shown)


maaikez (approved):
I tested with OCTT:
Test TC_050_2_CS is failing and was not in the past. It asks to plugin on the car side and then on the EV side and expects 'Occupied', but it stays to 'Preparing' and never sends an 'Occupied' status notification. This test was working in the past, but I don't know if this is on our side or that something changed in OCTT.

The point you mention about when to cancel the reservation might let TC_H_24_CS fail, but I did not look into it. It also already failed in the past.

There are no tests that do not pass after our changes which passed in the past.

dependencies.yaml
```diff
@@ -67,7 +67,7 @@ libevse-security:
 # OCPP
 libocpp:
   git: https://github.com/EVerest/libocpp.git
-  git_tag: 04059e2cb6ea34029b661b6803dce49727ea65c3
+  git_tag: bug/ocpp16-reservation-inoperative-multiple-state-change
```
hikinggrass, line 70: TODO: change to rev on main

hikinggrass (approved):

SebaLukas (approved):
:+1:

PR #1079, opened 2025-03-06
Session cost add message

maaikez:
## Describe your changes

Do not use the display message interface to send session cost messages, but use a specific session cost message type.


modules/DummyTokenValidator/main/auth_token_validatorImpl.cpp
```diff
@@ -21,8 +21,8 @@ auth_token_validatorImpl::handle_validate_token(types::authorization::ProvidedId
     types::authorization::ValidationResult ret;
     ret.authorization_status = types::authorization::string_to_authorization_status(config.validation_result);
     ret.reason = types::authorization::TokenValidationStatusMessage();
-    ret.reason->messages = std::vector<types::display_message::MessageContent>();
-    types::display_message::MessageContent content;
+    ret.reason->messages = std::vector<types::text_message::MessageContent>();
```
Pietfried, line 24: This can be dropped since the vector is default initialized
```suggestion
```

types/display_message.yaml
```diff
@@ -1,6 +1,14 @@
 description: >-
   Message to show on a display
 types:
+  IdentifierType:
+    description: >-
+      The type of 'identifier_id'. If omitted, type will be session id.
```
Pietfried, line 6: This type definition should not refer `identifier_id` and what happens if omitted when used as part of another type.
```suggestion
      Types of identifiers.
```

dependencies.yaml
```diff
@@ -67,7 +67,7 @@ libevse-security:
 # OCPP
 libocpp:
   git: https://github.com/EVerest/libocpp.git
-  git_tag: 6533694e8a25925985079bda33d5610f65c6d58f
+  git_tag: feature/session-cost-add-message
```
Pietfried, line 70: TODO: update when libocpp PR is merged.

Pietfried (changes requested):

modules/DummyTokenValidator/main/auth_token_validatorImpl.cpp
maaikez: 'messages' is optional, so it is initialized with std::nullopt.

types/display_message.yaml
maaikez: Already clear :). Changed it.

Pietfried (dismissed):
lgtm

interfaces/session_cost.yaml
```diff
@@ -3,6 +3,10 @@ description: >-
   provides cost of one session. If we have more than one EVESEID, we need to 
   instantiate this interface for each EVSE.
 vars:
+  session_cost_message:
+    description: Message to show to the user with information about the session cost.
+    type: object
+    $ref: /session_cost#/SessionCostMessage
```
dorezyuk, line 9: Does it make sense to send the session cost and session cost message independently? Would it not be easier to extend the existing session cost with an optional "text" field so the price and the info for a display are delivered at once?
also the name `session_cost_message` is really not ideal since we already have `session_cost` as message

Pietfried: I do agree that the naming is not ideal. The messages are seperated because they serve different purposes and are triggered at different times from the CSMS. This is required by the california pricing requirements and OCPP2.0.1. Im not sure if all required data is available at every time to actually combine the messages. @maaikez what do you think?

lib/staging/ocpp/ocpp_conversions.cpp
```diff
@@ -255,9 +272,9 @@ types::session_cost::SessionCost create_session_cost(const ocpp::RunningCost& ru
     cost.status = to_everest_running_cost_state(running_cost.state);
     cost.qr_code = running_cost.qr_code_text;
     if (running_cost.cost_messages.has_value()) {
-        cost.message = std::vector<types::display_message::MessageContent>();
+        cost.message = std::vector<types::text_message::MessageContent>();
         for (const ocpp::DisplayMessageContent& message : running_cost.cost_messages.value()) {
-            types::display_message::MessageContent m = to_everest_display_message_content(message);
+            types::text_message::MessageContent m = to_everest_display_message_content(message);
```
MarzellT, line 277: Just a minor note from me but wouldn't:
```
const auto m = to_everest_display_message_content(message);
```
be cleaner?

lib/staging/ocpp/ocpp_conversions.cpp
```diff
@@ -61,64 +61,64 @@ ocpp::v2::MessageStateEnum to_ocpp_201_display_message_state(const types::displa
     throw std::out_of_range("Could not convert types::display_message::MessageStateEnum to ocpp::v2::MessageStateEnum");
 }
 
-types::display_message::MessageFormat
+types::text_message::MessageFormat
```
MarzellT, line 64: I think you are explicitly using `types::text_message::MessageFormat` and `types::display_message::IdentifierType` everywhere. Wouldn't it be better to just declare them with `using`?

maaikez: That is possible but we have for example `types::display_message::IdentifierType` and `ocpp::IdentifierType` in this file. I think it is more confusing if a 'using' is added to be hones.

interfaces/session_cost.yaml
maaikez: Exactly what Piet says. In OCPP 2.0.1 the CSMS will send a cost message in a transaction message, without the pricing information (just as separate message). Same is done with a DataTransfer message in OCPP 1.6, only a message without pricing information. So for OCPP it doesn't make sense to have the whole `session_cost`, is we don't use all fields except for the messages.

I'm fine changing the name, is `SessionCostTextMessage` better or do you have a better idea?

maaikez: I thought tariff_message would probably be better.

Pietfried (approved):

PR #1093, opened 2025-03-17
fix: Yeti Simulator was generating invalid date and time strings

james-ctc:
this fix presents a standard format

## Describe your changes
fix: Yeti Simulator was generating invalid date and time strings
example was: "2025-03-17:15:07:46GMT"
It should be "2025-03-17T15:07:46Z"


modules/simulation/YetiSimulator/util/util.cpp
hikinggrass, line 1: Maybe not for this PR, but I think this whole file could be replaced with a call to https://github.com/EVerest/everest-framework/blob/main/lib/date.cpp

hikinggrass (approved):

maaikez (approved):

Pietfried (approved):

SebaLukas (approved):
Oops, I apparently didn't pay attention to that part in the review :sweat_smile: Thanks for discovering and fixing it

PR #1104, opened 2025-03-27
feat(ocpp16): Add test for connector 0 being replaced in db

mlitre:
Prior to the fix in libocpp, when a new profile would overwrite the connector id 0, the old profile would remain in the db and not be deleted. This new test makes sure that the db is now cleared properly.


Pietfried (approved):
lgtm!

Please update the commit id once https://github.com/EVerest/libocpp/pull/1021 is merged

PR #1105, opened 2025-03-27
Fix Auth module race condition on timeout

Pietfried:
## Describe your changes
Fixed race condition in Auth module that could be caused by rescheduling the timeout_timer when the execution of the task has already been started.

It could happen in the Auth module that:
1. EV was plugged in causing timeout_timer to schedule plug in timeout task
2. `on_token` was called from a seperate thread right before the timeout lambda fired
3. timeout lambda fires while `on_token` is executing. `on_token` holds the `lk`, timeout lambda is waiting for the `lk`
4. on_token rescheduled `timeout_timer.timeout(...)` potentially corrupting the memory of the timeout task that is waiting for the `lk` 

This change checks if a timeout task is still in progress and waits for it to finish before a new timeout task is scheduled.

Note: This is a workaround in the Auth module and potentially needs to be addressed in libtimer. @hikinggrass  


modules/Auth/lib/AuthHandler.cpp
```diff
@@ -562,14 +562,20 @@ void AuthHandler::notify_evse(int evse_id, const ProvidedIdToken& provided_token
                               validation_result.parent_id_token};
         this->evses.at(evse_id)->identifier.emplace(identifier);
 
+        // wait for potentially running timeout task to finish execution
+        this->cv.wait(lk, [&evse = this->evses.at(evse_id)]() { return !evse->timeout_in_progress.load(); });
```
AssemblyJohn, line 566: In the case we are waiting for the execution finish, and calling 'stop', shouldn't we be consistent and in each place where we call the timeout_timer.stop() wait for the task's execution?

For example before: https://github.com/EVerest/everest-core/blob/f66b4488bf42dc3bb25e2a4f3fa9a9e56bc60d77/modules/Auth/lib/AuthHandler.cpp#L746

AssemblyJohn (approved):

corneliusclaussen (approved):

PR #1109, opened 2025-03-30
Introduce charger information interface and example implementation

mhei:
## Describe your changes

This PR introduces a new 'charger information' interface and corresponding types. The idea is that this interface is a common base to bundle some meta/high-level information about the charger itself.
To illustrate the usage, the existing API and OCPP 1.6 modules are extended while keeping existing functionality and thus backwards compatibility if this new approach is not needed in existing installations.

Note: this PR depends on https://github.com/EVerest/libocpp/pull/1026

### DIFF
--- a/modules/API/API.cpp
+++ b/modules/API/API.cpp
@@ -736,6 +736,10 @@ void API::init() {
 void API::ready() {
     invoke_ready(*p_main);
 
+    if (not r_charger_information.empty()) {
+        this->charger_information = r_charger_information.at(0)->call_get_charger_information();
+    }
+
     std::string var_active_errors = this->api_base + "errors/var/active_errors";
     this->api_threads.push_back(std::thread([this, var_active_errors]() {
         auto next_tick = std::chrono::steady_clock::now();
--- a/modules/ChargerInfo/charger_info.yaml
+++ b/modules/ChargerInfo/charger_info.yaml
@@ -1,0 +1,6 @@
+manufacturer: "Pionix GmbH"
+manufacturer_url: "https://www.pionix.com/"
+vendor: "Pionix"
+model: "BelayBox"
+serial: "0123"
+firmware_version: "0.1.2"
--- a/modules/ChargerInfo/main/charger_informationImpl.cpp
+++ b/modules/ChargerInfo/main/charger_informationImpl.cpp
@@ -1,0 +1,69 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Pionix GmbH and Contributors to EVerest
+
+#include "charger_informationImpl.hpp"
+#include <fstream>
+#include <iostream>
+#include <string>
+#include <vector>
+
+namespace module {
+namespace main {
+
+void charger_informationImpl::init() {
+}
+
+void charger_informationImpl::ready() {
+    json info(handle_get_charger_information());
+
+    EVLOG_debug << "ChargerInformation: " << info.dump();
+}
+
+std::string charger_informationImpl::load_fw_version_from_file(const std::string& fn) {
+    std::ifstream fw_version_file(fn);
+    std::string first_line;
+
+    if (!fw_version_file)
+        return "";
+
+    if (!std::getline(fw_version_file, first_line))
+        return "";
+
+    return first_line;
+}
+
+types::charger_information::ChargerInformation charger_informationImpl::handle_get_charger_information() {
+    std::vector<std::string> keys = {"vendor",           "model",     "serial",   "friendly_name", "manufacturer",
+                                     "manufacturer_url", "model_url", "model_no", "revision",      "board_revision",
+                                     "firmware_version"};
+    json info = {};
+
+    // iterate over all linked key-value stores and merge all items,
+    // when a key exists in more than one kvs then the last one wins
+    for (const auto& kvs : mod->r_kvs) {
+        for (const auto k : keys)
+            if (kvs->call_exists(k)) {
+                const auto v = kvs->call_load(k);
+                info[k] = std::get<std::string>(v);
+            }
+    }
+
+    // finally check whether we should load the firmware version from a simple plain text file
+    if (!mod->config.firmware_version_file.empty()) {
+        info["firmware_version"] = load_fw_version_from_file(mod->config.firmware_version_file);
+    }
+
+    // generate fallback friendly_name
+    if (info.contains("vendor") and info.contains("model") and info.contains("serial") and
+        not info.contains("friendly_name")) {
+        info["friendly_name"] = info["vendor"].get<std::string>() + " " + info["model"].get<std::string>() + " [" +
+                                info["serial"].get<std::string>() + "]";
+    }
+
+    EVLOG_debug << "ChargerInformation: " << info.dump();
+
+    return info;
+}
+
+} // namespace main
+} // namespace module
--- a/modules/OCPP/OCPP.cpp
+++ b/modules/OCPP/OCPP.cpp
@@ -901,6 +901,13 @@ void OCPP::ready() {
         this->evse_ready_cv.wait(lk);
     }
 
+    // if charger information interface is connected, override only these specific properties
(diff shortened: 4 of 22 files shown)


modules/ChargerInfo/main/charger_informationImpl.cpp
```diff
@@ -0,0 +1,69 @@
...
+
+    // iterate over all linked key-value stores and merge all items,
+    // when a key exists in more than one kvs then the last one wins
+    for (const auto& kvs : mod->r_kvs) {
+        for (const auto k : keys)
+            if (kvs->call_exists(k)) {
+                const auto v = kvs->call_load(k);
+                info[k] = std::get<std::string>(v);
+            }
+    }
```
hikinggrass, line 49: Is your intention to have this information distributed in multiple modules providing the `kvs` interface?
Otherwise storing this as a complete json object may be a bit more efficient (you could still merge this afterwards)

modules/YamlStore/manifest.yaml
```diff
@@ -0,0 +1,13 @@
+description: Provides a (read-only) key-value store interface, backed by a simple YAML file.
```
hikinggrass, line 1: Maybe this can be reflected in the module name itself. So calling it `ReadOnlyYamlStore`, but I'm not 100% sure about this

hikinggrass, line 1:
```suggestion
description: Provides a read-only key-value store interface, backed by a simple YAML file.
```

modules/YamlStore/manifest.yaml
```diff
@@ -0,0 +1,13 @@
...
+config:
+  file:
+    description: Path to the YAML file used as persistent key-value store.
```
hikinggrass, line 4:
```suggestion
    description: Path to the YAML file used as a read-only key-value store.
```

modules/YamlStore/manifest.yaml
```diff
@@ -0,0 +1,13 @@
...
+    type: string
+provides:
+  main:
+    interface: kvs
+    description: Provides the key-value interface
```
hikinggrass, line 9:
```suggestion
    description: Provides a read-only key-value interface
```

types/charger_information.yaml
```diff
@@ -0,0 +1,82 @@
...
+      firmware_version:
+        description: >-
+          A string containing the current running firmware version of the complete system.
+          This is typically not the EVerest version itself, but a firmware version which appears
+          e.g. on the web frontend and as part of the filenames the customer can download from
+          the manufacturer website.
+          This is the only property which is typically not static for the product lifetime,
+          but determined at runtime.
+        type: string
+
```
hikinggrass, line 82: missing newline, instead there's additional whitespace
```suggestion

```

modules/OCPP/OCPP.cpp
```diff
@@ -901,6 +901,13 @@ void OCPP::ready() {
         this->evse_ready_cv.wait(lk);
     }
 
+    // if charger information interface is connected, override only these specific properties
+    // which were loaded from configuration file(s)
+    if (!this->r_charger_information.empty()) {
+        auto ci = this->r_charger_information.at(0)->call_get_charger_information();
+        this->charge_point->update_chargepoint_information(ci.vendor, ci.model, ci.serial, ci.firmware_version);
```
hikinggrass, line 908: In 1.6 (and 2.0.1 as well as 2.1) there's also some `modem` information you could transmit in the `BootNotification`, namely `ICCID` and `IMSI`. Maybe we want to set this information as well

types/charger_information.yaml
```diff
@@ -0,0 +1,82 @@
...
+      manufacturer_url:
+        description: >-
+          An URL to the vendor website, ideally not a deep link so that it is available for
+          the whole product lifetime.
+        type: string
+      model_url:
+        description: >-
+          An URL to the model website (if any); as above, ideally not a deep link.
+        type: string
+      model_no:
```
hikinggrass, line 57:
```suggestion
      model_number:
```

types/charger_information.yaml
```diff
@@ -0,0 +1,82 @@
...
+        description: >-
+          A model number as string (if any).
+        type: string
+      revision:
```
hikinggrass, line 61:
```suggestion
      product_revision:
```

or 

```suggestion
      model_revision:
```

hikinggrass (changes requested):

modules/ChargerInfo/charger_info.yaml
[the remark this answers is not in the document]
mhei: To be honest, I don't think the current naming of the keys is optimal, but I've already checked that there is a suitable replacement in the newer naming scheme. I added an explanation as to why there is no `vendor_url` for example. Since I don't know how many users the current file format has, I didn't want to touch it at all to maintain backward compatibility. I'd keep in for one LTS release, and then we could drop it.

modules/ChargerInfo/main/charger_informationImpl.cpp
mhei: The idea is to have different kvs modules which provide different keys and/or same keys but the later wins. Yes, having it as complete object and then merge all together would be an option, but the current kvs interface does not support such kind of access. To be honest, I'm not too worried about performance here, since this is almost static data and I assume that the number of hits is manageable.

hikinggrass (dismissed):
Sorry for the delay in review :sweat_smile: if you (squash &) rebase the PR we can merge this :+1:

hikinggrass:
> Sorry for my side, too. I was too busy with other things. I've squashed and rebased as requested, but compiled-tested only for now after rebase.

Thanks a lot! Looks like there are still 2 minor clang-format related things that the ci picks up. Once they are fixed I think this is ready to go for now

types/charger_information.yaml
```diff
@@ -0,0 +1,89 @@
...
+    properties:
+      vendor:
+        description: >-
+          Name of vendor, not including any legal form. Is typically used with `model` to
+          form a unique product name aka `vendor` + `single whitespace` + `model`.
+          Example: Pionix
+        type: string
+      model:
+        description: >-
+          A human-friedly name of the model, aka the product name; without vendor name,
```
barsnick: Typo "friedly"
```suggestion
          A human-friendly name of the model, aka the product name; without vendor name,
```

modules/API/API.cpp
```diff
@@ -709,13 +709,18 @@ void API::init() {
     std::string var_info = this->api_base + "info/var/info";
 
     if (this->config.charger_information_file != "") {
-        auto charger_information_path = std::filesystem::path(this->config.charger_information_file);
-        try {
-            this->charger_information = Everest::load_yaml(charger_information_path);
+        if (not r_charger_information.empty()) {
+            EVLOG_warning << "The configured charger information file (" << this->config.charger_information_file << ") "
+                          << "is ignored in favor of the charger information interface connection.";
```
hikinggrass:
```suggestion
            EVLOG_warning << "The configured charger information file (" << this->config.charger_information_file
                          << ") "
                          << "is ignored in favor of the charger information interface connection.";
```

modules/API/API.cpp
```diff
@@ -709,13 +709,18 @@ void API::init() {
...
+            auto charger_information_path = std::filesystem::path(this->config.charger_information_file);
+            try {
+                this->charger_information = Everest::load_yaml(charger_information_path);
 
-        } catch (const std::exception& err) {
-            EVLOG_error << "Error parsing charger information file at " << this->config.charger_information_file << ": "
-                        << err.what();
+            } catch (const std::exception& err) {
+                EVLOG_error << "Error parsing charger information file at " << this->config.charger_information_file << ": "
+                            << err.what();
```
hikinggrass:
```suggestion
                EVLOG_error << "Error parsing charger information file at " << this->config.charger_information_file
                            << ": " << err.what();
```

hikinggrass (approved):
I think those suggestions should cover the failing clang-format lint :)

PR #1115, opened 2025-04-07
EvseManager: Fix deadlock between hlc_mutex and state_machine_mutex

corneliusclaussen:
## Describe your changes

In rare occasions, a mutex deadlock could occur between hlc_mutex and state_machine_mutex. This PR removes hlc_mutex as it can be replaced by simple atomic_bools

### DIFF
--- a/modules/EvseManager/EvseManager.cpp
+++ b/modules/EvseManager/EvseManager.cpp
@@ -111,7 +111,7 @@ void EvseManager::init() {
     // apply sane defaults capabilities settings once on boot
     powersupply_capabilities = get_sane_default_power_supply_capabilities();
 
-    if (get_hlc_enabled()) {
+    if (hlc_enabled) {
         if (config.charge_mode == "DC") {
             // subscribe to run time updates for real initial values (and changes e.g. due to de-rating)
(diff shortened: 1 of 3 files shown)


modules/EvseManager/EvseManager.cpp
```diff
@@ -1329,13 +1315,7 @@ bool EvseManager::is_reserved() {
     return reserved;
 }
 
-bool EvseManager::get_hlc_enabled() {
-    Everest::scoped_lock_timeout lock(hlc_mutex, Everest::MutexDescription::EVSE_get_hlc_enabled);
-    return hlc_enabled;
-}
-
 bool EvseManager::get_hlc_waiting_for_auth_pnc() {
```
SebaLukas, line 1318: Is this get function necessary? `hlc_waiting_for_auth_pnc` is public

SebaLukas (commented):
In general, it looks good. One small question:

modules/EvseManager/EvseManager.cpp
Pietfried: Looks like `hlc_waiting_for_auth_pnc` is private

Pietfried (approved):

PR #1126, opened 2025-04-10
GenericPowermeter: improve error handling

barsnick:
## Describe your changes
This PR addresses a few error handling issues:
- If the power meter failed to deliver values initially, GenericPowerMeter would report incorrect values (`0`) instead of not reporting at all. Expected: Wait with reporting values until a complete value set has been successfully acquired.
- Reading of exponent values (which are optional) was not properly checked for errors from `serial_communication_hub`, silently falling back to the possibly incorrect value `0`. Expected: Ignore readings which return an error.
- Failure to read or parse the configuration would print an error, but the module would silently continue running without ever reporting any meter values. Expected: The module terminates if it is mis-configured.

The read failures are not hypothetical, as customers have observed power meters which occasionally fail to answer correctly.

All these issues are fixed in this PR. Furthermore, quite a bit of refactoring was done, to drop unneeded code and use structures which easier to understand and modify. Reading the PR commit by commit makes more sense, if the complete diff seems to large.

The delay of initial values was tested with a mockup of an Eastron meter (in place of SerialCommHub), returning errors:
https://github.com/chargebyte/everest-core/tree/feature/serialcomm_powermetersim

The handling of configuration parsing was tested with modified model configs.

### DIFF
--- a/modules/GenericPowermeter/main/powermeterImpl.cpp
+++ b/modules/GenericPowermeter/main/powermeterImpl.cpp
@@ -3,6 +3,8 @@
 
 #include "powermeterImpl.hpp"
 #include <fmt/core.h>
+#include <functional>
+#include <optional>
 #include <thread>
 #include <utils/date.hpp>
 #include <utils/yaml_loader.hpp>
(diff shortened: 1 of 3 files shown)


barsnick:
Friendliy ping for review.

modules/GenericPowermeter/main/powermeterImpl.hpp
```diff
@@ -113,17 +114,20 @@ class powermeterImpl : public powermeterImplBase {
...
     powermeterImpl::ModbusFunctionType select_modbus_function(const uint8_t function_code);
     void read_powermeter_values();
-    void readRegister(const RegisterData& register_config);
-    void process_response(const RegisterData& message_type,
-                          const types::serial_comm_hub_requests::Result register_message,
-                          const types::serial_comm_hub_requests::Result exponent_message);
+    bool read_register(const RegisterData& register_config);
+    // void process_response(const RegisterData& message_type,
+    //                       const types::serial_comm_hub_requests::Result register_message,
+    //                       const types::serial_comm_hub_requests::Result exponent_message);
```
hikinggrass, line 127:
```suggestion
```

modules/GenericPowermeter/main/powermeterImpl.cpp
```diff
@@ -325,216 +324,245 @@ powermeterImpl::ModbusFunctionType powermeterImpl::select_modbus_function(const
...
+    uint32_t value{0};
+    const auto& reg_value = reg_message.value.value();
+    if (reg_data.num_registers == 1 && reg_value.size() == 1) {
+        value = reg_value.at(0);
+    } else if (reg_data.num_registers == 2 && reg_value.size() == 2) {
+        value += reg_value.at(0) << 16;
+        value += reg_value.at(1);
     } else {
-        EVLOG_error << "Error! Received message is empty!\n";
+        throw std::runtime_error("Values of more than 2 registers in size are currently not supported!");
```
hikinggrass, line 559: In general it would be great to move towards proper error handling with the error framework (like in other powermeters in the codebase). But since this isn't a regression this doesn't have to be addressed in this PR

hikinggrass (dismissed):

hikinggrass (approved):

PR #1135, opened 2025-04-16
Adding a tls server state check before announcing security in the sdp response message

SebaLukas:
## Describe your changes
See title

We found out that EvseV2G falsely tells the car TLS is supported in the SDP message even if no V2G Leaf Certificate is available. With this PR a check is added to check if the tls server is available.

### DIFF
--- a/modules/EvseV2G/sdp.cpp
+++ b/modules/EvseV2G/sdp.cpp
@@ -138,9 +138,15 @@ int sdp_send_response(int sdp_socket, struct sdp_query* sdp_query) {
         return 1;
     }
 
+    using state_t = tls::Server::state_t;
+    const auto tls_server_state = sdp_query->v2g_ctx->tls_server->state();
+
+    const auto tls_server_available =
+        (tls_server_state == state_t::init_complete or tls_server_state == state_t::running);
+
     switch (sdp_query->security_requested) {
     case SDP_SECURITY_TLS:
-        if (sdp_query->v2g_ctx->local_tls_addr) {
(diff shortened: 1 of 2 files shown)


james-ctc (approved):
looks good to me

modules/EvseV2G/sdp.cpp
```diff
@@ -138,9 +138,15 @@ int sdp_send_response(int sdp_socket, struct sdp_query* sdp_query) {
         return 1;
     }
 
+    using state_t = tls::Server::state_t;
+    const auto tls_server_state = sdp_query->v2g_ctx->tls_server->state();
+
+    const auto tls_server_available =
+        (tls_server_state == state_t::init_complete or tls_server_state == state_t::running) ? true : false;
```
barsnick, line 145: A ternary `? true : false` is basically a nop. 😉
```suggestion
        (tls_server_state == state_t::init_complete or tls_server_state == state_t::running);
```

barsnick (changes requested):
I have always wondered (and this doesn't change much with this PR) what happens if neither of the TCP/TLS conditions are met? Only the log message (e.g. `"SDP requested NO-TLS, announcing nothing"`) will be misleading, right? The server will respond correctly once it is in fact in the correct state.

modules/EvseV2G/sdp.cpp
SebaLukas: Yeah you are right :+1: changed that

james-ctc (dismissed):

barsnick (dismissed):

Pietfried (approved):

james-ctc (approved):

PR #1137, opened 2025-04-17
Small changes to enable debugging

Abby-Wheelis:
## Describe your changes
Small changes introduced during testing and development related to #1017, separating to an independent PR for clarity, keeping as draft in case other changes arise as I'm wrapping up with #1017. 


Abby-Wheelis:
I think this was the only log change that I ended up needing to make that wasn't directly related to my work. I think this was useful enough in debugging that it would be worth adding it to the repository.

SebaLukas (dismissed):

SebaLukas:
Can you please update your branch to main? I think the ci jobs did not trigger because your commit is too old?

PR #1145, opened 2025-04-23
Auth: Fix interface

dorezyuk:
The auth interface uses the wrong type


Pietfried:
> The auth interface uses the wrong type

Thats a nice and interesting finding :+1: :smile:

hikinggrass (approved):

PR #1151, opened 2025-04-25
feat(OcppAuth): Updates to id_token_info for token in use on connector

mlitre:
We can now update id token info for a specific connector in case we receive updated info in a transaction event or start transaction message from OCPP. This allows to properly support new edges cases for the parent id tag. This is required for OCPP certification.

### DIFF
--- a/interfaces/auth_token_validator.yaml
+++ b/interfaces/auth_token_validator.yaml
@@ -13,5 +13,12 @@ cmds:
         Result object containing validation result
       type: object
       $ref: /authorization#/ValidationResult
+vars:
+  validate_result_update:
+    description: >-
+      Updated validation result for a given token due to some changes. These
+      can be triggered by TransactionEvent messages or StartTransaction.
+    type: object
+    $ref: /authorization#/ValidationResultUpdate
 errors:
   - reference: /errors/generic
--- a/modules/Auth/Auth.cpp
+++ b/modules/Auth/Auth.cpp
@@ -24,6 +24,11 @@ void Auth::init() {
             t.detach();
         });
     }
+    for (const auto& token_validator : this->r_token_validator) {
+        token_validator->subscribe_validate_result_update([this](ValidationResultUpdate validation_result_update) {
+            this->auth_handler->handle_token_validation_result_update(validation_result_update);
+        });
+    }
 }
 
 void Auth::ready() {
--- a/modules/Auth/lib/AuthHandler.cpp
+++ b/modules/Auth/lib/AuthHandler.cpp
@@ -125,6 +125,30 @@ TokenHandlingResult AuthHandler::on_token(const ProvidedIdToken& provided_token)
     return result;
 }
 
+void AuthHandler::handle_token_validation_result_update(const ValidationResultUpdate& validation_result_update) {
+    std::unique_lock<std::mutex> lk(this->event_mutex);
+    auto connector_id = validation_result_update.connector_id;
+    if (this->evses.find(connector_id) != this->evses.end() and this->evses.at(connector_id)->identifier.has_value()) {
+        EVLOG_info << "Updating validation result on connector: " << connector_id;
+        // Currently we only support updating the parent id token
+        this->evses.at(connector_id)->identifier->authorization_status =
+            validation_result_update.validation_result.authorization_status;
+        this->evses.at(connector_id)->identifier->parent_id_token =
+            validation_result_update.validation_result.parent_id_token;
+        types::authorization::ProvidedIdToken provided_token;
+        provided_token.id_token = this->evses.at(connector_id)->identifier->id_token;
+        provided_token.authorization_type = this->evses.at(connector_id)->identifier->type;
+        provided_token.parent_id_token = validation_result_update.validation_result.parent_id_token;
+        std::vector<int32_t> connectors_allowed{connector_id};
+        provided_token.connectors = connectors_allowed;
+        this->publish_token_validation_status_callback(provided_token,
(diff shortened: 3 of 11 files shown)


modules/Auth/lib/AuthHandler.cpp
```diff
@@ -124,6 +124,20 @@ TokenHandlingResult AuthHandler::on_token(const ProvidedIdToken& provided_token)
...
 }
 
+void AuthHandler::handle_token_validation_result_update(const ValidationResultUpdate& validation_result_update) {
+    auto connector_id = validation_result_update.connector_id;
+    if (this->evses.find(connector_id) != this->evses.end() and this->evses.at(connector_id)->identifier.has_value()) {
+        EVLOG_info << "Updating validation result on connector: " << connector_id;
+        this->evses.at(connector_id)->identifier->authorization_status =
+            validation_result_update.validation_result.authorization_status;
+        this->evses.at(connector_id)->identifier->parent_id_token =
+            validation_result_update.validation_result.parent_id_token;
```
Pietfried, line 134: You can maybe add a comment here that other parts of the `ValidationResult` are neglected for now

interfaces/auth_token_validator.yaml
```diff
@@ -13,5 +13,10 @@ cmds:
         Result object containing validation result
       type: object
       $ref: /authorization#/ValidationResult
+vars:
+  validate_result_update:
+    description: Updated validation result for a given token due to some changes.
```
Pietfried, line 18: You could add in the description that it was added because OCPP can do an update in `StartTransaction.req` or `TransactionEvent.conf` . It makes it a little easier to argue about it later if we need to

modules/Auth/lib/AuthHandler.cpp
```diff
@@ -124,6 +124,20 @@ TokenHandlingResult AuthHandler::on_token(const ProvidedIdToken& provided_token)
     return result;
 }
 
+void AuthHandler::handle_token_validation_result_update(const ValidationResultUpdate& validation_result_update) {
+    auto connector_id = validation_result_update.connector_id;
```
Pietfried, line 128: This should be protected using the `event_mutex` since `this->evses` is accessed and updated.  
```suggestion
    std::unique_lock<std::mutex> lk(this->event_mutex);
    auto connector_id = validation_result_update.connector_id;
```

modules/Auth/Auth.cpp
```diff
@@ -24,6 +24,14 @@ void Auth::init() {
             t.detach();
         });
     }
+    for (const auto& token_validator : this->r_token_validator) {
+        token_validator->subscribe_validate_result_update([this](ValidationResultUpdate validation_result_update) {
+            std::thread t([this, validation_result_update]() {
```
Pietfried, line 29: You dont have to run this in a new thread. It's done for `on_token` because `on_token` potentially blocks and/or takes some time to process the request. In order to handle multiple requests from the same provider, it's running it a seperate thread.

Since `handle_token_validation_result_update` can be processed and return without any waiting, its not required to run it in a separate thread.

Pietfried (changes requested):
Looks pretty good overall, I added some remarks inline. It would be nice to have 1-2 OCPP e2e tests for this

modules/Auth/Auth.cpp
mlitre: Done

mlitre:
> Looks pretty good overall, I added some remarks inline. It would be nice to have 1-2 OCPP e2e tests for this

I added a test for OCPP 1.6 and OCPP 2.0.1

modules/Auth/lib/AuthHandler.cpp
```diff
@@ -124,6 +124,22 @@ TokenHandlingResult AuthHandler::on_token(const ProvidedIdToken& provided_token)
...
+void AuthHandler::handle_token_validation_result_update(const ValidationResultUpdate& validation_result_update) {
+    std::unique_lock<std::mutex> lk(this->event_mutex);
+    auto connector_id = validation_result_update.connector_id;
+    if (this->evses.find(connector_id) != this->evses.end() and this->evses.at(connector_id)->identifier.has_value()) {
+        EVLOG_info << "Updating validation result on connector: " << connector_id;
+        // Currently we only support updating the parent id token
+        this->evses.at(connector_id)->identifier->authorization_status =
+            validation_result_update.validation_result.authorization_status;
+        this->evses.at(connector_id)->identifier->parent_id_token =
+            validation_result_update.validation_result.parent_id_token;
```
Pietfried, line 136: I think we should also call `publish_token_validation_status_callback` with the new status and updated provided id token here

mlitre: Done

tests/ocpp_tests/test_sets/ocpp201/transactions.py
```diff
@@ -453,3 +454,140 @@ async def test_two_parallel_transactions(
...
+        test_utility,
+        charge_point_v201,
+        "TransactionEvent",
+        {"eventType": "Ended"},
+    )
+
+    # stop charging session
+    test_controller.plug_out()
+
+    await asyncio.sleep(10)
```
Pietfried, line 593: Can we drop this?

Pietfried (dismissed):

tests/ocpp_tests/test_sets/ocpp201/transactions.py
mlitre: Ah yes we probably can, it came from the TC that I copied

Pietfried (approved):

PR #1159, opened 2025-04-30
fix(auth): Missing parent id token in token validation callback

mlitre:

### DIFF
--- a/modules/Auth/lib/AuthHandler.cpp
+++ b/modules/Auth/lib/AuthHandler.cpp
@@ -77,6 +77,7 @@ TokenHandlingResult AuthHandler::on_token(const ProvidedIdToken& provided_token)
     std::unique_lock<std::mutex> lk(this->event_mutex);
 
     TokenHandlingResult result;
+    ProvidedIdToken provided_token_copy = provided_token;
 
     // check if token is already currently processed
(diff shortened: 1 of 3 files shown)


modules/Auth/lib/AuthHandler.cpp
```diff
@@ -86,7 +87,7 @@ TokenHandlingResult AuthHandler::on_token(const ProvidedIdToken& provided_token)
         // process token if not already in process
         this->tokens_in_process.insert(provided_token);
         this->publish_token_validation_status_callback(provided_token, TokenValidationStatus::Processing);
-        result = this->handle_token(provided_token, lk);
+        result = this->handle_token(provided_token_copy, lk);
```
Pietfried, line 90: Did you consider to use the `provided_token_copy` also below in the `publish_token_validation_status_callback`s ? This looks like the cleaner solution to me, particularly for `TokenHandlingResult::USED_TO_STOP_TRANSACTION` it makes sense to do so . Maybe I am also missing a reason why the original token should be used

mlitre: Good catch I was missing that indeed

Pietfried (approved):

PR #1165, opened 2025-05-06
Fix/soft reset

mlitre:


modules/OCPP/OCPP.cpp
```diff
@@ -843,6 +843,10 @@ void OCPP::ready() {
     });
 
     this->charge_point->register_reset_callback([this](ocpp::v16::ResetType type) {
+        // small delay before stopping the charge point to make sure all responses are received
+        std::this_thread::sleep_for(std::chrono::seconds(2));
```
hikinggrass, line 847: should we make this configurable (in the manifest, with a default of 2s) ?

mlitre: Done

Pietfried (approved):

hikinggrass (dismissed):

modules/OCPP/OCPP.cpp
```diff
@@ -843,6 +843,10 @@ void OCPP::ready() {
     });
 
     this->charge_point->register_reset_callback([this](ocpp::v16::ResetType type) {
+        // small delay before stopping the charge point to make sure all responses are received
+        std::this_thread::sleep_for(std::chrono::seconds(this->config.ResetStopDelay));
```
Pietfried, line 847: Maybe we should only sleep for the soft reset? What do you think?

mlitre: Hmm I would agree, but we handle soft and hard the same way so I think it applies to both

Pietfried: The modules that implements `system` can also make a decision how to handle `Hard` and `Soft`, but I think on the OCPP side for the `Hard` reset we should trigger the reset_callback immediately

mlitre: I see, I made it so that we call the reset callback immediately without the timeout

Pietfried (dismissed):

Pietfried (approved):

PR #1166, opened 2025-05-06
chore(config): New 2 DC EVSE with OCPP config

mlitre:


config/config-sil-two-evse-dc-ocpp.yaml
volkert-fastned, line 1: First of all, thanks for working on this! This is very useful for us as well. 🙏🏽

I'm looking to add a similar SIL, but with OCPP 2.0.1 instead of OCPP 1.6.

I see existing files with either `ocpp` or `ocpp201` in the names, implying that `ocpp` means OCPP 1.6.

Going forward, maybe it would be better start naming OCPP 1.6 SIL files more explicitly?

So in this case `config/config-sil-two-evse-dc-ocpp16.yaml`?

We can then add another file called `config/config-sil-two-evse-dc-ocpp201.yaml`, and the distinction between the two will then immediately be obvious.

Thanks for considering!

volkert-fastned: Oh, wait... Is the OCPP 1.6 module in EVerest still called just `ocpp`? Yeah, in that case, it's best to disregard my suggestion, as long as EVerest doesn't rename that module to `ocpp16` either. Thanks.

config/config-sil-two-evse-dc-ocpp.yaml
```diff
@@ -0,0 +1,376 @@
+active_modules:
+  iso15118_charger_1:
+    module: EvseV2G
+    config_module:
+      device: br-e2d7d9df05a1
```
SebaLukas, line 5: I am still missing documentation on how the user creates such device. Until now, this was set to “auto” in the SIL and the modules selected and used the first suitable Ethernet device.

config/config-sil-two-evse-dc-ocpp.yaml
mlitre: Yes unfortunately the ocpp16 module is just called ocpp, but for the config files we could implement you suggest as it could make it clearer!

config/config-sil-two-evse-dc-ocpp.yaml
mlitre: Ah yes sorry I had hardcoded the docker network as the device, I have to see how to get this setup working properly and set correct names for the devices.

volkert-fastned:
Hello @mlitre. Any idea on what might be needed to get this PR our of Draft status and past the finish line?

Happy to help with testing. Thanks! 🙂

PR #1167, opened 2025-05-06
Limiting target_current if the ev does not check the evse max limits

SebaLukas:
## Describe your changes
Adding a check and limit target_current if the evse_max_limits are smaller than target_current value.

In the case that the charger reduces the current or power, it can happen that the EV ignores the new max limits. Due to the changes, the new limits are sent to the PowerSupply module as target values.

### DIFF
--- a/modules/EvseManager/EvseManager.cpp
+++ b/modules/EvseManager/EvseManager.cpp
@@ -375,47 +375,68 @@ void EvseManager::ready() {
             }
 
             // Car requests a target voltage and current limit
-            r_hlc[0]->subscribe_dc_ev_target_voltage_current([this](types::iso15118::DcEvTargetValues v) {
-                bool target_changed = false;
+            r_hlc[0]->subscribe_dc_ev_target_voltage_current(
+                [this](types::iso15118::DcEvTargetValues v) {
+                    bool target_changed = false;
 
-                // Hack for Skoda Enyaq that should be fixed in a different way
-                if (config.hack_skoda_enyaq and (v.dc_ev_target_voltage < 300 or v.dc_ev_target_current < 0))
-                    return;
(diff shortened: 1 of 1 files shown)


Pietfried (dismissed):

modules/EvseManager/EvseManager.cpp
```diff
@@ -394,6 +394,19 @@ void EvseManager::ready() {
...
 
+                bool car_breaks_limit{false};
+                const auto hlc_limits = charger->get_evse_max_hlc_limits();
+                if (v.dc_ev_target_current > hlc_limits.evse_maximum_current_limit) {
+                    v.dc_ev_target_current = hlc_limits.evse_maximum_current_limit;
+                    car_breaks_limit = true;
+                }
+
+                const auto target_power = v.dc_ev_target_current * v.dc_ev_target_voltage;
+                if (target_power > hlc_limits.evse_maximum_power_limit) {
```
corneliusclaussen, line 405: This will reduce the actual charging power quite a bit on some cars. I think we should use the actual voltage rather than the target voltage to calculate the power limit. This is what we do in other places in the code as well. If e.g. the target voltage is 420V (battery full) the actual voltage may just be 360v or so since it is empty. Then the current would be limited such that the power close to full is not higher than the max, which means it is significantly lower than the max power in the beginning of the charging session.

corneliusclaussen (changes requested):

modules/EvseManager/EvseManager.cpp
SebaLukas: You are right, I will change target_voltage to the actual voltage

corneliusclaussen (approved):

PR #1171, opened 2025-05-07
fix(ocpp16): Firmware update process + number of retries

mlitre:


robert-s-ubi:
Why does it say "(ocpp16)"? The change should (and I think it is?) be made for OCPP 2.x as well...

modules/System/main/systemImpl.cpp
```diff
@@ -425,6 +430,10 @@ void systemImpl::handle_reset(types::system::ResetType& type, bool& scheduled) {
     // channels in parallel when this call returns
     std::thread([this, type, scheduled] {
         EVLOG_info << "Reset request received: " << type << ", " << (scheduled ? "" : "not ") << "scheduled";
+        if (!this->mod->r_store.empty()) {
+            this->mod->r_store.at(0)->call_store(boot_reason_key,
+                                                 boot_reason_to_string(types::system::BootReason::RemoteReset));
+        }
```
Pietfried, line 436: Since `handle_reset` is now also called when a FW update has been successfully installed, `RemoteReset` is not always correct here, since in case of a FW update the BootReason shall be `FirmwareUpdate`.

We cant cover all reasons and since its just an example implementation of the `system` interface this is fine, but can support `RemoteReset` , `FirmwareUpdate` and `PowerUp` .

Pietfried (changes requested):

Pietfried (approved):

PR #1173, opened 2025-05-09
Skip HLC startup tests in GitHub CI

hikinggrass:
There seems to be an IPv6 issue with the current runner image: https://github.com/actions/runner-images/issues/12166
This skips the AC and DC HLC tests when the EXT_MOUNT environment variable is set to "/ext", which is the case for our CI.


SebaLukas (dismissed):
:+1:

SebaLukas (approved):

PR #1175, opened 2025-05-12
[OCPP 1.6] Update EvseManager Inoperative error to include details of the original cause

james-ctc:
## Describe your changes
Reformat EVSE Manager inopertative error to preserve original error

New configuration key added that optionally changes the vendor_id to be the ID from the original error rather than always being EVerest.

Added unit tests to check how different error types and sub-types will be presented.


Pietfried (approved):
lgtm

PR #1180, opened 2025-05-20
Feat/bazel 8

dorezyuk:


SirVer (approved):
Very exciting! Latest and greatest in build systems.

MODULE.bazel
```diff
@@ -0,0 +1,91 @@
+module(
+    name = "everest-core",
+    version = "0.0.0",
```
hikinggrass, line 3: This should follow what's in the CMakeLists.txt (and kept in sync)

MODULE.bazel
```diff
@@ -0,0 +1,91 @@
...
+)
+
+###############################################################################
+# B A Z E L  C E N T R A L  R E G I S T R Y # https://registry.bazel.build/
+###############################################################################
+
+bazel_dep(name = "everest-framework", version = "0.0.0")
```
hikinggrass, line 10: should probably be a real version tag

MODULE.bazel
```diff
@@ -0,0 +1,91 @@
...
+git_override(module_name="everest-framework", commit="aebffb45d9c1dc948ad31e718fc3efe9eac0a9ca", remote="https://github.com/EVerest/everest-framework")
```
hikinggrass, line 11: FIXME: rev @ main

MODULE.bazel
```diff
@@ -0,0 +1,91 @@
...
+
+bazel_dep(name = "everest-utils", version = "0.0.0")
```
hikinggrass, line 13: should probably be a real version tag
```suggestion
bazel_dep(name = "everest-utils", version = "0.5.2")
```

MODULE.bazel
```diff
@@ -0,0 +1,91 @@
...
+git_override(module_name="everest-utils", commit="23cb911efd607a8cd8cad11d9be7b3e67978262a", remote="https://github.com/Qwello/everest-utils")
```
hikinggrass, line 14: FIXME: rev @ main

MODULE.bazel
dorezyuk: main woudn't be reproducible

MODULE.bazel
dorezyuk: I've updated it manually - the number here does not make a huge difference but if we want to keep it in sync we would need some of the nice scripts you did in framework

MODULE.bazel
hikinggrass: I think I wrote that when it was still pointing to a rev on a feature-branch

Pietfried (approved):

PR #1183, opened 2025-05-21
fix: resolve buffer overrun by increasing buffer size

tomc-xsec:
`sprintf` writes four characters at a time, which causes a buffer overrun on the final iteration (`i = 5`). Buffer expanded to prevent overrun.

## Describe your changes
Increase the length of the buffer by 1 to accommodate the extra character. 


james-ctc:
a safer fix is to use snprintf with a size of 3 to ensure that only 3 characters are printed.
sprintf does include a terminating \0

```
snprintf(&buffer[i * 3], 3, "%02" PRIX8 ":", req->EVCCID.bytes[i]);
```

> The  functions snprintf() and vsnprintf() write at most size bytes (including the terminating null byte ('\0')) to str.

tomc-xsec:
> Thank you for your contribution. As far as I know, only 3 bytes are written to the buffer in the `sprintf` function, not 4. The` -1` is important here, that the last `:` is overwritten with` \0`.

`sprintf` outputs four characters per iteration: two digits, a ':', and a '\0'. When `i = 5`, `buffer[18]` is written to.

Compiling with `-fstack-protector-strong -fstack-clash-protection -fbounds-check` will trip the bug.

tomc-xsec:
I can send the bin and core if you would like to do your own analysis.

Is `snprintf` preferred or straight conversion through a hex digit table? `snprintf` seems overkill.

Either way I can make the change in the morning.

james-ctc:
I recommend using the snprintf solution for the following reasons:
`snprintf(&buffer[i * 3], 3, "%02" PRIX8 ":", req->EVCCID.bytes[i]);`
- a change in the definition of PRIX8 could change the number of bytes written
- no change in the size of the buffer is needed - some buffers are fixed sizes and a sizeof(buffer) could lead to an incorrect size being used elsewhere in the code (or on further code updates)
- in general use of `sprintf` is discouraged since no buffer overflow detection is done
- `snprintf` as suggested is a one-line change that adds overflow protection and reinforces the intent that each byte is represented as 2 hex digits followed by a :

tomc-xsec:
I've revised the commit in terms of `snprintf`. In addition I adjusted the loop structure to align buffer size parameters with the `snprintf` `size` parameter.

modules/EVSE/EvseV2G/iso_server.cpp
```diff
@@ -523,18 +523,15 @@ static bool publish_iso_certificate_installation_exi_req(struct v2g_context* ctx
...
-        sprintf(&buffer[i * 3], "%02" PRIX8 ":", req->EVCCID.bytes[i]);
+    for (std::size_t i = 0; i < req->EVCCID.bytesLen; i++) {
+        if (i > 0)
+            buffer[i * 3 - 1] = ':';
+        snprintf(&buffer[i * 3], std::size(buffer) - i * 3, "%02" PRIX8, req->EVCCID.bytes[i]);
     }
-    if (i)
-        buffer[i * 3 - 1] = '\0';
-    else
-        buffer[0] = '\0';
```
james-ctc, line 533: The existing code pattern is clear to follow and hence maintain:
1. create buffer
2. populate buffer
3. ensure string is terminated

The buffer overflow problem can be addressed with a 1 line change that keeps the original approach:
`snprintf(&buffer[i * 3], 3, "%02" PRIX8 ":", req->EVCCID.bytes[i]);`

The `if` inside the for loop is a confusing approach to add the colon when snprintf is being used anyway.
The `n` argument to snprintf is variable for no obvious reason.

This makes the code harder to follow.

tomc-xsec: I don't think this resolves the issue for two reasons:
1. The format `"%02" PRIX8 ":"`  outputs four characters, not three (terminating NUL is the fourth). `size = 3`, `snprintf` truncates the `:`.
2. If `snprintf(&buffer[i * 3], 4, "%02" PRIX8 ":", req->EVCCID.bytes[i]);`, there is an out of bounds access when `i = 5` (`snprintf` sets `buffer[18]='\0'`).

james-ctc: how about ...
```
char buffer[iso2_evccIDType_BYTES_SIZE * 3 + 1];
...
snprintf(&buffer[i * 3], 4, "%02" PRIX8 ":", req->EVCCID.bytes[i]);
```

SebaLukas (approved):

modules/EvseV2G/iso_server.cpp
```diff
@@ -523,13 +523,14 @@ static bool publish_iso_certificate_installation_exi_req(struct v2g_context* ctx
 static enum v2g_event handle_iso_session_setup(struct v2g_connection* conn) {
     struct iso2_SessionSetupReqType* req = &conn->exi_in.iso2EXIDocument->V2G_Message.Body.SessionSetupReq;
     struct iso2_SessionSetupResType* res = &conn->exi_out.iso2EXIDocument->V2G_Message.Body.SessionSetupRes;
-    char buffer[iso2_evccIDType_BYTES_SIZE * 3 - 1 + 1]; /* format: (%02x:) * n - (1x ':') + (1x NULL) */
+    char buffer[iso2_evccIDType_BYTES_SIZE * 3 + 1]; /* format: (%02x:) * n - (1x ':') + (1x NULL) + (1x workspace) */
     int i;
     enum v2g_event next_event = V2G_EVENT_NO_EVENT;
 
     /* format EVCC ID */
+    static_assert(std::size(buffer) > 3*std::extent_v<decltype(req->EVCCID.bytes)>);
```
james-ctc: formatting issue - it needs to be:
```
static_assert(std::size(buffer) > 3 * std::extent_v<decltype(req->EVCCID.bytes)>);
```

however the following is clearer
```
static_assert(sizeof(buffer) > (3 * sizeof(req->EVCCID.bytes)));
```

modules/EvseV2G/.iso_server.cpp.swp
james-ctc, line 1: shouldn't be included

modules/EvseV2G/.tools.hpp.swp
james-ctc, line 1: shouldn't be included

modules/EvseV2G/tools.cpp
```diff
@@ -210,3 +210,35 @@ convert_to_certificate_hash_data_info_vector(const types::evse_security::OCSPReq
     }
     return certificate_hash_data_info_vec;
 }
+
+/*!
+ * \brief Converts binary data to a colon-separated uppercase hexadecimal string.
+ */
+std::string
+hexify(const std::uint8_t *data, std::size_t datalen)
```
james-ctc: I recommend running via clang format as it won't pass the CI otherwise

modules/EvseV2G/tools.cpp
```diff
@@ -210,3 +210,35 @@ convert_to_certificate_hash_data_info_vector(const types::evse_security::OCSPReq
...
+{
+    constexpr char hexdigit[16] = {
```
james-ctc: missing null pointer check on data

modules/EvseV2G/tools.cpp
```diff
@@ -210,3 +210,38 @@ convert_to_certificate_hash_data_info_vector(const types::evse_security::OCSPReq
...
+
+    std::string str;
+
+    // Return empty string if either data is null or datalen is zero
+    if (datalen == 0)
+        return str;
+
+    // If datalen > 0, data must be non-null
+    if (!data)
+        throw std::invalid_argument("hexify: data is null but datalen > 0");
```
james-ctc: `if (data != nullptr)` is better
please don't use exceptions

james-ctc (changes requested):
Since you have added a new file and function there should also be some unit tests to check the operation of hexify

james-ctc (approved):
Thank you for the updates and code improvements.
The style changes required by lint (clang-format) will need to be made before the code can be merged. see https://github.com/EVerest/everest-core/actions/runs/17033641359/job/48400627450?pr=1183

james-ctc (changes requested):
Please leave history intact. It makes re-reviewing easier as your changes since last review are still visible.

> Handles incomplete input safely by filling missing octets with a specified character

This is a functionality change that should be a separate PR. This PR is about fixing a buffer overrun.

In particular the code is supposed to be logging what was received so adding extra characters is giving a misleading log message.

If this is what was sent: 0xDE, 0xAD, 0xBE
then this is what should be logged "DE:AD:BE"

james-ctc:
Buffer overflow issue addressed in this PR
https://github.com/EVerest/everest-core/pull/1553

PR #1192, opened 2025-05-27
Refactor(ocpp201): ComposedDeviceModel source lookup

Pietfried:
## Describe your changes
* ComposedDeviceModel does not cache full device model but instead just holds a map to lookup the source of a request by component, variable
* On `get_device_model` all sub device models are requested instead of returning the cached map (which was removed)


hikinggrass (dismissed):
One small nit, otherwise :+1:

PR #1194, opened 2025-05-27
feat(OCPP2.x): Dynamic device model initialization

Pietfried:
## Describe your changes
This PR extends to OCPP201 module to dynamically initialize the EVSE and Connector device model components based on the EVerest configuration.

Before this change, the correct number of EVSE and Connector components had to be defined and installed manually. [This PR](https://github.com/EVerest/libocpp/pull/1093) in libocpp ensures that the custom EVSE and Connector components are not installed anymore by default. 

_____________

**Note**: To migrate using the EVSE and Connector components managed by OCPP201 module instead of the libocpp device model, the Connector and EVSE components should be removed from the installation directory. When EVerest starts up, it will then remove these components from libocpp's device model storage.

### DIFF
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -79,7 +79,7 @@ libevse-security:
 # OCPP
 libocpp:
   git: https://github.com/EVerest/libocpp.git
-  git_tag: 663f59772c37ac7e83a051b3042504f4d013fe85
+  git_tag: 7b4ba2c8a106b2eefad4009b515148888ba1133a
   cmake_condition: "EVEREST_DEPENDENCY_ENABLED_LIBOCPP"
 # Josev
 Josev:
--- a/modules/OCPP201/OCPP201.cpp
+++ b/modules/OCPP201/OCPP201.cpp
@@ -150,6 +150,7 @@ void OCPP201::init_evse_maps() {
     for (size_t evse_id = 1; evse_id <= this->r_evse_manager.size(); evse_id++) {
         this->evse_ready_map[evse_id] = false;
         this->evse_soc_map[evse_id] = std::nullopt;
+        this->evse_hardware_capabilities_map[evse_id] = types::evse_board_support::HardwareCapabilities{};
     }
 }
 
@@ -367,6 +368,11 @@ void OCPP201::init() {
                 this->evse_ready_cv.notify_one();
             }
         });
+        this->r_evse_manager.at(evse_id - 1)
+            ->subscribe_hw_capabilities(
+                [this, evse_id](const types::evse_board_support::HardwareCapabilities& hw_capabilities) {
+                    this->evse_hardware_capabilities_map[evse_id] = hw_capabilities;
+                });
     }
 }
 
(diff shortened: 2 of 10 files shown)


modules/OCPP201/device_model/definitions.cpp
```diff
@@ -0,0 +1,111 @@
...
+} // namespace Characteristics
+} // namespace EvseDefinitions
+
+namespace ConnectorDefinitions {
+namespace Characteristics {
+
+const VariableCharacteristics AvailabilityState = [] {
+    VariableCharacteristics var;
+    var.dataType = DataEnum::OptionList;
+    var.supportsMonitoring = false;
```
mlitre, line 85: I had a question about supportsMonitoring, I guess which is more global than specific to this one. What determines whether a variable supports monitoring? I feel like there are many (like this one) where it would be interesting to have it. 
This variable is even more particular in the sense that, it definitely needs to support monitoring since the NotifyEvent is supposed to replace the StatusNotifications.

Pietfried: For the device model maintained in libocpp you can define `supportsMonitoring` individually for each variable within its component config.

Specifiying monitors for the EverestDeviceModel is currently not possible because the EverestDeviceModel is not (yet) backed with a persistant storage that could persist the monitors.

modules/OCPP201/OCPP201.cpp
```diff
@@ -367,6 +368,11 @@ void OCPP201::init() {
                 this->evse_ready_cv.notify_one();
             }
         });
+        this->r_evse_manager.at(evse_id - 1)
+            ->subscribe_hw_capabilities(
+                [this, evse_id](const types::evse_board_support::HardwareCapabilities& hw_capabilities) {
+                    this->evse_hardware_capabilities_map[evse_id] = hw_capabilities;
+                });
```
mlitre, line 375: Is there a reason why this is both in OCPP201.cpp and everest_device_model_storage.cpp? Couldn't we just use the subscribe in the everest_device_model_storage?

Pietfried: There is a reason. In OCPP2.0.1 we `subscribe_hw_capabilities` in the `init` function in order to not miss the initial publish. We instantiate the `EverestDeviceModel` in the `ready` function. This is required, since we should not call commands to `r_evse_manager` in the init, so we cant instantiate `EverestDeviceModel` in the `OCPP201::init`. If we only subscribe within the `EverestDeviceModel` we may miss the caps.

mlitre: Got it!

mlitre (dismissed):

dependencies.yaml
```diff
@@ -79,7 +79,7 @@ libevse-security:
 # OCPP
 libocpp:
   git: https://github.com/EVerest/libocpp.git
-  git_tag: 663f59772c37ac7e83a051b3042504f4d013fe85
+  git_tag: refactor/changes-to-device-model-components # FIXME
```
Pietfried:
```suggestion
  git_tag: 7b4ba2c8a106b2eefad4009b515148888ba1133a
```

PR #1195, opened 2025-05-27
Allow power on in ISO_POWER_READY state

wnagele:
We are switching into CP state C here, but do not call `allow_power_on` which leads to the contactors remaining open.


SebaLukas:
Thank you for contribution :)

I'm not sure whether the EV side should be allowed to close the contactors in the PowerDelivery state. This should only happen if the EvManager is really supposed to draw energy.
Basic Charging and HLC have slightly different times when the contactors should be opened/closed. Also, this is the EV side and not the charger side. On the charger side, the contactor must be closed when the EV side has sent the PowerDeliveryReq message.

That's why I wouldn't want to merge this PR.

wnagele:
Thanks, that makes sense. I missed that for the sim `iso_dc_power_on` had to be called once the power was ready. So there is no need to enable the power path in the ready state here.

PR #638, opened 2024-04-18
Beta release ISO15118-20

SebaLukas:
## Describe your changes
Adding Evse15118D20 module for iso15118-20 beta support
Adding the new EVerest/libiso15118 to everest-core

## Issue ticket number and link
Missing ISO15118-20 DC/DC BPT support in EVerest

## Checklist before requesting a review
- [x] I have performed a self-review of my code
- [ ] I have made corresponding changes to the documentation
- [x] I read the [contribution documentation](https://github.com/EVerest/EVerest/blob/main/CONTRIBUTING.md) and made sure that my changes meet its requirements

PR #1200, opened 2025-05-30
feat(OCPP2.1): Add AllowedEnergyTransferModes to auth validator

mlitre:
This list of allowed energy transfer modes has been added to the auth validator interface so that the evse manager can be aware of what energy transfer modes are allowed so that it can properly update the ISO15118 stack.

## Describe your changes

Added allowed transfer modes to be added to the auth interface so that they can be passed to the evse manager. Now we just need this to be implemented evse manager side to tell the hlc which modes are allowed. This is not yet supported. 
The updates should be done in `EvseManager/evse/evse_managerImpl.cpp`.

### DIFF
--- a/interfaces/ISO15118_charger.yaml
+++ b/interfaces/ISO15118_charger.yaml
@@ -7,18 +7,9 @@ cmds:
       evse_id:
         description: >-
           Set an ID that uniquely identifies the EVSE and the power outlet the
-          vehicle is connected to 
+          vehicle is connected to
         type: object
         $ref: /iso15118#/EVSEID
-      supported_energy_transfer_modes:
-        description: Available energy transfer modes supported by the EVSE
-        type: array
-        items:
-          description: The different energy modes supported by the SECC 
-          type: object
-          $ref: /iso15118#/SupportedEnergyMode
-        minItems: 1
-        maxItems: 6
       sae_j2847_mode:
         description: Charger is supporting SAE J2847 V2G/V2H version
         type: string
@@ -54,7 +45,7 @@ cmds:
         type: boolean
       central_contract_validation_allowed:
         description: >-
-          Indicates if the vehicle contract may be forwarded to and validated by a CSMS in case 
+          Indicates if the vehicle contract may be forwarded to and validated by a CSMS in case
           local validation was not successful
         type: boolean
 # Response messages to vars:
@@ -80,7 +71,7 @@ cmds:
       status:
         description: Set to true when contactor is closed, false when contactor is open
         type: boolean
-# Events 
+# Events
   dlink_ready:
     description: >-
       Signals dlink_ready from SLAC layer according to ISO15118-3
@@ -120,6 +111,22 @@ cmds:
         description: Set to true when to pause, set to false when to continue
         type: boolean
 # Update physical values
+  update_energy_transfer_modes:
+    description: >-
+      Update the supported energy transfer modes. Call at least once during start up.
+      Typically for ISO15118-2 and basic charging we expect unidirectional charging modes.
+      For ISO15118-20 the user can set what type of modes the hardware supports, may that be
+      bidirectional, unidirectional, wireless, etc charging.
+    arguments:
+      supported_energy_transfer_modes:
+        description: Available energy transfer modes supported by the EVSE
+        type: array
+        items:
+          description: The different energy modes supported by the SECC
+          type: object
+          $ref: /iso15118#/EnergyTransferMode
+        minItems: 1
+        maxItems: 6
   update_ac_max_current:
     description: Update the maximum allowed line current restriction per phase. Call at least once during start up.
     arguments:
@@ -143,7 +150,7 @@ cmds:
         type: object
         $ref: /iso15118#/DcEvseMinimumLimits
   update_isolation_status:
-    description: Update the isolation condition 
+    description: Update the isolation condition
     arguments:
       isolation_status:
         description: Result of the isolation monitoring
@@ -180,7 +187,7 @@ vars:
     description: An EIM authorization is requiered
     type: "null"
   require_auth_pnc:
-    description: >- 
+    description: >-
       The EVCC provides the payment details for a PnC authorization by sending
       the signature certificate chain and eMAID.
     type: object
--- a/interfaces/evse_manager.yaml
+++ b/interfaces/evse_manager.yaml
@@ -120,6 +120,26 @@ cmds:
         description: The plug and charge configuration object
         type: object
         $ref: /evse_manager#/PlugAndChargeConfiguration
+  update_allowed_energy_transfer_modes:
+    description: >-
+      Sets the supported energy transfer mode for ISO15118. It is expected that this command will update the ISO15118
+      software stack to only propose the list of allowed_energy_transfer_modes presented in this message.
+      If a transaction is already in progress, it is expected that this triggers a service renegotiation.
+      If no HLC is present, this will be accepted, but do nothing.
+    arguments:
+      allowed_energy_transfer_modes:
+        description: >-
+          The list of supported energy transfer modes. It cannot be empty as we need to propose something
+          to the EV.
+        type: array
+        items:
+          type: string
+          $ref: /iso15118#/EnergyTransferMode
+        minItems: 1
+    result:
+      description: Returns an enum indicating whether the update was successful or not.
+      type: string
+      $ref: /evse_manager#/UpdateAllowedEnergyTransferModesResult
 vars:
   session_event:
     description: Emits all events related to sessions
--- a/modules/Evse15118D20/charger/utils.hpp
+++ b/modules/Evse15118D20/charger/utils.hpp
@@ -5,9 +5,10 @@
 #include <generated/types/iso15118.hpp>
 #include <iso15118/message/type.hpp>
 
-static constexpr auto NUMBER_OF_SETUP_STEPS = 4;
+static constexpr auto NUMBER_OF_SETUP_STEPS = 5;
 
 enum class SetupStep {
+    SETUP,
     ENERGY_SERVICE,
     AUTH_SETUP,
     MAX_LIMITS,
--- a/modules/EvseManager/evse/evse_managerImpl.cpp
+++ b/modules/EvseManager/evse/evse_managerImpl.cpp
@@ -455,5 +455,18 @@ void evse_managerImpl::handle_set_plug_and_charge_configuration(
     }
 }
 
+types::evse_manager::UpdateAllowedEnergyTransferModesResult
+evse_managerImpl::handle_update_allowed_energy_transfer_modes(
+    std::vector<types::iso15118::EnergyTransferMode>& allowed_energy_transfer_modes) {
+    // if (incompatible) {
+    //   return "IncompatibleType";
+    // }
+    if (!mod->r_hlc.empty() && mod->r_hlc[0]) {
+        mod->r_hlc[0]->call_update_energy_transfer_modes(allowed_energy_transfer_modes);
+        return types::evse_manager::UpdateAllowedEnergyTransferModesResult::Accepted;
+    }
+    return types::evse_manager::UpdateAllowedEnergyTransferModesResult::NoHlc;
+}
+
 } // namespace evse
 } // namespace module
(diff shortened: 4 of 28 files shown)


modules/EvseManager/evse/evse_managerImpl.cpp
```diff
@@ -454,5 +454,18 @@ void evse_managerImpl::handle_set_plug_and_charge_configuration(
     }
 }
 
+void evse_managerImpl::handle_update_allowed_energy_transfer_modes(
+    std::vector<types::iso15118::EnergyTransferMode>& allowed_energy_transfer_modes) {
+    if (!mod->r_hlc.empty() && mod->r_hlc[0]) {
+        std::vector<types::iso15118::SupportedEnergyMode> modes{};
+        modes.reserve(allowed_energy_transfer_modes.size());
+        auto caps = mod->get_powersupply_capabilities();
```
SebaLukas: const auto caps

modules/Evse15118D20/charger/utils.hpp
```diff
@@ -5,9 +5,10 @@
 #include <generated/types/iso15118.hpp>
 #include <iso15118/message/type.hpp>
 
-static constexpr auto NUMBER_OF_SETUP_STEPS = 4;
+static constexpr auto NUMBER_OF_SETUP_STEPS = 5;
 
 enum class SetupStep {
+    SETUP,
```
james-ctc, line 11: This could be improved using EnumFlags (see lib/staging/util/include ...)
Since it manages NUMBER_OF_SETUP_STEPS and removes the need for casting:
```
    setup_steps_done.set(to_underlying_value(SetupStep::ENERGY_SERVICE));
// becomes
    setup_steps_done.set(SetupStep::ENERGY_SERVICE);
```

i.e. NUMBER_OF_SETUP_STEPS would no longer be needed.

an implementation for all() would need to be added. Something like
```
bool all() {
  return _value == (1ULL << (static_cast<std::underlying_type_t<T>>(T::last)) + 1) - 1;
}
```

modules/EvseV2G/charger/ISO15118_chargerImpl.cpp
```diff
@@ -78,6 +76,38 @@ void ISO15118_chargerImpl::handle_setup(
...
+        struct iso2_ServiceType sae_service;
+
+        init_iso2_ServiceType(&sae_service);
+
+        sae_service.FreeService = 1;
+        sae_service.ServiceCategory = iso2_serviceCategoryType_OtherCustom;
+
+        if (sae_j2847_mode == BidiMode::V2H) {
+            sae_service.ServiceID = 28472;
+            const std::string service_name = "SAE J2847/2 V2H";
```
james-ctc: std::string_view could be used which would remove the need for reinterpret_cast on line 92

modules/EvseV2G/charger/ISO15118_chargerImpl.cpp
```diff
@@ -78,6 +76,38 @@ void ISO15118_chargerImpl::handle_setup(
...
+            memcpy(sae_service.ServiceName.characters, reinterpret_cast<const char*>(service_name.data()),
+                   service_name.length());
+            sae_service.ServiceName.charactersLen = service_name.length();
+            sae_service.ServiceName_isUsed = true;
+        } else {
+            sae_service.ServiceID = 28473;
+            const std::string service_name = "SAE J2847/2 V2G";
+            memcpy(sae_service.ServiceName.characters, reinterpret_cast<const char*>(service_name.data()),
```
james-ctc: std::string_view should remove the need for reinterpret cast

james-ctc (commented):
some suggestions

modules/Evse15118D20/charger/utils.hpp
SebaLukas: We had already this discussion in #638. I would prefer to leave it at bitset

james-ctc: okay - there are non atomic versions in use elsewhere ... 
seems a shame not to use a consistent approach when using enums as bit flags.

SebaLukas: We can discuss this and then maybe open a separate PR. But I would not like to use this PR for that

james-ctc: that's fine

modules/EvseV2G/charger/ISO15118_chargerImpl.cpp
mlitre: Done!

interfaces/evse_manager.yaml
```diff
@@ -120,6 +120,19 @@ cmds:
...
         type: object
         $ref: /evse_manager#/PlugAndChargeConfiguration
+  update_allowed_energy_transfer_modes:
+    description: >-
+      Sets the supported energy transfer mode for ISO15118.
+    arguments:
+      allowed_energy_transfer_modes:
+        description: >-
+          Only used with OCPP 2.1, this allows the CSMS to transmit to the charging station
+          which energy transfer modes are allowed with the vehicle.
```
Pietfried: This description should be changed a little bit. Every module requiring the interface is able  to call this function. The description should be about what the function is / should be doing, not which module might be calling it.

interfaces/ISO15118_charger.yaml
```diff
@@ -120,6 +111,18 @@ cmds:
         description: Set to true when to pause, set to false when to continue
         type: boolean
 # Update physical values
+  update_energy_transfer_modes:
+    description: Update the supported energy transfer modes. Call at least once during start up.
```
Pietfried: Maybe it can be described in more detail what to expect from this function when a transaction is running (for -2, -20, PWM charging, etc.)

interfaces/evse_manager.yaml
```diff
@@ -120,6 +120,19 @@ cmds:
...
+        type: array
+        items:
+          type: string
+          $ref: /iso15118#/EnergyTransferMode
+        minItems: 1
```
Pietfried, line 138: Wouldnt it be useful to add a return type to indicate if updating the energy transfer mode fails? Think someone tries to set `AC_BPT` for a DC charger...

modules/EvseManager/evse/evse_managerImpl.cpp
```diff
@@ -455,5 +455,18 @@ void evse_managerImpl::handle_set_plug_and_charge_configuration(
...
 }
 
+void evse_managerImpl::handle_update_allowed_energy_transfer_modes(
+    std::vector<types::iso15118::EnergyTransferMode>& allowed_energy_transfer_modes) {
+    if (!mod->r_hlc.empty() && mod->r_hlc[0]) {
+        std::vector<types::iso15118::SupportedEnergyMode> modes{};
+        modes.reserve(allowed_energy_transfer_modes.size());
+        const auto caps = mod->get_powersupply_capabilities();
+        for (const auto& mode : allowed_energy_transfer_modes) {
+            modes.emplace_back(types::iso15118::SupportedEnergyMode{mode, caps.bidirectional});
```
Pietfried: This looks like it's possible to set a SupportedEnergyMode with:
- bidirectional : false
- energy_transfer_mode : AC_BPT 

This shouldnt be possible and for that reason I think we need a return type and reject the request in this case.

Also, it looks like the `SupportedEnergyMode` type doesnt need the `bidirectional` property.

modules/OCPP201/conversions.cpp
```diff
@@ -1513,5 +1516,48 @@ to_ocpp_clear_display_message_response(const types::display_message::ClearDispla
...
+    case ocpp::v2::EnergyTransferModeEnum::DC_ACDP_BPT:
+        return types::iso15118::EnergyTransferMode::DC_ACDP_BPT;
+    case ocpp::v2::EnergyTransferModeEnum::WPT:
+        return types::iso15118::EnergyTransferMode::WPT;
+    }
+    throw std::out_of_range("Could not convert EnergyTransferModeEnum");
+}
+
+std::optional<std::vector<types::iso15118::EnergyTransferMode>> to_everest_allowed_energy_transfer_modes(
+    const std::optional<std::vector<ocpp::v2::EnergyTransferModeEnum>>& allowed_energy_transfer_modes) {
```
Pietfried: More in line with the other implementations in this file it would be to not pass an optional but the actual vector here and check for `has_value` in the function that calls it.

modules/OCPP201/OCPP201.cpp
```diff
@@ -943,7 +939,35 @@ void OCPP201::ready() {
...
+                        info.type = "EVCCID";
+                        info_vector.push_back(info);
+                        tx_data->id_token->additionalInfo = info_vector;
+                    }
+                } else {
+                    ocpp::v2::IdToken token;
+                    token.additionalInfo = {};
+                    token.additionalInfo->push_back(
+                        ocpp::v2::AdditionalInfo{ev_info.evcc_id.value(), "EVCCID", std::nullopt});
+                    tx_data->id_token = token;
```
Pietfried: This could go into a separate member function, e.g. `void update_evcc_id_token(const std::string& evse_id, const std::string& evcc_id)`

types/authorization.yaml
```diff
@@ -191,6 +191,15 @@ types:
       reservation_id:
         description: The reservation id that is used with this validated token.
         type: integer
+      allowed_energy_transfer_modes:
+        description: >-
+          Only used with OCPP 2.1, this allows the CSMS to transmit to the charging station
+          which energy transfer modes are allowed with the vehicle.
```
Pietfried:
```suggestion
          which energy transfer modes are allowed with for the respective authorization request.
```

types/authorization.yaml
```diff
@@ -191,6 +191,15 @@ types:
       reservation_id:
         description: The reservation id that is used with this validated token.
         type: integer
+      allowed_energy_transfer_modes:
+        description: >-
+          Only used with OCPP 2.1, this allows the CSMS to transmit to the charging station
```
Pietfried: Same here like above, I wouldnt mention OCPP2.1 here

interfaces/ISO15118_charger.yaml
```diff
@@ -120,6 +111,27 @@ cmds:
...
+      supported_energy_transfer_modes:
+        description: Available energy transfer modes supported by the EVSE
+        type: array
+        items:
+          description: The different energy modes supported by the SECC
+          type: object
+          $ref: /iso15118#/EnergyTransferMode
+        minItems: 1
+        maxItems: 6
+    result:
```
Pietfried: Im not sure if a boolean is complex enough here or if we need something like `UpdateAllowedEnergyTransferModesResult` here as well.

I am also not sure if this function should block until the ServiceRenegotiation process turned out to be successful or not. @SebaLukas what do you think?

modules/OCPP201/OCPP201.cpp
```diff
@@ -1484,4 +1486,29 @@ void OCPP201::set_external_limits(const std::vector<ocpp::v2::CompositeSchedule>
     }
 }
 
+void OCPP201::update_evcc_id_token(const int& evse_id, const std::string& evcc_id) {
+    auto tx_data = this->transaction_handler->get_transaction_data(evse_id);
```
Pietfried, line 1505: `tx_data` could be `nullptr`, check before operating on it

interfaces/ISO15118_charger.yaml
```diff
@@ -7,18 +7,9 @@ cmds:
...
+          vehicle is connected to
         type: object
         $ref: /iso15118#/EVSEID
-      supported_energy_transfer_modes:
-        description: Available energy transfer modes supported by the EVSE
-        type: array
-        items:
-          description: The different energy modes supported by the SECC 
-          type: object
-          $ref: /iso15118#/SupportedEnergyMode
```
Pietfried, line 19: Looks like `SupportedEnergyMode` is not used anywhere anymore. We can remove the definition from `types/iso15118.yaml`

interfaces/ISO15118_charger.yaml
SebaLukas: I think we dont need a result to tell if the supported energy transfer modes were updated (On startup or dynamically updating)
Instead of blocking the call the d20 module should send an event (new publish var) if the service renegotiation was successfull or not.

SebaLukas (approved):
:+1:

PR #1201, opened 2025-05-30
Set up departure time handling

Abby-Wheelis:
## Describe your changes
Departure Time and Energy Amount have defaults specified in the EvManager manifest.yaml file, which will be used unless other values are passed in as part of the iso_start_v2g_session command

-1 can be passed into the iso_start_v2g_session command in order to enforce use of the defaults

[Issue #220 ](https://github.com/EVerest/EVerest/issues/220)

This replaces #1017


interfaces/ISO15118_ev.yaml
```diff
@@ -9,6 +9,16 @@ cmds:
           by the EVCC
         type: string
         $ref: /iso15118#/EnergyTransferMode
+      DepartureTime:
+        description: >-
+          The time when the EVCC wants to leave the charging station (in seconds)
+          (optional)
```
SebaLukas: Is this really optional? With the config option there is always a default value

modules/EvManager/main/car_simulatorImpl.cpp
```diff
@@ -199,8 +199,22 @@ void car_simulatorImpl::register_all_commands() {
         command_registry->register_command("iso_dc_power_on", 0, [this](const CmdArguments& arguments) {
             return this->car_simulation->iso_dc_power_on(arguments);
         });
-        command_registry->register_command("iso_start_v2g_session", 1, [this](const CmdArguments& arguments) {
-            return this->car_simulation->iso_start_v2g_session(arguments, mod->config.three_phases);
+        command_registry->register_command("iso_start_v2g_session", 3, [this](const CmdArguments& arguments) {
```
SebaLukas, line 203: Because of the changes the node red flows in everest-core/config/nodered/ should also be modified. Without this modification the register_command will fail.

modules/EvseV2G/iso_server.cpp
```diff
@@ -1147,32 +1147,50 @@ static enum v2g_event handle_iso_charge_parameter_discovery(struct v2g_connectio
...
+    struct linked_ac_params {
+        float max_current;
+        int64_t voltage;
+        int64_t pmax;
+    };
+
+    linked_ac_params default_params;
+
+    if (conn->ctx->is_dc_charger == false) {
+        int64_t departure_time_duration = req->AC_EVChargeParameter.DepartureTime;
```
SebaLukas: This makes no sense. `departure_time_duration` is here a local variable within the if. This is freed after the `if`

modules/EvseV2G/iso_server.cpp
```diff
@@ -1147,32 +1147,50 @@ static enum v2g_event handle_iso_charge_parameter_discovery(struct v2g_connectio
     res->EVSEChargeParameter_isUsed = 0;
     res->EVSEProcessing = (iso2_EVSEProcessingType)conn->ctx->evse_v2g_data.evse_processing[PHASE_PARAMETER];
 
+    struct linked_ac_params {
+        float max_current;
+        int64_t voltage;
+        int64_t pmax;
+    };
+
+    linked_ac_params default_params;
```
SebaLukas: Looking over your code I think you dont need `default_params` at all.

SebaLukas: I think you can calculate `voltage` and `pmax` here. But only `pmax` should be stored in a variable to access it twice.

modules/EvseV2G/iso_server.cpp
SebaLukas: Or is this for dlog?

SebaLukas (changes requested):
Please have look into my comments.

modules/EvseV2G/iso_server.cpp
Abby-Wheelis: Yes, it is used for the log.

modules/EvManager/main/car_simulatorImpl.cpp
Abby-Wheelis: I adjusted all of the flows with -1 -1 so the defaults are used and the command won't fail

Abby-Wheelis:
I have addressed the review comments and this is ready for further review

modules/EvseV2G/iso_server.cpp
```diff
@@ -1147,23 +1147,30 @@ static enum v2g_event handle_iso_charge_parameter_discovery(struct v2g_connectio
     res->EVSEChargeParameter_isUsed = 0;
     res->EVSEProcessing = (iso2_EVSEProcessingType)conn->ctx->evse_v2g_data.evse_processing[PHASE_PARAMETER];
 
+    int64_t pmax;
```
SebaLukas: `int64_t pmax{0}` instead of only `int64_t pmax`. Then `pmax` is initialized correctly and does not point anywhere

Abby-Wheelis: Thank you! Fixed this too

config/nodered/config-sil-dc-bpt-flow.json
```diff
@@ -1789,7 +1789,7 @@
         "options": [
             {
                 "label": "DC ISO15118-2",
-                "value": "sleep 1;iso_wait_slac_matched;iso_start_v2g_session DC;iso_wait_pwr_ready;iso_wait_for_stop 36000#iso_stop_charging;iso_wait_v2g_session_stopped;unplug#iso_pause_charging;iso_wait_for_resume#iso_start_bcb_toggle 3;iso_wait_pwm_is_running;iso_start_v2g_session DC;iso_wait_pwr_ready;sleep 36000;",
+                "value": "sleep 1;iso_wait_slac_matched;iso_start_v2g_session DC -1 -1;iso_wait_pwr_ready;iso_wait_for_stop 36000#iso_stop_charging;iso_wait_v2g_session_stopped;unplug#iso_pause_charging;iso_wait_for_resume#iso_start_bcb_toggle 3;iso_wait_pwm_is_running;iso_start_v2g_session DC -1 -1;iso_wait_pwr_ready;sleep 36000;",
```
SebaLukas: I had another look at `-1, -1`. I would like to have “correct” values in the node red flows as default values. I think nobody knows what to do with `-1, -1`.
So instead of `-1, -1` I would like to have `86400, 0`.

modules/EvManager/main/car_simulatorImpl.cpp
```diff
@@ -199,8 +199,22 @@ void car_simulatorImpl::register_all_commands() {
...
+            auto departure_time = std::stoi(arguments[1]);
+            auto e_amount = std::stoi(arguments[2]);
+            EVLOG_debug << "Before ladder departure time is " << departure_time << " and eamount is " << e_amount;
+            if (departure_time < 0) {
```
Abby-Wheelis, line 207: This code snippet handles the -1 dummy values, which forces the use of the values from the manifest of 86400 and 0.

config/nodered/config-sil-dc-bpt-flow.json
Abby-Wheelis: Given that the use of `-1` means the default in the manifest file is used, and the defaults in the manifest are `86400, 0`, do you still want to see that repeated in all of the nodered flows? I was thinking it would be easier to have the default stored in just one location, but if the `-1` values are too confusing I can change them.

modules/EvManager/main/car_simulatorImpl.cpp
SebaLukas: I know. I reviewed your changes

config/nodered/config-sil-dc-bpt-flow.json
SebaLukas: I think this is confusing. I dont know if everyone knows that -1 means that the default values are used. Why not directly use the default values? If someone have a custom node red flows and wants to use -1, -1 this is totally fine.

modules/EvManager/main/car_simulatorImpl.cpp
SebaLukas: The handling of -1, -1 can be left as it is.

Abby-Wheelis: Got it! I will change the nodred files

config/nodered/config-sil-dc-bpt-flow.json
Abby-Wheelis: I will make the change now!

config/config-sil-dc-d20.yaml
```diff
@@ -65,7 +65,11 @@ active_modules:
       connector_id: 1
       auto_enable: true
       auto_exec: false
+<<<<<<< HEAD
+      auto_exec_commands: sleep 3;iso_wait_slac_matched;iso_start_v2g_session DC 86400 0;iso_wait_pwr_ready;iso_wait_for_stop 15;iso_wait_v2g_session_stopped;unplug;
+=======
       auto_exec_commands: sleep 3;iso_wait_slac_matched;iso_start_v2g_session DC;iso_wait_pwr_ready;iso_dc_power_on;iso_wait_for_stop 15;iso_wait_v2g_session_stopped;unplug;
+>>>>>>> upstream/main
```
barsnick: There's a merge error here, and elsewhere.

Abby-Wheelis:
Ah, there is testing code in `everest-utils` that calls `iso_start_v2g_session` and has not had the extra arguments added to it.

config/nodered/config-sil-flow.json
```diff
@@ -3420,4 +3420,4 @@
             ]
         ]
     }
-]
+]
```
SebaLukas: Add again a new line

config/nodered/config-sil-two-evse-flow.json
```diff
@@ -3053,4 +3053,4 @@
             ]
         ]
     }
-]
+]
```
SebaLukas: See above

modules/EvManager/doc.rst
```diff
@@ -29,7 +29,7 @@ The module listens to the following MQTT topics:
     | Used to execute a charging session based on the semicolon separated provided command string.
     ::
 
-        "sleep 1;iso_wait_slac_matched;iso_start_v2g_session DC_extended;iso_wait_pwr_ready;sleep 36000"
+        "sleep 1;iso_wait_slac_matched;iso_start_v2g_session DC_extended 86400 0;iso_wait_pwr_ready;sleep 36000"
```
SebaLukas: Please change `DC_extended` to `DC`. Someone in the past forgot that to update :sweat_smile:

modules/EvManager/EvManager.hpp
```diff
@@ -47,6 +47,8 @@ struct Conf {
     int dc_discharge_v2g_minimal_soc;
     double max_current;
     bool three_phases;
+    int departure_time;
```
SebaLukas, line 51: Please add `departure_time` and `e_amount` to the EvManager `manifest.yaml` file.

modules/EvManager/main/car_simulation.cpp
```diff
@@ -268,19 +268,27 @@ bool CarSimulation::iso_dc_power_on(const CmdArguments& arguments) {
...
+    const auto& e_amount = std::stoi(arguments[2]);
+
+    EVLOG_debug << "energy mode: " << energy_mode << " departure time: " << departure_time << " eamount: " << e_amount
+                << " three phases: " << three_phases;
 
     if (energy_mode == constants::AC) {
         sim_data.energy_mode = EnergyMode::AC;
         if (three_phases == false) {
-            r_ev[0]->call_start_charging(types::iso15118::EnergyTransferMode::AC_single_phase_core);
-            charge_mode = ChargeMode::AC;
```
SebaLukas, line 304: Please add `charge_mode = ChargeMode::AC;` back

modules/EvseV2G/iso_server.cpp
SebaLukas: Then you can use `req->AC_EVChargeParameter.DepartureTime` directly in line 1163 calling dlog.

SebaLukas (changes requested):
Have a look in my comments

SebaLukas (dismissed):

dependencies.yaml
```diff
@@ -84,7 +84,7 @@ libocpp:
 # Josev
 Josev:
   git: https://github.com/EVerest/ext-switchev-iso15118.git
-  git_tag: 2025.9.0
+  git_commit_hash: c31c4e7
```
SebaLukas: This is not correct. Please use `git_tag: ` and the full hash commit

Abby-Wheelis:
There were too many merges to upstream in order to squash, superseded by #1468

SebaLukas:
Honestly I dont know where is the problem and why you produce so many commits.
With e.g. `git rebase -i HEAD~30` it is quite easy to squash the first 30 commits together: https://git-scm.com/docs/git-rebase#_interactive_mode
You can always work locally with your repo, check whether the squash worked, and don't have to force push your changes to the remote branch right away.

PR #1202, opened 2025-06-02
refactor: Changed certificate verification in openssl utils

Pietfried:
## Describe your changes
* load_contract_root_cert always uses MO and V2G trust anchors instead of only using V2G trust anchors if MO trust anchors are empty
* introduced new response code CertificateNotAllowed that applies in case a contract trust anchor is installed but does not match. This will result in responding with iso2_responseCodeType_FAILED_CertificateNotAllowedAtThisEVSE as required for the Hubject Plug&Charge certification. NoCertificateAvailable is only used in case no contract trust anchor is installed at all

### DIFF
--- a/modules/EvseV2G/crypto/crypto_openssl.cpp
+++ b/modules/EvseV2G/crypto/crypto_openssl.cpp
@@ -129,17 +129,17 @@ bool check_iso2_signature(const struct iso2_SignatureType* iso2_signature, EVP_P
 
 bool load_contract_root_cert(::openssl::certificate_list& trust_anchors, const char* V2G_file_path,
                              const char* MO_file_path) {
-    // note the file(s) may contain more than one certificate (hence must be PEM)
-    // try MO_file_path first then fallback to V2G_file_path
-
     trust_anchors.clear();
-    trust_anchors = ::openssl::load_certificates(MO_file_path);
+
+    auto mo_certs = ::openssl::load_certificates(MO_file_path);
+    trust_anchors = std::move(mo_certs);
+
+    auto v2g_certs = ::openssl::load_certificates(V2G_file_path);
+    trust_anchors.insert(trust_anchors.end(), std::make_move_iterator(v2g_certs.begin()),
+                         std::make_move_iterator(v2g_certs.end()));
+
     if (trust_anchors.empty()) {
-        log_error("Unable to load MO root(s)");
-        trust_anchors = ::openssl::load_certificates(V2G_file_path);
-        if (trust_anchors.empty()) {
-            log_error("Unable to load V2G root(s)");
-        }
+        log_error("Unable to load any MO or V2G root(s)");
     }
 
     return !trust_anchors.empty();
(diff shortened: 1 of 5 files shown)


modules/EvseV2G/crypto/crypto_openssl.cpp
```diff
@@ -129,17 +129,18 @@ bool check_iso2_signature(const struct iso2_SignatureType* iso2_signature, EVP_P
...
     trust_anchors.clear();
-    trust_anchors = ::openssl::load_certificates(MO_file_path);
+
+    auto mo_certs = ::openssl::load_certificates(MO_file_path);
+    trust_anchors.insert(trust_anchors.end(), std::make_move_iterator(mo_certs.begin()),
+                         std::make_move_iterator(mo_certs.end()));
+
+    auto v2g_certs = ::openssl::load_certificates(V2G_file_path);
+    trust_anchors.insert(trust_anchors.end(), std::make_move_iterator(v2g_certs.begin()),
+                         std::make_move_iterator(v2g_certs.end()));
```
Pietfried: @james-ctc @SebaLukas , according to hubject we can always use v2g and mo as trust anchors. I have not checked if this is in line with iso15118

AssemblyJohn (approved):
Code looks good to me. I can't commend on the ISO requirement, though.

modules/EvseV2G/crypto/crypto_openssl.cpp
SebaLukas: What does Hubject mean by that? The contract leaf and the sub CA 1 and 2 certificates can be based on the MO or on the V2G Root. That is why we load both roots here. If one is not available, the other can still be used.

SebaLukas: Ah, now I understand. Before the code change it was either or. Then Hubject is right. If there are several roots (MO and V2G), they should all be loaded. Not just MO or V2G

modules/EvseV2G/crypto/crypto_openssl.cpp
```diff
@@ -129,17 +129,18 @@ bool check_iso2_signature(const struct iso2_SignatureType* iso2_signature, EVP_P
...
 bool load_contract_root_cert(::openssl::certificate_list& trust_anchors, const char* V2G_file_path,
                              const char* MO_file_path) {
-    // note the file(s) may contain more than one certificate (hence must be PEM)
-    // try MO_file_path first then fallback to V2G_file_path
-
     trust_anchors.clear();
-    trust_anchors = ::openssl::load_certificates(MO_file_path);
+
+    auto mo_certs = ::openssl::load_certificates(MO_file_path);
+    trust_anchors.insert(trust_anchors.end(), std::make_move_iterator(mo_certs.begin()),
```
james-ctc: Line 134 could be 
```
trust_anchors = ::openssl::load_certificates(MO_file_path);
```
removing the need for the insert operation.
Alternatively
```
auto mo_certs = ::openssl::load_certificates(MO_file_path);
trust_anchors = std::move(mo_certs);
```
since mo_certs isn't needed anymore

Pietfried: addressed

PR #1217, opened 2025-06-06
Added TLS server suspend method to reject incoming connections

james-ctc:
## Describe your changes

feat: added TLS server suspend method
The TLS server can be suspended where it will reject all incoming connections until update() is called with new SSL configuration.

When suspended the state() is state_t::init_socket After update() the state() is state_t::init_complete and progresses to state_t::running following the next connection (or after a timeout)


Pietfried (approved):

SebaLukas (approved):
Looking good!

PR #1219, opened 2025-06-10
Fix(EvseManager): Removing inform_max_hlc_limits() when the power supply updates its capabilities

SebaLukas:
## Describe your changes
See title

In some cases hlc max limits was overwritten when the power supply updates its capabilities. If the previous max hlc limits are smaller than the capabilities, then the EvseManager was falsely overriding the new capabilities max limits to the powersupply. In this case the output of the powersupply was bigger than actually set by the EnergyManager.


Pietfried (approved):
LGTM.

Since the updated capabilities are not applied immediately, [this issue](https://github.com/EVerest/everest-core/issues/1223) can be used for a follow up PR.

PR #1224, opened 2025-06-12
 feat(IsoExtension/EvseManager): Add vars to interfaces

mlitre:
## Describe your changes

Added new vars to evse_manager interface so that we can broadcast the list of supported energy transfers as well as operation modes.

Added a new var to the iso15118_extensions interface to publish ev information. This is mainly used to update the device model for OCPP 2.1 requirements during V2X. 

Related to https://github.com/EVerest/libocpp/issues/1068. 

### DIFF
--- a/modules/EvseManager/EvseManager.cpp
+++ b/modules/EvseManager/EvseManager.cpp
@@ -2,6 +2,7 @@
 // Copyright Pionix GmbH and Contributors to EVerest
 #include "EvseManager.hpp"
 
+#include <algorithm>
 #include <fmt/color.h>
 #include <fmt/core.h>
 
@@ -9,6 +10,7 @@
 #include "SessionLog.hpp"
 #include "Timeout.hpp"
 #include "scoped_lock_timeout.hpp"
+#include "utils.hpp"
 
 using namespace std::literals::chrono_literals;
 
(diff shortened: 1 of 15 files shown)


SebaLukas:
Todo: `Describe your changes` & `Issue ticket number and link` are missing

modules/EvseManager/EvseManager.cpp
```diff
@@ -114,8 +116,17 @@ void EvseManager::init() {
...
-                [this](const auto& caps) { update_powersupply_capabilities(caps); });
+            r_powersupply_DC[0]->subscribe_capabilities([this](const auto& caps) {
+                update_powersupply_capabilities(caps);
+                const bool dc_was_updated = update_supported_energy_transfers(types::iso15118::EnergyTransferMode::DC);
+                bool dc_bpt_was_updated = false;
+                if (caps.bidirectional) {
+                    dc_bpt_was_updated = update_supported_energy_transfers(types::iso15118::EnergyTransferMode::DC_BPT);
+                }
+                if (dc_was_updated || dc_bpt_was_updated) {
+                    this->p_evse->publish_supported_energy_transfer_modes(supported_energy_transfers);
```
SebaLukas, line 127: I'm not sure if we should do this here.

mlitre: Should we only do it in the ready then?

modules/EvseManager/manifest.yaml
```diff
@@ -338,6 +338,20 @@ config:
...
+      Idle
+      ChargingOnly
+      CentralSetpoint
+      ExternalSetpoint
+      ExternalLimits
+      CentralFrequency
+      LocalFrequency
+      LocalLoadBalancing
+    type: string
+    default: Idle,ChargingOnly
```
Pietfried: I'd prefer to drop this config parameter definition for now until we know better how we integrate the supported operation modes. For now it is anyways just a placeholder and not part of any business logic

types/evse_manager.yaml
```diff
@@ -558,3 +558,24 @@ types:
       - IncompatibleEnergyTransfer
       - ServiceRenegotiationFailed
       - NoHlc
+  OperationModeEnum:
```
Pietfried: Same here, I would drop this for now, until we know if this belongs rather to OCPP, EnergyManager or the EvseManager

Pietfried (commented):
Overall this looks good to me.

As mentioned in line I would drop adding the `supportedOperationModes` for now. The integration of `supportedEnergyTransferModes` looks fine!

modules/EvseManager/EvseManager.cpp
SebaLukas: Ah, I had made a misunderstanding. I thought the supported_energy_transfers are sent to the ISO module. But that's not the case at all.

So you could do it that way. You might just have to synchronize it with the ISO module somehow. But that can happen later.

SebaLukas (approved):
Looking good :+1:

modules/EvseManager/EvseManager.cpp
mlitre: This is the list of supported energy transfers aka all physically possible, so it is not meant to be restrictive, but informative here.

types/evse_manager.yaml
mlitre: Ok I will remove these for now then and probably hardcode them into OCPP for now

PR #1233, opened 2025-06-28
EvseManager: log duty cycle only with single decimal digit

mhei:
## Describe your changes

This should normally be sufficient and is easier to read.

none

### DIFF
--- a/modules/EvseManager/Charger.cpp
+++ b/modules/EvseManager/Charger.cpp
@@ -998,7 +998,7 @@ void Charger::update_pwm_now(float dc) {
     session_log.evse(
         false,
         fmt::format(
-            "Set PWM On ({}%) took {} ms", dc * 100.,
+            "Set PWM On ({.1f}%) took {} ms", dc * 100.,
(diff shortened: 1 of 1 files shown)


SebaLukas (approved):
Looking good :+1: 

Have a look into the integration tests. One test case is failing

modules/EvseManager/Charger.cpp
```diff
@@ -998,7 +998,7 @@ void Charger::update_pwm_now(float dc) {
     session_log.evse(
         false,
         fmt::format(
-            "Set PWM On ({}%) took {} ms", dc * 100.,
+            "Set PWM On ({.1f}%) took {} ms", dc * 100.,
```
hikinggrass:
```suggestion
            "Set PWM On ({:.1f}%) took {} ms", dc * 100.,
```

Otherwise EvseManager crashes with:
`terminate called after throwing an instance of 'fmt::v10::format_error'
  what():  invalid format string`

hikinggrass (approved):

PR #1236, opened 2025-07-01
feat: add ISO15118D20 custom VAS functionality

markxoe:
## Describe your changes

Needs https://github.com/EVerest/libiso15118/pull/122

Add necessary interface to Evse15118D20 module. Add callbacks and functions to support ISO15118D20 custom VAS provided via ISO15118_vas interface

ToDo:
- [x] update `dependencies.yaml`

### DIFF
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -47,7 +47,7 @@ libcbv2g:
 # libiso15118
 libiso15118:
   git: https://github.com/EVerest/libiso15118.git
-  git_tag: 414d66b48cb89f1e721e340838497969d88b2a41
+  git_tag: v0.7.0
   cmake_condition: "EVEREST_DEPENDENCY_ENABLED_LIBISO15118"
 
 # LEM DCBM 400/600, IsabellenhuetteIemDcr modules


modules/Evse15118D20/charger/ISO15118_chargerImpl.cpp
```diff
@@ -239,6 +286,31 @@ void ISO15118_chargerImpl::init() {
             break;
         }
     });
+
+    supported_vas_services_per_provider.reserve(mod->r_iso15118_vas.size());
+
+    for (size_t i = 0; i < mod->r_iso15118_vas.size(); i++) {
+        supported_vas_services_per_provider.emplace_back();
+
+        this->mod->r_iso15118_vas[i]->subscribe_offered_vas([this, i](Array array) {
```
markxoe: I don't like that the payload is given as a "untyped" Array. The only solution I see would be to update the Interface so that the variable is a object and not an array

modules/Evse15118D20/charger/ISO15118_chargerImpl.cpp
```diff
@@ -298,13 +370,51 @@ void ISO15118_chargerImpl::ready() {
...
+void ISO15118_chargerImpl::update_supported_vas_services() {
+    iso15118::d20::SupportedVASs supported_vas_services;
+
+    for (const auto& provider_services : supported_vas_services_per_provider) {
+        for (const auto& service_id : provider_services) {
+            // Check for duplicate service IDs across all providers
+            if (std::find(supported_vas_services.begin(), supported_vas_services.end(), service_id) !=
+                supported_vas_services.end()) {
+                EVLOG_AND_THROW(
+                    Everest::EverestConfigError("Duplicate VAS service ID found: " + std::to_string(service_id)));
```
SebaLukas: The question is whether you should really throw an exception here and then need to restart the entire charger.
Perhaps only an error should appear in the log and the service ID and the incorrectly configured VAS module should be ignored

modules/Evse15118D20/charger/ISO15118_chargerImpl.cpp
SebaLukas: Then change the interface and add a new type :D As part of this PR.

modules/Evse15118D20/charger/ISO15118_chargerImpl.cpp
```diff
@@ -474,6 +584,56 @@ iso15118::session::feedback::Callbacks ISO15118_chargerImpl::create_callbacks()
...
+                EVLOG_warning << "Selected Service ID " << service.service_id
+                              << " is not supported by any VAS provider";
+                continue;
+            }
+
+            types::iso15118_vas::SelectedService converted_service;
+            converted_service.service_id = service.service_id;
+            converted_service.parameter_set_id = service.parameter_set_id;
+
+            services_by_provider[*provider_index].push_back(converted_service);
```
SebaLukas, line 600: Why not call `mod->r_iso15118_vas[provider_index]->call_selected_services(converted_service);` directly here?

modules/Evse15118D20/charger/ISO15118_chargerImpl.cpp
```diff
@@ -89,6 +89,53 @@ convert_dynamic_values(const iso15118::message_20::datatypes::BPT_Dynamic_DC_CLR
...
+    for (const auto& parameter : parameter_set.parameters) {
+        out.parameter.push_back(convert_parameter(parameter));
+    }
+
+    return out;
+}
+
+std::vector<iso15118::message_20::datatypes::ParameterSet>
+convert_parameter_set_list(const std::vector<types::iso15118_vas::ParameterSet>& parameter_set_list) {
+    std::vector<iso15118::message_20::datatypes::ParameterSet> out;
```
SebaLukas: You can use `dt::` instead of `iso15118::message_20::datatypes`

SebaLukas (commented):
Overall it looks good!
Have a look into my comments.

modules/Evse15118D20/charger/ISO15118_chargerImpl.cpp
markxoe: That is true, will change that

modules/Evse15118D20/charger/ISO15118_chargerImpl.cpp
markxoe: Will change 👍

modules/Evse15118D20/charger/ISO15118_chargerImpl.cpp
markxoe: I thought in some cases it might be useful that the provider receives the list of selected services in a single call. One could also just call all providers with the same payload, so that every provider knows about all selected services

modules/Evse15118D20/charger/ISO15118_chargerImpl.cpp
markxoe: 😄 👍

modules/Evse15118D20/charger/ISO15118_chargerImpl.cpp
SebaLukas: Sry I dont get it. There is no single call with all selected services. For every selected service there is an extra call to the provider.

SebaLukas: Right now there is one small difference:
Your approach:
```
provider[0]->call(34)
provider[0]->call(35)
provider[1]->call(6000)
```
My suggestion (Depends on selected_services):
```
provider[0]->call(34)
provider[1]->call(6000)
provider[0]->call(35)
```
I don't think that's a problem for the VAS modules.

markxoe: The selected services are first grouped by provider index, this is the code above. It is stored in the map `services_by_provider`. Your example would result in the map `{0: [34, 35], 1: [6000]}`. After grouping, the providers are notified in the code below, resulting in the two calls `provider[0]->call([34,35]); provider[1]->call([6000])`. The call itself is happening in the code below, lines 604-606

SebaLukas: Oh boy. For whatever reason, I thought it was a `std::map` with no vector. Doesn't make much difference though. Except that there are fewer calls. So your suggestion works!

What else I noticed: The `iso15118_vas` interface does not define how often the cmd `selected_services` may be called.
So once with all SelectedService or several calls with one SelectedService each.
It would make sense that you can only call the cmd once per ServiceSelectionReq/Session. Then it is also easier for the module implementation and does not have to cache the SelectedServices.

SebaLukas (approved):
:+1:

modules/Evse15118D20/charger/ISO15118_chargerImpl.cpp
markxoe: Good point; I've added a note to the interface specifying that `selected_services` shall only be called once per Request 👍

SebaLukas:
https://github.com/EVerest/libiso15118/pull/122 is merged

markxoe:
@SebaLukas should I update the dependency.yaml with the current libiso commit or wait for a new libiso release tag?

SebaLukas:
@markxoe You can update dependency.yaml with the current libiso commit

markxoe:
👍 Dependency is updated

PR #1238, opened 2025-07-03
feat(OCPP201): Include setpoints in OCPP201

Pietfried:
## Describe your changes
Add conversions from retrieved ChargingSchedulePeriod setpoints to EVerest external limits in OCPP201 module. With this change, the OCPP201 module is includes the setpoint information when calling the set_external_limits command at its evse_energy_sink(s)

### DIFF
--- a/modules/OCPP201/OCPP201.cpp
+++ b/modules/OCPP201/OCPP201.cpp
@@ -26,8 +26,13 @@ const std::string MASTER_PASS_GROUP_ID_VAR_NAME = "MasterPassGroupId";
 const std::string EV_CONNECTION_TIMEOUT_VAR_NAME = "EVConnectionTimeOut";
 const std::string CENTRAL_CONTRACT_VALIDATION_ALLOWED_VAR_NAME = "CentralContractValidationAllowed";
 const std::string CONTRACT_CERTIFICATE_INSTALLATION_ENABLED_VAR_NAME = "ContractCertificateInstallationEnabled";
+const std::string SETPOINT_PRIORITY_VAR_NAME = "SetpointPriority";
 const std::string TX_START_POINT_VAR_NAME = "TxStartPoint";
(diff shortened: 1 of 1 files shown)


modules/OCPP201/OCPP201.cpp
```diff
@@ -26,8 +26,13 @@ const std::string MASTER_PASS_GROUP_ID_VAR_NAME = "MasterPassGroupId";
 const std::string EV_CONNECTION_TIMEOUT_VAR_NAME = "EVConnectionTimeOut";
 const std::string CENTRAL_CONTRACT_VALIDATION_ALLOWED_VAR_NAME = "CentralContractValidationAllowed";
 const std::string CONTRACT_CERTIFICATE_INSTALLATION_ENABLED_VAR_NAME = "ContractCertificateInstallationEnabled";
+const std::string SETPOINT_PRIORITY_VAR_NAME = "SetpointPriority";
 const std::string TX_START_POINT_VAR_NAME = "TxStartPoint";
 const std::string TX_STOP_POINT_VAR_NAME = "TxStopPoint";
+const std::string SETPOINT_SOURCE = "OCPP";
+
+static constexpr int32_t LOWEST_SETPOINT_PRIORITY = 1000;
+static constexpr int32_t HIGHEST_SETPOINT_PRIORITY = 0;
```
mlitre, line 35: I did not see anything documented on the interface / type definition regarding priority. This confused me for a second because instinctively I expected the setpoint with the largest value as the setpoint with the highest priority, but it seems to be the opposite here. Could we either leave a comment or update the interface / type definition to explain this information?

Pietfried: It is documented as part of the `SetpointType` here: https://github.com/EVerest/everest-core/blob/feature/ocpp21-setpoint-support/types/energy.yaml#L186

mlitre (approved):

PR #1242, opened 2025-07-04
PhyVersoBSP: BSP rework alongside new MCU firmware release

rckstrh:
* MCU<>Everest protobuf extension/rework
    * Config structure for a charging port has changed -> more info under its own bullet point below
    * motor locking states locking/unlocking added, not only endstates (but not really usable at the moment as motor lock interface definition does not have lock state vars where we could communicate state to other modules)
    * KeepAlive packet has added configuration_done flag that signals everest the MCU has successfully been configured, booted and is in normal run mode
    * added error flags for PP fault, coil/relay feedback divergence from target state, heartbeat timeout

* Config extension
    * coil/relay feedback active high/low and pull directions can be configured
    * motor lock type defaults to no motor lock (as is the case for DC charging or AC charging without socket/detachable charging cable)
    * has_socket defines if we have a socketed charging port (with detachable cable)
    * if configuring a port for AC charging with a socket, critical errors will be thrown on everest/MCU side when not specifying motor lock type
    * Motor lock type only enforced on socketed AC ports
    * possible to completely disable charging port to do one charging port configs only
    * phyverso_cli parameters via json mirror whole everest module parameter set now
    * renamed cli_config_*.json files for more clarity where they could be used (phyverso_cli use only, no real-world application)

* Error handling
    * Bidirectional serial communication timeouts will be reported and throw respective errors / will get cleared after normal communcation resumes
    * new errors thrown on everest side that will also have already lead to an emergency stop on MCU side
        * PP fault
        * coil/relay feedback diverges from target 
        * heartbeat timeout (connection to/from Everest lost) 
        * diode fault

* Misc.
    * PP state logging fixed, was constantly spamming EVLOG with info messages when receiving periodic updates from MCU
    * sometimes serial communication gets completely mangled, leading to Everest not receiving config request packets -> MCU will not get config packets and will hang in boot loop
        * BSP will wait for MCU reporting that it is properly configured -> otherwise reset MCU / serial after some time and wait for new config request packet until MCU reports properly configured MCU, boot mode exit and normal operation
    * increased hard-reset nRST low time
    * added small delay in phyverso_cli input polling loop to not use 100% CPU

### DIFF
--- a/modules/HardwareDrivers/EVSE/PhyVersoBSP/PhyVersoBSP.cpp
+++ b/modules/HardwareDrivers/EVSE/PhyVersoBSP/PhyVersoBSP.cpp
@@ -22,23 +22,75 @@ void PhyVersoBSP::init() {
         return;
     }
 
+    serial.flush_buffers();
+
+    serial.signal_config_request.connect([&]() {
+        serial.send_config();
+        EVLOG_info << "Sent config packet to MCU";
+    });
+
+    serial.signal_connection_timeout.connect([this]() {
+        auto err = p_connector_1->error_factory->create_error("evse_board_support/CommunicationFault", "McuToEverest",
+                                                              "Serial connection to MCU timed out");
+        p_connector_1->raise_error(err);
+        err = p_connector_2->error_factory->create_error("evse_board_support/CommunicationFault", "McuToEverest",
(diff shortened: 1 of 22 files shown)


rckstrh:
Codacy issues are false positives.

PR #1247, opened 2025-07-08
OCPP201: add access to module configs using EVerest config service

hikinggrass:

### DIFF
--- a/modules/OCPP201/device_model/everest_device_model_storage.cpp
+++ b/modules/OCPP201/device_model/everest_device_model_storage.cpp
@@ -1,6 +1,8 @@
 // SPDX-License-Identifier: Apache-2.0
 // Copyright Pionix GmbH and Contributors to EVerest
 
+#include <everest/logging.hpp>
+
 #include <device_model/definitions.hpp>
 #include <device_model/everest_device_model_storage.hpp>
 #include <ocpp/v2/init_device_model_db.hpp>
@@ -22,6 +24,33 @@ using ocpp::v2::DbVariableAttribute;
 using ocpp::v2::DeviceModelVariable;
 
 namespace module::device_model {
+ocpp::v2::DataEnum to_ocpp_data_enum(const everest::config::Datatype& data_type) {
+    switch (data_type) {
+    case everest::config::Datatype::Unknown:
+        throw std::out_of_range("Could not convert Datatype::Unknown to DataEnum");
+    case everest::config::Datatype::String:
+        return ocpp::v2::DataEnum::string;
+    case everest::config::Datatype::Decimal:
+        return ocpp::v2::DataEnum::decimal;
+    case everest::config::Datatype::Integer:
+        return ocpp::v2::DataEnum::integer;
+    case everest::config::Datatype::Boolean:
+        return ocpp::v2::DataEnum::boolean;
+    }
+    throw std::out_of_range("Could not convert Datatype to DataEnum");
+}
+
+ocpp::v2::MutabilityEnum to_ocpp_mutability_enum(const everest::config::Mutability& mutability) {
+    switch (mutability) {
+    case everest::config::Mutability::ReadOnly:
+        return ocpp::v2::MutabilityEnum::ReadOnly;
+    case everest::config::Mutability::ReadWrite:
+        return ocpp::v2::MutabilityEnum::ReadWrite;
+    case everest::config::Mutability::WriteOnly:
+        return ocpp::v2::MutabilityEnum::WriteOnly;
+    }
+    throw std::out_of_range("Could not convert Mutability to MutabilityEnum");
+}
 
 namespace {
 
@@ -56,7 +85,8 @@ ComponentKey get_connector_component_key(const int32_t evse_id, const int32_t co
 
 // Helper function to construct DeviceModelVariable with common structure
 DeviceModelVariable make_variable(const std::string& name, const ocpp::v2::VariableCharacteristics& characteristics,
-                                  const std::string& value = "") {
+                                  const std::string& value = "",
+                                  ocpp::v2::MutabilityEnum mutability = ocpp::v2::MutabilityEnum::ReadOnly) {
     DeviceModelVariable var_data;
     var_data.name = name;
     var_data.characteristics = characteristics;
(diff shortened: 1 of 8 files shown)


modules/OCPP201/device_model/everest_device_model_storage.cpp
```diff
@@ -130,12 +204,35 @@ EverestDeviceModelStorage::EverestDeviceModelStorage(
...
+    // Component.evse.id/connector = mapping of module
+    // FIXME: what to do when there are different mappings from module & impl?
+    for (const auto& [module_id_type, module_config] : this->module_configs) {
+        ComponentKey component_key;
+        component_key.name = module_id_type.module_type;
+        component_key.instance = module_id_type.module_id;
+        const auto& mapping = this->mappings.at(module_id_type.module_id);
+        if (mapping.module.has_value()) {
+            const auto& module_mapping = mapping.module.value();
+            component_key.evse_id = module_mapping.evse;
```
Pietfried: we have to be careful here to not map to id `0`. EVerest allows it, OCPP2.0.1 doesnt.

image

modules/OCPP201/device_model/everest_device_model_storage.hpp
```diff
@@ -11,15 +11,28 @@
...
+    Variable,
+    ConfigEntry,
+};
+
 class EverestDeviceModelStorage : public ocpp::v2::DeviceModelStorageInterface {
+private:
+    std::shared_ptr<Everest::config::ConfigServiceClient> config_service_client;
+    std::map<Everest::config::ModuleIdType, everest::config::ModuleConfigurationParameters> module_configs;
+    std::map<std::string, ModuleTierMappings> mappings;
+    std::map<std::string, std::string> variable_name_impl_config;
```
Pietfried: nit: This can be moved to the private section of the class below

modules/OCPP201/device_model/everest_device_model_storage.cpp
```diff
@@ -130,12 +204,40 @@ EverestDeviceModelStorage::EverestDeviceModelStorage(
         }
     }
 
+    // build components from EVerest config
```
Pietfried: nit:
```suggestion
    // build OCPP2.x device model components from EVerest config
```

modules/OCPP201/device_model/everest_device_model_storage.cpp
```diff
@@ -130,12 +204,40 @@ EverestDeviceModelStorage::EverestDeviceModelStorage(
...
+    // This is our mapping strategy:
+    // Component.name = module_type
+    // Component.instance = module_id
+    // Component.evse.id/connector = mapping of module
+    // FIXME: what to do when there are different mappings from module & impl?
```
Pietfried: Since we don't represent the impls as individual OCPP components anyways, but just present the modules as OCPP components, I would just neglect these impl mappings for now

modules/OCPP201/device_model/everest_device_model_storage.cpp
```diff
@@ -130,12 +204,40 @@ EverestDeviceModelStorage::EverestDeviceModelStorage(
...
+    for (const auto& [module_id_type, module_config] : this->module_configs) {
+        ComponentKey component_key;
+        component_key.name = module_id_type.module_type;
+        component_key.instance = module_id_type.module_id;
+        const auto& mapping = this->mappings.at(module_id_type.module_id);
+        if (mapping.module.has_value()) {
+            const auto& module_mapping = mapping.module.value();
+            if (module_mapping.evse > 0) {
```
Pietfried, line 220: nit:
```suggestion
            // in OCPP2.x the id and connectorId of the EVSEType must be > 0
            if (module_mapping.evse > 0) {
```

modules/OCPP201/device_model/everest_device_model_storage.cpp
```diff
@@ -130,12 +204,40 @@ EverestDeviceModelStorage::EverestDeviceModelStorage(
...
+                component_key.evse_id = module_mapping.evse;
+                if (module_mapping.connector.has_value()) {
+                    const auto connector_id = module_mapping.connector.value();
+                    if (connector_id >= 0) {
```
Pietfried:
```suggestion
                    if (connector_id > 0) {
```

modules/OCPP201/device_model/everest_device_model_storage.hpp
```diff
@@ -11,15 +11,28 @@
...
 #include <ocpp/v2/device_model_storage_interface.hpp>
 #include <ocpp/v2/device_model_storage_sqlite.hpp>
+#include <utils/config_service.hpp>
 
 namespace module::device_model {
 
+enum class SourceType {
+    Variable,
+    ConfigEntry,
+};
```
Pietfried: It's probably just me, but I dont get the concept of this `SourceType` and the `source_mapping`

hikinggrass: replaced this with a set that keeps track of the variables that are backed by the config service

Pietfried (dismissed):
lgtm

Pietfried (approved):

modules/OCPP201/device_model/everest_device_model_storage.cpp
```diff
@@ -168,6 +270,41 @@ void EverestDeviceModelStorage::init_hw_capabilities(
     }
 }
 
+void EverestDeviceModelStorage::init_everest_config() {
```
Pietfried, line 273: Aren't we writing the everest config to the database already in the constructor? Looks like this function does the same thing again

hikinggrass: The build_...() one in the ctor prepares the these `component_configs` that seem to be necessary to initialize the db on libocpp side, while the `init_...()` the sets the initial value

Pietfried: The value is also set in `build...()` , https://github.com/EVerest/everest-core/pull/1247/files#diff-8bc11d4241cbc52477542c7d2312571c995009da1e52a92143b6a165c7b0fee1R168-R170

hikinggrass: I just tried just using the init() but that doesn't work because we have to call the `this->device_model_storage->set_variable_attribute_value()` on startup, otherwise we never update the values in the libocpp view of the device model :thinking:

PR #1254, opened 2025-07-16
feat(OCPPv2.x): Add support for missing conversions for v2x params

mlitre:
Added support for missing conversion of the notify ev charging needs message.

### DIFF
--- a/modules/OCPP201/conversions.cpp
+++ b/modules/OCPP201/conversions.cpp
@@ -822,28 +822,194 @@ ocpp::v2::EnergyTransferModeEnum to_ocpp_energy_transfer_mode(const types::iso15
     case types::iso15118::EnergyTransferMode::DC_unique:
         return ocpp::v2::EnergyTransferModeEnum::DC;
 
+    // OCPP 2.1+ enums
     case types::iso15118::EnergyTransferMode::AC_BPT:
+        return ocpp::v2::EnergyTransferModeEnum::AC_BPT;
     case types::iso15118::EnergyTransferMode::AC_BPT_DER:
+        return ocpp::v2::EnergyTransferModeEnum::AC_BPT_DER;
     case types::iso15118::EnergyTransferMode::AC_DER:
+        return ocpp::v2::EnergyTransferModeEnum::AC_DER;
     case types::iso15118::EnergyTransferMode::DC_BPT:
+        return ocpp::v2::EnergyTransferModeEnum::DC_BPT;
     case types::iso15118::EnergyTransferMode::DC_ACDP:
+        return ocpp::v2::EnergyTransferModeEnum::DC_ACDP;
     case types::iso15118::EnergyTransferMode::DC_ACDP_BPT:
+        return ocpp::v2::EnergyTransferModeEnum::DC_ACDP_BPT;
     case types::iso15118::EnergyTransferMode::WPT:
-        throw std::out_of_range("Could not convert EnergyTransferModeEnum: OCPP does not know this type");
+        return ocpp::v2::EnergyTransferModeEnum::WPT;
     }
 
     throw std::out_of_range("Could not convert EnergyTransferMode");
 }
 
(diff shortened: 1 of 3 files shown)


mlitre:
@Pietfried do we already have a util for transforming a float to rational number?
Currently even on the libiso side we do not transfer the energy offer, so we can either leave this as is or add this to future proof.

modules/OCPP201/conversions.cpp
```diff
@@ -822,28 +822,111 @@ ocpp::v2::EnergyTransferModeEnum to_ocpp_energy_transfer_mode(const types::iso15
...
+            }
+        }
+    }
+
+    return result;
+}
+
 ocpp::v2::NotifyEVChargingNeedsRequest
 to_ocpp_notify_ev_charging_needs_request(const types::iso15118::ChargingNeeds& charging_needs) {
     // The evseId is calculated outside of this function in the required OCPP201 module
```
Pietfried: in that case this function could simply be `to_ocpp_charging_needs(const types::iso15118::ChargingNeeds& charging_needs)`

modules/OCPP201/conversions.cpp
```diff
@@ -864,7 +947,43 @@ to_ocpp_notify_ev_charging_needs_request(const types::iso15118::ChargingNeeds& c
...
+        v2x_charging_params.maxDischargePower = v2x.max_discharge_power;
+        v2x_charging_params.maxDischargePower_L2 = v2x.max_discharge_power_l2;
+        v2x_charging_params.maxDischargePower_L3 = v2x.max_discharge_power_l3;
+        v2x_charging_params.minDischargePower = v2x.min_discharge_power;
+        v2x_charging_params.minDischargePower_L2 = v2x.min_discharge_power_l2;
+        v2x_charging_params.minDischargePower_L3 = v2x.min_discharge_power_l3;
+    }
+    if (charging_needs.der_charging_parameters.has_value()) {
+        const auto& der = charging_needs.der_charging_parameters.value();
+        auto& der_charging_params = _charging_needs.derChargingParameters.emplace();
```
Pietfried, line 1064: Looks like the `der` charging parameters are never populated to `der_charging_params` ?

modules/OCPP201/conversions.cpp
```diff
@@ -864,7 +947,43 @@ to_ocpp_notify_ev_charging_needs_request(const types::iso15118::ChargingNeeds& c
         dc_charging_parameters.evEnergyCapacity = dc.ev_energy_capacity;
         dc_charging_parameters.fullSoC = dc.full_soc;
         dc_charging_parameters.bulkSoC = dc.bulk_soc;
-    }
+    } else if (charging_needs.v2x_charging_parameters.has_value()) {
```
Pietfried, line 1033: is else if correct here, so v2x_charging_parameters can only be set if ac or dc are not set?

Pietfried (changes requested):

modules/OCPP201/conversions.cpp
mlitre: As far as I understood it yes, since ac/dc is for -2 and v2x for -20

modules/OCPP201/conversions.cpp
mlitre: My bad, it has been added now

modules/OCPP201/conversions.cpp
mlitre: Done, that makes more sense indeed

mlitre:
> > @Pietfried do we already have a util for transforming a float to rational number? Currently even on the libiso side we do not transfer the energy offer, so we can either leave this as is or add this to future proof.
> 
> We need to address this at some point, in libocpp and in EVerests type system, but I would not do it as part of this PR but rather create an issue for it

Ok that makes sense, I'll go ahead and create it.

Pietfried (approved):

PR #1255, opened 2025-07-17
Add parking bay interface

dorezyuk:


errors/parking_bay.yaml
```diff
@@ -0,0 +1,6 @@
+description: >-
```
corneliusclaussen: You could just point to generic.yaml for errors if there are no specific errors yet. This will also include VendorError/Warning which should be available at each HW driver interface to raise HW vendor specific errors.

PR #1269, opened 2025-07-24
Feat/ocpp21/q02

mlitre:


modules/EVSE/OCPP201/OCPP201.hpp
```diff
@@ -134,6 +134,10 @@ class OCPP201 : public Everest::ModuleBase {
     std::map<int32_t, bool> evse_ready_map;
     std::map<int32_t, std::optional<float>> evse_soc_map;
     std::map<int32_t, types::evse_board_support::HardwareCapabilities> evse_hardware_capabilities_map;
+    std::map<int32_t, std::vector<types::iso15118::EnergyTransferMode>> evse_supported_energy_transfer_modes;
+    std::map<int32_t, bool> evse_service_renegotiation_supported;
+    std::map<int32_t, std::string> evse_evcc_id;
```
Pietfried, line 139: Could we also update and keep track of the `evcc_id` as part of the `id_token` in the `TransactionData` of the `transaction_handler` ?  In that case we would not need this extra data structure

Pietfried (commented):
lgtm, please see one question in line

modules/EVSE/OCPP201/OCPP201.hpp
mlitre: That would not work as we always replace the id token used with what we receive from the auth interface

PR #1274, opened 2025-07-28
The charger can now pause the session if no energy is available

SebaLukas:
## Describe your changes
In IEC61851-23:2023 (CC.3.5.2 & CC.3.5.3) the charger has the opportunity to pause the session if no energy is available. In addition, two configuration options have been added that allow you to ignore the pause and enable the charger to ignore the fact that the EV is not entering a pause. However, neither option is standard-compliant

The EvseManager was a little too cautious. Because some cars do not accept 0W and 0A as the maximum limit.
With IEC61851-23:2023, it is now also possible to pause the session from the charger with ISO15118

### DIFF
--- a/modules/EvseManager/Charger.cpp
+++ b/modules/EvseManager/Charger.cpp
@@ -306,10 +306,11 @@ void Charger::run_state_machine() {
                 types::iso15118::DcEvseMaximumLimits evse_limit = shared_context.current_evse_max_limits;
                 if (not(evse_limit.evse_maximum_current_limit > 0 and evse_limit.evse_maximum_power_limit > 0)) {
                     if (not internal_context.no_energy_warning_printed) {
-                        EVLOG_warning << "No energy available, still retrying...";
+                        EVLOG_warning << "No energy available, still retrying... Some EVs dont like 0W and/or 0A in "
+                                         "ChargingParameterDiscoveryRes message";
                         internal_context.no_energy_warning_printed = true;
+                        signal_hlc_no_energy_available();
                     }
-                    break;
                 }
             }
 
@@ -516,6 +517,14 @@ void Charger::run_state_machine() {
             if (initialize_state) {
                 signal_simple_event(types::evse_manager::SessionEventEnum::PrepareCharging);
                 bcb_toggle_reset();
+
+                if (config_context.charge_mode == ChargeMode::DC) {
+                    // Create a copy of the atomic struct
+                    types::iso15118::DcEvseMaximumLimits evse_limit = shared_context.current_evse_max_limits;
+                    if (not(evse_limit.evse_maximum_current_limit > 0 and evse_limit.evse_maximum_power_limit > 0)) {
+                        signal_hlc_no_energy_available();
+                    }
+                }
             }
 
             if (config_context.charge_mode == ChargeMode::DC) {
@@ -896,6 +905,7 @@ void Charger::process_cp_events_state(CPEvent cp_event) {
         } else if (cp_event == CPEvent::CarRequestedStopPower) {
             shared_context.iec_allow_close_contactor = false;
             signal_dc_supply_off();
+            shared_context.current_state = EvseState::ChargingPausedEVSE;
         }
         break;
 
--- a/modules/EvseManager/manifest.yaml
+++ b/modules/EvseManager/manifest.yaml
@@ -338,6 +338,20 @@ config:
       - UUID_BASE64
       - SHORT_BASE64
     default: UUID
+  zero_power_ignore_pause:
+    description: >-
+      If no energy is available the charger can signal a pause to the car before cable check.
+      With this option the charger ignores that and continues as if energy were available.
+      This is not standard compliant!
+    type: boolean
+    default: false
+  zero_power_allow_ev_to_ignore_pause:
+    description: >-
+      If no energy is available the charger can signal a pause to the car before cable check.
+      With this option, the charger ignores that the ev goes to the energy phase instead of pausing
+      the session. This is not standard compliant!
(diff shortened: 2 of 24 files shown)


modules/EvseV2G/iso_server.cpp
```diff
@@ -1180,7 +1180,9 @@ static enum v2g_event handle_iso_charge_parameter_discovery(struct v2g_connectio
                 .RelativeTimeInterval.duration_isUsed = 1;
             conn->ctx->evse_v2g_data.evse_sa_schedule_list.SAScheduleTuple.array[0]
                 .PMaxSchedule.PMaxScheduleEntry.array[0]
-                .RelativeTimeInterval.duration = SA_SCHEDULE_DURATION;
+                .RelativeTimeInterval.duration =
+                conn->ctx->evse_v2g_data.no_energy_pause.status == NoEnergyPauseStatus::None ? SA_SCHEDULE_DURATION
+                                                                                             : 60 * 30;
```
Pietfried: Where do these numbers come from?

modules/EvseManager/manifest.yaml
```diff
@@ -338,6 +338,20 @@ config:
       - UUID_BASE64
       - SHORT_BASE64
     default: UUID
+  zero_power_ignore_pause:
+    description: >-
+      If no energy is a available the charger can signal a pause to the car before cable check.
```
Pietfried:
```suggestion
      If no energy is available the charger can signal a pause to the car before cable check.
```

modules/EvseManager/manifest.yaml
```diff
@@ -338,6 +338,20 @@ config:
...
+      With this option the charger ignores that and continues as if energy were available.
+      This is not standard compliant!
+    type: boolean
+    default: false
+  zero_power_allow_ev_to_ignore_pause:
+    description: >-
+      If no energy is a available the charger can signal a pause to the car before cable check.
```
Pietfried:
```suggestion
      If no energy is available the charger can signal a pause to the car before cable check.
```

modules/EvseV2G/v2g.hpp
```diff
@@ -163,6 +163,13 @@ struct SAE_Bidi_Data {
     bool discharging;
 };
 
+enum NoEnergyPauseStatus {
+    None,
+    AllowEvToIgnorePause,
+    AfterCablecCheckPreCharge,
+    BeforeCablecCheck,
```
Pietfried: usage of enums has to be fixed as well
```suggestion
    AfterCableCheckPreCharge,
    BeforeCableCheck,
```

modules/EvseManager/manifest.yaml
```diff
@@ -338,6 +338,20 @@ config:
...
+      With this option, the charger ignores that the ev goes to the energy phase instead of pause
```
Pietfried:
```suggestion
      With this option, the charger ignores that the ev goes to the energy phase instead of pausing
```

modules/EvseV2G/iso_server.cpp
SebaLukas: That's the pause duration I assumed. But I put 60 * 30 (1800 seconds) in a constexpr.

modules/EvseManager/manifest.yaml
SebaLukas: Thx for noticing

modules/EvseV2G/v2g.hpp
SebaLukas: Thx for noticing

modules/EvseManager/manifest.yaml
SebaLukas: Thx for noticing

modules/EvseManager/manifest.yaml
SebaLukas: Thx for noticing

modules/EvseManager/Charger.cpp
```diff
@@ -306,10 +306,11 @@ void Charger::run_state_machine() {
                 types::iso15118::DcEvseMaximumLimits evse_limit = shared_context.current_evse_max_limits;
                 if (not(evse_limit.evse_maximum_current_limit > 0 and evse_limit.evse_maximum_power_limit > 0)) {
                     if (not internal_context.no_energy_warning_printed) {
-                        EVLOG_warning << "No energy available, still retrying...";
+                        EVLOG_warning << "No energy available, still retrying... Some EVs dont like 0W and/or 0A in "
+                                         "ChargingParameterDiscoveryRes message";
                         internal_context.no_energy_warning_printed = true;
+                        signal_hlc_no_energy_available();
                     }
-                    break;
```
Pietfried, line 312: Can we safely remove the break here?

SebaLukas: This is the `break` this caused the whole too cautious and stops the session in the cable check.
In the first draft I had a config option to ignore this break. But following `IEC61851-23:2023` this break is not necessary anymore.

Pietfried (approved):

modules/EvseV2G/iso_server.cpp
```diff
@@ -1418,6 +1441,13 @@ static enum v2g_event handle_iso_power_delivery(struct v2g_connection* conn) {
                     pthread_mutex_unlock(&conn->ctx->mqtt_lock);
                 }
             }
+        } else if (conn->ctx->is_dc_charger and ev_should_pause) {
```
james-ctc: The Codacy working on line 1444 is correct `conn->ctx->is_dc_charger` is always true because of the check on line 1414 that it is false.

james-ctc (dismissed):

PR #1277, opened 2025-07-28
EnergyManager to use steady_clock for valid until timeouts.

james-ctc:
## Describe your changes

fix: The valid_until part of EnforcedLimits is a date and time string
This identifies when a limit is no longer valid. The problem occurs when there is a system clock update because the relationship between the current time and valid_until is no longer correct.

If the clock jumps forward then the most recent EnforcedLimits can be considered expired. If the clock jumps backwards the EnforcedLimits will be consider valid for too long.

This change changes valid_until to a number of seconds that the limit is valid.
The Evse Manager then uses the steady_clock (a monotonic clock that only ever increments) to calculate the valid_until time.

### DIFF
--- a/modules/EnergyManager/EnergyManagerImpl.cpp
+++ b/modules/EnergyManager/EnergyManagerImpl.cpp
@@ -10,8 +10,6 @@
 #include "BrokerFastCharging.hpp"
 #include "Market.hpp"
 
-using namespace std::literals::chrono_literals;
-
 namespace module {
 
 static BrokerFastCharging::Switch1ph3phMode to_switch_1ph3ph_mode(const std::string& m) {
@@ -185,8 +183,6 @@ std::vector<types::energy::EnforcedLimits> EnergyManagerImpl::run_optimizer(type
         const auto sold_energy = local_market.get_sold_energy();
 
         if (sold_energy.size() > 0) {
-            using namespace std::chrono;
-
             types::energy::EnforcedLimits l;
             l.uuid = local_market.energy_flow_request.uuid;
             l.valid_for = config.update_interval * 10;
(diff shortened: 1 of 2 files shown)


SebaLukas (approved):
Looking good :+1:

modules/EvseManager/Charger.cpp
```diff
@@ -1077,14 +1078,14 @@ float Charger::ampere_to_duty_cycle(float ampere) {
     return dc;
 }
 
-bool Charger::set_max_current(float c, std::chrono::time_point<date::utc_clock> validUntil) {
+bool Charger::set_max_current(float c, std::chrono::time_point<std::chrono::steady_clock> validUntil) {
     if (c >= 0.0 and c <= CHARGER_ABSOLUTE_MAX_CURRENT) {
 
         // is it still valid?
-        if (validUntil > date::utc_clock::now()) {
+        if (validUntil > steady_clock::now()) {
```
barsnick: Inconsistently using `std::chrono::steady_clock` above, and `steady_clock` here.

modules/EnergyManager/EnergyManagerImpl.cpp
```diff
@@ -184,11 +185,11 @@ std::vector<types::energy::EnforcedLimits> EnergyManagerImpl::run_optimizer(type
         const auto sold_energy = local_market.get_sold_energy();
 
         if (sold_energy.size() > 0) {
+            using namespace std::chrono;
```
barsnick: Using namespace introduced here, but no new code requiring it.

SebaLukas (approved):

barsnick (approved):
I only reviewed the remarks I gave previously. No functional check.

PR #1256, opened 2025-07-17
Refactor OVM interface and include MCS limits

corneliusclaussen:
## Describe your changes

Refactor over_voltage_monitoring interface with the following changes:

1. Add support for 6.3.1.106.3 (and not only 6.3.1.106.2) (lower speed over voltage error shutdown)
2. Use medium/high severity on errors to distiguish between error and emergency shutdown when an error is raised
3. Include MCS over voltage thresholds
4. Update voltage limits if they change during power transfer

All errors with medium/low severity are treated as error shutdowns from now on, and only errors with high severity  trigger an emergency shutdown.

## Checklist before requesting a review
- [x] I have performed a self-review of my code
- [ ] I have made corresponding changes to the documentation
- [x] I read the [contribution documentation](https://github.com/EVerest/EVerest/blob/main/CONTRIBUTING.md) and made sure that my changes meet its requirements

PR #1278, opened 2025-07-29
Trigger emergency shutdown when EV voltage limit < EV target voltage

d-u-d-e:
## Describe your changes
According to IEC61851_23 CC.6.2 (reported here for ease):

**Protection against overvoltage of battery**
_The d.c. supply shall trigger a d.c. supply initiated emergency shutdown according to CC.3.4
in order to prevent overvoltage at the battery, if output voltage exceeds maximum voltage limit
sent by the vehicle for 400 ms. (See 6.4.3.107)._

We should trigger an emergency shutdown in this case. Without this change, it was not possible for us to pass functional tests at TUV Rheinland. What they do is basically the following:
- start DC charging with maximum voltage limit set to 500V
- set target voltage to 350V
- update the new limit voltage to 250V
- verify that the emergency shutdown is triggered according to CC.3.4

Instead currently the current target voltage is adjusted to the new limit. 

### DIFF
--- a/modules/EvseManager/EvseManager.cpp
+++ b/modules/EvseManager/EvseManager.cpp
@@ -526,6 +526,20 @@ void EvseManager::ready() {
                 EVLOG_info << "Received EV maximum limits: " << l;
                 Everest::scoped_lock_timeout lock(ev_info_mutex,
                                                   Everest::MutexDescription::EVSE_subscribe_dc_ev_maximum_limits);
+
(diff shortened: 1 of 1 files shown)


d-u-d-e:
@SebaLukas I signed off my commit. 
I have no doubt PR #1256  will solve the same issue in a better way. Mine was just a straightforward solution to pass the ISO tests.

PR #1299, opened 2025-08-09
Feature/dc external derate interface

corneliusclaussen:

### DIFF
--- a/modules/EVSE/EvseManager/EvseManager.cpp
+++ b/modules/EVSE/EvseManager/EvseManager.cpp
@@ -2033,4 +2033,54 @@ void EvseManager::apply_new_target_voltage_current() {
     }
 }
 
+static std::optional<float> min_optional(std::optional<float> a, std::optional<float> b) {
+    // if both a and b have values, return the smaller one.
+    if (a.has_value() and b.has_value()) {
+        return (b.value() < a.value() ? b.value() : a.value());
+    }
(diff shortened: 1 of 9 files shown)


modules/EVSE/EvseManager/EvseManager.cpp
```diff
@@ -2033,4 +2033,54 @@ void EvseManager::apply_new_target_voltage_current() {
     }
 }
 
+static std::optional<float> min_optional(std::optional<float> a, std::optional<float> b) {
```
djchhp, line 2036: That could be covered by 
```
std::optional<float> min_optional(std::optional<float> const& a, std::optional<float> const& b) {
    return std::min(a, b, [](std::optional<float> const& aa, std::optional<float> const& bb) {
        return aa.value_or(std::numeric_limits<float>::max()) < bb.value_or(std::numeric_limits<float>::max());
    });
}

```

or even more generic in helpers or utils

```
template <class T>
std::optional<T> numeric_min_optional(std::optional<T> const& a, std::optional<T> const& b) {
    return std::min(a, b, [](std::optional<T> const& aa, std::optional<T> const& bb) {
        return aa.value_or(std::numeric_limits<T>::max()) < bb.value_or(std::numeric_limits<T>::max());
    });
}

auto const min_optional = numeric_min_optional<float>;
```

Pietfried (approved):

PR #1308, opened 2025-08-14
RsPaymentTerminal: Refactor the card disabled handling

dorezyuk:
## Describe your changes

Refactors the RsPaymentTerminal ignore card handling


modules/RsPaymentTerminal/src/main.rs
```diff
@@ -252,12 +228,40 @@ impl PaymentTerminalModule {
...
+                {
                     log::debug!("Reading is disabled, waiting...");
                     std::thread::sleep(Duration::from_secs(1));
                     continue;
                 }
 
+                // Attempting to get an invoice token
+                if token.is_none() && is_enabled_for_bank_card != CARD_DISABLED {
+                    if let Some(publisher) = publishers.bank_session_token_slots.get(0) {
+                        if timeout.elapsed() > Duration::from_secs(0) {
```
ydankner, line 246: Nice catch. Nit - these two ifs can be unified to a single if.

modules/RsPaymentTerminal/src/main.rs
```diff
@@ -361,17 +367,12 @@ impl PaymentTerminalModule {
...
-        map_guard.values().any(|v| !v.is_empty()) // If any connector has set any of the card types, it's enabled
-    }
-
-    fn connectors_for_card_type(&self, card_type: CardType) -> Vec<i64> {
+    /// Returns the connectors enabled for the given `CardType`.
+    ///
+    /// Following the token convention Some(Vec::new()) means disabled and
+    /// `None` means enabled for all. A non-empty list enabled the card-type
+    /// only for specific connectors.
+    fn is_enabled_for(&self, card_type: CardType) -> Option<Vec<i64>> {
```
ydankner: Nit - the naming is a bit strange, hinting that the returned value should be bool. Maybe something like `connectors_enabled_for`?

modules/RsPaymentTerminal/src/main.rs
```diff
@@ -202,6 +202,7 @@ impl PaymentTerminalModule {
     /// don't flag the token as pre-validated to allow the consumers to add
     /// custom validation steps on top.
     fn begin_transaction(&self, publishers: &ModulePublisher) -> Result<()> {
+        const CARD_DISABLED: Option<Vec<i64>> = Some(Vec::new());
```
ydankner: If you choose to rename `is_enabled_for` to `connectors_enabled_for`, this could also be changed to `NO_CONNECTORS` which is more natural

modules/RsPaymentTerminal/src/main.rs
```diff
@@ -284,38 +288,40 @@ impl PaymentTerminalModule {
...
-        }
+            CardInfo::MembershipCard(id_token) => {
+                let is_enabled_for_rfid_card = self.is_enabled_for(CardType::RfidCard);
+                if is_enabled_for_rfid_card == CARD_DISABLED {
+                    log::info!("Rfid cards disabled, ignoring");
+                }
+                ProvidedIdToken::new(id_token, AuthorizationType::RFID, is_enabled_for_rfid_card)
+            }
+        };
+        publishers.token_provider.provided_token(provided_token)?;
```
ydankner, line 328: So now we always send the provided token, but sometimes with an empty enabled connectors list? Does that make sense in terms of the Auth module's interface? Why send a token at all if it's not enabled for any connector?

modules/RsPaymentTerminal/src/main.rs
```diff
@@ -284,38 +288,40 @@ impl PaymentTerminalModule {
...
-            CardInfo::MembershipCard(id_token) => Some(ProvidedIdToken::new(
-                id_token,
-                AuthorizationType::RFID,
-                None,
-            )),
-        } {
-            publishers.token_provider.provided_token(provided_token)?;
-        }
+            CardInfo::MembershipCard(id_token) => {
+                let is_enabled_for_rfid_card = self.is_enabled_for(CardType::RfidCard);
```
ydankner: So did we not respect connectors with disabled RFID card until now? Nice catch

modules/RsPaymentTerminal/src/main.rs
dorezyuk: yes right now this is the expected behaviour

modules/RsPaymentTerminal/src/main.rs
dorezyuk: chaining if let and normal lets is yet not in our rust version

hikinggrass (approved):

Issue #1263, opened 2025-07-22
LEMDCBM400600: bind to network interface


ManuelJanocha:
### Describe the problem

Currently, the LEM driver does not offer a way to explicitly bind to a specific network interface. For systems with multiple interfaces, this can lead to unpredictable behavior or routing issues.

The LEM configuration should provide a dedicated option that allows users to specify which interface the driver should use for communication. 

[0001_lem_add_network_interface.patch](https://github.com/user-attachments/files/21358377/0001_lem_add_network_interface.patch)

### EVerest Domain

Hardware Drivers

### Affected EVerest Module

LEMDCBM400600

### Describe your solution

Please review the attached patch and add it to everest-core.

ManuelJanocha:
Danke 😀

PR #1310, opened 2025-08-20
Integrate poposed fix for #1263: dedicated option for LEM which interface the driver uses for communication

florinmihut:
#1263

### DIFF
--- a/modules/LemDCBM400600/main/http_client.hpp
+++ b/modules/LemDCBM400600/main/http_client.hpp
@@ -62,7 +62,7 @@ private:
     bool tls_enabled;
     std::string dcbm_tls_certificate;
     int command_timeout_ms = 5000; // default timeout in milliseconds
-    std::string network_interface; // New member variable for the network interface
+    std::string network_interface; // network interface
 
     [[nodiscard]] CURL* create_curl_handle_and_setup_url(const std::string& path) const;
     HttpResponse perform_request(CURL* connection, const std::string& request_body, const char* method_name,


modules/LemDCBM400600/LemDCBM400600.hpp
```diff
@@ -22,6 +22,7 @@ namespace module {
 struct Conf {
     std::string ip_address;
     int port;
+    std::string interface;
     std::string meter_tls_certificate;
```
florinmihut, line 26: False positive.

florinmihut:
There are 2 possible dangers with this PR.
1. The interface that it is set in the configuration is not existing at start time. I assume that could be mitigated by delaying starting everest until the interface is available - since anyway for a successful charging LEM is critical
2. The name of the interface changes due to enumeration timing for example - in this case the module will not start. However this could be as well mitigated over udev rules.
This PR allows to use LEM modules out of the box connected over different physical interfaces without the need for a manufacturer to do a customization/configuration at the end of the production line.

modules/LemDCBM400600/main/http_client.hpp
```diff
@@ -59,6 +62,7 @@ class HttpClient : public HttpClientInterface {
     bool tls_enabled;
     std::string dcbm_tls_certificate;
     int command_timeout_ms = 5000; // default timeout in milliseconds
+    std::string network_interface; // New member variable for the network interface
```
barsnick: The "New member variable for" part of the comment is pretty useless, and should be omitted.
```suggestion
    std::string network_interface; // network interface
```
or better even, since the variable name speaks for itself:
```suggestion
    std::string network_interface;
```

PR #1312, opened 2025-08-20
AuthHandler: Report used evse-id

dorezyuk:

### DIFF
--- a/lib/everest/everest_api_types/tests/expected_types_file_hashes.csv
+++ b/lib/everest/everest_api_types/tests/expected_types_file_hashes.csv
@@ -1,6 +1,6 @@
 # Add all *.yaml files that the everest_api_types library depends upon here.
 # Add the SHA256 hashes.
-types/authorization.yaml,41f947d0f7d1c4b7116ead52913868ab38d2671c41b6ff19faa6b3d9348429bc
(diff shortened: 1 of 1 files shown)


barsnick:
I also had some issues with the logging, mildly related:
https://github.com/EVerest/everest-core/pull/1314

types/authorization.yaml
```diff
@@ -34,6 +34,7 @@ types:
       - Rejected
       # Occurs when the token has been authorized but no connector is selected within connection_timeout seconds
       - TimedOut
+      - Used
```
dorezyuk: @hikinggrass @Pietfried should we be here even more explicit and say UsedToStart and UsedToStop?

Pietfried: Yes we should be explicit and use `UsedToStart` and `UsedToStop`.  We could drop `Accepted` in that case

dorezyuk: accepted still is there to report that it was accepted by the validation

Pietfried: I see, would be indeed better if we keep it and report e.g.

`Processing` > `Accepted` > `UsedToStartTransaction`,
`Processing` > `Accepted` > `TimeOut`

Pietfried (approved):

PR #1313, opened 2025-08-20
Move modules into subdirectories that describe their function

hikinggrass:
## Describe your changes
Add the following subdirs and move modules into these subdirs:

API
EnergyManagement
EV
EVSE
Examples
HardwareDrivers
Misc
Simulation
Testing


rckstrh (approved):

MODULE.bazel.lock
james-ctc, line 1: this file can be removed

james-ctc: .lock files could be added to .gitignore

hikinggrass: what's your take on this @dorezyuk ? I think you added it initially in the bazel8 migration

hikinggrass:
> > BoardSupportPackages
> 
> BTW, I always thought of these as "hardware adaptation layers", but perhaps that's just me. "BSP" isn't incorrect, I just sometimes confuse it with BSPs for controller platforms, which deliver the OS and the drivers and the SDK and the control APIs...

I wasn't quite sure how to call it since the interface these drivers implement use this term. But a better name is certainly appreciated. What do you think about adding an "EVSE" subdir into "HardwareDrivers" like this:
HardwareDrivers/EVSE
and that add all these "bsps" like the YetiDriver etc. into itt?

MODULE.bazel.lock
dorezyuk: yes it is prone to git conflicts - lets kill it

hikinggrass: :+1:

corneliusclaussen (dismissed):

Pietfried (approved):

dorezyuk:
looks great - go for it

PR #1324, opened 2025-08-27
RpcApi: add new JSON RPC API module

barsnick:
## Describe your changes
This introduces a new API module RpcApi, which uses JSON RPC, and WebSockets as a transport layer.

It uses both RPC Methods and Notifications.

This is the implementation according to the proposal with was [discussed here on Zulip](https://lfenergy.zulipchat.com/#narrow/channel/455067-EVerest.3A-Energy-Mgmt/topic/.E2.9C.94.20external.20API.20module) and [documented here](https://chargebyte.atlassian.net/wiki/external/NDkzYjg5MzAxY2U3NDg0MmE0N2MyOGQwZTcyOWYxNTA). The original proposal what slightly modified, as several details proved inconsistent, impractical, or just mis-specified.

Documentation of the current state of implementation is included in the module's `docs/` subdirectory. The current (or the final merged) state will be manifested as API version 1.0.0. Future updates to the external API definition will require version bumps.

A Python GUI test client is included under `modules/API/RpcApi/tools/python-client-gui/`.

Not implemented in the current state of the API:

- DC charging
- TLS
- API authentication

This will be added later, with corresponding bumps of the API version (where applicable).

A non-squashed version of this work is found in the branch [feature/json-rpc-api-nosquash](https://github.com/EVerest/everest-core/tree/feature/json-rpc-api-nosquash). for reference regarding small change steps, commit messages with motivations, and so on.

Co-authored-by: Fabian Hartung <fabian.hartung@chargebyte.com>
Co-authored-by: Moritz Barsnick <moritz.barsnick@chargebyte.com>


modules/API/RpcApi/data/GenericInfoStore.hpp
```diff
@@ -0,0 +1,87 @@
...
+            // create a copy of the data object
+            T data_copy = this->dataobj;
+            // unlock explicitly before entering callback
+            data_lock.unlock();
+            this->notification_callback(data_copy);
+        }
+    }
+
+    // TBD: Do we need to be able to return an error if we cannot set the data?
+    // virtual void set_data(const nlohmann::json& in) = 0;
```
hikinggrass:
```suggestion
```

modules/API/RpcApi/data/DataStore.hpp
```diff
@@ -0,0 +1,142 @@
...
+            }
+        }
+        EVLOG_error << "EVSE index " << evse_index << " not found in data store.";
+        return nullptr;
+    }
+};
+
+} // namespace data
+
+#endif // DATASTORE_HPP
```
hikinggrass:
```suggestion
#endif // DATASTORE_HPP

```

modules/API/RpcApi/data/GenericInfoStore.cpp
hikinggrass, line 1: is this needed?

modules/API/RpcApi/RpcApiRequestHandler.hpp
```diff
@@ -0,0 +1,49 @@
...
+    const std::vector<std::unique_ptr<evse_managerIntf>>& evse_managers;
+    const std::vector<std::unique_ptr<external_energy_limitsIntf>>& evse_energy_sink;
+
+    struct {
+        std::optional<float> evse_limit; ///< Maximum current or power limit for the EVSE
+        bool is_current_set = false;     ///< Flag to indicate if current or power limit is set
+    } configured_limits;
+};
+
+#endif // RPCAPIREQUESTHANDLER_HPP
```
hikinggrass:
```suggestion
#endif // RPCAPIREQUESTHANDLER_HPP

```

modules/API/RpcApi/data/SessionInfo.cpp
```diff
@@ -0,0 +1,202 @@
...
+    }
+    return 0.0f; // No discharged energy if export values are not set
+}
+
+std::chrono::seconds SessionInfoStore::get_charging_duration_s() const {
+    std::lock_guard<std::mutex> lock(this->session_info_mutex);
+
+    return std::chrono::duration_cast<std::chrono::seconds>(this->end_time_point - this->start_time_point);
+}
+} // namespace data
```
hikinggrass:
```suggestion
} // namespace data

```

modules/API/RpcApi/helpers/LimitDecimalPlaces.cpp
```diff
@@ -0,0 +1,37 @@
...
+    if (j.is_object() || j.is_array()) {
+        for (auto& el : j) {
+            roundFloatsInJson(el, precision);
+        }
+    } else if (j.is_number_float()) {
+        double value = j.get<double>();
+        j = formatAndRoundDouble(value, step, precision);
+    }
+}
+} // namespace helpers
```
hikinggrass:
```suggestion
} // namespace helpers

```

modules/API/RpcApi/helpers/LimitDecimalPlaces.hpp
```diff
@@ -0,0 +1,14 @@
...
+#define LIMIT_DECIMAL_PLACES_HPP
+
+#include <nlohmann/json.hpp>
+
+namespace helpers {
+double formatAndRoundDouble(double value, int precision = 3);
+void roundFloatsInJson(nlohmann::json& j, int precision = 3);
+} // namespace helpers
+
+#endif // LIMIT_DECIMAL_PLACES_HPP
```
hikinggrass:
```suggestion
#endif // LIMIT_DECIMAL_PLACES_HPP

```

modules/API/RpcApi/rpc/RpcHandler.cpp
```diff
@@ -0,0 +1,355 @@
...
+        }
+        // Store message in a map with client_id as key
+        std::lock_guard<std::mutex> lock(m_mtx);
+        messages[client_id].data.push_back(request);
+        messages[client_id].transport_interface = transport_interface;
+        EVLOG_debug << "Received message from client " << client_id << ": " << request.dump();
+
+        auto elapsed = duration_cast<milliseconds>(now - m_last_req_notification);
+        if (elapsed >= REQ_COLLECTION_TIMEOUT) {
+            m_last_req_notification = now; // Timer neu starten
```
hikinggrass:
```suggestion
            m_last_req_notification = now; // restart timer
```

modules/API/RpcApi/tests/helpers/JsonRpcUtils.hpp
```diff
@@ -0,0 +1,56 @@
...
+            result_obj[json_key_value.begin().key()] == json_key_value.begin().value()) {
+            return true;
+        }
+    }
+    return false;
+}
+
+} // namespace json_rpc_utils
+
+#endif // JSON_RPC_UTILS_HPP
```
hikinggrass:
```suggestion
#endif // JSON_RPC_UTILS_HPP

```

modules/API/RpcApi/tools/python-client-gui/README.md
```diff
@@ -0,0 +1,72 @@
...
+- `settings.json`: Stores the last used IP and port.
+- `saved_calls.json`: Stores saved method calls with their parameters.
+
+---
+
+## 📘 Compatibility Matrix
+
+| GUI Version | JSON-RPC Server API Version | Notes                    |
+|-------------|-----------------------------|--------------------------|
+| `0.0.1`     | `0.0.1`                     | Initial stable version   |
```
hikinggrass: JSON-RPC Server API should probably be `1.0.0` (as that's the API version set in the CMakeLists.txt

The GUI doesn't seem to have or report a version at the moment, can you add this?

modules/API/RpcApi/tools/python-client-gui/requirements.txt
```diff
@@ -0,0 +1 @@
+websockets
```
hikinggrass: probably a good idea to pin this to a known-working version

modules/API/RpcApi/tools/python-client-gui/saved_calls.json
hikinggrass, line 1: shoud this be checked in?

modules/API/RpcApi/tools/python-client-gui/settings.json
hikinggrass, line 1: shoud this be checked in?

modules/API/RpcApi/data/GenericInfoStore.cpp
FaHaGit: No, its not needed. Removed it.

modules/API/RpcApi/tools/python-client-gui/README.md
FaHaGit: Now the version is in the title of the application. Fixed the number to 1.0.0.

modules/API/RpcApi/tools/python-client-gui/saved_calls.json
FaHaGit: Yes, it should. All method calls with parameters are already pre-initialized there. This makes it possible to use the API directly.

modules/API/RpcApi/tools/python-client-gui/settings.json
FaHaGit: Not strictly necessary, but for the sake of completeness, I would leave it in.

barsnick (commented):
Discovery of json-rpc-cxx in non-EDM builds (e.g. Yocto) is broken.

This is pending a fix.

barsnick:
> This is pending a fix.

Fixed.

modules/API/RpcApi/RpcApiRequestHandler.hpp
```diff
@@ -0,0 +1,48 @@
+#ifndef RPCAPIREQUESTHANDLER_HPP
```
hikinggrass: missing license header
```suggestion
// SPDX-License-Identifier: Apache-2.0
// Copyright Pionix GmbH and Contributors to EVerest
#ifndef RPCAPIREQUESTHANDLER_HPP
```

modules/API/RpcApi/data/SessionInfo.cpp
```diff
@@ -0,0 +1,202 @@
+#include "SessionInfo.hpp"
```
hikinggrass: missing license header
```suggestion
// SPDX-License-Identifier: Apache-2.0
// Copyright Pionix GmbH and Contributors to EVerest
#include "SessionInfo.hpp"
```

modules/API/RpcApi/manifest.yaml
```diff
@@ -0,0 +1,57 @@
...
+  websocket_port:
+    description: Port for the websocket server to listen on.
+    type: integer
+    default: 80
+  websocket_interface:
+    description: >-
+      Network interface/device for the websocket server to listen on. E.g. 'lo' or 'eth1'.
+      Listen on all interfaces if empty.
+    type: string
+    default: ""
```
hikinggrass: I'm not sure if listening on all interfaces should be the default, since this could easily lead to the service being available on unintended interface. What do you think about a more defensive approach here, eg.
"" (empty) = listen on no interface
"auto" or "all" = listen on all interfaces?

modules/API/RpcApi/server/WebsocketServer.hpp
```diff
@@ -0,0 +1,52 @@
+#ifndef WEBSOCKETSERVER_H
```
hikinggrass: missing license header
```suggestion
// SPDX-License-Identifier: Apache-2.0
// Copyright Pionix GmbH and Contributors to EVerest
#ifndef WEBSOCKETSERVER_H
```

modules/API/RpcApi/tests/helpers/JsonRpcUtils.hpp
```diff
@@ -0,0 +1,56 @@
+#ifndef JSON_RPC_UTILS_HPP
```
hikinggrass: missing license header
```suggestion
// SPDX-License-Identifier: Apache-2.0
// Copyright Pionix GmbH and Contributors to EVerest
#ifndef JSON_RPC_UTILS_HPP
```

modules/API/RpcApi/tests/helpers/WebSocketTestClient.cpp
```diff
@@ -0,0 +1,175 @@
+#include "WebSocketTestClient.hpp"
```
hikinggrass: missing license header
```suggestion
// SPDX-License-Identifier: Apache-2.0
// Copyright Pionix GmbH and Contributors to EVerest
#include "WebSocketTestClient.hpp"
```

modules/API/RpcApi/tests/helpers/WebSocketTestClient.hpp
```diff
@@ -0,0 +1,92 @@
+#ifndef WEBSOCKETTESTCLIENT_HPP
```
hikinggrass: missing license header
```suggestion
// SPDX-License-Identifier: Apache-2.0
// Copyright Pionix GmbH and Contributors to EVerest
#ifndef WEBSOCKETTESTCLIENT_HPP
```

modules/API/RpcApi/tests/rpc/RpcHandlerTests.cpp
```diff
@@ -0,0 +1,864 @@
+#include "../data/DataStore.hpp"
```
hikinggrass: missing license header
```suggestion
// SPDX-License-Identifier: Apache-2.0
// Copyright Pionix GmbH and Contributors to EVerest
#include "../data/DataStore.hpp"
```

modules/API/RpcApi/tests/server/WebsocketServerTests.cpp
```diff
@@ -0,0 +1,160 @@
+#include <everest/logging.hpp>
```
hikinggrass: missing license header
```suggestion
// SPDX-License-Identifier: Apache-2.0
// Copyright Pionix GmbH and Contributors to EVerest
#include <everest/logging.hpp>
```

modules/API/RpcApi/tools/python-client-gui/everest-json-rpc-websocket-client.py
```diff
@@ -0,0 +1,296 @@
+# SPDX-License-Identifier: Apache-2.0
+# This code was generated and enhanced with the help of an AI assistant (ChatGPT by OpenAI).
```
hikinggrass: not sure about this kind of statement in a license header

hikinggrass (changes requested):

modules/API/RpcApi/manifest.yaml
barsnick: Having this module configured, but asking it to listen to _no_ interface, only makes sense once we have other transports - which might not even need an interface.

We discussed this WebSocket configuration issue and think it makes sense to:
- default to interface `lo` - this is sufficient for some applications and provides a quite safe default
- throw an error if nothing/empty is configured and `websocket_enabled` is true
- introduce alias `all` to actually listen on all interfaces

modules/API/RpcApi/tools/python-client-gui/everest-json-rpc-websocket-client.py
barsnick: Dropped.

modules/API/RpcApi/manifest.yaml
barsnick: Done.

BTW, libwebsockets also allows binding to an address here. We won't document this (yet), because while it does work, it goes into an endless loop (with log flood) if it cannot bind to that particular address, in versions pre 4.4.0. (Fixed [here](https://github.com/warmcat/libwebsockets/commit/93ae4cb628578b917092bc9e546d7472ff8dcd2c), even for IPv4 - tested by backporting to 4.3.x.)

types/json_rpc_api.yaml
Pietfried, line 1: Imo these types should not be defined as part of the EVerest type system, because they are specific to the module, not used as part of the EVerest internal interfaces and not supposed to be used within any other EVerest module. This couples the internal YAML definitions / type system to this external API. At the same time, it would introduce quite some duplicates of existing YAML types.

The best way would probably to re(use) the EVerest API types for the translation layer: https://github.com/EVerest/everest-core/tree/main/lib/everest/everest_api_types

modules/API/RpcApi/manifest.yaml
```diff
@@ -0,0 +1,58 @@
...
+    min_connections: 1
+    max_connections: 128
+  evse_energy_sink:
+    interface: external_energy_limits
+    min_connections: 0
+    max_connections: 128
+  energy_listener:
+    interface: energy
+    min_connections: 1
+    max_connections: 128
```
Pietfried: If I understand it correctly this requirement is only needed to identify the `evse_active_phase_count`. Maybe this information could be added to the `evse_manager` interface to avoid this extra requirement.

barsnick: > If I understand it correctly this requirement is only needed to identify the `evse_active_phase_count`. Maybe this information could be added to the `evse_manager` interface to avoid this extra requirement.

True. The value was even already there, but was not changing accordingly. I tracked that down to a bug, which I fixed in https://github.com/EVerest/everest-core/pull/1533.

I will rework RpcApi to use evse_manager only.

modules/API/RpcApi/tools/python-client-gui/README.md
```diff
@@ -0,0 +1,72 @@
...
+```bash
+pip install -r requirements.txt
+```
+
+---
+
+## ▶️ How to Run
+
+```bash
+python3 everest-json-rpc-api.py
```
mhei: This should probably be:
```suggestion
python3 everest-json-rpc-websocket-client.py
```

config/config-sil-rpcapi.yaml
```diff
@@ -0,0 +1,199 @@
...
+    connections:
+      energy_consumer:
+        - implementation_id: energy_grid
+          module_id: api_sink_1_evsemgr
+    module: EnergyNode
+  iso15118_car:
+    config_module:
+      device: auto
+      supported_ISO15118_2: true
+    connections: {}
```
mhei: Not needed, I'd drop this line.

config/config-sil-rpcapi.yaml
```diff
@@ -0,0 +1,199 @@
...
+    module: PyEvJosev
```
mhei: Nit: I'd sort this as first after the instance name.

config/config-sil-rpcapi.yaml
barsnick: Yeah, this was just slightly adapted from `config-sil-rpcapi.yaml`, with all its shortcomings. I will clean up nevertheless.

PR #1327, opened 2025-08-27
GenericPowermeter: use exponent value instead of throwing it away

hikinggrass:


hikinggrass:
@barsnick I stumbled upon this while looking at compiler warnings, it did not look like this was intended in the original code

Pietfried (approved):

barsnick:
> @barsnick I stumbled upon this while looking at compiler warnings, it did not look like this was intended in the original code

Good catch. I must switch on warnings more often, and also actually go through the CI logs.

PR #1342, opened 2025-09-02
feat(EvseV2G): Adding iso15118_vas requirement to EvseV2G 

SebaLukas:
## Describe your changes
- Custom VAS can be connected to the EvseV2G module and offered.
- Internet service config option to PyEvJosev

VAS such as Internet service could not yet be offered.

### DIFF
--- a/modules/EVSE/EvseV2G/iso_server.cpp
+++ b/modules/EVSE/EvseV2G/iso_server.cpp
@@ -699,15 +699,83 @@ static enum v2g_event handle_iso_service_detail(struct v2g_connection* conn) {
         if (req->ServiceID == conn->ctx->evse_v2g_data.evse_service_list[idx].ServiceID) {
             service_id_found = true;
 
+            init_iso2_ServiceParameterListType(&res->ServiceParameterList);
             /* Fill parameter list of the requested service id [V2G2-549] */
-            for (uint8_t idx2 = 0; idx2 < conn->ctx->evse_v2g_data.service_parameter_list[idx].ParameterSet.arrayLen;
-                 idx2++) {
-                res->ServiceParameterList.ParameterSet.array[idx2] =
-                    conn->ctx->evse_v2g_data.service_parameter_list[idx].ParameterSet.array[idx2];
+            if (conn->ctx->evse_v2g_data.service_parameter_list.find(req->ServiceID) !=
+                conn->ctx->evse_v2g_data.service_parameter_list.end()) {
+                res->ServiceParameterList.ParameterSet =
+                    conn->ctx->evse_v2g_data.service_parameter_list.at(req->ServiceID).ParameterSet;
+                res->ServiceParameterList_isUsed = true;
+                break;
             }
-            res->ServiceParameterList.ParameterSet.arrayLen =
-                conn->ctx->evse_v2g_data.service_parameter_list[idx].ParameterSet.arrayLen;
-            res->ServiceParameterList_isUsed = (res->ServiceParameterList.ParameterSet.arrayLen != 0) ? 1 : 0;
+
+            for (size_t i = 0; i < conn->ctx->supported_vas_services_per_provider.size(); i++) {
+                const auto& provider_services = conn->ctx->supported_vas_services_per_provider.at(i);
+                if (std::find(provider_services.begin(), provider_services.end(), req->ServiceID) !=
+                    provider_services.end()) {
+                    const auto vas_parameters = conn->ctx->r_vas.at(i)->call_get_service_parameters(req->ServiceID);
+
+                    if (vas_parameters.empty()) {
+                        break;
+                    }
+
+                    iso2_ServiceParameterListType vas_parameter_list{};
+                    init_iso2_ServiceParameterListType(&vas_parameter_list);
+                    vas_parameter_list.ParameterSet.arrayLen = vas_parameters.size() <= 5 ? vas_parameters.size() : 5;
+
+                    for (size_t j = 0; j < vas_parameters.size() and j <= 5; j++) {
+                        const auto& vas_parameter = vas_parameters.at(j);
+
+                        vas_parameter_list.ParameterSet.array[j].ParameterSetID =
+                            static_cast<int16_t>(vas_parameter.set_id);
+
+                        size_t t{0};
+                        for (const auto& parameter : vas_parameter.parameters) {
+                            auto& out_parameter = vas_parameter_list.ParameterSet.array[j].Parameter.array[t++];
+
+                            init_iso2_ParameterType(&out_parameter);
+
+                            parameter.name.copy(out_parameter.Name.characters, parameter.name.length());
+                            out_parameter.Name.charactersLen = parameter.name.length();
+
+                            if (parameter.value.int_value.has_value()) {
+                                out_parameter.intValue = parameter.value.int_value.value();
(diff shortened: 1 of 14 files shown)


modules/EVSE/EvseV2G/connection/connection.cpp
```diff
@@ -401,6 +401,8 @@ static void* connection_handle_tcp(void* data) {
 
     dlog(DLOG_LEVEL_INFO, "Started new TCP connection thread");
 
+    remove_service_from_service_list_if_exists(conn->ctx, V2G_SERVICE_ID_CERTIFICATE);
```
mlitre: Is there a reason why we remove this here? I know certificate handling is not allowed with a TCP connection, but why is this only being added here now? Is this not handled somewhere else already?

modules/EVSE/EvseV2G/iso_server.cpp
```diff
@@ -699,15 +699,80 @@ static enum v2g_event handle_iso_service_detail(struct v2g_connection* conn) {
...
+                    conn->ctx->evse_v2g_data.service_parameter_list.at(req->ServiceID).ParameterSet;
+                res->ServiceParameterList_isUsed = true;
+                break;
+            }
+
+            for (size_t i = 0; i < conn->ctx->supported_vas_services_per_provider.size(); i++) {
+                const auto& provider_services = conn->ctx->supported_vas_services_per_provider.at(i);
+                if (std::find(provider_services.begin(), provider_services.end(), req->ServiceID) !=
+                    provider_services.end()) {
+                    const auto vas_parameters = conn->ctx->r_vas.at(i)->call_get_service_parameters(req->ServiceID);
```
mlitre, line 716: Would it not be interesting to get these parameters at startup? This avoids a blocking call during EV communication?

modules/EVSE/EvseV2G/iso_server.cpp
```diff
@@ -699,15 +699,80 @@ static enum v2g_event handle_iso_service_detail(struct v2g_connection* conn) {
...
+                            init_iso2_ParameterType(&out_parameter);
+
+                            parameter.name.copy(out_parameter.Name.characters, parameter.name.length());
+                            out_parameter.Name.charactersLen = parameter.name.length();
+
+                            if (parameter.value.int_value.has_value()) {
+                                out_parameter.intValue = parameter.value.int_value.value();
+                                out_parameter.intValue_isUsed = true;
+                            } else if (parameter.value.finite_string.has_value()) {
+                                // TODO(SL): Check if string is too big to copy char array
```
mlitre, line 745: Maybe for now we can just add a warning log?

modules/EVSE/EvseV2G/iso_server.cpp
SebaLukas: This is the point to have the parameters on runtime/during the hlc session. I know this is a blocking call, I have the same concern, but for the most cases this should be fine.

mlitre: Got it, we can keep this concern for later if it becomes a real issue.

modules/EVSE/EvseV2G/connection/connection.cpp
SebaLukas: See #1322 I fixed this here already and also added the fix in #1322

modules/EVSE/EvseV2G/tests/v2g_ctx_test.cpp
```diff
@@ -59,7 +64,7 @@ class V2gCtxTest : public testing::Test {
     }
 
     void SetUp() override {
-        auto ptr = v2g_ctx_create(&charger, &extensions, &security);
+        auto ptr = v2g_ctx_create(&charger, &extensions, &security, {&vas_list.at(0)});
```
SebaLukas: @james-ctc Do you think this is okay? I am not sure. Feels hacky :sweat_smile:

modules/EVSE/EvseV2G/iso_server.cpp
SebaLukas: Done :+1:

mlitre (approved):

modules/EVSE/EvseV2G/tests/v2g_ctx_test.cpp
james-ctc: it depends on whether you need tests where `r_vas` needs to contain more than one element.
as it stands I would probably go for:
```
module::stub::iso15118_vasIntfStub vas_item;
...
 V2gCtxTest() : charger(adapter), security(adapter), vas_item(adapter) {
...
auto ptr = v2g_ctx_create(&charger, &extensions, &security, {&vas_item});
```

modules/EVSE/EvseV2G/tests/v2g_ctx_test.cpp
```diff
@@ -161,7 +166,10 @@ TEST(valgrind, memcheck) {
     module::stub::ISO15118_chargerImplStub charger(adapter);
     module::stub::evse_securityIntfStub security(adapter);
     module::stub::iso15118_extensionsImplStub extensions;
-    auto ptr = v2g_ctx_create(&charger, &extensions, &security);
+
+    std::vector<module::stub::iso15118_vasIntfStub> vas_list = {module::stub::iso15118_vasIntfStub(adapter)};
+
+    auto ptr = v2g_ctx_create(&charger, &extensions, &security, {&vas_list.at(0)});
```
james-ctc: could be simplified to:
```
module::stub::iso15118_vasIntfStub vas_item(adapter);
auto ptr = v2g_ctx_create(&charger, &extensions, &security, {&vas_item});
```

modules/EVSE/EvseV2G/tests/v2g_main.cpp
```diff
@@ -128,8 +129,9 @@ int main(int argc, char** argv) {
     module::stub::ISO15118_chargerImplStub charger(adapter);
     EvseSecurity security(adapter);
     module::stub::iso15118_extensionsImplStub extensions;
+    std::vector<module::stub::iso15118_vasIntfStub> vas_list = {module::stub::iso15118_vasIntfStub(adapter)};
 
-    auto* ctx = v2g_ctx_create(&charger, &extensions, &security);
+    auto* ctx = v2g_ctx_create(&charger, &extensions, &security, {&vas_list.at(0)});
```
james-ctc: could be simplified to:
```
module::stub::iso15118_vasIntfStub vas_item(adapter);
auto* ctx = v2g_ctx_create(&charger, &extensions, &security, {&vas_item});
```

modules/EVSE/EvseV2G/tests/v2g_ctx_test.cpp
SebaLukas: I like that, thx :+1:

MatthiasContribute:
Hi, thank you for that PR. I am testing the Vas with PyJosev setup using  the config-sil-dc.yaml and add the vas service configuration.

Is that possible to trigger the VAS service using the pyJosev? In the moment, it does look like it is not happening. 
[config-sil-dc-vas.yaml](https://github.com/user-attachments/files/23545600/config-sil-dc-vas.yaml)

MatthiasContribute:
Another question, is it possible to dump the pyjosev - charger communication? Thank you.

PR #1353, opened 2025-09-10
FIX: Changing DC capabilities for 'current' not applied

djchhp:
## Describe your changes
Updated capabilities for a DC power supply were not always taken into account. Especially changing only current capabilities during a charging session, did not transition to into the capabilities reported to the car.

- [ - ] I have made corresponding changes to the documentation
- [ x ] I read the [contribution documentation](https://github.com/EVerest/EVerest/blob/main/CONTRIBUTING.md) and made sure that my changes meet its requirements

### DIFF
--- a/modules/EVSE/EvseManager/energy_grid/energyImpl.cpp
+++ b/modules/EVSE/EvseManager/energy_grid/energyImpl.cpp
@@ -112,6 +112,7 @@ void energyImpl::clear_request_schedules() {
 
 void energyImpl::ready() {
     hw_caps = mod->get_hw_capabilities();
+    last_powersupply_capabilities = mod->get_powersupply_capabilities();
     clear_request_schedules();
 
     // request energy now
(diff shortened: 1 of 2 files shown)


modules/EVSE/EvseManager/energy_grid/energyImpl.cpp
```diff
@@ -498,6 +531,7 @@ void energyImpl::handle_enforce_limits(types::energy::EnforcedLimits& value) {
                     last_enforced_limits_watt = watt_leave_side;
                     last_target_voltage = target_voltage;
                     last_actual_voltage = actual_voltage;
+                    last_powersupply_capabilities = powersupply_capabilities;
 
                     auto powersupply_capabilities = mod->get_powersupply_capabilities();
```
Pietfried: looks like you dont have to call `get_powersupply_capabilities()` again since its already set [here](https://github.com/EVerest/everest-core/pull/1353/files#diff-99c56b5f65ac448c10f47dcb178dd0db7d85b6f5cd36111f6784883405286558R518) 
```suggestion
```

Pietfried (dismissed):

modules/EVSE/EvseManager/energy_grid/energyImpl.cpp
djchhp: Your are right, I missed to delete that line.

SebaLukas (approved):
Looking good :+1: 
Perhaps you can also move the local static functions into an empty namespace:
```
namespace {
bool almost_eq(float a, float b) { [....] }
} // namespace
```

PR #1355, opened 2025-09-10
Feature Internet VAS module

mlitre:
## Describe your changes

Add a new module capable of letting an ISO15118-2 module that it supports the internet service VAS.
This module requires, iptables, ip6tables, ip, udhcpd and radvd. All of these things are necessary to be able to route packets from the EV to the internet correctly.


modules/EVSE/Iso15118InternetVas/CMakeLists.txt
```diff
@@ -0,0 +1,45 @@
...
+# ev@bcc62523-e22b-41d7-ba2f-825b493a3c97:v1
+
+target_sources(${MODULE_NAME}
+    PRIVATE
+        "iso15118_vas/ISO15118_vasImpl.cpp"
+)
+
+# ev@c55432ab-152c-45a9-9d2e-7281d50c69c3:v1
+# insert other things like install cmds etc here
+target_compile_features(${MODULE_NAME} PRIVATE cxx_std_17)
```
SebaLukas: Isn't that generally active for all EVerest modules?

modules/EVSE/Iso15118InternetVas/doc.rst
SebaLukas, line 1: Do both `.rst` have the same content? It seems so. If that is the case, then one .rst file can be removed.

modules/EVSE/Iso15118InternetVas/manifest.yaml
```diff
@@ -0,0 +1,26 @@
...
+config:
+  ev_interface:
+    description: >-
+      Name of the interface which is connected to the EV.
+    type: string
+  modem_interface:
+    description: >-
+      Name of the interface which is connected to the internet.
+    type: string
+
```
SebaLukas: Remove new line

modules/EVSE/Iso15118InternetVas/manifest.yaml
```diff
@@ -0,0 +1,26 @@
...
+provides:
+  iso15118_vas:
+    interface: ISO15118_vas
+    description: The internet VAS
+requires:
+  evse_manager:
+    interface: evse_manager
+    min_connections: 0
+    max_connections: 1
+
```
SebaLukas: See above

modules/EVSE/Iso15118InternetVas/iso15118_vas/ISO15118_vasImpl.cpp
```diff
@@ -0,0 +1,188 @@
...
+        this->stop_internet_service();
+    }
+    if (internet_service_thread.joinable()) {
+        internet_service_thread.join();
+    }
+}
+
+void ISO15118_vasImpl::init() {
+    this->scripts_path = mod->info.paths.libexec;
+    this->internet_service_running = false;
```
SebaLukas: You can remove this line if you initialise `internet_service_running` in .hpp with `internet_service_running{false};`

modules/EVSE/Iso15118InternetVas/iso15118_vas/ISO15118_vasImpl.cpp
```diff
@@ -0,0 +1,188 @@
...
+                }
+            });
+    }
+}
+
+void ISO15118_vasImpl::ready() {
+    types::iso15118_vas::OfferedServices services;
+    services.service_ids.reserve(1);
+    services.service_ids.emplace_back(InternetAccessServiceIdD2);
+    this->publish_offered_vas(services);
```
SebaLukas: For only one offered_service I would change the code to:
```cpp
this->publish_offered_vas({InternetAccessServiceIdD2});
```

modules/EVSE/Iso15118InternetVas/iso15118_vas/ISO15118_vasImpl.cpp
```diff
@@ -0,0 +1,188 @@
...
+                    selected_ports.push_back(HTTPS_PORT);
+                }
+            }
+        }
+    }
+    return selected_ports;
+}
+
+void ISO15118_vasImpl::handle_selected_services(std::vector<types::iso15118_vas::SelectedService>& selected_services) {
+    std::vector<int> ports_to_open = this->get_selected_internet_ports(selected_services);
```
SebaLukas: Why not `const auto ports_to_open`?

modules/EVSE/Iso15118InternetVas/iso15118_vas/ISO15118_vasImpl.cpp
```diff
@@ -0,0 +1,188 @@
...
+#include <sys/types.h>
+
+namespace module {
+namespace iso15118_vas {
+
+constexpr int InternetAccessServiceIdD2 = 3;
+// constexpr int InternetAccessServiceIdD20 = 65;
+const std::string INTERNET_SETUP_SCRIPT = "vas-internet-setup.sh";
+
+constexpr int HTTP_PARAM_SET_ID = 3;
```
SebaLukas, line 17: Right now both param_set_ids or http/s ports are supported. In the case that only one port should be supported the code must be changed. Maybe trough a config option?

modules/EVSE/Iso15118InternetVas/CMakeLists.txt
mlitre: Yes I guess that's true, I think I had copied this from the system module

modules/EVSE/Iso15118InternetVas/doc.rst
mlitre: Done

modules/EVSE/Iso15118InternetVas/manifest.yaml
mlitre: Done

modules/EVSE/Iso15118InternetVas/manifest.yaml
mlitre: Done

modules/EVSE/Iso15118InternetVas/iso15118_vas/ISO15118_vasImpl.cpp
mlitre: Done

modules/EVSE/Iso15118InternetVas/iso15118_vas/ISO15118_vasImpl.cpp
mlitre: Done

modules/EVSE/Iso15118InternetVas/iso15118_vas/ISO15118_vasImpl.cpp
mlitre: Done

modules/EVSE/Iso15118InternetVas/iso15118_vas/ISO15118_vasImpl.cpp
mlitre: That's a good idea, I will do that.

modules/EVSE/Iso15118InternetVas/iso15118_vas/ISO15118_vasImpl.cpp
james-ctc, line 1: is 15118-2 requirement [V2G2-636] "If a V2G Communication Session without TLS is used, the SECC shall not provide the Message Set VAS" handled? I can't see any check here.
Is this done in the EvseV2G module?

modules/EVSE/Iso15118InternetVas/docs/index.rst
```diff
@@ -0,0 +1,77 @@
...
+
+This module implements the ISO 15118-2 Value Added Service (VAS) for providing internet access to a connected electric vehicle (EV). When an EV requests this service, the module configures the charger's networking to share its internet connection with the EV over the power line communication (PLC) interface.
+
+How it works
+============
+
+1.  **Service Announcement**: The module advertises the availability of the Internet Access service (Service ID 3) to the EV, offering HTTP and HTTPS access.
+2.  **Service Selection**: If the EV selects this service, the module initiates the network setup.
+3.  **Network Setup**: The module executes a single helper script, ``vas-internet-setup.sh``, to configure all necessary networking components:
+    - It starts a DHCPv4 server (``udhcpd``) on the EV-facing network interface (``ev_interface``) to provide an IPv4 address to the EV.
```
james-ctc: This might be a departure from 15118. I can't see any mention of supporting IPv4. There is mention of DHCPv6 as an optional IP address allocation mechanism rather than SLAAC.

I think the intent is that the EVSE is IPv6 only and the SECC (or more likely router behind the SECC) does IPv6/IPv4 translation.
This could lead to interoperability issues where an EV is IPv6 only and expecting IPv6 to IPv4 conversion to happen in the network.

> To access IPv4 from an IPv6-only network, you need a NAT64/DNS64 gateway to act as a translator. DNS64 handles name resolution by embedding IPv4 addresses into synthetic IPv6 addresses, which are then routed to a NAT64 device. The NAT64 device performs network address translation, converting the IPv6 packets into IPv4 packets for the destination IPv4 server and then converting the IPv4 response back into IPv6 for the client.

james-ctc: Perhaps the setup of IPv4 should be optional?

james-ctc: Following 7.6.3 IP Addressing
The EVCC will have a link local IP address and a global IP address: "if router is present in local link" which I don't believe is the case.

I think the intention is that the SECC provides network address translation so that the rest of the internet never sees the EVCC IP address. This would also then support the IPv6 only EVCC communicating with the internet via an IPv4 internet connection.

modules/EVSE/Iso15118InternetVas/docs/index.rst
```diff
@@ -0,0 +1,77 @@
...
+Requirements
+============
+
+Software
+--------
+
+The module relies on the ``vas-internet-setup.sh`` script, which requires the following external tools to be available in the system's PATH:
+
+- ``radvd``: For IPv6 Router Advertisements.
+- ``udhcpd``: For the DHCPv4 server (typically provided by BusyBox).
```
james-ctc: is there a DNS server configured?

modules/EVSE/Iso15118InternetVas/iso15118_vas/ISO15118_vasImpl.cpp
```diff
@@ -0,0 +1,187 @@
...
+                ports_str += ",";
+            }
+        }
+        start_internet_service(ports_str);
+    }
+}
+
+void ISO15118_vasImpl::start_script(const std::string& script_name, const std::vector<std::string>& args) {
+    const auto script_path = this->scripts_path / script_name;
+    everest::run_application::run_application(script_path, args, [](const std::string& message) {
```
james-ctc: another option is to get the result from run_application and perhaps only log when there is a problem.
CmdOutput includes the exit code as well as the command output as a single string or a vector of lines.

I would only use a lambda where some output would lead to a Terminate result ...

modules/EVSE/Iso15118InternetVas/iso15118_vas/ISO15118_vasImpl.cpp
```diff
@@ -0,0 +1,187 @@
...
+        return everest::run_application::CmdControl::Continue;
+    });
+}
+
+void ISO15118_vasImpl::start_internet_service(const std::string& ports) {
+    if (this->internet_service_running) {
+        EVLOG_warning << "Internet service is already running.";
+        return;
+    }
+    this->internet_service_running = true;
```
james-ctc: There is a remote chance that there would be an issue were start_internet_service() called twice very quickly.
a safer pattern is to use a mutex
```
{
    std::lock_guard lock(mutex);
    if (internet_service_running) {
        EVLOG_warning << "Internet service is already running.";
        return;
    }
    internet_service_running = true;
}
```

it makes the test and set into a immutable operation
(internet_service_running would no longer need to be atomic)

another alternative would be to use exchange()
```
const auto previous = internet_service_running.exchange(true);
if (previous) {
    EVLOG_warning << "Internet service is already running.";
    return;
}

```

modules/EVSE/Iso15118InternetVas/iso15118_vas/ISO15118_vasImpl.cpp
```diff
@@ -0,0 +1,187 @@
...
+                    std::vector<std::string>{"up", this->mod->config.ev_interface, this->mod->config.modem_interface,
+                                             this->active_ports});
+}
+
+void ISO15118_vasImpl::stop_internet_service() {
+    if (!this->internet_service_running) {
+        EVLOG_warning << "Internet service is not running.";
+        return;
+    }
+    this->internet_service_running = false;
```
james-ctc:
```
{
    std::lock_guard lock(mutex);
    if (!internet_service_running) {
        EVLOG_warning << "Internet service is not running.";
        return;
    }
    internet_service_running = false;
}
```

modules/EVSE/Iso15118InternetVas/iso15118_vas/ISO15118_vasImpl.cpp
```diff
@@ -0,0 +1,187 @@
...
+    if (this->internet_service_running) {
+        EVLOG_warning << "Internet service is already running.";
+        return;
+    }
+    this->internet_service_running = true;
+    this->active_ports = ports;
+    EVLOG_info << "Starting internet service for ports: " << this->active_ports;
+
+    if (internet_service_thread.joinable()) {
+        internet_service_thread.join();
```
james-ctc: a couple of potential issues:
- if run_application never returns then you will block here until it does
- if clean shutdown of modules is implemented then you could have a thread that is never joined.
One option is to detach the thread so it never needs to be joined.

```
std::thread(&ISO15118_vasImpl::start_script, this, INTERNET_SETUP_SCRIPT,
                    std::vector<std::string>{"up", this->mod->config.ev_interface, this->mod->config.modem_interface,
                    this->active_ports}).detach();
```

modules/EVSE/Iso15118InternetVas/vas-internet-setup.sh
```diff
@@ -0,0 +1,224 @@
+#!/bin/bash
+
+# This script manages the full network setup to provide internet access
+# to an EV. It handles IP forwarding, NAT, a DHCP server, and a
+# router advertisement daemon.
+
+set -euo pipefail
+
+# --- Configuration ---
+DHCP_STATIC_IP="192.168.28.1"
```
james-ctc: perhaps use a less common 172.18.200.1
192.168. and 10. are commonly used.
one solution in shell scripts is to make use of an optional file in /etc/default

```
DHCP_STATIC_IP="192.168.28.1
[ -f /etc/default/vas-internet ] && . /etc/default/vas-internet
```
then  /etc/default/vas-internet can override DHCP_STATIC_IP if needed

modules/EVSE/Iso15118InternetVas/vas-internet-setup.sh
```diff
@@ -0,0 +1,224 @@
+#!/bin/bash
```
james-ctc: I tend to use /bin/sh there are license considerations when using bash (GPL3).

modules/EVSE/Iso15118InternetVas/vas-internet-setup.sh
```diff
@@ -0,0 +1,224 @@
...
+
+  # Allow return traffic for established connections
+  iptables -C FORWARD -i "$WAN_IFACE" -o "$LAN_IFACE" -m state --state RELATED,ESTABLISHED -j ACCEPT 2>/dev/null || \
+    iptables -A FORWARD -i "$WAN_IFACE" -o "$LAN_IFACE" -m state --state RELATED,ESTABLISHED -j ACCEPT
+ip6tables -C FORWARD -i "$WAN_IFACE" -o "$LAN_IFACE" -m state --state RELATED,ESTABLISHED -j ACCEPT 2>/dev/null || \
+    ip6tables -A FORWARD -i "$WAN_IFACE" -o "$LAN_IFACE" -m state --state RELATED,ESTABLISHED -j ACCEPT
+
+  # Add specific forwarding rules for outbound traffic from LAN interface
+  if [ -z "$PORTS" ]; then
+    echo "    - Allowing all outgoing traffic (no port restriction)"
```
james-ctc: I'm not sure this is a sensible default from a security perspective.
15118 seems to want specific ports, so this should never be empty.

modules/EVSE/Iso15118InternetVas/vas-internet-setup.sh
```diff
@@ -0,0 +1,224 @@
...
+# --- Core Logic Functions ---
+
+start_services() {
+  local LAN_IFACE=$1
+  local WAN_IFACE=$2
+  local PORTS=${3:-}
+
+  echo "[+] Starting all internet VAS services for $LAN_IFACE -> $WAN_IFACE"
+
+  # 1. Enable IP Forwarding
```
james-ctc, line 76: some basic firewall concepts
- block everything by default
- enable the bare minimum
- be careful about destinations (broadcast, multicast, SECC network, other EVs)
- remember there are other protocols beyond TCP/UDP

e.g.
ICMP is needed but only on the EV link
DHCP/DNS to the SECC only (DNS is proxied to a real server ...)
TCP is needed for the allowed ports only via the external internet interface
i.e. no HTTP/HTTPS to the SECC
no routing protocols (EVCC just needs the default route)
15118 ports need to be available: SDP and TCP/TLS

modules/EVSE/Iso15118InternetVas/vas-internet-setup.sh
```diff
@@ -0,0 +1,224 @@
...
+  echo "  [1/4] Enabling IPv4 & IPv6 forwarding"
+  sysctl -w net.ipv4.ip_forward=1 >/dev/null
+  sysctl -w net.ipv6.conf.all.forwarding=1 >/dev/null
+
```
james-ctc, line 82: forwarding doesn't appear to be disabled (and could be a problem if it was)

My recommendation is that forwarding is a documented requirement with examples for adding to /etc/sysctl.d
I don't think this script should be enabling forwarding. This could create all sorts of security issues if other network devices are not properly protected.

modules/EVSE/Iso15118InternetVas/vas-internet-setup.sh
```diff
@@ -0,0 +1,224 @@
...
+    iptables -C FORWARD -i "$LAN_IFACE" -o "$WAN_IFACE" -j ACCEPT 2>/dev/null || \
+      iptables -A FORWARD -i "$LAN_IFACE" -o "$WAN_IFACE" -j ACCEPT
+    ip6tables -C FORWARD -i "$LAN_IFACE" -o "$WAN_IFACE" -j ACCEPT 2>/dev/null || \
+      ip6tables -A FORWARD -i "$LAN_IFACE" -o "$WAN_IFACE" -j ACCEPT
+  else
+    echo "    - Allowing outgoing traffic on TCP/UDP ports: $PORTS"
+    # TCP
+    iptables -C FORWARD -i "$LAN_IFACE" -o "$WAN_IFACE" -p tcp -m multiport --dports "$PORTS" -j ACCEPT 2>/dev/null || \
+      iptables -A FORWARD -i "$LAN_IFACE" -o "$WAN_IFACE" -p tcp -m multiport --dports "$PORTS" -j ACCEPT
+    # UDP
```
james-ctc: the c++ only seems to support http and https (both TCP) - so UDP shouldn't be enabled.

modules/EVSE/Iso15118InternetVas/vas-internet-setup.sh
```diff
@@ -0,0 +1,224 @@
...
+  echo "[+] Starting all internet VAS services for $LAN_IFACE -> $WAN_IFACE"
+
+  # 1. Enable IP Forwarding
+  echo "  [1/4] Enabling IPv4 & IPv6 forwarding"
+  sysctl -w net.ipv4.ip_forward=1 >/dev/null
+  sysctl -w net.ipv6.conf.all.forwarding=1 >/dev/null
+
+  # 2. Setup NAT and Forwarding Rules
+  echo "  [2/4] Setting up NAT and FORWARD rules on $WAN_IFACE"
+  # MASQUERADE rule for outbound traffic on WAN interface
```
james-ctc, line 100: to simplify maintenance I suggest using user defined chains.
removing later is then much simpler and less error prone.
Not only that but this section can start with the remove and hence the check -C won't be needed

modules/EVSE/Iso15118InternetVas/vas-internet-setup.sh
```diff
@@ -0,0 +1,224 @@
...
+    kill "$(cat "$RADVD_PID_FILE")" || echo "RADVD server already stopped."
+    rm -f "$RADVD_PID_FILE"
+  else
+    echo "RADVD PID file not found. Maybe it was not running."
+  fi
+
+  # 2. Remove NAT and FORWARD rules
+  echo "  [3/3] Removing NAT and FORWARD rules"
+  iptables -t nat -D POSTROUTING -o "$WAN_IFACE" -j MASQUERADE 2>/dev/null || true
+  ip6tables -t nat -D POSTROUTING -o "$WAN_IFACE" -j MASQUERADE 2>/dev/null || true
```
james-ctc, line 221: an edge case:
what if the charger is talking to another device and wanting to run NAT ... removing this could cause issues.
e.g. the CS is connected via another ethernet port to a local lan (power meters, battery walls ...) and has been configured to be the gateway to the internet.

modules/EVSE/Iso15118InternetVas/vas-internet-setup.sh
james-ctc, line 1: I think it is best that this script is clearly identified as an example since there are many edge cases to consider that will be manufacturer specific.

Comments can be added to highlight areas that others might want to modify.
Along with a disclaimer that this should not be used in a production environment.
(if a manufacturer is happy with it then they can just remove the disclaimer)

modules/EVSE/Iso15118InternetVas/iso15118_vas/ISO15118_vasImpl.cpp
mlitre: This should be done in EvseV2G, but I just saw that it wasn't so we should definitely add it there.

modules/EVSE/Iso15118InternetVas/docs/index.rst
mlitre: Ok so if I understood it correctly, we should do IPv6 only by default and make IPv4 optional?

modules/EVSE/Iso15118InternetVas/docs/index.rst
mlitre: No it would be the one provided by the EVSE's router

modules/EVSE/Iso15118InternetVas/iso15118_vas/ISO15118_vasImpl.cpp
mlitre: Ok got it thanks for the information

modules/EVSE/Iso15118InternetVas/iso15118_vas/ISO15118_vasImpl.cpp
mlitre: Ah I see, I'll use the mutex I think

modules/EVSE/Iso15118InternetVas/iso15118_vas/ISO15118_vasImpl.cpp
mlitre: Yeah I was wondering between using detach or not, I guess that is the most suitable here

modules/EVSE/Iso15118InternetVas/docs/index.rst
james-ctc: my reading of 15118 is IPv4 shouldn't be used at all. An example script could include how to enable it if needed. But I wouldn't support IPv4 by default.

modules/EVSE/Iso15118InternetVas/docs/index.rst
james-ctc: The EVCC will need to have DNS configured. 

> [V2G2-053] If the SECC provides the Message Sets Certificate Install, Certificate Update, or Value Added
Services as defined in subclause 8.6.2 it shall support IETF RFC 6106 for providing a DNS
server address

part of radvd perhaps?

james-ctc (approved):
I suggest unaddressed comments are added as TODO items.

james-ctc (approved):

SebaLukas (approved):

PR #1361, opened 2025-09-10
Feature: add Linux_Systemd_Rauc module

Pietfried:
## Describe your changes

This PR adds the Linux_Systemd_Rauc module . It implements the `system` interface using rauc and base linux system commands.

The module requires a custom library for safe system commands, rauc and sdbus. This has been added as part of this PR to lib/everest/system


lib/everest/system/src/rauc-dbus-base.cpp
```diff
@@ -0,0 +1,455 @@
...
+    // checking the success or failure of an OTA update following reboot
+    //
+    //   get_primary_slot() may not be helpfull since it is updated as a result
+    //   of rauc status mark-active and isn't a good indicator that we have
+    //   rebooted into the newly updated partition
+    //
+    //    get_boot_slot() should be more reliable since it indicates the slot
+    //    we booted from. This will change on a successful OTA + reboot
+
+    std::string cmd = "/usr/bin/check_system_health.sh";
```
Pietfried: @james-ctc is this a file we may want to check in as part of the module and/or make the path to it configurable?

Pietfried:
TODO: add sdbus requirement to ci build kit image

lib/everest/system/src/rauc-dbus-base.cpp
james-ctc: I recommend making it a configuration item with an empty default.
`/usr/bin/check_system_health.sh` was provided in a Yocto meta layer. Changing to a config item removes that dependency.

lib/everest/system/src/rauc-dbus-base.cpp
```diff
@@ -0,0 +1,455 @@
...
+
+    // TODO(james-ctc): should the failure of this prevent the OTA being marked good?
+    const auto res = everest::lib::system::safe_system(cmd, {"check_system_health.sh"});
```
james-ctc: with the command being a config variable; the todo should be addressed.
i.e. if the script returns an error then the update is not marked as good.
the method decide_if_good() is the best place to implement that.

if the config script is not set then the result code should be 0 (success)
if the config script is set but the script file doesn't exist (or is not executable) then I recommend having some specific rc values that decide_if_good() could check.
Some may want to assume success unless the script runs and finds an issue.

res.status provides useful information ...
e.g. res.status == CMD_SETUP_FAILED then set res.code to a known value
process exit codes are usually 8-bit so -1 or > 255 would work well

lib/everest/system/src/rauc-dbus-base.cpp
```diff
@@ -0,0 +1,455 @@
...
+    return call_method_async(handler, proxy, interface::Installer, method::GetSlotStatus, timeout_us);
+}
+
+sdbus::PendingAsyncCall RaucBase::get_progress(property_cb handler) const {
+    return get_property_async(handler, proxy, property::Progress, interface::Installer);
+}
+
+bool RaucBase::decide_if_good(const rauc_messages::UpdateTransaction& saved, const CurrentState& current) {
+    // by default just check that the boot slot has changed
+    // The return code from /usr/bin/check_system_health.sh could be used to
```
james-ctc: with the script being a config item the return code should be checked.

lib/everest/system/include/everest/system/dbus-base.hpp
```diff
@@ -0,0 +1,37 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Pionix GmbH and Contributors to EVerest
+
+/**
+ * \file sdbus-c++ version independent interfaces
+ *
+ * The sdbus-c++ library is used to provide a C++ interface to DBus.
+ * The functions defined here enable code to compile and run against
+ * version 1.6 or version 2.1.0 of sdbus-c++.
```
james-ctc: comment needs updating:
`version 1.4 or higher or version 2.0.1 or higher`

lib/everest/system/include/everest/system/safe-system.hpp
```diff
@@ -0,0 +1,84 @@
...
+                                 int timeout_s = 30) {
+    return safe_system(-1, pathname, args, timeout_s);
+}
+
+/// @brief splits a given command line (cmd + args) into its components, respecting quoting
+/// @param command - the commandline to be split
+/// @return a vector
+std::pair<std::string, std::vector<std::string>> split_command_line(const std::string& command);
+
+/// @brief update the exit status from a chile process
```
james-ctc: chile -> child

lib/everest/system/src/rauc-dbus-base.cpp
```diff
@@ -0,0 +1,455 @@
...
+    //   rebooted into the newly updated partition
+    //
+    //    get_boot_slot() should be more reliable since it indicates the slot
+    //    we booted from. This will change on a successful OTA + reboot
+
+    std::string cmd = "/usr/bin/check_system_health.sh";
+
+    // TODO(james-ctc): should the failure of this prevent the OTA being marked good?
+    const auto res = everest::lib::system::safe_system(cmd, {"check_system_health.sh"});
+    if (res.code != 0) {
```
james-ctc: This check should have been updated when safe system was updated.
it should check res.status as well as res.code
EVLOG_error should also include the values.

modules/Misc/Linux_Systemd_Rauc/Linux_Systemd_Rauc.hpp
```diff
@@ -0,0 +1,83 @@
...
+
+namespace module {
+
+struct Conf {
+    double DefaultRetries;
+    double DefaultRetryInterval;
+    std::string OCPPLogPath;
+    std::string SessionLogPath;
+    std::string RebootCommand;
+};
```
james-ctc, line 33: add config for the OTA check script

modules/Misc/Linux_Systemd_Rauc/manifest.yaml
```diff
@@ -0,0 +1,36 @@
...
+    default: /var/log/everest/ocpp
+  SessionLogPath:
+    description: Output directory for session log files (see evse_manager session_logging_path)
+    type: string
+    default: /var/log/everest/session
+  RebootCommand:
+    description: Command to execute for rebooting the system
+    type: string
+    default: /sbin/reboot
+provides:
```
james-ctc, line 32: Add OTA check script as a string specifying the script to run to check if an OTA was successful.
The default should be an empty string

modules/Misc/Linux_Systemd_Rauc/manifest.yaml
```diff
@@ -0,0 +1,36 @@
...
+    interface: system
+requires:
+  store:
+    interface: kvs
+enable_external_mqtt: false
+metadata:
+  license: https://opensource.org/licenses/Apache-2.0
+  authors:
+    - Piet Gömpel
+    - Cornelius Claussen
```
james-ctc, line 44: probably should include me :-)
`- James Chapman`

modules/Misc/Linux_Systemd_Rauc/manifest.yaml
```diff
@@ -0,0 +1,36 @@
...
+    type: string
+    default: /var/log/everest/ocpp
+  SessionLogPath:
+    description: Output directory for session log files (see evse_manager session_logging_path)
+    type: string
+    default: /var/log/everest/session
+  RebootCommand:
+    description: Command to execute for rebooting the system
+    type: string
+    default: /sbin/reboot
```
james-ctc, line 23: Add OTA check script as a string specifying the script to run to check if an OTA was successful.
The default should be an empty string

lib/everest/system/src/rauc-dbus-base.cpp
```diff
@@ -0,0 +1,473 @@
...
+        EVLOG_error << "Health check script '" << verify_update_script_path << "' terminated by signal: " << res.code;
+        return HealthCheckStatus::ScriptError;
+    case everest::lib::system::CMD_SUCCESS:
+        if (res.code == 0) {
+            EVLOG_debug << "Health check script '" << verify_update_script_path << "' returned success";
+            return HealthCheckStatus::Success;
+        } else {
+            EVLOG_error << "Health check script '" << verify_update_script_path
+                        << "' returned error code: " << res.code;
+            return HealthCheckStatus::ScriptError;
```
james-ctc: I think it would be useful to separate the script exiting with an error and the script terminating due to a signal (line 224). Since the process exit code is no longer returned I think that each branch should return a unique enum value (including the should not reach here value)

lib/everest/system/src/safe-system.cpp
```diff
@@ -0,0 +1,235 @@
...
+std::string cmd_execution_status_to_string(const CommandExecutionStatus& status) {
+    switch (status) {
+    case CommandExecutionStatus::CMD_SUCCESS:
+        return "CMD_SUCCESS";
+    case CommandExecutionStatus::CMD_TERMINATED:
+        return "CMD_TERMINATED";
+    case CommandExecutionStatus::CMD_SETUP_FAILED:
+        return "CMD_SETUP_FAILED";
+    default:
+        throw std::runtime_error("No known conversion for the given CommandExecutionStatus");
```
james-ctc: looking at where cmd_execution_status_to_string() is used I suggest removing the exception and returning a suitably worded string instead. The alternative is to add try/catch around the use of cmd_execution_status_to_string()

modules/Misc/Linux_Systemd_Rauc/rauc-dbus.cpp
```diff
@@ -0,0 +1,162 @@
...
+            }
+        });
+}
+
+bool Rauc::decide_if_good(const rauc_dbus::rauc_messages::UpdateTransaction& saved, const CurrentState& current) {
+    // The original approach uses the primary slot, however this might not be
+    // as reliable as hoped. A change in boot slot should be more reliable
+    // however prior to this change the boot slot wasn't saved
+
+    bool result;
```
james-ctc: This should also use the updated decide_if_good() from the base class
```
bool result{false};
if (RaucBase::decide_if_good(saved, current)) {
   if (saved.boot_slot.empty()) {
   ...
   } else {
   ...
   }
}
```

james-ctc (dismissed):

lib/everest/system/include/everest/system/dbus_base.hpp
hikinggrass, line 1: These files should use `_` instead of `-` in their name

lib/everest/system/include/everest/system/rauc_dbus_base.hpp
hikinggrass, line 1: These files should use `_` instead of `-` in their name

lib/everest/system/include/everest/system/safe_system.hpp
hikinggrass, line 1: These files should use `_` instead of `-` in their name

lib/everest/system/src/dbus_base.cpp
hikinggrass, line 1: These files should use `_` instead of `-` in their name

lib/everest/system/src/rauc_dbus_base.cpp
hikinggrass, line 1: These files should use `_` instead of `-` in their name

lib/everest/system/src/rauc_dbus_base.cpp
```diff
@@ -0,0 +1,473 @@
...
+
+std::string RaucBaseSync::get_primary_slot(uint64_t timeout_us) {
+    std::string result;
+    auto method = dbus::createMethodCall(proxy, interface::Installer, method::GetPrimary);
+    call_method(result, proxy, method, method::GetPrimary, timeout_us);
+    return result;
+}
+
+std::map<std::string, std::map<std::string, std::string>> RaucBaseSync::get_slot_status(uint64_t timeout_us) {
+    // a(sa{sv})
```
hikinggrass, line 315:
```suggestion
```

lib/everest/system/src/safe_system.cpp
hikinggrass, line 1: These files should use `_` instead of `-` in their name

modules/Misc/Linux_Systemd_Rauc/diagnostics_handler.cpp
hikinggrass, line 1: These files should use `_` instead of `-` in their name

modules/Misc/Linux_Systemd_Rauc/diagnostics_handler.hpp
hikinggrass, line 1: These files should use `_` instead of `-` in their name

modules/Misc/Linux_Systemd_Rauc/diagnostics_collector.sh
hikinggrass, line 1: These files should use `_` instead of `-` in their name

modules/Misc/Linux_Systemd_Rauc/rauc_dbus.cpp
hikinggrass, line 1: These files should use `_` instead of `-` in their name

modules/Misc/Linux_Systemd_Rauc/rauc_dbus.hpp
hikinggrass, line 1: These files should use `_` instead of `-` in their name

lib/everest/system/src/rauc_dbus_base.cpp
```diff
@@ -312,7 +312,6 @@ std::string RaucBaseSync::get_primary_slot(uint64_t timeout_us) {
 }
 
 std::map<std::string, std::map<std::string, std::string>> RaucBaseSync::get_slot_status(uint64_t timeout_us) {
-    // a(sa{sv})
```
james-ctc: That is an important comment - it describes the dbus format of the message
see https://dbus.freedesktop.org/doc/dbus-specification.html

hikinggrass: ah, reverted the change :+1:

james-ctc (approved):
looks good to me

PR #1366, opened 2025-09-15
Feature/infypower dc powersupply driver

djchhp:
## Describe your changes
Integrate InfyPower DC powersupply driver


modules/HardwareDrivers/PowerSupplies/InfyPower/can_driver_acdc/CanBus.cpp
```diff
@@ -0,0 +1,134 @@
...
+    std::stringstream ss;
+    ss << std::hex << std::setfill('0');
+    for (size_t i = 0; i < bytes.size(); ++i) {
+        ss << std::setw(2) << static_cast<unsigned>(bytes[i]);
+    }
+    return ss.str();
+}
+
+bool CanBus::_tx(uint32_t can_id, const std::vector<uint8_t>& payload) {
+    // EVLOG_debug << "CAN frame sent using ID:" << std::hex << can_id << "#" << bytes_to_hex(payload);
```
Pietfried: could comment in or remove

modules/HardwareDrivers/PowerSupplies/InfyPower/can_driver_acdc/CanBus.hpp
```diff
@@ -0,0 +1,44 @@
...
+
+private:
+    std::unique_ptr<can::socket_can> can_bus;
+    std::atomic_bool on_error{false};
+    event::fd_event_handler ev_handler;
+    event::timer_fd recovery_timer;
+    event::timer_fd poll_status_timer;
+    std::atomic_bool exit_rx_thread;
+    std::thread rx_thread_handle;
+    std::condition_variable rx_thread_cv;
```
Pietfried: it doesnt look like `rx_thread_cv` is really used. There is a `notify_one` in the souce file, but no `wait`

djchhp:
@florinmihut Please address the review comments

modules/HardwareDrivers/PowerSupplies/InfyPower/main/power_supply_DCImpl.cpp
```diff
@@ -0,0 +1,268 @@
...
+    } else {
+        EVLOG_info << "Infy: Updating voltage/current via CAN: " << exportVoltage.load() << "V / "
+                   << exportCurrentLimit.load() << "A (but no active modules detected)";
+    }
+    mod->acdc->set_voltage_current(exportVoltage.load(), exportCurrentLimit.load());
+};
+
+void power_supply_DCImpl::handle_setImportVoltageCurrent(double& voltage, double& current) {
+    if (caps.min_import_voltage_V.has_value() && caps.max_import_current_A.has_value()) {
+        if (voltage > caps.max_import_voltage_V.value())
```
Pietfried, line 186: Looks like this can lead to bad optional access for `max_import_voltage_V`

modules/HardwareDrivers/PowerSupplies/InfyPower/can_driver_acdc/CanBus.cpp
florinmihut: Removed.

modules/HardwareDrivers/PowerSupplies/InfyPower/can_driver_acdc/CanBus.hpp
florinmihut: Good find, thank you. It was some garbage left, I deleted it and its usage.

modules/HardwareDrivers/PowerSupplies/InfyPower/main/power_supply_DCImpl.cpp
florinmihut: Aaaa, true, but import is invalid since the device supports (for now) only export and no energy import. 
However, if somebody would have called the function it would have missbehaved. 
I initialized the max and min values for import voltage and current to 0 so even if the import is called, the values will be set to 0 anyway.

Pietfried: i'd still add a check in the if condition for `max_import_voltage_V.has_value()` allowing the funciton to be more self contained

Issue #1075, opened 2025-03-06
EvseManager: The `limits` variable does not take PP values into account


matteocarnelos:
### Describe the bug

The `limits` variable of the `evse_manager` interface published by the EvseManager module does not take into account PP values published by the driver. This affects the call of the `ac_set_overcurrent_limit_A` of the `evse_board_support` interface too. Moreover, according to @corneliusclaussen, this affects AC ISO-2 charging as well.

The overcurrent detection mechanism is not affected, meaning that it will trigger correctly in case of consumption over cable capacity.

### EVerest Domain

Charge Control

### Affected EVerest Module

EvseManager

### To Reproduce

1. Run EVerest with an EvseManager-Driver configuration with 32A maximum import current and hardware capabilities
2. Connect a cable with a different PP value (e.g. 13A)
3. The published `limits` variable, under the `everest/<module-name>/evse/var` MQTT topic does not change to 13A

### Anything else?

- Reproduced on EVerest `2024.10.0`
- Preliminary discussion on Zulip: https://lfenergy.zulipchat.com/#narrow/channel/417675-EVerest-General/topic/EVerest.20not.20adjusting.20current.20limits.20according.20to.20PP.20values

PR #1377, opened 2025-09-22
EvseManager: fix limits variable does not take PP values into account

FaHaGit:
## Describe your changes

Fix bug "EvseManager: fix limits variable does not take PP values into account". Already tested the bug fix in the SIL.

#1075 

### DIFF
--- a/modules/EVSE/EvseManager/Charger.cpp
+++ b/modules/EVSE/EvseManager/Charger.cpp
@@ -1088,8 +1088,9 @@ bool Charger::set_max_current(float c, std::chrono::time_point<std::chrono::stea
                 shared_context.max_current = c_abs;
                 shared_context.max_current_valid_until = validUntil;
             }
+            c_abs = get_max_current_internal();
             bsp->set_overcurrent_limit(c_abs);
-            signal_max_current(c);
+            signal_max_current(c < 0.0f ? -c_abs : c_abs);
             return true;
         }
     }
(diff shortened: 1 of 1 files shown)


modules/EVSE/EvseManager/Charger.cpp
```diff
@@ -1077,18 +1077,20 @@ float Charger::ampere_to_duty_cycle(float ampere) {
 }
 
 bool Charger::set_max_current(float c, std::chrono::time_point<std::chrono::steady_clock> validUntil) {
-    if (c >= -CHARGER_ABSOLUTE_MAX_CURRENT and c <= CHARGER_ABSOLUTE_MAX_CURRENT) {
+    float c_abs{std::fabs(c)};
```
SebaLukas, line 1080: I dont like the variable name `c_abs` but I could not find a better alternative :sweat_smile:

modules/EVSE/EvseManager/Charger.cpp
```diff
@@ -1077,18 +1077,20 @@ float Charger::ampere_to_duty_cycle(float ampere) {
...
             {
                 Everest::scoped_lock_timeout lock(state_machine_mutex,
                                                   Everest::MutexDescription::Charger_set_max_current);
-                shared_context.max_current = std::fabs(c);
+                shared_context.max_current = c_abs;
                 shared_context.max_current_valid_until = validUntil;
             }
-            bsp->set_overcurrent_limit(std::fabs(c));
-            signal_max_current(c);
+            c_abs = get_max_current_internal();
```
SebaLukas, line 1093: Add a comment, that `c` and `c_abs` are not overwritten because `shared_context.max_current` was set with `c_abs` and will be checked in the `get_max_current_internal()` with the pp current value.
Without the comment, one might quickly assume that `c` is overwritten and no longer used.

SebaLukas (commented):
Overall looking good :+1:

modules/EVSE/EvseManager/Charger.cpp
FaHaGit: I have added some comments

SebaLukas (approved):
:+1:

PR #1389, opened 2025-09-26
Move error-framework example modules

mhei:
## Describe your changes

Move the error framework examples one directory higher since there are also Python error examples included.

While at, add the correct condition to prevent the Python modules to be built if not enabled.

none


modules/Examples/error-framework/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+ev_add_module(ExampleErrorRaiser)
+ev_add_module(ExampleErrorSubscriber)
+ev_add_module(ExampleErrorGlobalSubscriber)
+
+if(EVEREST_ENABLE_PY_SUPPORT)
```
hikinggrass: This should not be necessary since ev_add_module already checks if the language support of the module is activated or not. Did you have issues with this? Then the check in everest-generate.cmake would be broken and needs fixing

mhei: You're right... it is working as expected. Then probably a mistake on my side. Anyway, moving the error examples is still reasonable IMHO. I'll update the PR.

PR #1394, opened 2025-09-30
Refactoring Evse15118D20

SebaLukas:
## Describe your changes
Move some anoynm functions in utils.hpp/cpp. And a few clang-tidy suggestions were also implemented.

The high amount of local functions in Evse15118D20 made the module confusing. Therefore, the module was tidied up a bit in this PR.

### DIFF
--- a/modules/EVSE/Evse15118D20/charger/utils.hpp
+++ b/modules/EVSE/Evse15118D20/charger/utils.hpp
@@ -2,13 +2,18 @@
 // Copyright Pionix GmbH and Contributors to EVerest
 #pragma once
 
+#include <vector>
+
 #include <generated/types/iso15118.hpp>
+#include <generated/types/iso15118_vas.hpp>
+
 #include <iso15118/d20/ev_information.hpp>
+#include <iso15118/message/service_detail.hpp>
 #include <iso15118/message/type.hpp>
 
 static constexpr auto NUMBER_OF_SETUP_STEPS = 5;
 
-enum class SetupStep {
+enum class SetupStep : std::uint8_t {
     SETUP,
     ENERGY_SERVICE,
     AUTH_SETUP,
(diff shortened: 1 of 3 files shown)


modules/EVSE/Evse15118D20/charger/utils.hpp
```diff
@@ -2,13 +2,18 @@
 // Copyright Pionix GmbH and Contributors to EVerest
 #pragma once
 
+#include <vector>
```
james-ctc, line 7: #include &lt;optional&gt;

modules/EVSE/Evse15118D20/charger/utils.hpp
```diff
@@ -2,13 +2,18 @@
...
+#include <generated/types/iso15118_vas.hpp>
+
 #include <iso15118/d20/ev_information.hpp>
+#include <iso15118/message/service_detail.hpp>
 #include <iso15118/message/type.hpp>
 
 static constexpr auto NUMBER_OF_SETUP_STEPS = 5;
 
-enum class SetupStep {
+enum class SetupStep : std::uint8_t {
```
james-ctc, line 25: this should be in a namespace - module::charger perhaps?

modules/EVSE/Evse15118D20/charger/utils.hpp
```diff
@@ -105,6 +110,24 @@ constexpr types::iso15118::V2gMessageId convert_v2g_message_type(iso15118::messa
 }
 
 namespace module::charger {
+
+template <typename T, typename U> std::optional<T> convert_from_optional(const std::optional<U>& in);
+
 types::iso15118::AppProtocol convert_app_protocol(const iso15118::message_20::SupportedAppProtocol& app_protocol);
 types::iso15118::EvInformation convert_ev_info(const iso15118::d20::EVInformation& ev_info);
+
+template <typename T> types::iso15118::DcChargeDynamicModeValues convert_dynamic_values(const T& in);
```
james-ctc: There only appear to be two implementations:
```
types::iso15118::DcChargeDynamicModeValues convert_dynamic_values(const dt::Dynamic_DC_CLReqControlMode& in);
types::iso15118::DcChargeDynamicModeValues convert_dynamic_values(const dt::BPT_Dynamic_DC_CLReqControlMode& in); 
```
so function overloading could be used instead - that has the advantage detecting incorrect arguments rather than a function link error with templates

Other uses of templates below could also use overloaded functions ...

james-ctc (commented):
looks okay - templates could be replaced with function overloading which would pickup issues at compile rather than link time.

modules/EVSE/Evse15118D20/charger/utils.hpp
SebaLukas: My only problem with function overloading is that I also have to include several other libiso15118 header files in `utils.hpp`. But never mind, I can also use the overloaded functions.

james-ctc: You should be able to use opaque types so long as it is a reference or pointer
e.g.

In the header file
```
class classA;

bool function(classA &parameter);
```

The CPP file would need to include the actual headers (as it currently does)

SebaLukas: I tried that and I got an compile error. For me including the header files in this header file is ok

james-ctc: yes, including is fine.

james-ctc (approved):

PR #1397, opened 2025-10-01
Refactoring DC BPT

SebaLukas:
## Describe your changes
- Refactoring handle_enforce_limits to handle min & max charge and discharge limits
- Refactoring `subscribe_d20_dc_dynamic_charge_mode()` lambda to support d20 bidi  

Without this refactoring dc bpt does not work for ISO15118-20

### DIFF
--- a/modules/EVSE/EvseManager/EvseManager.cpp
+++ b/modules/EVSE/EvseManager/EvseManager.cpp
@@ -15,13 +15,12 @@
 
 using namespace std::literals::chrono_literals;
 
-namespace {
-static const std::vector<std::unique_ptr<powermeterIntf>> EMPTY_POWERMETER_VECTOR;
-}
-
 namespace module {
 
-static const types::power_supply_DC::Capabilities get_sane_default_power_supply_capabilities() {
+namespace {
+const std::vector<std::unique_ptr<powermeterIntf>> EMPTY_POWERMETER_VECTOR;
+
+types::power_supply_DC::Capabilities get_sane_default_power_supply_capabilities() {
     // Init power supply capabilities with safe defaults
     types::power_supply_DC::Capabilities psu_caps;
     psu_caps.bidirectional = false;
@@ -35,11 +34,11 @@ static const types::power_supply_DC::Capabilities get_sane_default_power_supply_
     return psu_caps;
 }
 
-inline static void trim_colons_from_string(std::string& text) {
+void trim_colons_from_string(std::string& text) {
     text.erase(remove(text.begin(), text.end(), ':'), text.end());
 }
 
-inline static types::authorization::ProvidedIdToken create_autocharge_token(std::string token, int connector_id) {
+types::authorization::ProvidedIdToken create_autocharge_token(std::string token, int connector_id) {
     types::authorization::ProvidedIdToken autocharge_token;
     autocharge_token.authorization_type = types::authorization::AuthorizationType::Autocharge;
     trim_colons_from_string(token);
@@ -47,6 +46,13 @@ inline static types::authorization::ProvidedIdToken create_autocharge_token(std:
     autocharge_token.connectors.emplace({connector_id});
     return autocharge_token;
 }
+
+bool almost_eq(double a, double b) {
+    constexpr auto eps = 1e-6;
+    return std::fabs(a - b) <= eps;
+}
+
+} // namespace
 
 void EvseManager::init() {
 
@@ -638,9 +644,11 @@ void EvseManager::ready() {
                 // Setting  power to actual voltage on the powermeter
                 const auto actual_voltage = ev_info.present_voltage.value_or(latest_target_voltage);
 
-                target_current = std::min((max_charge_power / actual_voltage), max_charge_current);
-
-                if (target_voltage != latest_target_voltage or target_current != latest_target_current) {
+                target_current = (actual_voltage <= 0.0)
+                                     ? max_charge_current
+                                     : std::min((max_charge_power / actual_voltage), max_charge_current);
+                if (not(almost_eq(target_voltage, latest_target_voltage) and
+                        almost_eq(target_current, latest_target_current))) {
                     latest_target_current = target_current;
                     latest_target_voltage = target_voltage;
                     target_changed = true;


modules/EVSE/EvseManager/EvseManager.cpp
```diff
@@ -471,18 +471,15 @@ void EvseManager::ready() {
...
                     if (config.hack_present_current_offset > 0) {
-                        present_values.evse_present_current =
-                            present_values.evse_present_current.value() + config.hack_present_current_offset;
+                        const auto current_offset = std::fabs(present_values.evse_present_current.value()) +
+                                                    static_cast<float>(config.hack_present_current_offset);
+                        present_values.evse_present_current = (m.current_A >= 0) ? current_offset : -current_offset;
                     }
 
-                    if (config.hack_pause_imd_during_precharge and m.voltage_V * m.current_A > 1000) {
+                    if (config.hack_pause_imd_during_precharge and m.voltage_V * std::fabs(m.current_A) > 1000) {
```
Pietfried, line 488: An explaining comment would help here and the magic `1000` could be made a const

modules/EVSE/EvseManager/EvseManager.cpp
```diff
@@ -566,31 +563,90 @@ void EvseManager::ready() {
...
-                if (latest_target_power <= PRE_CHARGE_MAX_POWER or values.min_charge_power > latest_target_power or
-                    values.max_charge_power < latest_target_power) {
-                    latest_target_current = static_cast<double>(values.max_charge_power) / latest_target_voltage;
-                    if (values.max_charge_current < latest_target_current) {
-                        latest_target_current = values.max_charge_current;
-                    }
+                // Setting  power to actual voltage on the powermeter
+                const auto actual_voltage = ev_info.present_voltage.value_or(latest_target_voltage);
+
+                target_current = std::min((max_charge_power / actual_voltage), max_charge_current);
```
Pietfried: Add a guard to prevent division-by-zero of `actual_voltage`.

modules/EVSE/EvseManager/EvseManager.cpp
```diff
@@ -566,31 +563,90 @@ void EvseManager::ready() {
                 });
 
             r_hlc[0]->subscribe_d20_dc_dynamic_charge_mode([this](types::iso15118::DcChargeDynamicModeValues values) {
-                constexpr auto PRE_CHARGE_MAX_POWER = 800.0f;
+                static bool last_is_actually_exporting_to_grid{false};
```
Pietfried, line 572: `last_is_actually_exporting_to_grid` looks like it should rather be a member of `EvseManager.hpp`. It is used in the `subscribe_d20_dc_dynamic_charge_mode` and in `powersupply_DC_set` and they look like they check the same concept (“was I exporting last cycle?”) but like this they are completely independent variables.

modules/EVSE/EvseManager/EvseManager.cpp
```diff
@@ -566,31 +563,90 @@ void EvseManager::ready() {
...
-                    latest_target_current = static_cast<double>(values.max_charge_power) / latest_target_voltage;
-                    if (values.max_charge_current < latest_target_current) {
-                        latest_target_current = values.max_charge_current;
-                    }
+                // Setting  power to actual voltage on the powermeter
+                const auto actual_voltage = ev_info.present_voltage.value_or(latest_target_voltage);
+
+                target_current = std::min((max_charge_power / actual_voltage), max_charge_current);
+
+                if (target_voltage != latest_target_voltage or target_current != latest_target_current) {
```
Pietfried: Comparing doubles/floats for exact equality is brittle, use an epsilon based comparison instead, e.g.

```cpp
static inline bool almost_equal(double a, double b, double eps = 1e-6) {
    return std::fabs(a - b) <= eps;
}
```

or there is already an `almost_eq` function in energyImpl.cpp

modules/EVSE/EvseManager/energy_grid/energyImpl.cpp
```diff
@@ -538,96 +538,110 @@ void energyImpl::handle_enforce_limits(types::energy::EnforcedLimits& value) {
...
                         // current limit is too low, i.e. charging will not reach the actual watt value.
                         // FIXME: we could use some magic here that involves actual measured voltage as well.
                         if (actual_voltage > 10) {
-                            evse_max_limits.evse_maximum_current_limit =
-                                value.limits_root_side.total_power_W.value().value / actual_voltage;
+                            total_current = watt_leave_side / actual_voltage;
                         } else {
-                            evse_max_limits.evse_maximum_current_limit =
-                                value.limits_root_side.total_power_W.value().value / target_voltage;
+                            total_current = watt_leave_side / target_voltage;
```
Pietfried, line 562: Again risk of division by zero

modules/EVSE/EvseManager/EvseManager.cpp
SebaLukas: I don't know where the `1000` comes from :man_shrugging:

modules/EVSE/EvseManager/energy_grid/energyImpl.cpp
SebaLukas: A very small risk. This will only happen if line 554 will be deleted.

modules/EVSE/EvseManager/EvseManager.cpp
SebaLukas: Then we have to use almost_eq relatively often. There are lots of places where floats/doubles are compared. But I'll try to use almost_eq here.

djchhp: @SebaLukas in https://github.com/EVerest/everest-core/pull/1360 I factored out almost_eq to lib/util
https://github.com/EVerest/everest-core/blob/a9506aa4d9f58f5343d7750287fdce5e808c1e49/lib/everest/util/include/everest/util/comparison.hpp

If it's usage becomes more prominent, I suggest taking that file and make it part of this PR. I also suggest subdirectory 'math' in the util lib for this

modules/EVSE/EvseManager/EvseManager.cpp
SebaLukas: Hmm, I'd rather leave it as it is.

modules/EVSE/EvseManager/EvseManager.cpp
SebaLukas: For now I added the `almost_eq()` function to an anonymous namespace in `EvseManager.cpp`. In the future we should use your refactored `comparsion.hpp` file.
But for me right now it is easier to add this small function.

modules/EVSE/EvseManager/EvseManager.cpp
Pietfried: If I understand the usage of `last_is_actually_exporting_to_grid` correctly it is actually an issue that should be adressed. The lambdas should use the same member state `last_is_actually_exporting_to_grid`. Defining them as static in the lambdas only makes them different variables with the same name

SebaLukas: > Defining them as static in the lambdas only makes them different variables with the same name

Yes, that's the idea behind it. I could also rename the variable in `subscribe_d20_dc_dynamic_charge_mode` lambda, but I think `last_is_actually_exporting_to_grid` actually fits well.
`is_actually_exporting_to_grid` is already a member variable and the static variables are used to check whether the energy flow has changed.

Pietfried (approved):

PR #1402, opened 2025-10-06
EvseV2G: ensure that connections are only accepted when possible

hikinggrass:
## Describe your changes

If a connection is already established another cannot be handled, so we should not allocate resources (like handler threads) for these connection attempts

### DIFF
--- a/modules/EVSE/EvseV2G/connection/connection.cpp
+++ b/modules/EVSE/EvseV2G/connection/connection.cpp
@@ -432,6 +432,8 @@ static void* connection_handle_tcp(void* data) {
     }
     dlog(DLOG_LEVEL_INFO, "TCP connection closed gracefully");
 
+    conn->ctx->connection_initiated = false;
+
     if (rv != ERROR_SESSION_ALREADY_STARTED) {
         /* cleanup and notify lower layers */
         connection_teardown(conn);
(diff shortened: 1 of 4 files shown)


james-ctc:
This is probably needed in ISOMux as well.
This PR does solve the resource issue - I do wonder whether we should consider allowing multiple connections and manage the resources appropriately.

modules/EVSE/EvseV2G/connection/tls_connection.cpp
james-ctc, line 88: `ctx->connection_initiated = false;` probably needed if the creation of the thread fails

modules/EVSE/EvseV2G/connection/connection.cpp
```diff
@@ -494,6 +496,16 @@ static void* connection_server(void* data) {
...
+            connection_teardown(conn);
+            free(conn);
+            conn = NULL;
+            continue;
+        }
+        ctx->connection_initiated = true;
+
         if (pthread_create(&conn->thread_id, &attr, connection_handle_tcp, conn) != 0) {
             dlog(DLOG_LEVEL_ERROR, "pthread_create() failed: %s", strerror(errno));
             continue;
```
james-ctc, line 513: ctx->connection_initiated = false; probably needed if the creation of the thread fails

modules/EVSE/EvseV2G/connection/connection.cpp
james-ctc, line 453: perhaps `conn->ctx->connection_initiated = false;` is needed to ensure the first connection is allowed.
Depends on whether connection_server() can be called more than once.

james-ctc (changes requested):

modules/EVSE/EvseV2G/connection/connection.cpp
hikinggrass: added this

SebaLukas (approved):

PR #1404, opened 2025-10-06
EvseV2G: change terminate_connection_on_failed_response default to true

hikinggrass:
This is very likely a safer default behavior since it keeps control of the V2G connection at the EVSE

### DIFF
--- a/modules/EVSE/EvseV2G/manifest.yaml
+++ b/modules/EVSE/EvseV2G/manifest.yaml
@@ -31,7 +31,7 @@ config:
       V2G connection is terminated immediately on a failed response code, otherwise
       the EV is responsible for closing of the V2G communication session with SessionStop.
     type: boolean
-    default: false
+    default: true
   tls_key_logging:
     description: >-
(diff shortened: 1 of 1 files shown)


SebaLukas:
The problem here is that some EVs dont like the immediately tcp/tls server shutdown and the cars wont charge about 5-30mins after this. Because the ECU crashed.
So the default should be false and not true

kwoyhope:
> The problem here is that some EVs dont like the immediately tcp/tls server shutdown and the cars wont charge about 5-30mins after this. Because the ECU crashed. So the default should be false and not true

How about setting SO_RCVTIMEO to short(ex, 1.5s) value on V2G_Resp_FAILED situation?
ie,  not terminate socket connection immediately by EVSE on error but by socket recv timeout only when EV doesn't handle error immediately/correctly(ex, send SessionStopReq, or disconnect tcp/tls from EV(client) side first).
IMHO, this could help for both normal and abnormal(including PLC link fail) case of EV behavior.

PR #1405, opened 2025-10-06
EvseV2G/IsoMux: remove insecure cipher ECDH-ECDSA-AES128-SHA256

hikinggrass:
## Describe your changes

This cipher isn't supported by OpenSSL anymore anyway, so we can also just drop it from the cipher lists


james-ctc:
My only concern is that it is a requirement from 15118-2 section 7.7.3.4
`For interoperability, of course, the SECC shall support all cipher suites as listed in Table 7.`

SebaLukas:
Yeah it is a requirement of ISO-2 but newer openssl versions dont support the cipher suite anyway. So removing it, should be not a problem.

PR #1413, opened 2025-10-08
PacketSniffer improvements

barsnick:
## Describe your changes
Some improvements to PacketSniffer logging and error handling.

### DIFF
--- a/modules/Misc/PacketSniffer/PacketSniffer.cpp
+++ b/modules/Misc/PacketSniffer/PacketSniffer.cpp
@@ -58,7 +58,7 @@ void PacketSniffer::ready() {
 
 void PacketSniffer::capture(const std::string& logpath, const std::string& session_id) {
     already_started = true;
-    const std::string fn = fmt::format("{}/ethernet-traffic.dump", logpath);
+    const std::string fn = fmt::format("{}/ethernet-traffic.pcap", logpath);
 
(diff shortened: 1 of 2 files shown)


modules/Misc/PacketSniffer/PacketSniffer.cpp
```diff
@@ -16,17 +16,22 @@ void PacketSniffer::init() {
...
     if (p_handle == nullptr) {
-        EVLOG_error << fmt::format("Could not open device {}. Sniffing disabled.", config.device);
+        EVLOG_error << fmt::format("Could not open device \"{}\", Sniffing disabled.{}", config.device,
+                                   errb.size() > 0 ? (std::string(" Error: ") + errb) : "");
         return;
     }
 
-    if (pcap_datalink(p_handle) != DLT_EN10MB) {
-        EVLOG_error << fmt::format("Device {} doesn't provide Ethernet headers - not supported. Sniffing disabled.",
+    if (config.device != "any" && pcap_datalink(p_handle) != DLT_EN10MB) {
```
SebaLukas, line 26: Why "any"?

mhei: See commit message. The special word "any" is understood by pcap to capture not only on a specific interface. But to allow this specific use-case, the extra check for this special word guards the existing datalink header check as we know for sure that it will not be DLT_EN10MB in this case and it would fail.

PR #1415, opened 2025-10-09
Feat/async util

djchhp:
## Describe your changes
This adds a new monitor class template to everest/lib/util implementing the monitor pattern

The monitor wraps up some data structure and give pointer like access via a handle, that ensures exclusive access.

Depending on the used mutex type additional features, like getting the handle with a timeout (std::timed_mutex) or conditional wait (std::mutex) are available.

See the included doxygen documentation and the testcases for details

### DIFF
--- a/lib/everest/tls/include/everest/tls/tls_types.hpp
+++ b/lib/everest/tls/include/everest/tls/tls_types.hpp
@@ -6,7 +6,7 @@
 
 #include <cstdint>
 
-#include <everest/util/EnumFlags.hpp>
+#include <everest/util/enum/EnumFlags.hpp>
 
 struct ocsp_response_st;
 struct ssl_ctx_st;
@@ -26,7 +26,7 @@ private:
         last = trusted_ca_keys,
     };
 
-    everest::util::AtomicEnumFlags<flags_t> flags;
+    everest::lib::util::AtomicEnumFlags<flags_t> flags;
 
 public:
     void status_request_received() {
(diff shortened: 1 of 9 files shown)


lib/everest/tls/include/everest/tls/tls_types.hpp
```diff
@@ -6,7 +6,7 @@
 
 #include <cstdint>
 
-#include <everest/util/EnumFlags.hpp>
+#include <everest/util/enum/EnumFlags.hpp>
```
james-ctc, line 9: Note: changing the location of header files has impacts beyond everest-core.
The lib/everest libraries are used by other repos.
The result is dependant code will now need conditional compilation depending on which version of everest-core they are being compiled against.
So as a minimum there needs to be a cmake variable available to support this.

djchhp: Thank you for that insight. I just did not expect this small lib to be used anywhere else.

Thinking about it now:
What code could that be? I don't think that external code is allowed to rely on such details. Changing interfaces or types happens quite frequently and that breaks potentially everything.

djchhp: If in this case changing the location triggers an cascade of errors, I would prefer to move things back to their original location to a conditional compilation.

djchhp: added library version as discussed.

Pietfried (approved):

PR #1422, opened 2025-10-14
fix: detect musl-based systems and increase stack size

markxoe:
## Describe your changes

musl based systems have a default stack size that is far smaller that glibc based systems. This leads to stack overflows on some systems, including Alpine Linux.
This PR adds a check to the root `CMakeLists.txt` that adds a linker flag to increase the stack size to 8MiB if necessary.

This change has been tested on Alpine Linux 3.18 and Ubuntu 24.04

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -43,16 +43,22 @@ endif()
 # test whether we are building on a musl based system.
 # On these we need to increase the stack size as it is very low by default
 find_program(LDD_EXECUTABLE ldd)
-if(LDD_EXECUTABLE)
-    # we use /bin/ls as it is very likely to exist on any system. ldd prints the dynamic linked libraries
+find_program(LS_EXECUTABLE ls)
+if(LDD_EXECUTABLE AND LS_EXECUTABLE)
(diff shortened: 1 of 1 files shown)


SebaLukas (approved):

CMakeLists.txt
```diff
@@ -40,6 +40,21 @@ else()
     set(ATOMIC_LIBS "")
 endif()
 
+# test whether we are building on a musl based system.
+# On these we need to increase the stack size as it is very low by default
+find_program(LDD_EXECUTABLE ldd)
+if(LDD_EXECUTABLE)
+    # we use /bin/ls as it is very likely to exist on any system. ldd prints the dynamic linked libraries
+    # of the given executable, if musl is used it will print it as a linked object
+    execute_process(COMMAND ${LDD_EXECUTABLE} /bin/ls OUTPUT_VARIABLE LDD_LS_OUTPUT ERROR_QUIET)
```
hikinggrass: Probably using find_program for `ls` would be needed as well, otherwise I could imagine this using a host binary in eg a cross compiling setup

markxoe: good point :+1:

hikinggrass (approved):

SebaLukas (approved):

PR #1438, opened 2025-10-17
Adding more utilities to lib 'util'

djchhp:
This adds more utilities to the header only 'util' library

## Describe your changes
New templates for concurrency

* simple_queue
  * A simple wrapper around std::queue with a more convenient interface
* thread_safe_queue
  * A thread safe queue, especially for producer/consumer scenarios
* async_wrapper
  * Wrapping up a resource for serialized access. The resource has it's own associated thread and all access is performed in that thread in order. 

For details refer to the doxygen documentation and the unit tests. Both are fairly exhaustive

### DIFF
--- a/lib/everest/util/include/everest/util/async/async_wrapper.hpp
+++ b/lib/everest/util/include/everest/util/async/async_wrapper.hpp
@@ -1,0 +1,268 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2020 - 2025 Pionix GmbH and Contributors to EVerest
+
+/**
+ * @file Wrapping of a resource and serializing the access to the resource
+ * @brief The content is based on the conference talk 'Concurrency-and-Parallelism'
+ * held by Herb Sutter at C-and-Beyond-2012.
+ * The original pattern has been extend with exhaustive handling of corner cases and errors.
+ * Policies allow for the adaptation to common use cases.
+ */
+
+#pragma once
+
+#include <atomic>
+#include <everest/util/queue/thread_safe_queue.hpp>
+#include <exception>
+#include <functional>
+#include <future>
+#include <memory>
+#include <thread>
+#include <type_traits>
+
+namespace everest::lib::util::testing_interface {
+//  Forward declaration of the test fixture used as a friend
+class AsyncWrapperTest;
+} // namespace everest::lib::util::testing_interface
+
+namespace everest::lib::util {
+
+// --- EXCEPTION POLICIES (Policy 1) ---
+
+/**
+ * @brief Policy that triggers a global shutdown if a user-supplied task throws an exception.
+ * @details Used for guarded resources where an exception implies resource corruption.
+ */
+struct GlobalFailurePolicy {
+    /**
+     * @brief Sets the global promise with the given exception pointer, permanently failing the executor.
+     */
+    static void handle_user_exception(std::shared_ptr<std::promise<void>> const& global_promise_ptr,
+                                      std::exception_ptr current_exception_ptr) {
+        try {
+            global_promise_ptr->set_exception(current_exception_ptr);
+        } catch (const std::future_error&) {
+            // Ignore: promise was already satisfied by an earlier thread/task
+        }
+    }
+};
+
+/**
+ * @brief Policy that contains a user-supplied exception locally (does not cause global failure).
+ * @details Used for background tasks where exceptions do not corrupt the resource state.
+ */
+struct LocalFailurePolicy {
+    /**
+     * @brief Handles the user exception by doing nothing to the global promise.
+     */
+    static void handle_user_exception([[maybe_unused]] std::shared_ptr<std::promise<void>> const& global_promise_ptr,
+                                      [[maybe_unused]] std::exception_ptr current_exception_ptr) {
+    }
+};
+
+// --- SHUTDOWN POLICIES (Policy 2) ---
+
+/**
+ * @brief Destructor policy that waits for all queued tasks to finish execution before joining the worker thread.
+ */
+struct WaitToFinishPolicy {
+    /**
+     * @brief Performs a graceful shutdown by pushing a sentinel task and joining.
+     */
+    template <typename QueueT, typename ThreadT>
+    static void shutdown(QueueT& queue, ThreadT& worker, std::atomic_bool& done_flag) {
+        // Push the final stop signal, which will be executed after all pending tasks.
+        queue.push([&done_flag] { done_flag = true; });
+        worker.join();
+    }
+};
+
+/**
+ * @brief Destructor policy that signals the worker to exit immediately, potentially dropping queued tasks.
+ */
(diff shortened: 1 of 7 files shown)


lib/everest/util/include/everest/util/async/async_wrapper.hpp
```diff
@@ -0,0 +1,272 @@
...
+ * @brief Policy that contains a user-supplied exception locally (does not cause global failure).
+ * @details Used for background tasks where exceptions do not corrupt the resource state.
+ */
+struct LocalFailurePolicy {
+    /**
+     * @brief Handles the user exception by doing nothing to the global promise.
+     */
+    static void handle_user_exception(std::shared_ptr<std::promise<void>> const& global_promise_ptr,
+                                      std::exception_ptr current_exception_ptr) {
+        (void)global_promise_ptr;
```
james-ctc: from recent reviews of clang-tidy recommendations the following is preferred. i.e. comment out the unused argument:
```
static void handle_user_exception(std::shared_ptr<std::promise<void>> const& /*global_promise_ptr*/,
                                      std::exception_ptr /*current_exception_ptr*/) {
```

lib/everest/util/include/everest/util/async/async_wrapper.hpp
```diff
@@ -0,0 +1,272 @@
...
+     * @brief Performs an immediate shutdown by setting the flag and unblocking the thread.
+     */
+    template <typename QueueT, typename ThreadT>
+    static void shutdown(QueueT& queue, ThreadT& worker, std::atomic_bool& done_flag) {
+        // Signal termination immediately
+        done_flag = true;
+
+        // Send a dummy task to unblock the worker if it's currently blocking on pop()
+        queue.push([] {});
+
```
james-ctc: remove blank lines - so similar to lines 77 and 78

lib/everest/util/include/everest/util/async/async_wrapper.hpp
```diff
@@ -0,0 +1,272 @@
...
+ * @tparam T The resource type being wrapped.
+ * @tparam ExceptionPolicy Defines global failure behavior (e.g., GlobalFailurePolicy).
+ * @tparam ShutdownPolicy Defines destructor behavior (e.g., WaitToFinishPolicy).
+ * @tparam QueueT The underlying thread-safe queue implementation.
+ */
+template <typename T, typename ExceptionPolicy, typename ShutdownPolicy,
+          template <class> typename QueueT = thread_safe_queue>
+class async_wrapper_impl {
+public:
+    friend class everest::lib::util::testing_interface::AsyncWrapperTest; ///< Allows GTest fixture to access
```
james-ctc, line 110: an alternative is to use protected rather than private and then unit tests etc. use a wrapper class that inherits async_wrapper_impl and will have access to the protected members and methods.

lib/everest/util/include/everest/util/queue/thread_safe_queue.hpp
```diff
@@ -0,0 +1,102 @@
...
+    value_type pop() {
+        return pop_impl(-1).value();
+    }
+
+private:
+    std::optional<value_type> pop_impl(int timeout_ms) {
+        std::unique_lock lock(m_mtx);
+        auto wait_predicate = [this]() { return not m_queue.empty(); };
+
+        if (timeout_ms < 0) {
```
james-ctc, line 85: perhaps a check for timeout_ms == 0 could be added to avoid unlocking and then relocking the mutex

lib/everest/util/include/everest/util/queue/thread_safe_queue.hpp
```diff
@@ -0,0 +1,102 @@
...
+            m_cv.wait(lock, wait_predicate);
+        } else {
+            auto element_found = m_cv.wait_for(lock, std::chrono::milliseconds(timeout_ms), wait_predicate);
```
james-ctc: should be const auto ... however there isn't a need for a temporary variable
```
if (not m_cv.wait_for(lock, std::chrono::milliseconds(timeout_ms), wait_predicate)) {
  return std::nullopt;
}
```

lib/everest/util/include/everest/util/queue/thread_safe_queue.hpp
djchhp: Thanks for the suggestions, it's a nice optimization.

lib/everest/util/include/everest/util/queue/thread_safe_queue.hpp
djchhp: I made this `auto const`. You are right, it's not necessary, but I think it's easier to read. I am sure we can rely on the optimizer and there is no impact on performance at all.

lib/everest/util/include/everest/util/async/async_wrapper.hpp
djchhp: Yes, that code came from old habits. I use this most of the time now. https://en.cppreference.com/w/cpp/language/attributes/maybe_unused.html 

That states clearly what the intend is and looks superior to the old fashioned void cast and also commenting out parameter names. What do you think?

lib/everest/util/include/everest/util/async/async_wrapper.hpp
djchhp: done

lib/everest/util/include/everest/util/async/async_wrapper.hpp
djchhp: Yes, you are right. I was trying around quite a bit with that, too. With all the template parameters, that turned out to be a bit unpractical. Also when using that approach you are no longer testing the original class (well, the behavior should be exactly the same). Also things, that are supposed to be an implementation detail would be accessible by any class inheriting from the async_wrapper_impl. That is an even bigger encapsulation breach.
I think this comes down to taste, since all the approaches have their own draw backs. 

In general I think tests should rely on the public interface only, whenever that is possible. But this template class is somewhat special, since many execution paths cannot possible be triggered by anything but real failure of the system. Since handling those cases gracefully is a major feature, there is no way around at least a limited breach of encapsulation here.

lib/everest/util/include/everest/util/queue/thread_safe_queue.hpp
james-ctc, line 91: Since m_queue.pop() checks if the queue is empty then there is a further simplification possible.

```
        } else if (timeout_ms > 0) {
            m_cv.wait_for(lock, std::chrono::milliseconds(timeout_ms), wait_predicate);
        }
```
The check for found or not is whether the queue is empty.

lib/everest/util/include/everest/util/async/async_wrapper.hpp
james-ctc: Fine by me - I suggest letting Kai know as there might be a clang tidy rule that can recommend that approach.

PR #1442, opened 2025-10-20
EvseManager: Don't skip events

### LINKED ISSUE #1441: IECStateMachine may skip bsp event
### Describe the bug

IECStateMachine may skip bsp events and does not guarantee any ordering of the events. 

When receiving a bsp event, the `process_bsp_event` writes the event into the member `cp_state` and offloads the handling into a thread. Now, if bsp events arrive in close succession this does not guarantee that every `cp_state` is processed before being overwritten. Instead the threads may only see the last event. It does also not guarantee that the events generated by the `state_machine` are processed sequentially since the `events` queues are processed in different threads 

### EVerest Domain

Charge Control

### Affected EVerest Module

EvseManager

### To Reproduce

_No response_

### Anything else?

_No response_

dorezyuk:
Fixes #1441

### DIFF
--- a/modules/EVSE/EvseManager/EvseManager.cpp
+++ b/modules/EVSE/EvseManager/EvseManager.cpp
@@ -933,7 +933,7 @@ void EvseManager::ready() {
             r_over_voltage_monitor[0]->call_reset_over_voltage_error();
         }
 
-        charger->process_event(event);
+        charger->bsp_event_queue.push(event);
 
         // Forward some events to HLC
         if (hlc_enabled) {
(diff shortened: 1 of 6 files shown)


modules/EVSE/EvseManager/EvseManager.cpp
```diff
@@ -869,7 +869,7 @@ void EvseManager::ready() {
             r_over_voltage_monitor[0]->call_reset_over_voltage_error();
         }
 
-        charger->process_event(event);
```
dorezyuk, line 936: my understanding is that the initial deadlock reported by cornelius occurred in the following screnario. 

1. In Charger::process_event we aquire the Charger::state_machine_mutex
2. The charger then can call bsp (e.x IECStateMachine::set_pwm), the IECStateMachine then calls IECStateMachine::feed_state_machine, where we call the signal_event, which will try to call again Charger::process_event. From there we deadlock on the Chager::state_machine_mutex.

The solution here is to offload the Charger::process_event strictly into the main_thread of the Charger

PR #1462, opened 2025-10-27
RsPaymentTerminal: Use AuthTokenInterface

dorezyuk:
Changes the payment terminal interface to use the auth-token-validator. This prevents from reserving bank card money if no connector is available

### DIFF
--- a/modules/HardwareDrivers/Payment/RsPaymentTerminal/src/main.rs
+++ b/modules/HardwareDrivers/Payment/RsPaymentTerminal/src/main.rs
@@ -201,7 +201,7 @@ impl From<ErrorMessages> for AuthorizationStatus {
             ErrorMessages::PaymentMethodNotSupported => AuthorizationStatus::Blocked,
             ErrorMessages::AbortViaTimeoutOrAbortKey => AuthorizationStatus::Invalid,
             #[cfg(feature = "with_lavego_error_codes")]
-            ErrorMessages::Declined => AuthorizationStatus::Blocked,
+            ErrorMessages::Declined => AuthorizationStatus::Timeout,
             ErrorMessages::ReceiverNotReady
                 | ErrorMessages::SystemError
                 | ErrorMessages::ErrorFromDialUp   // error from dial-up/communication fault
@@ -444,7 +444,7 @@ impl AuthTokenValidatorServiceSubscriber for PaymentTerminalModule {
         }
 
         if &provided_token.id_token.value == INVALID_BANK_TOKEN {
-            log::warn!("Received an {INVALID_BANK_TOKEN}");
+            log::warn!("Validating a `BankCard` without an invoice token");
             return Ok(AuthorizationStatus::Invalid.into());
         }
 


types/authorization.yaml
```diff
@@ -14,6 +14,8 @@ types:
       - NotAtThisLocation
       - NotAtThisTime
       - Unknown
+      - PinRequired
```
dorezyuk, line 17: @hikinggrass @Pietfried does this make sense? I see that we're mapping those to ocpp and this won't work. I'm also not sure if the everest_api_types can be changed or if you want to keep them stable

modules/HardwareDrivers/Payment/RsPaymentTerminal/src/main.rs
```diff
@@ -459,26 +368,63 @@ impl SessionCostClientSubscriber for PaymentTerminalModule {
...
+            );
+            return Ok(AuthorizationStatus::Invalid.into());
+        }
 
-    fn allow_all_cards_for_every_connector(&self, _context: &Context) -> ::everestrs::Result<()> {
-        let mut map_guard = self.connector_to_card_type.lock().unwrap();
-        map_guard.clear();
-        Ok(())
+        if &provided_token.id_token.value == INVALID_BANK_TOKEN {
+            log::warn!("Received an {INVALID_BANK_TOKEN}");
```
wku12: Not very descriptive, can you write here that we are trying to validate bank card without invoice token?

doc/everest_api_specs/auth_token_validator_API/asyncapi.yaml
```diff
@@ -184,6 +184,8 @@ components:
         - NotAtThisLocation
         - NotAtThisTime
         - Unknown
+        - PinRequired
+        - Timeout
```
ydankner, line 194: Timeout is a bit generic - are we indicating here payment terminal operation timeout specifically?

types/payment_terminal.yaml
```diff
@@ -30,19 +30,3 @@ types:
...
-    type: string
-    enum:
-        - BankCard
-        - RfidCard
-  RejectionReason:
-    description: Indicates the reason why the card was rejected
-    type: string
-    enum:
-        - PinRequired
-        - InsufficientFunds
```
ydankner, line 44: I noticed `InsufficientFunds` and others are not present in `AuthorizationStatus` which replaces this one, is there no need for those

modules/HardwareDrivers/Payment/RsPaymentTerminal/src/main.rs
```diff
@@ -32,27 +32,34 @@ include!(concat!(env!("OUT_DIR"), "/generated.rs"));
 use anyhow::Result;
```
ydankner, line 33: I somehow can't comment on lines above - line 15-16 of this file contain information that will no longer hold true:
```
//! In case a bank card is presented, it will pre-authorize the
//! `pre_authorization_amount` and issue a token on the `auth_token_provider`
```
Please update this :)

modules/HardwareDrivers/Payment/RsPaymentTerminal/src/main.rs
```diff
@@ -159,100 +166,119 @@ impl ProvidedIdToken {
...
     ///
-    /// In case of a bank card we start the transaction on the payment terminal.
-    /// In case of membership cards we just issue the token. In both cases we
-    /// don't flag the token as pre-validated to allow the consumers to add
-    /// custom validation steps on top.
-    fn begin_transaction(&self, publishers: &ModulePublisher) -> Result<()> {
-        const NO_CONNECTORS: Option<Vec<i64>> = Some(Vec::new());
+    /// Regardless of the card type we don't flag the token as pre-validated to
+    /// allow the consumers to add custom validation steps on top. For bank
+    /// cards use the auth_token_validation to pre-authorize money.
```
ydankner: You probably mean `auth_token_validator`, not validation

modules/HardwareDrivers/Payment/RsPaymentTerminal/src/main.rs
```diff
@@ -159,100 +166,119 @@ impl ProvidedIdToken {
...
-                if get_connectors_for_rfid_card == NO_CONNECTORS
-                    && get_connectors_for_bank_card == NO_CONNECTORS
-                {
-                    log::debug!("Reading is disabled, waiting...");
-                    std::thread::sleep(Duration::from_secs(1));
+                    let inner: PTError = inner.into();
+                    publishers.payment_terminal.raise_error(inner.into());
                     continue;
+                } else {
+                    publishers.payment_terminal.clear_all_errors();
```
ydankner, line 284: Nice!

modules/HardwareDrivers/Payment/RsPaymentTerminal/src/main.rs
```diff
@@ -159,100 +166,119 @@ impl ProvidedIdToken {
...
                 }
 
                 // Attempting to get an invoice token
-                if token.is_none() && get_connectors_for_bank_card != NO_CONNECTORS {
```
ydankner, line 253: Is it intentional that we attempt to get the invoice token now even if we don't allow bank cards?

modules/HardwareDrivers/Payment/RsPaymentTerminal/src/main.rs
```diff
@@ -271,76 +297,43 @@ impl PaymentTerminalModule {
...
-                                publishers
-                                    .payment_terminal
-                                    .rejection_reason((*rejection_reason).into())?;
-                            }
-                            return Err(anyhow::anyhow!("Failed to read card: {e:?}"));
+                            log::warn!("Failed to read a card {e:?}");
+                            // Cleared in the next loop if we can configure the
+                            // feig.
+                            let inner: PTError = e.into();
+                            publishers.payment_terminal.raise_error(inner.into());
```
ydankner, line 328: We need to remember to change the way the display logic detects the failure to rely on everest errors here instead of rejection_reason. Do we have a Jira ticket with this information?

modules/HardwareDrivers/Payment/RsPaymentTerminal/src/main.rs
```diff
@@ -271,76 +297,43 @@ impl PaymentTerminalModule {
...
-                }
-                ProvidedIdToken::new(
-                    id_token,
-                    AuthorizationType::RFID,
-                    get_connectors_for_rfid_card,
-                )
-            }
+            CardInfo::Bank => ProvidedIdToken::new(
+                // If we don't have a valid token, still issue the token but
+                // reject it in the validation.
```
ydankner: Maybe document here - `this is done for transparency, to allow external modules (i.e. display), to be notified of the payment failure`

modules/HardwareDrivers/Payment/RsPaymentTerminal/src/main.rs
```diff
@@ -271,76 +297,43 @@ impl PaymentTerminalModule {
...
+                token.unwrap_or(INVALID_BANK_TOKEN.to_string()),
+                AuthorizationType::BankCard,
+                map_guard
+                    .entry(AuthorizationType::BankCard)
+                    .or_default()
```
ydankner, line 346: I don't know this rust idiom - this means an `Some` with empty vector in this case? Or `None`?

modules/HardwareDrivers/Payment/RsPaymentTerminal/src/main.rs
```diff
@@ -459,27 +409,87 @@ impl SessionCostClientSubscriber for PaymentTerminalModule {
...
         _context: &Context,
         connector_id: i64,
-        supported_cards: Vec<CardType>,
+        supported_cards: Vec<AuthorizationType>,
     ) -> ::everestrs::Result<()> {
-        let mut map_guard = self.connector_to_card_type.lock().unwrap();
-        map_guard.insert(connector_id, supported_cards);
+        let mut map_guard = self.card_type_to_connector.lock().unwrap();
+
+        for (key, val) in map_guard.iter_mut() {
```
ydankner: Maybe more indicative names here than key, val?

modules/HardwareDrivers/Payment/RsPaymentTerminal/src/main.rs
```diff
@@ -459,27 +409,87 @@ impl SessionCostClientSubscriber for PaymentTerminalModule {
...
+            match supported_cards.iter().find(|card| *card == key) {
+                Some(_) => {
+                    // Card is allowed -> add it to the connector
```
ydankner: Nit - `add it to the connector` -> `Add the connector under the card type`

modules/HardwareDrivers/Payment/RsPaymentTerminal/src/main.rs
```diff
@@ -459,27 +409,87 @@ impl SessionCostClientSubscriber for PaymentTerminalModule {
...
+                    let connector_ids = val.get_or_insert_default();
+                    if let Err(idx) = connector_ids.binary_search(&connector_id) {
+                        connector_ids.insert(idx, connector_id);
+                    }
+                }
+                None => {
+                    // Card is forbidden -> remove the connector from that
```
ydankner: Nit - `remove the connector from that` -> `Remove the connector from the card type`

modules/HardwareDrivers/Payment/RsPaymentTerminal/src/main.rs
```diff
@@ -502,14 +512,18 @@ fn main() -> Result<()> {
     let pt_module = Arc::new(PaymentTerminalModule {
         tx,
         feig: SyncFeig::new(pt_config),
-        connector_to_card_type: Mutex::new(HashMap::new()),
+        card_type_to_connector: Mutex::new(HashMap::from_iter([
+            (AuthorizationType::BankCard, None),
+            (AuthorizationType::RFID, None),
+        ])),
```
ydankner: Nit - to avoid slight code repetition, you can initialize it here to the default value (empty hash map), and immediately after call `pt_module.allow_all_cards_for_every_connector`. Otherwise, if we ever add other authorization type, this change will have to be copied in both places.

doc/everest_api_specs/auth_token_validator_API/asyncapi.yaml
dorezyuk: my thought was that it can also be used for other token validator - e.x. if the call to the csms times times out

types/payment_terminal.yaml
dorezyuk: I think the `InsufficientFunds` is mapped to `NoCredit`

modules/HardwareDrivers/Payment/RsPaymentTerminal/src/main.rs
dorezyuk: default of Option is None

modules/HardwareDrivers/Payment/RsPaymentTerminal/src/main.rs
wku12: bump

modules/HardwareDrivers/Payment/RsPaymentTerminal/src/main.rs
```diff
@@ -159,101 +167,137 @@ impl ProvidedIdToken {
...
-            ErrorMessages::CreditNotSufficient => RejectionReason::InsufficientFunds,
-            ErrorMessages::PaymentMethodNotSupported => RejectionReason::CardNotSupported,
-            ErrorMessages::AbortViaTimeoutOrAbortKey => RejectionReason::Aborted,
+                | ErrorMessages::NecessaryDeviceNotPresentOrDefective => AuthorizationStatus::PinRequired,
+            ErrorMessages::CreditNotSufficient => AuthorizationStatus::NoCredit,
+            ErrorMessages::PaymentMethodNotSupported => AuthorizationStatus::Blocked,
+            ErrorMessages::AbortViaTimeoutOrAbortKey => AuthorizationStatus::Invalid,
             #[cfg(feature = "with_lavego_error_codes")]
-            ErrorMessages::Declined => RejectionReason::Timeout,
+            ErrorMessages::Declined => AuthorizationStatus::Blocked,
```
wku12: This is timeout according to lavego docs

modules/HardwareDrivers/Payment/RsPaymentTerminal/src/main.rs
```diff
@@ -159,101 +167,137 @@ impl ProvidedIdToken {
...
+    /// Sender for the `ModulePublisher` -> to get the publisher from `on_ready`
+    /// into the main thread.
+    tx: Sender<ModulePublisher>,
+
+    /// The Feig interface.
+    feig: SyncFeig,
+
+    /// The mapping of supported connectors for every card type. None means
+    /// no restrictions, Some(vec![]) means no connector.
+    card_type_to_connector: Mutex<HashMap<AuthorizationType, Option<Vec<i64>>>>,
```
wku12, line 257: Having a hash map for few auth types seems excessive
Plus, the searches are quite weird. You first check the enum in the map and than do binary search on 2 element vector...
I don't think this change make the code easier

modules/HardwareDrivers/Payment/RsPaymentTerminal/src/main.rs
```diff
@@ -272,76 +316,44 @@ impl PaymentTerminalModule {
...
-                )
-            }
+            CardInfo::Bank => ProvidedIdToken::new(
+                // If we don't have a valid token, still issue the token but
+                // reject it in the validation - so e.x. display can still
+                // react to cards being read.
+                token.unwrap_or(INVALID_BANK_TOKEN.to_string()),
+                AuthorizationType::BankCard,
+                map_guard
+                    .entry(AuthorizationType::BankCard)
```
wku12, line 345: Did you check what Auth module does with None as connectors for provided token?
Last time I checked internal tests, when connectors are published as None, Auth grants auth to all connectors
I thought the whole idea here which we discussed was to still stop Auth module from getting the authentication attempt through bank cards if they are disabled

dorezyuk: yes None means no limits. It should never be none - default is here to not unwrap. Should we return Some(Vec::new())?

modules/HardwareDrivers/Payment/RsPaymentTerminal/src/main.rs
dorezyuk: the idea was to make the "read" part easier since we generate/provide tokens more often than updating the enabled connectors

modules/HardwareDrivers/Payment/RsPaymentTerminal/src/main.rs
dorezyuk: keeping None as discussed in person

modules/HardwareDrivers/Payment/RsPaymentTerminal/src/main.rs
```diff
@@ -541,415 +568,276 @@ mod tests {
     use mockall::predicate::eq;
     use zvt_feig_terminal::feig::TransactionSummary;
 
+    impl From<SyncFeig> for PaymentTerminalModule {
```
ydankner, line 571: Nit - since I see the values below are test values (pre_authorization_amount: 11), to prevent someone from using this accidentally and not changing the pre auth amount, you can add `#[cfg(test)]` above the `impl ...` statement

dorezyuk: that is actually a blunder - will fix

PR #1463, opened 2025-10-27
Added external SLAC API

Pietfried:
## Describe your changes

Added support for external slac API module:
* Added slac_API module - External API for SLAC
* Added BUSlac module - BringUp module for SLAC
* Added SLAC API documentation


modules/API/slac_API/slac_API.cpp
```diff
@@ -0,0 +1,161 @@
...
+    });
+}
+
+void slac_API::generate_api_var_state() {
+    subscribe_api_topic("state", [=](const std::string& data) {
+        API_types_ext::State val;
+        if (deserialize(data, val)) {
+            p_main->publish_state(serialize(val));
+            return true;
+        }
```
djchhp: There is the conversion step missing between the internal and external enum. Compoare with'mode' var in power_supply_DC for example.

modules/API/slac_API/slac_API.cpp
```diff
@@ -0,0 +1,161 @@
...
+    });
+}
+
+void slac_API::generate_api_var_request_error_routine() {
+    subscribe_api_topic("request_error_routine", [=](const std::string& data) {
+        bool val = false;
+        if (deserialize(data, val)) {
+            p_main->publish_request_error_routine(nullptr);
+            return true;
+        }
```
djchhp: This does not look right. Do you expect data to contain a serialized bool, and if so, what is it used for. looking at the asyncapi.yaml I would expect empty data.

modules/API/slac_API/slac_API.cpp
```diff
@@ -0,0 +1,161 @@
...
+
+namespace ev_API_types = ev_API::V1_0::types;
+namespace API_types_ext = ev_API_types::slac;
+namespace API_generic = ev_API_types::generic;
+using ev_API::deserialize;
+
+void slac_API::init() {
+    invoke_init(*p_main);
+
+    topics.setTargetApiModuleID(info.id, "slac");
```
djchhp: When rebasing on current main, this needs to become 
`topics.setup(info.id, "slac", 1);`

modules/API/slac_API/manifest.yaml
```diff
@@ -0,0 +1,25 @@
...
+config:
+  cfg_communication_check_to_s:
+    description: "Maximum time between two communication check events. Values <= 0 disables communication checks."
+    type: integer
+    default: 5
+  cfg_heartbeat_interval_ms:
+    description: "Interval between two heartbeat messages send by the API. Values <= 0 disable heartbeat"
+    type: integer
+    default: 1000
+  cfg_request_reply_to_s:
```
djchhp: Remove this, since this interface has no request/reply at all

modules/API/slac_API/main/slacImpl.cpp
```diff
@@ -0,0 +1,55 @@
...
+#include <everest_api_types/slac/codec.hpp>
+
+namespace module {
+namespace main {
+
+using namespace everest::lib::API;
+namespace generic = everest::lib::API::V1_0::types::generic;
+
+void slacImpl::init() {
+    timeout_s = mod->config.cfg_request_reply_to_s;
```
djchhp: This can be remove, since there is no request/reply used

modules/API/slac_API/slac_API.hpp
```diff
@@ -0,0 +1,100 @@
...
+class slac_API : public Everest::ModuleBase {
+public:
+    slac_API() = delete;
+    slac_API(const ModuleInfo& info, Everest::MqttProvider& mqtt_provider, std::unique_ptr<slacImplBase> p_main,
+             Conf& config) :
+        ModuleBase(info),
+        mqtt(mqtt_provider),
+        p_main(std::move(p_main)),
+        config(config),
+        comm_check("slac/CommunicationFault", "Bridge to implementation connection lost", this->p_main) {
```
djchhp: is this slac/ oer generic/ ?

modules/BringUp/BUSlac/doc.rst
djchhp, line 1: remove this file?

modules/BringUp/BUSlac/docs/index.rst
djchhp, line 1: remove this file?

modules/API/slac_API/docs/index.rst
djchhp, line 1: remove this file?

modules/API/slac_API/doc.rst
djchhp, line 1: reference ths asyncapi.yaml

doc/everest_api_specs/slac_API/asyncapi.yaml
```diff
@@ -0,0 +1,369 @@
...
+            - MATCHED
+    DLinkStatus:
+      type: object
+      properties:
+        dlink_ready:
+          type: boolean
+          description: >-
+            Inform higher layers about a change in data link status. Emits true
+            if link was set up and false when the link is shut down.
+    NullType:
```
djchhp: I am not sure about this

modules/API/slac_API/slac_API.cpp
Pietfried: Internally the `state` variable is just a string so I dont think a conversion is required here. The enum values are defined in the interface directly and not as a separate EVerest type: 
https://github.com/EVerest/everest-core/blob/main/interfaces/slac.yaml#L20

modules/API/slac_API/slac_API.cpp
Pietfried: Ah, the bool conversion was wrong indeed. publishing the `nullptr` should be fine, since its done in the same way here: https://github.com/EVerest/everest-core/blob/main/modules/EVSE/EvseSlac/main/slacImpl.cpp#L62

modules/API/slac_API/slac_API.cpp
djchhp: They are required, because that triggers a compilation failure if something changes on the enum definition. It is the also style used thoughout all API modules and that should remain consistend.

Pietfried: There is no internal EVerest enum I can convert to since the enum values are only specified in the internal interface but not as a seperate type. I would need to change the internal interface for this.

djchhp: Ah, now I get it, sorry. Well then we need to live with it.

felix-ulonska:
**⚠️ PR Affected by Security Incident**

This PR was affected by a security incident where a compromised account force-pushed to this branch and closed the PR.

**✅ Branch History Restored**

We've run an automation that restored the complete history of this branch to the time before the attack. All your commits are preserved. Nothing has been updated in the respective branch in case someone force-pushed after the attack but before our automation.

**🔄 Action Required**

GitHub won't allow reopening this PR after the close and force-push. Please:
 - Open a new PR from your branch to the target branch. You can use the following link to open a new PR: [Open PR](https://github.com/EVerest/everest-core/compare/main...feature/slac-api)
 - Link back to this PR in the description to include previous discussions

We apologize for the inconvenience.

PR #1469, opened 2025-10-29
Add compatibility with Boost 1.89 while retaining backwards compat

hikinggrass:
Since Boost 1.69 Boost.system is a header-only library, but retained compatibility by defining link targets

This compatibility was removed with Boost 1.89.

This adds a simple compatibility layer since a few EVerest (everest-framework and libtimer) dependencies rely on the old behavior until these dependencies are patched and integrated as well


SebaLukas (approved):

andistorm (approved):
LGTM 🐧

PR #1472, opened 2025-11-03
Implementation of the isolation monitor, voltage to earth processing

florinmihut:
Implementation of the isolation monitor, voltage to earth processing in EvseManager - raise error if outside borders and stop charging

### DIFF
--- a/modules/EVSE/EvseManager/EvseManager.cpp
+++ b/modules/EVSE/EvseManager/EvseManager.cpp
@@ -456,11 +456,53 @@ void EvseManager::ready() {
                 r_imd[0]->subscribe_isolation_measurement([this](types::isolation_monitor::IsolationMeasurement m) {
                     // new DC isolation monitoring measurement received
 
-                    // Are we in charge loop?
+                    // Check for isolation errors
                     if (charger->get_current_state() == Charger::EvseState::Charging and
                         not check_isolation_resistance_in_range(m.resistance_F_Ohm)) {
                         error_handling->raise_isolation_resistance_fault(
-                            fmt::format("Isolation resistance too low during charging: {} Ohm", m.resistance_F_Ohm));
+                            fmt::format("Isolation resistance too low during charging: {} Ohm", m.resistance_F_Ohm),
+                            "Resistance");
+                    }
+                    // Check L1e and L2e to PE as defined by IEC 61851-23:2023, $6.3.1.112.2
+                    if (charger->get_current_state() == Charger::EvseState::Charging) {
+                        bool is_in_range = check_voltage_to_protective_earth_in_range(m);
+                        auto now = std::chrono::steady_clock::now();
+
+                        if (not is_in_range) {
+                            // Voltage to earth is out of range
+                            if (voltage_to_earth_failure_count == 0) {
+                                // First failure - record the time
+                                first_voltage_to_earth_failure_time = now;
+                                voltage_to_earth_failure_count = 1;
+                                EVLOG_debug << "Voltage to earth out of range (first failure)";
+                            } else {
+                                // Subsequent failure - increment count
+                                voltage_to_earth_failure_count++;
+                                auto time_since_first_failure = std::chrono::duration_cast<std::chrono::seconds>(
+                                    now - first_voltage_to_earth_failure_time);
+
+                                EVLOG_debug << "Voltage to earth out of range. Failure count: "
+                                            << voltage_to_earth_failure_count
+                                            << ", time since first: " << time_since_first_failure.count() << "s";
+
+                                // Only raise error if we have at least 2 failures AND at least 2 seconds between first
+                                // and last
+                                if (voltage_to_earth_failure_count >= REQUIRED_CONSECUTIVE_FAILURES and
+                                    time_since_first_failure >= MIN_TIME_BETWEEN_FIRST_AND_LAST_FAILURE) {
+                                    error_handling->raise_isolation_resistance_fault(
+                                        fmt::format("Voltage to earth too high during charging L1e: {} V, L2e: {} V",
+                                                    m.voltage_to_earth_l1e_V.value_or(NAN),
+                                                    m.voltage_to_earth_l2e_V.value_or(NAN)),
+                                        "VoltageToEarth");
+                                }
+                            }
+                        } else {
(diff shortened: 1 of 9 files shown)


modules/EVSE/EvseManager/EvseManager.cpp
```diff
@@ -450,11 +450,20 @@ void EvseManager::ready() {
...
                         error_handling->raise_isolation_resistance_fault(
-                            fmt::format("Isolation resistance too low during charging: {} Ohm", m.resistance_F_Ohm));
+                            fmt::format("Isolation resistance too low during charging: {} Ohm", m.resistance_F_Ohm),
+                            "Resistance");
+                    }
+                    // Check L1e and L2e to PE as defined by IEC 61851-23:2023, $6.3.1.112.2
+                    if (charger->get_current_state() == Charger::EvseState::Charging and
+                        not check_voltage_to_protective_earth_in_range(m)) {
+                        error_handling->raise_isolation_resistance_fault(
+                            fmt::format("Voltage to earth too low during charging L1e: {} V, L2e: {} V",
```
corneliusclaussen: too high instead of too low

modules/EVSE/EvseManager/EvseManager.cpp
```diff
@@ -1694,6 +1702,24 @@ bool EvseManager::cable_check_should_exit() {
     return charger->get_current_state() not_eq Charger::EvseState::PrepareCharging;
 }
 
+bool EvseManager::check_voltage_to_protective_earth_in_range(types::isolation_monitor::IsolationMeasurement m) {
+    static const double MAX_VOLTAGE_STATIC = 550.0; // defined by IEC 61851-23:2023, $6.3.1.112.2
+    if (m.voltage_V.has_value() and m.voltage_to_earth_l1e_V.has_value() and m.voltage_to_earth_l2e_V.has_value()) {
+        if (powersupply_capabilities.max_export_voltage_V > 500) {
+            double max_allowed_value = m.voltage_V.value() * 1.1;
+            return (abs(m.voltage_to_earth_l1e_V.value()) < max_allowed_value and
```
corneliusclaussen: abs() is poorly defined in C++: it may be int or float. Maybe cleaner to use std::abs() from  cmath

modules/EVSE/EvseManager/EvseManager.cpp
```diff
@@ -1694,6 +1702,24 @@ bool EvseManager::cable_check_should_exit() {
...
+                    abs(m.voltage_to_earth_l2e_V.value()) < max_allowed_value);
+        } else {
+            return (abs(m.voltage_to_earth_l1e_V.value()) < MAX_VOLTAGE_STATIC and
+                    abs(m.voltage_to_earth_l2e_V.value()) < MAX_VOLTAGE_STATIC);
+            ;
```
corneliusclaussen: extra ;

modules/EVSE/EvseManager/EvseManager.cpp
```diff
@@ -1694,6 +1702,24 @@ bool EvseManager::cable_check_should_exit() {
     return charger->get_current_state() not_eq Charger::EvseState::PrepareCharging;
 }
 
+bool EvseManager::check_voltage_to_protective_earth_in_range(types::isolation_monitor::IsolationMeasurement m) {
+    static const double MAX_VOLTAGE_STATIC = 550.0; // defined by IEC 61851-23:2023, $6.3.1.112.2
```
corneliusclaussen: constexpr?

modules/EVSE/EvseManager/EvseManager.cpp
```diff
@@ -450,11 +450,20 @@ void EvseManager::ready() {
...
+                    // Check for isolation errors
                     if (charger->get_current_state() == Charger::EvseState::Charging and
                         not check_isolation_resistance_in_range(m.resistance_F_Ohm)) {
                         error_handling->raise_isolation_resistance_fault(
-                            fmt::format("Isolation resistance too low during charging: {} Ohm", m.resistance_F_Ohm));
+                            fmt::format("Isolation resistance too low during charging: {} Ohm", m.resistance_F_Ohm),
+                            "Resistance");
+                    }
+                    // Check L1e and L2e to PE as defined by IEC 61851-23:2023, $6.3.1.112.2
+                    if (charger->get_current_state() == Charger::EvseState::Charging and
```
corneliusclaussen: We probably need to add an additional check here: this may trigger while ramping down voltage: For 1000V capabilities, if the actual voltage is e.g. 60V because it is currently in ramp down while switching off, the limit is only 66V. If the UL1e/2e is poorly synchronized to the actual voltage measurement it may trigger errors. So we should probably check if power supply is switched on and if voltage is above a useful limit  (e.g. powersupply_capabilities.min_export_voltage() or so)

corneliusclaussen (changes requested):

modules/EVSE/EvseManager/EvseManager.cpp
```diff
@@ -1733,6 +1774,32 @@ bool EvseManager::cable_check_should_exit() {
...
 }
 
+bool EvseManager::check_voltage_to_protective_earth_in_range(types::isolation_monitor::IsolationMeasurement m) {
+    static constexpr double MAX_VOLTAGE_STATIC = 550.0; // defined by IEC 61851-23:2023, $6.3.1.112.2
+    if (m.voltage_V.has_value() and m.voltage_to_earth_l1e_V.has_value() and m.voltage_to_earth_l2e_V.has_value()) {
+        // This may trigger while ramping down voltage so we should try to check against the minium valid voltage the
+        // the power supply can deliver. If the UL1e/2e is poorly synchronized to the actual voltage measurement it may
+        // trigger false positives thus limiting the checks to min_export_voltage
+        if (m.voltage_V.value() > powersupply_capabilities.min_export_voltage_V) {
+            // the standard defines the charger rating (powersupply_capabilities.max_export_voltage_V > 500) as
```
corneliusclaussen: This needs to be changed back to the capabilties instead of the actual voltage

florinmihut: Reimplemented as requested by the standard now as agreed.

corneliusclaussen (changes requested):
Testing on real HW produces ModBus read/write errors sometimes.

PR #1490, opened 2025-11-07
Remove ac_read_pp_ampacity from evse_board_support interface

Pietfried:
## Describe your changes
Remove `ac_read_pp_ampacity` command from evse_board_support interface.

BSPs shall use the var `ac_pp_ampacity` instead to report the pp_ampacity.

The existing BSPs should continue to be functional because they are already publishing `ac_pp_ampacity`:
- AdAcEvse22KwzKitBSP
- PhyVersoBSP
- TIDA010939
- YetiDriver

The MicroMegaWattBSP doesn't support socket and is therefore not affected.

_______________

The changes in this PR include a proper implementation of the `ac_pp_ampacity` handling in the EvseManager. This was missing. The EvseManager only implemented PP handling in case the command `ac_read_pp_ampacity` was used.


modules/Simulation/YetiSimulator/board_support/evse_board_supportImpl.cpp
```diff
@@ -69,29 +69,6 @@ void evse_board_supportImpl::handle_evse_replug(int& _) {
     EVLOG_error << "Replugging not supported";
 }
 
-types::board_support_common::ProximityPilot evse_board_supportImpl::handle_ac_read_pp_ampacity() {
```
SebaLukas, line 72: The YetiSimulator does not publish `ac_pp_ampacity`. Maybe add this publish with this PR?

SebaLukas (commented):
Looking good :) Please have a look into my comment about the YetiSimulator

modules/Simulation/YetiSimulator/board_support/evse_board_supportImpl.cpp
Pietfried: The YetiSimulator publishes its connector type as a fixed cable here https://github.com/EVerest/everest-core/blob/main/modules/Simulation/YetiSimulator/board_support/evse_board_supportImpl.cpp#L20 so the pp value would not be used and doesn't make sense with fixed cable.

It would be nice to allow the YetiSimulator to be configurable for socket / cable but I'd prefer to do this in a subsequent PR

PR #1512, opened 2025-11-12
Fix EIM with PlugEvents and multiple EVSEs

Pietfried:
## Describe your changes

Using Authorized event of EvseManager to `call_session_setup` at ISO15118 module to remove the contract payment option. 

This is required to fix Plug&Charge in case  multiple EVSEs. If a card is swiped, the EvseManager only receives authorization after a plug in is registered, but `call_session_setup` is called before the authorization is recieved. This leads to the fact that `call_session_setup` may be called including the payment option Contract. Using the Authorized event to `call_session_setup` again will remove this option so that the ISO15118 stack doesnt offer contract if authorization has been previously received

`call_session_setup` is not called on `SessionEvent::SessionFinished` anymore.

### DIFF
--- a/modules/EVSE/EvseManager/EvseManager.cpp
+++ b/modules/EVSE/EvseManager/EvseManager.cpp
@@ -1089,27 +1089,39 @@ void EvseManager::ready() {
             p_evse->publish_ev_info(ev_info);
         }
 
+        if (not hlc_enabled) {
+            return;
+        }
+
+        if (s != types::evse_manager::SessionEventEnum::Authorized and
+            s != types::evse_manager::SessionEventEnum::SessionFinished) {
+            return;
+        }
+
         std::vector<types::iso15118::PaymentOption> payment_options;
         // if pnc is disabled, disable contract installation and central contract validation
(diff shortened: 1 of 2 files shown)


modules/EVSE/EvseManager/EvseManager.cpp
```diff
@@ -1089,27 +1089,35 @@ void EvseManager::ready() {
...
-                payment_options.push_back(types::iso15118::PaymentOption::ExternalPayment);
-            }
-            r_hlc[0]->call_session_setup(payment_options, _contract_certificate_installation_enabled,
-                                         _central_contract_validation_allowed);
+        if (config.payment_enable_eim) {
+            payment_options.push_back(types::iso15118::PaymentOption::ExternalPayment);
+        }
+        if (pnc_enabled and s != types::evse_manager::SessionEventEnum::Authorized) {
+            // In this case we have already received EIM authentication,
+            // no need for Contract payment option.
```
SebaLukas: Comment is wrong -> `In this case we have not received EIM auth [...]`

PR #1526, opened 2025-11-16
Extended evse_manager_consumerAPI module with SessionInfo

Pietfried:
## Describe your changes
Extended evse_manager_consumerAPI module with SessionInfo:
* Additional route e2m/session_info was added that contains information about a charging session
* Data of e2m/session_info is not simply forwared from internal APIs. The data from internal APIs is processed and converted into the external API format This requires a small state machine to track the EVSE status and meter values

Note: The implementation is based on the implementation of SessionInfo in the [API module](https://github.com/EVerest/everest-core/tree/main/modules/API/API). Some parts were reused and modified.

### DIFF
--- a/modules/API/evse_manager_consumer_API/evse_manager_consumer_API.cpp
+++ b/modules/API/evse_manager_consumer_API/evse_manager_consumer_API.cpp
@@ -80,6 +80,7 @@ void evse_manager_consumer_API::ready() {
     generate_api_cmd_random_delay_set_duration_s();
 
     generate_api_var_session_event();
+    generate_api_var_session_info(); // special, not just forwarded
     generate_api_var_limits();
     generate_api_var_ev_info();
     generate_api_var_car_manufacturer();
@@ -440,4 +441,31 @@ void evse_manager_consumer_API::subscribe_api_topic(std::string const& var,
         }
     });
 }
+
+void evse_manager_consumer_API::generate_api_var_session_info() {
+    this->session_info.handle()->set_publish_callback(
+        [this](const everest::lib::API::V1_0::types::evse_manager::SessionInfo& external) {
+            auto topic = topics.everest_to_extern("session_info");
+            try {
+                auto&& payload = serialize(external);
+                mqtt.publish(topic, payload);
+            } catch (const std::exception& e) {
+                EVLOG_warning << "Variable: '" << topic << "' failed with -> " << e.what();
+            } catch (...) {
+                EVLOG_warning << "Invalid data: Cannot convert internal to external or serialize it.\n" << topic;
+            }
+        });
+
+    this->r_evse_manager->subscribe_session_event([this](types::evse_manager::SessionEvent const& session_event) {
+        session_info.handle()->update_state(session_event);
+    });
+
+    this->r_evse_manager->subscribe_powermeter([this](types::powermeter::Powermeter const& powermeter) {
+        session_info.handle()->update_powermeter(powermeter);
+    });
+
+    this->r_evse_manager->subscribe_selected_protocol(
+        [this](std::string const& protocol) { session_info.handle()->update_selected_protocol(protocol); });
+}
+
 } // namespace module
--- a/modules/API/evse_manager_consumer_API/evse_manager_consumer_API.hpp
+++ b/modules/API/evse_manager_consumer_API/evse_manager_consumer_API.hpp
@@ -29,8 +29,11 @@
 
 // ev@4bf81b14-a215-475c-a1d3-0a484ae48918:v1
 // insert your custom include headers here
+#include <everest/util/async/monitor.hpp>
 #include <everest_api_types/utilities/CommCheckHandler.hpp>
 #include <everest_api_types/utilities/Topics.hpp>
+
+#include "session_info.hpp"
 
 namespace ev_API = everest::lib::API;
 
@@ -119,6 +122,7 @@ private:
     void generate_api_cmd_random_delay_set_duration_s();
 
     void generate_api_var_session_event();
+    void generate_api_var_session_info();
     void generate_api_var_limits();
     void generate_api_var_ev_info();
     void generate_api_var_car_manufacturer();
@@ -148,6 +152,8 @@ private:
     ev_API::Topics topics;
     ev_API::CommCheckHandler<generic_errorImplBase> comm_check;
     size_t hb_id{0};
+    everest::lib::util::monitor<SessionInfo> session_info;
+
     // ev@211cfdbe-f69a-4cd6-a4ec-f8aaa3d1b6c8:v1
 };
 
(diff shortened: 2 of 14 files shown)


modules/API/evse_manager_consumer_API/evse_manager_consumer_API.cpp
```diff
@@ -440,4 +454,61 @@ void evse_manager_consumer_API::subscribe_api_topic(std::string const& var,
...
+                if (transaction_finished.meter_value.energy_Wh_export.has_value()) {
+                    auto energy_Wh_export = transaction_finished.meter_value.energy_Wh_export.value().total;
+                    session_info->set_end_energy_export_wh(energy_Wh_export);
+                } else {
+                    session_info->end_energy_export_wh_was_set = false;
+                }
+            }
+        }
+        publish_session_info();
+    });
```
djchhp, line 461: I suggest moving the logic into the session_info class and keep this function simple.

modules/API/evse_manager_consumer_API/evse_manager_consumer_API.cpp
```diff
@@ -440,4 +454,61 @@ void evse_manager_consumer_API::subscribe_api_topic(std::string const& var,
...
+        }
+        if (powermeter.power_W.has_value()) {
+            session_info->set_latest_total_w(powermeter.power_W.value().total);
+        }
+
+        if (session_info->is_session_running()) {
+            publish_session_info();
+        }
+    });
+
```
djchhp, line 466: Same here, let session_info object decide what to do.

modules/API/evse_manager_consumer_API/evse_manager_consumer_API.hpp
```diff
@@ -101,6 +103,7 @@ class evse_manager_consumer_API : public Everest::ModuleBase {
     void subscribe_api_topic(std::string const& var, ParseAndPublishFtor const& parse_and_publish);
 
     auto forward_api_var(std::string const& var);
+    void publish_session_info();
```
djchhp: Why the new name? generate_api_var_session_info would be consistent

modules/API/evse_manager_consumer_API/session_info.cpp
```diff
@@ -0,0 +1,222 @@
...
+    start_energy_export_wh(0),
+    end_energy_export_wh(0),
+    latest_total_w(0),
+    state(State::Unknown) {
+    this->start_time_point = date::utc_clock::now();
+    this->end_time_point = this->start_time_point;
+}
+
+everest::lib::API::V1_0::types::evse_manager::SessionInfo SessionInfo::to_external_api() {
+    std::lock_guard<std::mutex> lock(this->session_info_mutex);
```
djchhp: What is the state you are tryning to protect. Maybe a monitor from lib/util would do the job.

It looks like you are protecting the whole object. In that case removing all mutexes here and lock the session_info object in the API with a monitor should be cleaner

modules/API/evse_manager_consumer_API/session_info.cpp
```diff
@@ -0,0 +1,222 @@
...
+    everest::lib::API::V1_0::types::evse_manager::SessionInfo result;
+
+    auto charged_energy_wh = this->end_energy_import_wh - this->start_energy_import_wh;
+    int32_t discharged_energy_wh{0};
+    if ((this->start_energy_export_wh_was_set == true) && (this->end_energy_export_wh_was_set == true)) {
```
djchhp: == true is not necessary

djchhp (changes requested):

modules/API/evse_manager_consumer_API/evse_manager_consumer_API.cpp
Pietfried: I restructured SessionInfo and added it in separate handlers for the respective events

modules/API/evse_manager_consumer_API/evse_manager_consumer_API.hpp
Pietfried: I moved this into a callback of SessionInfo

modules/API/evse_manager_consumer_API/session_info.cpp
Pietfried: I used the monitor provided by lib/util

modules/API/evse_manager_consumer_API/evse_manager_consumer_API.hpp
```diff
@@ -148,6 +152,8 @@ class evse_manager_consumer_API : public Everest::ModuleBase {
     ev_API::Topics topics;
     ev_API::CommCheckHandler<generic_errorImplBase> comm_check;
     size_t hb_id{0};
+    everest::lib::util::monitor<std::unique_ptr<SessionInfo>> session_info;
```
djchhp: Letting monitor handle a std::unique_ptr is not necessary. It should be just
```
everest::lib::util::monitor<SessionInfo> session_info([](auto){});
```

making it a unique_ptr is essentially a workaround on monitor not having proper swap capabilities. I fixed this in [PR 1542](https://github.com/EVerest/everest-core/pull/1542) .

With swap in place your code later on becomes 

``` 
    everest::lib::util::monitor<SessionInfo> tmp{
        [this](const everest::lib::API::V1_0::types::evse_manager::SessionInfo& external) {
            auto topic = topics.everest_to_extern("session_info");
            try {
                auto&& payload = serialize(external);
                mqtt.publish(topic, payload);
                EVLOG_warning << "Variable: '" << topic << "' failed with -> " << e.what();
            } catch (...) {
                EVLOG_warning << "Invalid data: Cannot convert internal to external or serialize it.\n" << topic;
            }
        }
    };
    swap(tmp, session_info);
```

But I would argue that is also still a workaround on the design of `session_info`.

To fix this properly I suggest the following (which does not need the PR mentioned above)

1. Remove the callback as parameter of the constructor or give it a default value of `[](auto){}`.
2. Declare the monitor member as `everest::lib::util::monitor<SessionInfo> session_info`
3. Add a function  `set_callback` that can set the callback
4. grab a handle of the monitor in `generate_api_var_session_info` and call set_callback

modules/API/evse_manager_consumer_API/evse_manager_consumer_API.cpp
```diff
@@ -440,4 +441,43 @@ void evse_manager_consumer_API::subscribe_api_topic(std::string const& var,
...
+                EVLOG_warning << "Invalid data: Cannot convert internal to external or serialize it.\n" << topic;
+            }
+        }));
+
+    this->r_evse_manager->subscribe_session_event([this](types::evse_manager::SessionEvent const& session_event) {
+        try {
+            session_info.handle()->update_state(session_event);
+        } catch (const std::exception& e) {
+            EVLOG_warning << "Session event handling failed with -> " << e.what();
+        }
```
djchhp: The try/catch protects from failures of `session_info` member functions. `monitor::handle` essentially never fails or only in situation you cannot possible recover from anyway. This allows to move the exception handling into `session_info` and keep make the intention of the code cleaner.

djchhp (changes requested):

djchhp (approved):

PR #1527, opened 2025-11-17
Adding heartbeat counter to every API

djchhp:
## Describe your changes
This adds an increasing ID to the payload of all the hearbeat messages. It can be used to verify, that no messages have been lost.

### DIFF
--- a/doc/everest_api_specs/external_energy_limits_consumer_API/asyncapi.yaml
+++ b/doc/everest_api_specs/external_energy_limits_consumer_API/asyncapi.yaml
@@ -240,6 +240,13 @@ components:
       name: receive_heartbeat
       title: 'Receive heartbeat'
       summary: Heartbeat produced by EVerest as configured via cfg_heartbeat_interval_ms in the EVerest configuration
+      contentType: application/json
+      payload:
+        $ref: '#/components/schemas/HeartBeatId'
+      examples:
+        - summary: "Heartbeat"
+          payload: 42
+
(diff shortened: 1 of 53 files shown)


doc/everest_api_specs/external_energy_limits_consumer_API/asyncapi.yaml
```diff
@@ -240,6 +240,12 @@ components:
       name: receive_heartbeat
       title: 'Receive heartbeat'
       summary: Heartbeat produced by EVerest as configured via cfg_heartbeat_interval_ms in the EVerest configuration
```
cburandt, line 242:
```suggestion
      summary: Heartbeat produced by EVerest as configured via cfg_heartbeat_interval_ms in the EVerest configuration
      contentType: application/json
```
Not introduced by this PR, but still missing.

modules/API/ocpp_consumer_API/ocpp_consumer_API.hpp
cburandt: Also not introduced in this PR, but why is that public?

cburandt (dismissed):
Two things - since they were not introduced here, we can also fix them in a new PR.

modules/API/ocpp_consumer_API/ocpp_consumer_API.hpp
djchhp: For no good reason. It is private now.

PR #1528, opened 2025-11-17
Adds the ev termination code to the module

martinlukas84:
## Describe your changes
Adds the ev termination code to the module to be available on iso15118_charger interface.
This information can be used to display the EV error code in process view etc.

- [x ] I read the [contribution documentation](https://github.com/EVerest/EVerest/blob/main/CONTRIBUTING.md) and made sure that my changes meet its requirements


types/iso15118.yaml
```diff
@@ -1846,3 +1846,14 @@ types:
       target_reactive_power_L3:
         description: Target reactive power requested by the EVSE on L3 in W
         type: number
+  EvTermination:
+    description: Information concerning TLS certificates used by EV, as well as supported protocols.
```
mhei: Looks like copy&paste error

types/iso15118.yaml
```diff
@@ -1846,3 +1846,14 @@ types:
...
+    type: object
+    additionalProperties: false
+    properties:
+      ev_termination_code:
+        description: Defines the EV termination code
+        type: string
+      ev_termination_explanation :
```
mhei: Pls remove whitespace before :

types/iso15118.yaml
```diff
@@ -1846,3 +1846,14 @@ types:
...
+        description: Defines the EV termination explanation
```
mhei: maybe "contains" instead of "defines"? same for termination code above.

mhei (changes requested):

types/iso15118.yaml
martinlukas84: Thanks, will be updated...

types/iso15118.yaml
martinlukas84: Thanks, will be updated...

types/iso15118.yaml
martinlukas84: You are right, will be updated

mhei (approved):

SebaLukas (changes requested):
Various things:
- How is `ev_termination` used? The `EvseManager` should subscribe to the new var. Otherwise, there is no example here. Therefore, an explanation in the form of code or text would be great.
- In `dependincies.yaml`, the git hash of `libiso15118` must also be updated with the necessary changes. Otherwise, the build will always fail.
- Due to the changes to the type, the everest api must also be adapted: https://github.com/EVerest/everest-core/blob/main/lib/everest/everest_api_types/tests/expected_types_file_hashes.csv
Otherwise, a unit test will always fail. However, since I cannot find an Iso15118 api spec in everest_api, the adaptation for the unit test is probably sufficient.

SebaLukas:
> I can add a subscription to evse_manager, but this would be just an information to print to the log. But this also done in the libiso. Another usecase is not in my mind.

Then the question is, why do we need the changes in the interface if it is not being used?

SebaLukas:
> Regarding the git hash, is this not senseful to do it when the libiso is merged?

Usually we add the current git hash from the libs into the PR already. Without it the user need to checkout manually the lib itself and the ci is always failing

martinlukas84:
> > I can add a subscription to evse_manager, but this would be just an information to print to the log. But this also done in the libiso. Another usecase is not in my mind.
> 
> Then the question is, why do we need the changes in the interface if it is not being used?

But it's being used... In our usecase we have an Profinet module where we send the EVTerminationCode to the PLC for process view. This could also be a usecase for other implementation to display this information.

SebaLukas:
This information would be great in the description explaining why these changes have been made.
Anyway, updating the git hash is only missing right now

dependencies.yaml
```diff
@@ -44,10 +44,11 @@ libcbv2g:
   git: https://github.com/EVerest/libcbv2g.git
   git_tag: v0.3.1
   cmake_condition: "EVEREST_DEPENDENCY_ENABLED_LIBCBV2G"
+
```
SebaLukas: Remove this line

felix-ulonska:
**⚠️ PR Affected by Security Incident**

This PR was affected by a security incident where a compromised account force-pushed to this branch and closed the PR.

**✅ Branch History Restored**

We've run an automation that restored the complete history of this branch to the time before the attack. All your commits are preserved. Nothing has been updated in the respective branch in case someone force-pushed after the attack but before our automation.

**🔄 Action Required**

GitHub won't allow reopening this PR after the close and force-push. Please:
 - Open a new PR from your branch to the target branch. You can use the following link to open a new PR: [Open PR](https://github.com/EVerest/everest-core/compare/main...feature/ev_termination_code)
 - Link back to this PR in the description to include previous discussions

We apologize for the inconvenience.

PR #1549, opened 2025-11-21
feature: extend/improve the VAS interface

mhei:
## Describe your changes

Some (proprietary) VAS services expect the ServiceName to be set, so this PR extends the internal interface.
While at, let's fix some generic issues with the current implementation.

For SIL testing, I also modified the PyJosev module which requires this PR to be merged: https://github.com/EVerest/ext-switchev-iso15118/pull/52

### DIFF
--- a/modules/EV/PyEvJosev/manifest.yaml
+++ b/modules/EV/PyEvJosev/manifest.yaml
@@ -44,6 +44,14 @@ config:
     description: If true, the ev will ask for internet service
     type: boolean
     default: false
+  request_all_service_details:
+    description: If true, the ev will ask for details about all offered services
+    type: boolean
+    default: false
+  select_all_vas_services:
+    description: If true, the ev will select all offered services
+    type: boolean
+    default: false
   supported_d20_energy_services:
     description: >-
       The supported ISO15118-20 energy services (DC, DC_BPT, AC, AC_BPT) the EV supports,
@@ -51,7 +59,7 @@ config:
       and is the most likely to be selected. The services should be separated only with commas.
     type: string
     default: "DC,DC_BPT"
-provides:  
+provides:
   ev:
     interface: ISO15118_ev
     description: This module implements the ISO15118-2 implementation of an EV
--- a/modules/EV/PyEvJosev/module.py
+++ b/modules/EV/PyEvJosev/module.py
@@ -63,6 +63,8 @@ class PyEVJosevModule():
             enable_tls_1_3()
 
         self._es.internet_service_needed = self._setup.configs.module['is_internet_service_needed']
+        self._es.all_service_details = self._setup.configs.module['request_all_service_details']
+        self._es.all_vas_services = self._setup.configs.module['select_all_vas_services']
 
         # setup publishing callback
         def publish_callback(variable_name: str, value: any):
--- a/modules/EVSE/Evse15118D20/charger/ISO15118_chargerImpl.cpp
+++ b/modules/EVSE/Evse15118D20/charger/ISO15118_chargerImpl.cpp
@@ -1,6 +1,8 @@
 // SPDX-License-Identifier: Apache-2.0
 // Copyright Pionix GmbH and Contributors to EVerest
 #include "ISO15118_chargerImpl.hpp"
+#include <fmt/core.h>
+#include <fmt/ranges.h>
 
 #include "session_logger.hpp"
 #include "utils.hpp"
@@ -144,13 +146,16 @@ void ISO15118_chargerImpl::init() {
         this->mod->r_iso15118_vas[i]->subscribe_offered_vas(
             [this, i](const types::iso15118_vas::OfferedServices& offered_services) {
                 std::vector<uint16_t> service_ids;
+                std::scoped_lock lock(vas_mutex);
 
-                service_ids.reserve(offered_services.service_ids.size());
-                for (const auto& item : offered_services.service_ids) {
-                    service_ids.push_back(item);
+                service_ids.reserve(offered_services.services.size());
+                for (const auto& item : offered_services.services) {
+                    service_ids.push_back(item.service_id);
                 }
 
-                EVLOG_verbose << "Updated Supported VAS services for provider #" << i;
+                EVLOG_verbose << fmt::format("Updated Supported VAS services for provider #{} ({} service{})", i,
+                                             offered_services.services.size(),
+                                             offered_services.services.size() > 1 ? "s" : "");
                 supported_vas_services_per_provider[i] = service_ids;
 
                 // report to controller if it exists, also check for duplicate service ids
@@ -225,7 +230,10 @@ void ISO15118_chargerImpl::ready() {
 
     // if the vas providers report their supported vas services before the controller exists,
     // we need to update the controller with the supported vas services after instantiation
-    update_supported_vas_services();
+    {
+        std::scoped_lock lock(vas_mutex);
+        update_supported_vas_services();
+    }
 
     try {
         controller->loop();
@@ -252,13 +260,16 @@ void ISO15118_chargerImpl::update_supported_vas_services() {
     }
 
     if (this->controller) {
+        EVLOG_verbose << fmt::format("Updated controller VAS list: {}", fmt::join(supported_vas_services, ","));
         this->controller->update_supported_vas_services(supported_vas_services);
     } else {
         EVLOG_verbose << "Controller not initialized, skipping setting supported VAS services.";
     }
 }
 
-std::optional<size_t> ISO15118_chargerImpl::get_vas_provider_index(uint16_t service_id) const {
+std::optional<size_t> ISO15118_chargerImpl::get_vas_provider_index(uint16_t service_id) {
+    std::scoped_lock lock(vas_mutex);
+
     for (size_t i = 0; i < supported_vas_services_per_provider.size(); i++) {
         const auto& provider_services = supported_vas_services_per_provider[i];
         if (std::find(provider_services.begin(), provider_services.end(), service_id) != provider_services.end()) {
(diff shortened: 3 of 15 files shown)


modules/EV/PyEvJosev/manifest.yaml
```diff
@@ -44,14 +44,22 @@ config:
     description: If true, the ev will ask for internet service
     type: boolean
     default: false
+  request_all_service_details:
+    description: If true, the ev will ask for details about all offered services
+    type: boolean
+    default: false
+  select_all_vas_services:
+    description: If true, the ev will ask for details about all offered services
```
FaHaGit: I think you forgot to adapt the description

modules/EVSE/Evse15118D20/charger/ISO15118_chargerImpl.cpp
```diff
@@ -143,15 +144,11 @@ void ISO15118_chargerImpl::init() {
...
+                // remember all offered services
+                supported_vas_services_per_provider[i] = offered_services;
 
-                service_ids.reserve(offered_services.service_ids.size());
-                for (const auto& item : offered_services.service_ids) {
-                    service_ids.push_back(item);
-                }
-
-                EVLOG_verbose << "Updated Supported VAS services for provider #" << i;
-                supported_vas_services_per_provider[i] = service_ids;
```
barsnick, line 154: BTW: IMHO the previous implementation was missing a lock here, but now it seems no longer needed and you added proper locks everywhere.

modules/EVSE/Evse15118D20/charger/ISO15118_chargerImpl.cpp
```diff
@@ -235,19 +232,23 @@ void ISO15118_chargerImpl::ready() {
...
 
 void ISO15118_chargerImpl::update_supported_vas_services() {
+    std::scoped_lock lock(vas_mutex);
     iso15118::d20::SupportedVASs supported_vas_services;
+    vas_service_id_to_provider_map.clear();
 
-    for (const auto& provider_services : supported_vas_services_per_provider) {
-        for (const auto& service_id : provider_services) {
+    for (size_t idx = 0; idx < supported_vas_services_per_provider.size(); ++idx) {
+        for (auto& service : supported_vas_services_per_provider.at(idx).services) {
```
barsnick: Can this no longer be `const`? Because of the new `std::ref()` below?

modules/EVSE/EvseV2G/iso_server.cpp
```diff
@@ -721,34 +714,42 @@ static enum v2g_event handle_iso_service_detail(struct v2g_connection* conn) {
...
                                 out_parameter.intValue = parameter.value.int_value.value();
                                 out_parameter.intValue_isUsed = true;
                             } else if (parameter.value.finite_string.has_value()) {
-                                // TODO(SL): Check if string is too big to copy char array
                                 const auto& temp = parameter.value.finite_string.value();
-                                if (temp.length() > sizeof(out_parameter)) {
-                                    dlog(DLOG_LEVEL_WARNING, "Paramater String is too long to copy into char array");
+                                if (temp.length() > sizeof(out_parameter.stringValue.characters)) {
+                                    dlog(DLOG_LEVEL_WARNING,
+                                         "Parameter String is too long to copy into char array -> truncated");
```
barsnick: Feel free to print the offending parameter string. This helps in debugging.

modules/EVSE/EvseV2G/v2g_ctx.cpp
```diff
@@ -148,8 +147,10 @@ void v2g_ctx_init_charging_values(struct v2g_context* const ctx) {
     if (ctx->hlc_pause_active != true) {
         ctx->evse_v2g_data.charge_service.ServiceCategory = iso2_serviceCategoryType_EVCharging;
         ctx->evse_v2g_data.charge_service.ServiceID = (uint16_t)1;
-        memcpy(ctx->evse_v2g_data.charge_service.ServiceName.characters, init_service_name, sizeof(init_service_name));
-        ctx->evse_v2g_data.charge_service.ServiceName.charactersLen = sizeof(init_service_name);
+        strncpy(ctx->evse_v2g_data.charge_service.ServiceName.characters, init_service_name,
+                sizeof(ctx->evse_v2g_data.charge_service.ServiceName.characters));
```
barsnick: We might need to add a comment, so that Codacy makes an exception. (We know what we are doing with `strncpy()`.)

Codacy:
> Easily used incorrectly; doesn't always \0-terminate or check for invalid pointers [MS-banned] (CWE-120).

modules/EVSE/Evse15118D20/charger/ISO15118_chargerImpl.cpp
mhei: I did not really change the behavior at this position. Since the element/space was assigned before the actuall callback is called, the vector access is not racy here. Only when we iterate later in `update_supported_vas_service`. This is why I added a mutex there.

modules/EVSE/EvseV2G/v2g_ctx.cpp
mhei: Do you have an idea how this could look like?

barsnick: I thought I had seen use of this possibility somewhere in EVerest's code base already, but it turns out I am wrong.

This check seems to come from flawfinder, and there's no way to disable it.

Perhaps `memcpy()` is the better choice after all (just not the original code)?

modules/EV/PyEvJosev/module.py
SebaLukas, line 1: To use `all_service_details` and `all_vas_services` update `dependencies.yaml` with the git hash from your Josev branch

modules/EVSE/Evse15118D20/charger/ISO15118_chargerImpl.hpp
```diff
@@ -14,6 +14,9 @@
 
 // ev@75ac1216-19eb-4182-a85c-820f1fc2c091:v1
 #include <bitset>
+#include <functional>
```
SebaLukas: Why you need to include functional?

modules/EVSE/Evse15118D20/charger/ISO15118_chargerImpl.hpp
```diff
@@ -14,6 +14,9 @@
...
+#include <mutex>
+#include <utility>
```
SebaLukas: Why you need to include utility?

modules/EVSE/Evse15118D20/charger/ISO15118_chargerImpl.cpp
```diff
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 // Copyright Pionix GmbH and Contributors to EVerest
 #include "ISO15118_chargerImpl.hpp"
+#include <mutex>
```
SebaLukas: Not needed because you already include mutex in the header file

modules/EVSE/EvseV2G/v2g_ctx.cpp
mhei: `strncpy` is actually designed for exactly this use case, but its unfortunate misuse by many developers in recent years has given it a bad reputation. It's a shame we have to deal with these consequences now. Anyway, I'll rework this and stick with a fixed memcpy-based solution.

mhei: I added a new tools function `strncpy_to_v2g` and unified the usage.

barsnick (approved):

FaHaGit (approved):
LGTM

modules/EVSE/Evse15118D20/charger/ISO15118_chargerImpl.hpp
```diff
@@ -92,10 +95,13 @@ class ISO15118_chargerImpl : public ISO15118_chargerImplBase {
     iso15118::d20::EvseSetupConfig setup_config;
     std::bitset<NUMBER_OF_SETUP_STEPS> setup_steps_done{0};
 
-    std::vector<iso15118::d20::SupportedVASs> supported_vas_services_per_provider;
+    std::vector<types::iso15118_vas::OfferedServices> supported_vas_services_per_provider;
+    std::map<uint16_t, std::pair<std::size_t, std::reference_wrapper<types::iso15118_vas::OfferedService>>>
+        vas_service_id_to_provider_map;
```
SebaLukas: I don't quite understand why `supported_vas_services_per_provider` has been changed here. D20 only needs the ServiceIDs from the updated OfferedServices type. So why can't we continue to use the old vector?

I also don't understand the need for the new `vas_service_id_to_provider_map` variable.

mhei: The original idea was to store the complete offered service instance in case it might be used later. But you're right, this does not make sense for D20 since the data is not used at all. I'll revert to a minimal adaption.

mhei:
I've rebased and squashed some of the feedbacks.

modules/EVSE/Evse15118D20/charger/ISO15118_chargerImpl.hpp
```diff
@@ -93,9 +94,10 @@ class ISO15118_chargerImpl : public ISO15118_chargerImplBase {
     std::bitset<NUMBER_OF_SETUP_STEPS> setup_steps_done{0};
 
     std::vector<iso15118::d20::SupportedVASs> supported_vas_services_per_provider;
+    std::mutex vas_mutex;
 
     void update_supported_vas_services();
-    std::optional<size_t> get_vas_provider_index(uint16_t service_id) const;
+    std::optional<size_t> get_vas_provider_index(uint16_t service_id);
```
SebaLukas, line 100: Why no `const`?

mhei: This method must ensure that the list is not currently being modified and therefore needs to obtain the lock. But this "modifies" the object and thus the "const" cannot be used anymore.

PR #1565, opened 2025-11-27
While testing MCS feature an authentication after plugged-in failed

martinlukas84:
## Describe your changes
After the car is plugged in and the authentication was done, Everest goes to state "prepare_charging", while the car stucks at authentication. The reason is that the car is never notified about authentication finished by the EVSE.
After debugging, a timing problem between HLC and bsp cp_event was determinated. When car was plugged, method hlc[0]->subscribe_require_auth_eim() is called milliseconds earlier than the bsp cp_event notifies about plug in event to evsemanager. This leads to a reset of variable "hlc_waiting_for_auth_eim" and avoidance of calling r_hlc[0]->call_authorization_response(...::Accepted,...). In MCS no Slac is used. In addition, ChargeSom based products from chargebyte are using a safetyController to set and read CP states. There are latencies in the communication with EVerest. Assuming that the code in CPEvent::CarPluggedIn should set a clean state at the car plugged in event. We can't imaging any usecase where variable "hlc_waiting_for_auth_eim" should be true when car is not plugged. This is why we tending to call the code in cp_event "CarUnplugged" rather than in "CarPluggedIn".

Re-submission of https://github.com/EVerest/everest-core/pull/1514

### DIFF
--- a/modules/EVSE/EvseManager/EvseManager.cpp
+++ b/modules/EVSE/EvseManager/EvseManager.cpp
@@ -940,7 +940,7 @@ void EvseManager::ready() {
         // Forward some events to HLC
         if (hlc_enabled) {
             // Reset HLC auth waiting flags on new session
-            if (event == CPEvent::CarPluggedIn) {
+            if (event == CPEvent::CarUnplugged) {
                 r_hlc[0]->call_reset_error();
                 r_hlc[0]->call_ac_contactor_closed(false);
                 r_hlc[0]->call_stop_charging(false);


SebaLukas (commented):
I dont know if this is the best solution for your problem.
I can imagine that problems will arise if the reset is performed at the end of a session rather than at the beginning.
In the best case scenario, the reset will be performed in the `CarPluggedIn` & `CarUnplugged` event.

corneliusclaussen:
Btw do you have an implementation for the "slac" requirement for T1S? I think it should fix the ordering issue if implemented correctly, i.e. wait for the enter_bcd() / leave_bcd() commands and publish dlink_ready() etc accordingly.
Iso stack should only accept connections when dlink is ready, that may also need to be fixed.

PR #1579, opened 2025-11-27
feat: Dold RN5893 IMD Hardware Driver

markxoe:
## Describe your changes

- add Dold RN5893 IMD Hardware Driver
- add BringUp config for this driver

Original PR: #1395


modules/HardwareDrivers/IsolationMonitors/DoldRN5893/main/isolation_monitorImpl.cpp
```diff
@@ -0,0 +1,500 @@
...
+
+    // Note that we continue in this initial setup even if anything fails as the main loop reconfigures the device if
+    // necessary
+
+    const auto success = configure_device();
+    if (not success) {
+        EVLOG_error << "Failed to configure device";
+    } else {
+        EVLOG_info << "Device configured successfully";
+    }
```
barsnick, line 25: Shouldn't the module _not_ go into the main loop if it fails to configure?

Something like this suggestion:
```suggestion
    if (not success) {
        EVLOG_error << "Failed to configure device"; // FIXME probably obsolete
        throw std::runtime_error("Failed to configure device");
    }

    EVLOG_info << "Device configured successfully";
```
or:

```suggestion
    if (not success) {
        EVLOG_AND_THROW(Everest::EverestConfigError("Failed to configure device"));
    }

    EVLOG_info << "Device configured successfully";
```

modules/HardwareDrivers/IsolationMonitors/DoldRN5893/main/isolation_monitorImpl.cpp
```diff
@@ -0,0 +1,500 @@
...
+        // is handled here
+        if (self_test_triggered and not self_test_running and device_state == DeviceState_30002::SelfTesting) {
+            EVLOG_info << "Device has started selftesting";
+            self_test_running = true;
+        }
+
+        // Once the device has entered self test mode AND then left it again (either device state is reset or some fault
+        // is raised)
+        if (self_test_triggered and self_test_running and
+            (device_state != DeviceState_30002::SelfTesting or device_fault != DeviceFault_30001::NoFailure)) {
```
barsnick, line 93: Wouldn't this part always be true?
`(device_state != DeviceState_30002::SelfTesting or device_fault != DeviceFault_30001::NoFailure)`

modules/HardwareDrivers/IsolationMonitors/DoldRN5893/manifest.yaml
```diff
@@ -0,0 +1,123 @@
+description: DOLD RN5893 IMD driver module
+config:
+  device_id:
+    description: Modbus device address of DOLD RN5893, set via potentiometers on device
+    type: number
+
+  self_test_timeout_s:
+    description: Maximum time in seconds a self-test is allowed to take
+    type: number
```
barsnick: If you make this an `integer`, you wouldn't need to cast it to int in the code.

modules/HardwareDrivers/IsolationMonitors/DoldRN5893/manifest.yaml
```diff
@@ -0,0 +1,123 @@
...
+
+  timeout_release:
+    description: 40002 - Timeout release - whether the timeout on the IMD side should be enabled. This enables writing the Timeout register every cycle
+    type: boolean
+    default: false
+  timeout_s:
+    description: |
+      40003 - Timeout - the time in seconds after the device will trigger a fault if no communication is received.
+      This should not be smaller than 2 seconds, as this module writes the register every second
+    type: number
```
barsnick, line 35: If you make this an `integer`, you wouldn't need to cast it to int in the code.

modules/HardwareDrivers/IsolationMonitors/DoldRN5893/manifest.yaml
```diff
@@ -0,0 +1,123 @@
...
+    default: DE_ENERGIZED_ON_TRIP
+  power_supply_type:
+    description: 42004 - Power supply type - Type of connected power network. See datasheet for details
+    type: string
+    enum:
+      - AC
+      - DC
+      - 3NAC
+    default: DC
+  response_value_alarm_kohm:
```
barsnick, line 67: If you make this an `integer`, you wouldn't need to cast it to int in the code. (Same for most other `number` configs.)

modules/HardwareDrivers/IsolationMonitors/DoldRN5893/docs/index.rst
barsnick, line 1: Perhaps capitalize the file name like all others in the repo: README.md

modules/HardwareDrivers/IsolationMonitors/DoldRN5893/manifest.yaml
markxoe: Good catch, this works for almost all numbers here, will change

modules/HardwareDrivers/IsolationMonitors/DoldRN5893/manifest.yaml
markxoe: The timeout register uses milliseconds, see https://github.com/EVerest/everest-core/blob/175c853361b3dc6c05252f9fb0586bcfd438c210/modules/HardwareDrivers/IsolationMonitors/DoldRN5893/main/isolation_monitorImpl.cpp#L469
 I would avoid using milliseconds here as seconds make configurations more readable. Would this be okay or should I change the config to milliseconds instead?

modules/HardwareDrivers/IsolationMonitors/DoldRN5893/main/isolation_monitorImpl.cpp
markxoe: When a self test has started completely (i.e. triggered and running) the device state is `SelfTesting` and the device fault is `NoFault` (at least in a normal case) which results in `false`. This `if` is intended to run, when the self test is running and either the state has changed (i.e. `state != SelfTesting`. This also happens when the device is done self testing) or a fault was raised (i.e. `fault != NoFailure`). Note that a fault may change the fault register first, before the state register changes, as described here: https://github.com/EVerest/everest-core/blob/175c853361b3dc6c05252f9fb0586bcfd438c210/modules/HardwareDrivers/IsolationMonitors/DoldRN5893/readme.md?plain=1#L40 which is why the device fault check is included.
Am I missing something else?

(Technically `self_test_running` implies that `self_test_triggered` is true but IMO this is more readable this way)

modules/HardwareDrivers/IsolationMonitors/DoldRN5893/main/isolation_monitorImpl.cpp
markxoe: As the configuration is written whenever a communication fault happens this would be catched lated on (I.e. a communication fault is raised, which makes the EvseManager inoperable, and if the communication is established later, the configuration is written again).
I think this is useful in scenarios where linux + everest are faster to boot than the device or when the device is faulty  (if I return here, I think everest would boot-loop)

barsnick: Okay, got this, thanks.

modules/HardwareDrivers/IsolationMonitors/DoldRN5893/manifest.yaml
barsnick: Ah, perhaps I missed the use of this one. Of course, fractions of seconds are perfectly acceptable, and number/float/double can be used.

modules/HardwareDrivers/IsolationMonitors/DoldRN5893/main/isolation_monitorImpl.cpp
barsnick: I failed to read that this is `device_state` and `device_fault`, not twice the same var. Ignore my ignorance, sorry.

f43i4n:
Can someone of the code owners take a look at this?

modules/HardwareDrivers/IsolationMonitors/DoldRN5893/docs/index.rst
cburandt, line 1: If this content was in docs/index.rst, it would automatically be included into the autogenerated html documentation. Since it's user documentation it should really be part of that.

modules/HardwareDrivers/IsolationMonitors/DoldRN5893/README.md
```diff
@@ -0,0 +1,48 @@
...
+
+If the device reports *Communication Fault Modbus* in the *device fault* register, the driver will try to reset the device by writing `1` to the *control word 1* register. After that, the driver resets the control word 1 to the previous value.
+
+The driver will try to write the device reset command every cycle until the device fault is cleared.
+
+## Measurement and publishing modes
+
+This driver supports three modes of operation:
+- **Standard mode**: The device pauses measurements upon startup and when `stop()` is called. Measurements are started by calling `start()`
+- **Continous measurement mode**: The device continuously performs measurements, bit 8 of the "control word 1" register (address 40001) is not set at any time. This is useful if the device should always alarm on isolation faults, even when no EV is connected. Measurements are still only published when `start()` is called, until `stop()` is called
```
cburandt:
```suggestion
- **Continuous measurement mode**: The device continuously performs measurements, bit 8 of the "control word 1" register (address 40001) is not set at any time. This is useful if the device should always alarm on isolation faults, even when no EV is connected. Measurements are still only published when `start()` is called, until `stop()` is called
```

modules/HardwareDrivers/IsolationMonitors/DoldRN5893/README.md
```diff
@@ -0,0 +1,48 @@
...
+  - This mode is enabled by setting `keep_measurement_active` to `true`
+- **Always publish mode**: Like in continous measurement mode, the device continuously performs measurements. Additionally, all measurements are published, even when `stop()` is called. This is only useful in very specific scenarios, e.g. if a special module needs measurements all the time
+  - This mode is enabled by setting both `always_publish_measurements` and `keep_measurement_active` to `true`
+
+## Device instructions
+
+- For applications following UL 2231 the parameter `automatic_self_test` has to be disabled (i.e. set to `false`)
+- Changes to modbus registers triggered by modbus messages may have a short delay before they can be read back over the bus. The device's internal reaction is faster than what is reported via Modbus, however
+- The device state may not always report "Error" when a device fault is present (i.e. the device has an internal fault or the device fault register reports a fault), because of internal prioritization of faults. Because of this, we only check the device fault register to determine if a fault is present and report that to Everest
```
cburandt:
```suggestion
- The device state may not always report "Error" when a device fault is present (i.e. the device has an internal fault or the device fault register reports a fault), because of internal prioritization of faults. Because of this, we only check the device fault register to determine if a fault is present and report that to EVerest
```

modules/HardwareDrivers/IsolationMonitors/DoldRN5893/main/isolation_monitorImpl.cpp
```diff
@@ -0,0 +1,498 @@
...
+
+            if (not update_control_word1(ControlWord1Action::ResetFaults)) {
+                EVLOG_error << "Failed to reset device faults";
+                continue;
+            }
+
+            if (not update_control_word1()) {
+                EVLOG_error << "Failed to update control word 1";
+                continue;
+            };
```
cburandt:
```suggestion
            }
```

modules/HardwareDrivers/IsolationMonitors/DoldRN5893/main/isolation_monitorImpl.cpp
```diff
@@ -0,0 +1,498 @@
...
+        }
+
+        // If a communication fault was previously raised, we can clear it now, as we were able to read from the device.
+        // Also upload the configuration again, as the device might have reset/restarted
+        if (error_state_monitor->is_error_active("isolation_monitor/CommunicationFault", "")) {
+            // reconfigure device after communication fault, if this fails, keep the communication fault
+            if (not configure_device()) {
+                EVLOG_error << "Failed to reconfigure device after communication fault";
+                continue;
+            };
```
cburandt:
```suggestion
            }
```

modules/HardwareDrivers/IsolationMonitors/DoldRN5893/main/isolation_monitorImpl.cpp
```diff
@@ -0,0 +1,498 @@
...
+    } else if (mod->config.indicator_relay_k2_function == "INSULATION_FAULT_ON_DC-_OR_DEVICE_FAULT") {
+        indicator_relay_k2_function = (1U << 5);
+    } else {
+        EVLOG_error << "Invalid indicator relay K2 function configuration: " << mod->config.indicator_relay_k2_function;
+        return false;
+    }
+
+    new_settings[8] = indicator_relay_k1_function; // 2008
+    new_settings[9] = indicator_relay_k2_function; // 2009
+
```
cburandt, line 319: Looks like module configuration automatic_self_test is not included here(?)

markxoe: Thank you! fixed in e2d0ffa

modules/HardwareDrivers/IsolationMonitors/DoldRN5893/docs/index.rst
```diff
@@ -0,0 +1,60 @@
+.. _everest_modules_handwritten_DoldRN5893:
+
+********************************************************************************
+Dold RN5893 IMD with UL 2231 approval - especially for DC charging stations
+********************************************************************************
```
cburandt:
```suggestion
.. ********************************************************************************
.. Dold RN5893 IMD with UL 2231 approval - especially for DC charging stations
.. ********************************************************************************
```

cburandt: If you include this change, this file should perfectly blend into the autogenerated module page.

modules/HardwareDrivers/IsolationMonitors/DoldRN5893/docs/index.rst
markxoe: Good point, file has been converted to rst and relocated

modules/HardwareDrivers/IsolationMonitors/DoldRN5893/docs/index.rst
markxoe: No worries and thank you for reviewing!

PR #1580, opened 2025-11-27
EvseManager: also propagate no. of phases changed at runtime via limits var

barsnick:
This value in the limits var of the evse_manager interface was only being updated at init(), while the value in the limits of the energy interface was uptodate.

## Describe your changes
Propagate the change to the evse_manager interface when the internal value changes.
This is a new request after incorrect closure of https://github.com/EVerest/everest-core/pull/1533.


barsnick:
> Please note that we recently discussed removing `limits` from the `evse_manager` interface and plan to only support `enforced_limits` moving forward, because it is a more complex and therefore the more appropriate type for AC and DC.

Thanks for the reminder. This is good to know.

I will merge this PR, just because it does fix a bug, and rework RpcApi to use `enforced_limits`, which have apparently been available long enough. I will also check whether they properly propagate the number of phases "in use".

PR #1583, opened 2025-11-27
RpcApi: add new JSON RPC API module

barsnick:
## Describe your changes
This introduces a new API module RpcApi, which uses JSON RPC, and WebSockets as a transport layer.

It uses both RPC Methods and Notifications.

This is the implementation according to the proposal with was [discussed here on Zulip](https://lfenergy.zulipchat.com/#narrow/channel/455067-EVerest.3A-Energy-Mgmt/topic/.E2.9C.94.20external.20API.20module) and [documented here](https://chargebyte.atlassian.net/wiki/external/NDkzYjg5MzAxY2U3NDg0MmE0N2MyOGQwZTcyOWYxNTA). The original proposal was slightly modified, as several details proved inconsistent, impractical, or just mis-specified.

Documentation of the current state of implementation is included in the module's `docs/` subdirectory. The current (or the final merged) state will be manifested as API version 1.0.0. Future updates to the external API definition will require API version bumps.

A Python GUI test client is included under `modules/API/RpcApi/tools/python-client-gui/`.

Not implemented in the current state of the API:
- DC charging
- TLS
- API authentication

This will be added later, with corresponding bumps of the API version (where applicable).

A non-squashed version of this work is found in the branch [feature/json-rpc-api-nosquash](https://github.com/EVerest/everest-core/tree/feature/json-rpc-api-nosquash), for reference regarding small change steps, commit messages with motivations, and so on.

Co-authored-by: Fabian Hartung <fabian.hartung@chargebyte.com>
Co-authored-by: Moritz Barsnick <moritz.barsnick@chargebyte.com>

This is a new request after incorrect closure of https://github.com/EVerest/everest-core/pull/1324.


modules/API/RpcApi/RpcApi.cpp
```diff
@@ -0,0 +1,458 @@
...
+    hw_caps_data_new.min_phase_count_export = hwcaps.min_phase_count_export;
+    hw_caps_data_new.min_phase_count_import = hwcaps.min_phase_count_import;
+    hw_caps_data_new.phase_switch_during_charging = hwcaps.supports_changing_phases_during_charging;
+
+    // submit changes
+    hw_caps_data.set_data(hw_caps_data_new);
+}
+
+bool RpcApi::check_evse_mapping() {
+    // Interate over the configured EVSE mapping and configure the data store accordingly
```
cburandt:
```suggestion
    // Iterate over the configured EVSE mapping and configure the data store accordingly
```

modules/API/RpcApi/RpcApi.cpp
```diff
@@ -0,0 +1,458 @@
...
+            if (evse_manager->get_mapping().value().connector.has_value()) {
+                // Initialize connector index
+                connector.index = evse_manager->get_mapping().value().connector.value();
+                types::evse_manager::Evse evse = evse_manager->call_get_evse();
+                if (!evse.connectors.empty() && evse.connectors[0].type.has_value()) {
+                    try {
+                        connector.type = types::json_rpc_api::string_to_connector_type_enum(
+                            types::evse_manager::connector_type_enum_to_string(
+                                evse.connectors[0]
+                                    .type.value())); // evse.connectors[0].type; // use the first connector type
```
cburandt:
```suggestion
                                    .type.value()));  // use the first connector type
```

modules/API/RpcApi/RpcApi.cpp
```diff
@@ -0,0 +1,458 @@
...
+                    EVLOG_debug << "No connector type determined for connector index " << connector.index;
+                }
+                evse_data->evseinfo.set_available_connector(connector);
+                evse_data->evsestatus.set_active_connector_index(connector.index); // TODO: support multiple connectors
+            } else {
+                EVLOG_debug << "No connector index configured in the EVSE mapping, using default connector index "
+                            << connector.index;
+            }
+        } else {
+            // FIXME no mapping may be incorrect
```
cburandt: Seeing a "FIXME" raises the question if sth. relevant is missing here?
Not sure what the exact problem is. 
EVerest does not enforce mappings: So my question from a users perspective would be, if the RpcApi module does not work (in a meaningful way) if mappings are missing. In that case it should fail harder (maybe stop working and EVLOG_error to tell why).
If it can be used without the mappings set, then a EVLOG_warning could tell about the restrictions to be expected.

modules/API/RpcApi/docs/index.rst
```diff
@@ -0,0 +1,901 @@
+.. _everest_modules_handwritten_RpcApi:
+
+*******************************************
+RpcApi
+*******************************************
```
cburandt:
```suggestion
.. *******************************************
.. RpcApi
.. *******************************************
```

Adapt to new documentation.

modules/API/RpcApi/docs/index.rst
```diff
@@ -0,0 +1,901 @@
...
+* **EVSE** – relates to a specific EVSE (Electric Vehicle Supply Equipment)
+
+Methods
+~~~~~~~
+This includes all calls that can be executed by the client. Please note that if authentication is active,
+a token must be included in the request except for API.Hello. Otherwise, the request will be rejected.
+The following chapter headings represent the method name of the JSON-RPC protocol. The response and
+requests objects shown in the following chapters map the “params” value in the JSON-RPC object. If
+"params" is marked as "{}" it means that no parameters are required and the "params" object can be omitted.
+An example JSON-RPC request without "params" is shown in the :ref:`API.Hello request section <example-json-rpc-request-with-params>`
```
cburandt:
```suggestion
An example JSON-RPC request without "params" is shown in the :ref:`API.Hello request section <example-json-rpc-request-without-params>`
```
?

modules/API/RpcApi/types/README.md
```diff
@@ -0,0 +1,13 @@
...
+
+To change the JSON-RPC types, edit `json_rpc_api.yaml`.
+
+To generate and use an updated header in `modules/API/RpcApi/types/json_rpc_api/`:
+
+1. Copy `json_rpc_api.yaml` into the top-level `types/` directory.
+2. Start a build.
+3. The generated header appears at `build/generated/include/generated/types/json_rpc_api.hpp`.
+4. Move that header to `modules/API/RpcApi/types/json_rpc_api/`.
+5. Remove the build folder `build/generated` before your next compile.
```
cburandt:
```suggestion
5. Remove the copied `json_rpc_api.yaml` from the `/types` directory.
6. Remove the build folder `build/generated` before your next compile.
```

modules/API/API/API.hpp
```diff
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 // Copyright Pionix GmbH and Contributors to EVerest
-#ifndef API_HPP
-#define API_HPP
+#ifndef METHODS_API_HPP
+#define METHODS_API_HPP
```
mlitre: Is there a reason for this change? Additionally this seems to be a manual change rather than generated or did I miss something?

modules/API/RpcApi/data/GenericInfoStore.hpp
```diff
@@ -0,0 +1,82 @@
...
+            return this->dataobj;
+        } else {
+            return std::nullopt;
+        }
+    }
+    // set the data object. This method must be overridden if the data object is not a simple type
+    //   e.g. pointers
+    // Note: all setters, also in derived classes, must use the data mutex
+    //   e.g. std::unique_lock<std::mutex> data_lock(this->data_mutex)
+    void set_data(const T& in) {
```
mlitre: Shouldn't the method be marked as virtual then if you want to allow the method to be overridden

modules/API/RpcApi/data/DataStore.cpp
```diff
@@ -0,0 +1,398 @@
...
+
+// we currently don't get this info from the system yet, so allow setting to unknown
+void ChargerInfoStore::set_unknown() {
+    this->dataobj.vendor = "unknown";
+    this->dataobj.model = "unknown";
+    this->dataobj.serial = "unknown";
+    this->dataobj.firmware_version = "unknown";
+    // pretend we got something
+    this->data_is_valid = true;
+}
```
mlitre, line 38: Shouldn't this be wrapped in a mutex? I know it is to fill with "garbage data", but it is mentioned to make sure that the mutex is always used when accessing dataobj

modules/API/RpcApi/data/DataStore.cpp
```diff
@@ -0,0 +1,398 @@
...
+void EVSEStatusStore::update_data_is_valid() {
+    if (this->data_is_valid) {
+        return; // No need to update if data is already valid
+    }
+
+    // Check if all fields are set
+    for (const auto& [field, is_set] : field_status) {
+        if (!is_set) {
+            this->data_is_valid = false;
+            EVLOG_debug << "EVSEStatusStore: Field " << static_cast<int>(field) << " is not set, data is invalid.";
```
mlitre: It would probably be a good idea for long term maintenance and readability to have a dedicated helper to string function here instead of just casting the field to an int

modules/API/RpcApi/data/DataStore.cpp
```diff
@@ -0,0 +1,398 @@
...
+}
+
+std::optional<RPCDataTypes::ACChargeParametersObj>& EVSEStatusStore::get_ac_charge_param() {
+    std::unique_lock<std::mutex> data_lock(this->data_mutex);
+    return this->dataobj.ac_charge_param;
+}
+
+std::optional<RPCDataTypes::DCChargeParametersObj>& EVSEStatusStore::get_dc_charge_param() {
+    std::unique_lock<std::mutex> data_lock(this->data_mutex);
+    return this->dataobj.dc_charge_param;
```
mlitre: Is it intentional to return a reference here? If so is it not possible that you can now get concurrent accesses on dataobj.xx_charge_param as there is no more locking mechanism outside of this method?

modules/API/RpcApi/data/SessionInfo.cpp
```diff
@@ -0,0 +1,205 @@
...
+    case Event::PluginTimeout:
+        this->state = State::AuthTimeout;
+        break;
+    case Event::ReservationStart:
+        this->state = State::Reserved;
+        break;
+    case Event::ReservationEnd:
+        [[fallthrough]];
+    case Event::SessionFinished:
+        this->state = State::Unplugged;
```
mlitre, line 111: Is this always true? Reservation ends when the ID is swiped or authorized, you could be plugged in at that time.
Here is SessionFinished the EVerest transaction events (aka the actual plugout or can it also just be the end of HLC?)

modules/API/RpcApi/data/SessionInfo.cpp
```diff
@@ -0,0 +1,205 @@
...
+                                                 const int active_priority) {
+    std::lock_guard<std::mutex> lock(this->session_info_mutex);
+    this->active_enable_disable_source = active_source;
+    this->active_enable_disable_state = active_state;
+    this->active_enable_disable_priority = active_priority;
+}
+
+static void to_json(json& j, const SessionInfoStore::Error& e) {
+    j = json{{"type", e.type}, {"description", e.description}, {"severity", e.severity}};
+}
```
mlitre: nit would be to have this at the start of the file, instead of in the middle of a class' methods

modules/API/RpcApi/data/SessionInfo.cpp
```diff
@@ -0,0 +1,205 @@
...
+
+float SessionInfoStore::get_charged_energy_wh() const {
+    return end_energy_import_wh - start_energy_import_wh;
```
mlitre: Isn't a mutex required here?

modules/API/RpcApi/helpers/LimitDecimalPlaces.hpp
```diff
@@ -0,0 +1,13 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright chargebyte GmbH and Contributors to EVerest
+#ifndef LIMIT_DECIMAL_PLACES_HPP
+#define LIMIT_DECIMAL_PLACES_HPP
+
+#include <nlohmann/json.hpp>
+
+namespace helpers {
+double formatAndRoundDouble(double value, int precision = 3);
+void roundFloatsInJson(nlohmann::json& j, int precision = 3);
```
mlitre: nit: Could we harmonize the naming convention here

modules/API/RpcApi/rpc/methods/Api.hpp
```diff
@@ -0,0 +1,48 @@
...
+    RPCDataTypes::HelloResObj hello();
+    void set_authentication_required(bool required);
+    bool is_authentication_required() const;
+    void set_api_version(const std::string& version);
+    const std::string& get_api_version() const;
+    void set_authenticated(bool authenticated);
+    bool is_authenticated() const;
+
+private:
+    DataStoreCharger& m_dataobj;
```
mlitre, line 39: Is there a specific reason why a reference to the DataStoreCharger was used instead of a shared pointer or even only having the Api object hold the object itself? It seems like access to this object might not be properly handled

modules/API/RpcApi/rpc/methods/ChargePoint.hpp
```diff
@@ -0,0 +1,37 @@
...
+    explicit ChargePoint(DataStoreCharger& dataobj) : m_dataobj(dataobj){};
+
+    ~ChargePoint() = default;
+
+    // Methods
+    RPCDataTypes::ChargePointGetEVSEInfosResObj getEVSEInfos();
+    RPCDataTypes::ChargePointGetActiveErrorsResObj getActiveErrors();
+
+private:
+    DataStoreCharger& m_dataobj;
```
mlitre, line 32: If this data store is shared shouldn't there be a better way of sharing it between classes rather than just a reference?

modules/API/RpcApi/rpc/methods/Evse.cpp
```diff
@@ -0,0 +1,202 @@
...
+        res.error = RPCDataTypes::ResponseErrorEnum::ErrorInvalidEVSEIndex;
+        return res;
+    }
+
+    if (!evse->evseinfo.get_data().has_value()) {
+        res.error = RPCDataTypes::ResponseErrorEnum::ErrorNoDataAvailable;
+        return res;
+    }
+
+    res.info = evse->evseinfo.get_data().value();
```
mlitre: nit: you can probably just copy the data once to avoid locking and unlocking mutexes all the time. This comment can be applied to the whole PR

modules/API/RpcApi/rpc/notifications/ChargePoint.hpp
```diff
@@ -0,0 +1,40 @@
...
+    // Constructor and Destructor
+    // Deleting the default constructor to ensure the class is always initialized with a DataStoreCharger object
+    ChargePoint() = delete;
+    // This needs to take a copy of rpc_server for reference counting, not a reference to it
+    ChargePoint(std::shared_ptr<rpc::JsonRpc2ServerWithClient> rpc_server, data::DataStoreCharger& dataobj,
+                int precision = 3);
+    ~ChargePoint() = default;
+
+    // Notifications
+    void sendActiveErrorsChanged(const std::vector<RPCDataTypes::ErrorObj>& active_errors);
```
mlitre: nit: apply to all, let's keep naming the same

modules/API/RpcApi/rpc/RpcHandler.hpp
```diff
@@ -0,0 +1,140 @@
...
+class ClientConnector : public jsonrpccxx::IClientConnector {
+public:
+    explicit ClientConnector(std::vector<std::shared_ptr<TransportInterface>>& interfaces,
+                             std::unordered_map<TransportInterface::ClientId, bool>& api_hello_received) :
+        hello_received(api_hello_received), transport_interfaces(interfaces) {
+    }
+    std::string Send(const std::string& notification) override {
+        const std::vector<uint8_t> notif_char_array{notification.begin(), notification.end()};
+        for (const auto& interface : transport_interfaces) {
+            // FIXME: hello_received changes with client connects and disconnects, so requires a mutex
```
mlitre: This seems important to fix

modules/API/RpcApi/tests/helpers/JsonRpcUtils.hpp
```diff
@@ -0,0 +1,58 @@
...
+#define JSON_RPC_UTILS_HPP
+
+#include "../../helpers/LimitDecimalPlaces.hpp"
+#include <nlohmann/json.hpp>
+
+namespace json_rpc_utils {
+
+inline nlohmann::json create_json_rpc_request(const std::string& method, const nlohmann::json& params, int id) {
+    nlohmann::json request;
+    request["jsonrpc"] = "2.0";
```
mlitre: For future maintainability, it might be better to set this as a constant and all methods use that constant to avoid having to change the value everywhere

modules/API/RpcApi/RpcApi.cpp
```diff
@@ -0,0 +1,458 @@
...
+        auto& evse_data = this->data.evses.emplace_back(std::make_unique<data::DataStoreEvse>());
+        // subscribe to evse_manager interface variables
+        this->subscribe_evse_manager(evse_manager, *evse_data);
+    }
+
+    if (r_evse_energy_sink.empty()) {
+        EVLOG_warning << "No EVSE energy sinks configured. Configuration of EVSE external limits will not be possible.";
+    }
+
+    // Check if all evse_energy_sink(s) have a mapping to an EVSE
```
mlitre: This comment seems wrong as there is no checking of the energy sinks only evse manager

modules/API/RpcApi/docs/index.rst
barsnick: This comments this part out. I think we intended to put a title on top, so probably
```
======
RpcApi
======
```
is correct?

modules/API/RpcApi/docs/index.rst
barsnick: Wow, no idea how this happened. Likely copy/waste.

modules/API/RpcApi/docs/index.rst
cburandt: As a matter of fact it's supposed to be commented out. You can use the suggestion - the new documentation includes the handwritten module description into another file (the autogenerated reference page for the same module) and therefore the headlines would clash. Leaving it there but commenting it out makes the source file look a bit nicer for readers who do not use the html docs but read the rst files directly.

modules/API/API/API.hpp
barsnick: Looks like some peculiar merge error. No intention to touch the API module!

modules/API/RpcApi/data/DataStore.cpp
barsnick: Correct, this function is missing the initial lock.

modules/API/RpcApi/data/DataStore.cpp
barsnick: This was more for debugability ("why isn't the EVSE Status ready?", which happened due to buggy initializations), and creating an extra string to enum mapping seemed overkill. I was thinking of using a `std::bitset` anyway, will see if I can remodel this.

Oh, and it was missing a lock on the mutex as well.

modules/API/RpcApi/rpc/methods/Evse.cpp
barsnick: I totally missed that checking an optional existence as return value and then getting the value calls the function (and its lock) twice. Good catch.

Strictly speaking, this could also be a race condition, leading to a crash.

modules/API/RpcApi/tests/helpers/JsonRpcUtils.hpp
barsnick: We can't just change the JSON RPC API version. But anyway, yeah, magic numbers and strings deserve a constant.

modules/API/RpcApi/data/GenericInfoStore.hpp
barsnick: Good catch. We didn't use it because we do have the special (generated) assignment operators for the complex types. Will fix.

modules/API/RpcApi/data/SessionInfo.cpp
barsnick: Copied from the original API module, but I agree.

modules/API/RpcApi/RpcApi.cpp
barsnick: The API is usable, as it will still present status values from the charger. You cannot set limits though. (This is the same restriction as in the API module.)

On attempts to access something which cannot, a warning is emitted at run-time. Also, the JSON-RPC response has a meaningful error response (unlike the API module).

I will expand the warning here to clarify this.

modules/API/RpcApi/data/SessionInfo.cpp
barsnick: This is one of the riddles of trying to construct a charger "state" out of the `types::evse_manager::SessionEventEnum` vents. I have done this in other (proprietary) modules previously. It's dangerous.

Is there a more clear interface to the charger state? I would prefer if EvseManager owned the state (and we could filter it down if there are too many to be useful) and provided it via its interface.

To answer your question to this particular case here: You are probably right. We did a lean effort to map things down (and new events keep getting added). Should we rather fall through to the block `/// explicitly fall through all the SessionEventEnum values we are not handling`?

modules/API/RpcApi/docs/index.rst
```diff
@@ -0,0 +1,901 @@
...
+     - Description
+   * - 1.0.0
+     - Initial version of the RpcApi module
+
+Table of Contents
+=====================
+
+.. contents:: Table of Contents
+   :depth: 3
+   :local:
```
cburandt: The rendered html will automatically include a toc of the page in a sidebar (e.g. see https://everest.github.io/nightly/reference/modules/EVSE/EvseManager/autogenerated.html)

modules/API/RpcApi/docs/index.rst
```diff
@@ -0,0 +1,901 @@
...
+     "evse_index": "int"
+   }
+
+.. _example-json-rpc-request-with-params:
+
+**Example JSON RPC Request:**
+
+.. code-block:: json
+
+{"jsonrpc": "2.0", "method": "EVSE.GetInfo", "id": 1, "params": {"evse_index": 1}}
```
cburandt:
```suggestion
   {"jsonrpc": "2.0", "method": "EVSE.GetInfo", "id": 1, "params": {"evse_index": 1}}
```

modules/API/RpcApi/docs/index.rst
```diff
@@ -0,0 +1,901 @@
...
+"params" is marked as "{}" it means that no parameters are required and the "params" object can be omitted.
+An example JSON-RPC request without "params" is shown in the :ref:`API.Hello request section <example-json-rpc-request-without-params>`
+and an example JSON-RPC request with "params" is shown in the :ref:`EVSE.GetInfo request section <example-json-rpc-request-with-params>`.
+
+⚠️ Note: Configuring charging parameters (such as charging current limits) via the API interface cannot
+always be guaranteed, since these values may also be influenced by other sources within EVerest. For example,
+if EVerest has been configured with an upper limit of 12A, this value cannot be exceeded via the API interface.
+In this specific case, if a client attempts to configure 16A through the API, the applied value will
+still be limited to 12A. Therefore, it is important to always observe the configured values returned
+via the EVSE.StatusChanged notification.
```
cburandt: Consider using rst admonitions (see https://documatt.com/restructuredtext-reference/admonitions.html) to render hints, warnings, etc. as nice boxes (see https://everest.github.io/nightly/how-to-guides/getting-started/get-started-sw.html to see how they are rendered in the theme used for the everest documentation)

modules/API/RpcApi/docs/index.rst
```diff
@@ -0,0 +1,901 @@
...
+The response message contains basic information from the EVerest ChargePoint, such as the API version
+in use, to enable further communication.
+While the Hello call does not necessarily require a token, it may be called with one. If a token is
+provided, the server verifies it, and the reply includes information about the token’s validity as well
+as the associated user and permissions.
+
+In case authentication is required, the optional parameter permission_scopes can be used to indicate
+the permissions (e.g., read/write access) the client has when using the given token.
+
+⚠️ Note: The fields authenticated, permission_scopes and everest_version are currently not supported.
```
cburandt:
```suggestion
.. warning::

   The fields authenticated, permission_scopes and everest_version are currently not supported.
```

Maybe use rst admonition to make this a nice box in html output.

modules/API/RpcApi/manifest.yaml
```diff
@@ -0,0 +1,54 @@
...
+    default: false
+  max_decimal_places_other:
+    description: Maximum number of decimal places for all floating point values. Ignored if value is 0.
+    type: integer
+    default: 2
+    minimum: 0
+provides:
+  main:
+    description: EVerest JSON-RPC API
+    interface: empty
```
cburandt: This is no longer required (introduced recently by https://github.com/EVerest/everest-core/commit/1167784e4c1c29c9a2914fee3fabf8290f944f80)

Associated cpp/hpp can also be removed.

modules/API/RpcApi/cmake/Findjson-rpc-cxx.cmake
```diff
@@ -0,0 +1,101 @@
...
+  The directory containing ``jsonrpccxx/common.hpp``.
+
+#]=======================================================================]
+
+# Find nlohmann_json dependency first
+find_package(nlohmann_json QUIET)
+if(NOT nlohmann_json_FOUND)
+    # Try to find nlohmann/json with alternative names
+    find_package(nlohmann-json QUIET)
+endif()
```
cburandt, line 45: What if nlohmann-json wasn't found at all?

modules/API/RpcApi/cmake/Findjson-rpc-cxx.cmake
```diff
@@ -0,0 +1,101 @@
...
+    PATHS
+        ${CMAKE_PREFIX_PATH}/include
+        /usr/local/include
+        /usr/include
+        ${json-rpc-cxx_ROOT}/include
+        $ENV{json-rpc-cxx_ROOT}/include
+    PATH_SUFFIXES
+        jsonrpccxx
+    DOC "json-rpc-cxx include directory"
+)
```
cburandt, line 69: Did you try to not specify any PATHs and did it fail then? I think CMake should be looking in all these places anyways, so it should be o.k. to not specify PATHS here. (https://cmake.org/cmake/help/v3.16/command/find_path.html)
(also, not sure how these hard-coded paths will behave when trying to cross-compile)

PATH_SUFFIXES should not be required: find_path is looking for NAME "jsonrpccxx/common.hpp" anyways and is likely not meant to search for "jsonrpccxx/jsonrpccxx/common.hpp" as well.

modules/API/RpcApi/cmake/Findjson-rpc-cxx.cmake
```diff
@@ -0,0 +1,101 @@
...
+
+include(FindPackageHandleStandardArgs)
+find_package_handle_standard_args(json-rpc-cxx
+    FOUND_VAR json-rpc-cxx_FOUND
+    REQUIRED_VARS
+        json-rpc-cxx_INCLUDE_DIR
+)
```
cburandt, line 77: Require nlohmann to be found?

modules/API/RpcApi/CMakeLists.txt
```diff
@@ -0,0 +1,81 @@
...
+        "rpc/notifications/Evse.cpp"
+        "server/TransportInterface.cpp"
+        "server/WebsocketServer.cpp"
+        "RpcApiRequestHandler.cpp"
+)
+
+target_compile_options(${MODULE_NAME}
+    PRIVATE
+        -Wimplicit-fallthrough
+#        -Werror=switch-enum
```
cburandt:
```suggestion
        -Werror=switch-enum
```
Turn this on by default?

cburandt (commented):
Some comments regarding the CMake files

modules/API/RpcApi/manifest.yaml
barsnick: Will remove after merging `main` (later today).

modules/API/RpcApi/data/DataStore.cpp
barsnick: That's true, good catch! That's exactly why we have a locking `set_*()` pendant. (These are currently unused functions, though.) Will fix.

modules/API/RpcApi/rpc/methods/Api.hpp
barsnick: Yes. This seemed like the most obvious implementation: RpcApi "owns" the data object globally, and RequestHandler/methods/notifications have access during its lifetime ("observers", "users"). RpcApi outlives them all - strictly speaking, none of them ever get destructed.

Would you agree? Or do you see a point in converting to shared pointers?

modules/API/RpcApi/rpc/methods/ChargePoint.hpp
barsnick: As mentioned in the other comment, this is owned and shared by RpcApi, and used here.

modules/API/RpcApi/data/SessionInfo.cpp
mlitre: This would probably be a good feature to add, since events don't (and shouldn't) give state

modules/API/RpcApi/rpc/methods/Api.hpp
mlitre: I see your point, that makes sense. 
Is there protection for concurrent access though on this specific object?

modules/API/RpcApi/CMakeLists.txt
barsnick: Took some fiddling, but I managed to make it work.

modules/API/RpcApi/cmake/Findjson-rpc-cxx.cmake
barsnick: Thanks for these suggestions.

I'm no CMake expert, and examples for these cases were a bit hard to find.

I tested with and without my fix in a minimum bare environment, e.g. with json-rpc-cxx and with/without nlohmann-json, and it now behaves as desired, while it was indeed broken before. Should be okay now.

modules/API/RpcApi/rpc/methods/Api.hpp
barsnick: All accesses into the shared DataStoreCharger are protected, and no sequential accesses exist which need to be grouped under a lock.

Additionally, the transport instances which represent the clients are handled in a single threaded manner (which is of course not a condition which can be expected to be met when designing these data classes).

modules/API/RpcApi/data/SessionInfo.cpp
barsnick: I would rather push this to a separate ticket and discussion. And there's already a lot of refactoring going on inside EvseManager, this would conflict with that work inside Charger.cpp.

At this point, I just need a good replacement mapping here. **I need some advice**, or I will try to come up with something more useful - like ignoring this event.

BTW, here's where we "borrowed" the bug:
https://github.com/EVerest/everest-core/blob/2026.02.0-rc1/modules/API/API/API.cpp#L166-L169

barsnick: Additionally reservation events do not correlate with the session:
https://github.com/EVerest/everest-core/blob/2026.02.0-rc1/modules/EVSE/EvseManager/evse/evse_managerImpl.cpp#L89-L93

We could just ignore them, but then wouldn't be able to display a reserved state. At this point, I'd probably leave the implementation as it is right now.

mlitre: Probably a good idea to leave as is then

barsnick:
@pietfried @corneliusclaussen We need a kind approval from (one of) you, because we modified the `CODEOWNERS` file:
```
/.github/ @pietfried @corneliusclaussen
```

Pietfried (approved):
Approved based on review from @cburandt and @mlitre Please squash your commits before merging

PR #1587, opened 2025-11-28
fix(EvseV2G): Set Responsecode to FAULT if auth times out

corneliusclaussen:
## Describe your changes

If auth is revoked e.g. due to a plugin timeout, set the ISO-2/DIN response code to FAILED instead of OK.

### DIFF
--- a/modules/EVSE/EvseV2G/iso_server.cpp
+++ b/modules/EVSE/EvseV2G/iso_server.cpp
@@ -1187,6 +1187,8 @@ static enum v2g_event handle_iso_authorization(struct v2g_connection* conn) {
         } else {
             res->ResponseCode = iso2_responseCodeType_FAILED;
         }
+    } else if (conn->ctx->session.authorization_rejected == true) {
+        res->ResponseCode = iso2_responseCodeType_FAILED;
     }
 
 error_out:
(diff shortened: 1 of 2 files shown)


modules/EVSE/EvseV2G/iso_server.cpp
```diff
@@ -1187,6 +1187,8 @@ static enum v2g_event handle_iso_authorization(struct v2g_connection* conn) {
         } else {
             res->ResponseCode = iso2_responseCodeType_FAILED;
         }
+    } else if (conn->ctx->session.authorization_rejected == true) {
```
FaHaGit, line 1190: The handling of the "authorization_rejected" parameter is implemented slightly differently between DIN and ISO, if I understand correctly. In ISO, "authorization_rejected" is only evaluated if processing has been set to "finished". In DIN it is independed of the "evse_processing". Is that how it's intended to work? It will probably not make any difference, as the "evse_processing" is configured together with “authorization_rejected.”

corneliusclaussen: It should not make a difference here, the purpose is to set both finished and failed as response code at the same time so that the EV knows auth will not happen anymore in this session.

PR #1597, opened 2025-12-02
EvseV2G: Wait for peer FIN before closing TCP socket

FaHaGit:
## Describe your changes

- Send a half-close (SHUT_WR) instead of full SHUT_RDWR so the peer can still deliver its FIN.
- Poll the socket for up to 3 seconds and drain any remaining data/FIN before closing, reducing premature connection tear-down.

**How it was already tested:**

1. Established a TCP session with "socat -6 - TCP6:[IPv6 %eth0]:PORT,forever"
(Allowed the connection to idle until the timeout expired.)
3. Observed logs (Wireshark + EVerest) showing the client FIN after ~500 ms and the code waiting the same ~500 ms in wait_for_peer_close() before completing shutdown. 

Here is a EVerest log with additional debugging log messages for and after wait_for_peer_close() call with timestamps:

```shell
2025-12-02 08:29:56.271759 [INFO] iso15118_charge  :: v2g_handle_connection exited with -1
2025-12-02 08:29:56.271764 [INFO] iso15118_charge  :: Closing TCP connection
2025-12-02 08:29:58.272380 [INFO] iso15118_charge  :: Before wait_for_peer_close()
2025-12-02 08:29:58.773837 [INFO] iso15118_charge  :: After wait_for_peer_close()
2025-12-02 08:29:58.773922 [INFO] iso15118_charge  :: TCP connection closed gracefully
``` 

pcap trace:
grafik

### DIFF
--- a/modules/EVSE/IsoMux/connection/connection.cpp
+++ b/modules/EVSE/IsoMux/connection/connection.cpp
@@ -34,6 +34,7 @@
 #define DEFAULT_TCP_PORT              61342
 #define DEFAULT_TLS_PORT              64110
 #define ERROR_SESSION_ALREADY_STARTED 2
+#define CLIENT_FIN_TIMEOUT            3000
 
 /*!
  * \brief connection_create_socket This function creates a tcp/tls socket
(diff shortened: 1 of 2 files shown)


FaHaGit:
Regarding the sleeps while terminating the tcp connection: what is the two second sleep before shutdown() call for?
I would prefer to remove that as well..

SebaLukas:
> In this case, we should add an explaining comment. Do you know the specific car or at least the brands?

I dont know the cars anymore. We introduced this change in an US Testival 2023. After this changes we didnt had any issues anymore with the tcp shutdown.

barsnick:
> Some cars did not like the immediate shutdown. Therefore a 2 second sleep was built in. I would prefer to leave the sleep in there.

So, before any sleeps were introduced, was it so that EvseV2G first sent a FIN (via `shutdown()`), and then mostly immediately an RST (on `close()`), because the TCP connection was still alive?

If so, the changes should resolve this, shouldn't they? The new code sends a FIN, waits for the EV's FIN-ACK and for the EV's FIN, thereby completing the four-way handshake. An RST shouldn't even appear anymore, the EV wouldn't the see final `close()` at all.

> In this case, we should add an explaining comment.

I second this.

modules/EVSE/IsoMux/connection/connection.cpp
```diff
@@ -388,6 +388,7 @@ void* connection_handle_tcp(void* data) {
     /* tear down connection gracefully */
     dlog(DLOG_LEVEL_INFO, "Multiplexer: Closing TCP connection");
 
+    /* some EV's did not like the immediate shutdown. Therefore we sleep for 2 seconds */
```
lategoodbye, line 391: The commit log is about EvseV2G, but you changed the IsoMux. Is there possibly something missing?

FaHaGit: Thanks. Yes, I accidentally adapted the wrong module. Now it should be correct.

PR #1598, opened 2025-12-02
Refactor Event Queue for OCPP1.6

Pietfried:
## Describe your changes
refactor(OCPP): Added dedicated type for Event Queue in OCPP module to allow easier enhancements with new event types

### DIFF
--- a/modules/EVSE/OCPP/OCPP.cpp
+++ b/modules/EVSE/OCPP/OCPP.cpp
@@ -482,8 +482,7 @@ void OCPP::init() {
             const auto error_info = get_error_info(error);
             this->charge_point->on_error(evse_id, error_info);
         } else {
-            ErrorRaised error_raised{error};
-            this->event_queue.emplace(evse_id, error_raised);
+            this->event_queue.emplace(evse_id, ErrorRaised{error});
         }
     };
 
@@ -493,8 +492,7 @@ void OCPP::init() {
         if (this->started) {
             this->charge_point->on_error_cleared(evse_id, error.uuid.uuid);
         } else {
-            ErrorCleared error_cleared{error};
-            this->event_queue.emplace(evse_id, error_cleared);
+            this->event_queue.emplace(evse_id, ErrorCleared{error});
         }
     };
 


modules/EVSE/OCPP/OCPP.cpp
```diff
@@ -325,8 +325,7 @@ void OCPP::init_evse_subscriptions() {
             if (!this->started) {
                 EVLOG_info << "OCPP not fully initialized, but received a session event on evse_id: " << evse_id
                            << " that will be queued up: " << session_event.event;
-                std::scoped_lock lock(this->session_event_mutex);
-                this->event_queue[evse_id].push(session_event);
+                this->event_queue.push(Event(evse_id, EventType::SessionEvent, session_event));
```
james-ctc: should be `event_queue.emplace({evse_id, EventType::SessionEvent, session_event});`

modules/EVSE/OCPP/OCPP.cpp
```diff
@@ -471,21 +470,20 @@ void OCPP::init() {
 
     const auto error_handler = [this](const Everest::error::Error& error) {
         const auto evse_id = error.origin.mapping.has_value() ? error.origin.mapping.value().evse : 0;
-        const auto error_info = get_error_info(error);
         if (this->started) {
+            const auto error_info = get_error_info(error);
             this->charge_point->on_error(evse_id, error_info);
         } else {
-            this->event_queue[evse_id].push(error_info);
+            this->event_queue.push(Event(evse_id, EventType::ErrorRaised, error));
```
james-ctc: should be `event_queue.emplace({evse_id, EventType::ErrorRaised, error});`

modules/EVSE/OCPP/OCPP.cpp
```diff
@@ -471,21 +470,20 @@ void OCPP::init() {
...
     };
 
     const auto error_cleared_handler = [this](const Everest::error::Error& error) {
         const auto evse_id = error.origin.mapping.has_value() ? error.origin.mapping.value().evse : 0;
-
         if (this->started) {
             this->charge_point->on_error_cleared(evse_id, error.uuid.uuid);
         } else {
-            this->event_queue[evse_id].push(error.uuid.uuid);
+            this->event_queue.push(Event(evse_id, EventType::ErrorCleared, error));
```
james-ctc: should be `event_queue.emplace({evse_id, EventType::ErrorCleared, error});`

modules/EVSE/OCPP/OCPP.cpp
```diff
@@ -587,7 +585,7 @@ void OCPP::init() {
             this->charge_point->on_log_status_notification(
                 log_status.request_id, types::system::log_status_enum_to_string(log_status.log_status));
         } else {
-            this->event_queue[0].push(log_status);
+            this->event_queue.push(Event(0, EventType::LogStatus, log_status));
         }
```
james-ctc, line 599: see other examples of using emplace

modules/EVSE/OCPP/OCPP.cpp
```diff
@@ -598,7 +596,7 @@ void OCPP::init() {
                     firmware_update_status.request_id,
                     conversions::to_ocpp_firmware_status_notification(firmware_update_status.firmware_update_status));
             } else {
-                this->event_queue[0].push(firmware_update_status);
+                this->event_queue.push(Event(0, EventType::FirmwareUpdateStatus, firmware_update_status));
```
james-ctc: see other examples of using emplace

modules/EVSE/OCPP/OCPP.cpp
```diff
@@ -1051,40 +1049,51 @@ void OCPP::ready() {
...
-                    EVLOG_info << "Processing queued cleared error event for evse_id: " << evse_id
-                               << ", error id: " << cleared_uuid;
-                    this->charge_point->on_error_cleared(evse_id, cleared_uuid);
-                }
-                evse_event_queue.pop();
+        // process event queue
+        std::vector<Event> events;
+        while (auto ev = this->event_queue.try_pop()) {
+            events.push_back(std::move(ev.value()));
+        }
```
james-ctc: Comments needed to explain why a copy of the queued events is needed rather than having a single loop:

```
while (auto ev = this->event_queue.try_pop()) {
  switch(ev.type)
}
```

modules/EVSE/OCPP/OCPP.hpp
```diff
@@ -39,19 +39,35 @@
...
+    FirmwareUpdateStatus,
+    LogStatus
+};
+
+using EventData = std::variant<types::evse_manager::SessionEvent, Everest::error::Error, types::system::LogStatus,
+                               types::system::FirmwareUpdateStatus>;
+
+struct Event {
+    EventType type;
+    EventData data;
```
james-ctc, line 56: Ideally EventType shouldn't be needed. std::variant manages that for you.
Error being used for raised and cleared could be an issue though.

A better solution is to have separate ErrorRaised and ErrorCleared types so that all the advantages of std::variant can be used.

e.g.

```
struct ErrorRaised : public Everest::error::Error {};
struct ErrorCleared : public Everest::error::Error {};


using EventData = std::variant<types::evse_manager::SessionEvent, ErrorRaised, ErrorCleared,
  types::system::LogStatus, types::system::FirmwareUpdateStatus>;
```

modules/EVSE/OCPP/OCPP.cpp
```diff
@@ -1051,40 +1049,51 @@ void OCPP::ready() {
...
+
+        for (const auto& queued_event : events) {
+            switch (queued_event.type) {
```
james-ctc: rather than switch std::visit should be used.
e.g.

```
[std::visit](https://en.cppreference.com/w/cpp/utility/variant/visit.html)(overloaded{
            [](const types::evse_manager::SessionEvent& arg) {
                …
            },
            [](const ErrorRaised& arg) {
                …
            },
            [](const ErrorCleared& arg) {
                …
            },
            [](types::system::LogStatus& arg) {
                …
            },
            [](types::system::FirmwareUpdateStatus& arg) {
                …
            },
        }, queued_event.data);
```

modules/EVSE/OCPP/OCPP.hpp
Pietfried: I introduced `EventType` to solve error raised and error cleared (+ upcoming new event for PublicKey string). I went with your suggestion to make better use of the variant

modules/EVSE/OCPP/OCPP.cpp
Pietfried: I switched to using an std::queue + a custom mutex instead of the thread_safe_queue to lock the queue while the events are being processed.

modules/EVSE/OCPP/OCPP.cpp
Pietfried: `emplace` was not available in https://github.com/EVerest/everest-core/blob/main/lib/everest/util/include/everest/util/queue/thread_safe_queue.hpp but I switched to using `std::queue` + `emplace` and a custom mutex that allows more control over the locking when the queued events are processed (see later comment)

modules/EVSE/OCPP/OCPP.cpp
```diff
@@ -469,21 +476,25 @@ void OCPP::init() {
...
     const auto error_handler = [this](const Everest::error::Error& error) {
+        std::lock_guard<std::mutex> lg(this->event_mutex);
         const auto evse_id = error.origin.mapping.has_value() ? error.origin.mapping.value().evse : 0;
         if (this->started) {
             const auto error_info = get_error_info(error);
             this->charge_point->on_error(evse_id, error_info);
         } else {
-            this->event_queue.push(Event(evse_id, EventType::ErrorRaised, error));
+            ErrorRaised error_raised{error};
+            this->event_queue.emplace(evse_id, error_raised);
```
james-ctc: The following should work, removing the need for a temporary object
```
this->event_queue.emplace({evse_id, ErrorRaised{error_raised});
```

modules/EVSE/OCPP/OCPP.cpp
```diff
@@ -469,21 +476,25 @@ void OCPP::init() {
...
 
     const auto error_cleared_handler = [this](const Everest::error::Error& error) {
+        std::lock_guard<std::mutex> lg(this->event_mutex);
         const auto evse_id = error.origin.mapping.has_value() ? error.origin.mapping.value().evse : 0;
         if (this->started) {
             this->charge_point->on_error_cleared(evse_id, error.uuid.uuid);
         } else {
-            this->event_queue.push(Event(evse_id, EventType::ErrorCleared, error));
+            ErrorCleared error_cleared{error};
+            this->event_queue.emplace(evse_id, error_cleared);
```
james-ctc:
```
this->event_queue.emplace({evse_id, ErrorCleared{error_raised});
```

james-ctc (dismissed):
minor suggestion re emplace, other than that looking good!

modules/EVSE/OCPP/OCPP.cpp
Pietfried: `this->event_queue.emplace(evse_id, ErrorRaised{error_raised});` does the job

mlitre (approved):

PR #1600, opened 2025-12-02
Fix unexpected d-link action in case of an error

FaHaGit:
## Describe your changes

Currently, as soon as V2G communication is unexpectedly terminated by a failed response code and the EV does not send a SessionStopReq, the dlink action parameter is incorrectly left at the initial value "dlink_error". This means that a connected EvseManager is not notified when V2G communication is terminated in this error case. However, the enum value dlink_error does not make sense within the EvseV2G module, as a data-link error cannot be detected by this module.


corneliusclaussen (approved):
LGTM

SebaLukas (approved):

PR #1609, opened 2025-12-03
Add configurable plug-in timeout behavior to Auth module 

Pietfried:
## Describe your changes

This PR adds a new configuration parameter `plug_in_timeout_enabled` to the Auth module, which controls the authorization-timeout behavior after a plug-in event.

When enabled, an internal timer is started immediately after an EV  is detected as plugged in. During the time defined by connection_timeout, the user must present a valid authorization token.

If no valid token is received within the specified connection_timeout, the authorization attempt is considered as timed out and a re-plug by the user is required to start a new authorization process.
      
This setting is particularly useful for charging stations with multiple EVSEs and a shared, non-EVSE-specific authorization interface (e.g., a single RFID reader). It prevents authorization tokens from being incorrectly associated with an EVSE where an EV is plugged in but has not been authorized, by ensuring that expired or ambiguous plug-ins are not considered for EVSE assignments for future authorization attempts.

So far the behavior was `plug_in_timeout_enabled` implicitly being set to true and no check how many EVSEs are present was performed. This is therefore a breaking change with `plug_in_timeout_enabled` defaulting to `false`.

### DIFF
--- a/modules/EVSE/Auth/lib/AuthHandler.cpp
+++ b/modules/EVSE/Auth/lib/AuthHandler.cpp
@@ -45,10 +45,11 @@ std::string token_handling_result_to_string(const TokenHandlingResult& result) {
 } // namespace conversions
 
 AuthHandler::AuthHandler(const SelectionAlgorithm& selection_algorithm, const int connection_timeout,
-                         bool prioritize_authorization_over_stopping_transaction, bool ignore_faults,
-                         const std::string& id, kvsIntf* store) :
+                         bool plug_in_timeout_enabled, bool prioritize_authorization_over_stopping_transaction,
+                         bool ignore_faults, const std::string& id, kvsIntf* store) :
     selection_algorithm(selection_algorithm),
     connection_timeout(connection_timeout),
+    plug_in_timeout_enabled(plug_in_timeout_enabled),
     prioritize_authorization_over_stopping_transaction(prioritize_authorization_over_stopping_transaction),
     ignore_faults(ignore_faults),
     reservation_handler(evses, id, store) {
@@ -567,7 +568,7 @@ AuthHandler::SelectEvseResult AuthHandler::select_evse(const std::vector<int>& s
             // transaction is active
             for (const auto& evse_id : selected_evses) {
                 const auto& evse = this->evses.at(evse_id);
-                if (!evse->transaction_active) {
+                if (evse->is_available()) {
                     result.status = SelectEvseReturnStatus::EvseSelected;
                     result.evse_id = evse_id;
                     return result;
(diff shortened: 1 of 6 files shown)


modules/EVSE/Auth/manifest.yaml
```diff
@@ -53,6 +53,22 @@ config:
...
+      
+      If no valid token is received within the specified connection_timeout, the authorization attempt is considered as timed out and a
+      re-plug by the user is required to start a new authorization process.
+      
+      This setting is particularly useful for charging stations with multiple EVSEs and a shared, non-EVSE-specific authorization interface
+      (e.g., a single RFID reader). It prevents authorization tokens from being incorrectly associated with an EVSE where an EV is
+      plugged in but has not been authorized, by ensuring that expired or ambiguous plug-ins are not considered for EVSE assignments
+      for future authorization attempts.
+    type: boolean
+    default: false
```
mlitre, line 71: Maybe the default should be true so it stays the same as the current implementation

modules/EVSE/Auth/lib/AuthHandler.cpp
```diff
@@ -567,7 +568,7 @@ AuthHandler::SelectEvseResult AuthHandler::select_evse(const std::vector<int>& s
             // transaction is active
             for (const auto& evse_id : selected_evses) {
                 const auto& evse = this->evses.at(evse_id);
-                if (!evse->transaction_active) {
+                if (evse->is_available()) {
```
mlitre, line 571: available and transaction_active are not exactly the same thing, a particular reason why we should change this?

modules/EVSE/Auth/lib/AuthHandler.cpp
```diff
@@ -777,6 +778,12 @@ void AuthHandler::handle_session_event(const int evse_id, const SessionEvent& ev
             this->evses.at(evse_id)->plugged_in = true;
             EVLOG_info << "Plug In event for evse#" << evse_id << ", starting auth";
 
+            // only set timeout if there are multiple evses managed by this auth handler
+            // or plug in timeout is explicitly enabled
+            if (this->evses.size() == 1 or !this->plug_in_timeout_enabled) {
+                return;
```
mlitre: Shouldn't we break instead here so that we still do the checks on reservation after the switch case?

modules/EVSE/Auth/lib/AuthHandler.cpp
Pietfried: `is_available()` checks for the `plug_in_timeout` while `transaction_active` doesn't. For `FindFirst` that lead to the fact that transactions could start even when a plug in timeout occured. This bug existed also before the PR.

modules/EVSE/Auth/manifest.yaml
Pietfried: The behavior changed slightly anyways because for a charging station with a single EVSE a plug in timeout doesnt exist anymore, and I consider `false` to be the more reasonable default, because usually authorization requests are bound to an EVSE and in these cases a plug in timeout should not trigger.

PR #1610, opened 2025-12-03
Configure ISO settings based on capabilities

Pietfried:
## Describe your changes

* Set EnergyTransferMode based on `hw_capabilities.max_phase_count_import` and verify `hw_capabilities` support discharging (export)

### DIFF
--- a/modules/EVSE/EvseManager/EvseManager.cpp
+++ b/modules/EVSE/EvseManager/EvseManager.cpp
@@ -324,10 +324,24 @@ void EvseManager::ready() {
             setup_physical_values.ac_nominal_voltage = config.ac_nominal_voltage;
             r_hlc[0]->call_set_charging_parameters(setup_physical_values);
 
-            transfer_modes.push_back(types::iso15118::EnergyTransferMode::AC_three_phase_core);
+            switch (hw_capabilities.max_phase_count_import) {
+            case 3:
+                transfer_modes.push_back(types::iso15118::EnergyTransferMode::AC_three_phase_core);
+                [[fallthrough]];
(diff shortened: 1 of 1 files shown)


modules/EVSE/EvseManager/EvseManager.cpp
```diff
@@ -324,10 +324,24 @@ void EvseManager::ready() {
             setup_physical_values.ac_nominal_voltage = config.ac_nominal_voltage;
             r_hlc[0]->call_set_charging_parameters(setup_physical_values);
 
-            transfer_modes.push_back(types::iso15118::EnergyTransferMode::AC_three_phase_core);
+            switch (hw_capabilities.max_phase_count_import) {
+            case 3:
+                transfer_modes.push_back(types::iso15118::EnergyTransferMode::AC_three_phase_core);
+                [[fallthrough]];
+            case 2:
+                transfer_modes.push_back(types::iso15118::EnergyTransferMode::AC_two_phase);
```
SebaLukas, line 332: This could cause a problem with ISO-2/EvseV2G. There is no AC_two_phase for ISO-2 definied. Only Three and Single

SebaLukas: I looked at the EvseV2G module and it shouldn't be a problem to send `AC_two_phase`. It is ignored in the module, but a warning may be displayed:
https://github.com/EVerest/everest-core/blob/main/modules/EVSE/EvseV2G/charger/ISO15118_chargerImpl.cpp#L353

SebaLukas (approved):

PR #1642, opened 2025-12-12
Integrate libevse-security v0.10.1

hikinggrass:
## Describe your changes
This contains the libevse-security source code from its last v0.10.1 release.
This release can be found at https://github.com/EVerest/libevse-security/releases/tag/v0.10.1

### DIFF
--- a/lib/everest/evse_security/include/evse_security/certificate/x509_wrapper.hpp
+++ b/lib/everest/evse_security/include/evse_security/certificate/x509_wrapper.hpp
@@ -135,9 +135,7 @@ private:
     // Relevant file in which this certificate resides
     std::optional<fs::path> file;
 
-#ifdef DEBUG_MODE_EVSE_SECURITY
     std::string debug_common_name;
-#endif
 };
 
 } // namespace evse_security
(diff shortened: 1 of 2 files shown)


lib/everest/evse_security/.clang-format
Pietfried, line 1: can be removed

lib/everest/evse_security/.eslintrc.json
Pietfried, line 1: can be removed

lib/everest/evse_security/.clang-format
hikinggrass: removed

lib/everest/evse_security/.eslintrc.json
hikinggrass: removed

lib/everest/evse_security/include/evse_security/certificate/x509_wrapper.hpp
```diff
@@ -135,9 +135,7 @@ class X509Wrapper {
     // Relevant file in which this certificate resides
     std::optional<fs::path> file;
 
-#ifdef DEBUG_MODE_EVSE_SECURITY
```
Pietfried: If you remove it here you can also remove the `DEBUG_MODE_EVSE_SECURITY` completely in the tests that use it and drop the cmake compile definitions

hikinggrass: I don't think that would be correct since that would set the debug_common_name fields to actual values, which we apparently don't want in non-debug builds. Otherwise this ifdef would not be there

Pietfried: I see, I assumed you want to remove the complete guard but just checked the commit message with your reasoning of the change :+1:

Pietfried (approved):

PR #1645, opened 2025-12-12
feat: huawei v100 telemetry data

markxoe:
## Describe your changes

- add a generic telemetry publisher interface to the fusion charger library
- fusion charger library now publishes different registers via said interface
- Everest module implements an Everest-specific implementation (`TelemetryPublisherEverest`) which publishes the data via Everests MqttProvider to a configurable mqtt base topic
- update the mock to publish a few more simulated registers
- Note that the TelementryPublisher can be replaced in the future once an Everest-wide telemetry system is implemented

### DIFF
--- a/modules/HardwareDrivers/PowerSupplies/Huawei_V100R023C10/Huawei_V100R023C10.cpp
+++ b/modules/HardwareDrivers/PowerSupplies/Huawei_V100R023C10/Huawei_V100R023C10.cpp
@@ -76,6 +76,20 @@ void Huawei_V100R023C10::init() {
             std::runtime_error("OVMs are necessary but number of OVMs does not match number of connectors in use"));
     }
 
+    if (config.telemetry_topic_prefix.empty()) {
+        this->telemetry_publisher = std::make_shared<fusion_charger::telemetry::TelemetryPublisherNull>();
+    } else {
+        this->telemetry_publisher = std::make_shared<TelemetryPublisherEverest>(
+            [this](const std::string& topic, const nlohmann::json& data) {
+                try {
+                    mqtt.publish(topic, data.dump());
+                } catch (std::exception& e) {
+                    EVLOG_error << "Failed publishing telemetry data to MQTT topic " << topic << ": " << e.what();
+                }
+            },
+            config.telemetry_topic_prefix);
+    }
+
     // Initialize all connectors. After that the config was loaded and we can initialize the dispenser
     for (int i = 0; i < number_of_connectors_used; i++) {
         invoke_init(*implementations[i]);
@@ -97,6 +111,8 @@ void Huawei_V100R023C10::init() {
     dispenser_config.verify_secure_goose_hmac = config.verify_secure_goose;
     dispenser_config.module_placeholder_allocation_timeout =
         std::chrono::seconds(config.module_placeholder_allocation_timeout_s);
+
+    dispenser_config.telemetry_publisher = this->telemetry_publisher;
 
     if (config.tls_enabled) {
         tls_util::MutualTlsClientConfig mutual_tls_config;
--- a/modules/HardwareDrivers/PowerSupplies/Huawei_V100R023C10/Huawei_V100R023C10.hpp
+++ b/modules/HardwareDrivers/PowerSupplies/Huawei_V100R023C10/Huawei_V100R023C10.hpp
@@ -21,6 +21,7 @@
 
 // ev@4bf81b14-a215-475c-a1d3-0a484ae48918:v1
 // insert your custom include headers here
+#include "telemetry_publisher_everest.hpp"
 #include <dispenser.hpp>
 #include <vector>
 // ev@4bf81b14-a215-475c-a1d3-0a484ae48918:v1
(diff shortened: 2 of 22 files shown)


modules/HardwareDrivers/PowerSupplies/Huawei_V100R023C10/Huawei_V100R023C10.hpp
```diff
@@ -99,6 +104,8 @@ class Huawei_V100R023C10 : public Everest::ModuleBase {
     };
     // PSU upstream voltage source
     UpstreamVoltageSource upstream_voltage_source;
+
+    std::shared_ptr<fusion_charger::telemetry::TelemetryPublisherBase> telemetry_publisher;
```
Pietfried: To me it looks like this doens't need to be a member of this class but can instead be a unique_ptr in `DispenserConfig`

modules/HardwareDrivers/PowerSupplies/Huawei_V100R023C10/connector_base/base.cpp
```diff
@@ -75,6 +80,9 @@ void ConnectorBase::ev_init() {
                 this->external_provided_data.contactor_status = ContactorStatus::OFF;
             }
 
+            mod->telemetry_publisher->datapoint_changed(telemetry_subtopic, "bsp_event",
```
Pietfried: The name of the datapoints e.g. "bsp_event", "output_voltage" etc. could be defined as constants and reused

modules/HardwareDrivers/PowerSupplies/Huawei_V100R023C10/telemetry_publisher_everest.cpp
```diff
@@ -0,0 +1,99 @@
...
+                return false;
+            });
+
+            if (stop_publisher_thread) {
+                return;
+            }
+        }
+
+        // wait for a few ms to batch changes
+        std::this_thread::sleep_for(std::chrono::milliseconds(10));
```
Pietfried, line 69: I think this could lead to intermediate values being lost, probably acceptable but could be problematic for e.g. `psu_running_mode`, `bsp_event` etc.

modules/HardwareDrivers/PowerSupplies/Huawei_V100R023C10/Huawei_V100R023C10.cpp
```diff
@@ -76,6 +76,14 @@ void Huawei_V100R023C10::init() {
             std::runtime_error("OVMs are necessary but number of OVMs does not match number of connectors in use"));
     }
 
+    if (config.telemetry_topic_prefix.empty()) {
+        this->telemetry_publisher = std::make_shared<fusion_charger::telemetry::TelemetryPublisherNull>();
+    } else {
+        this->telemetry_publisher = std::make_shared<TelemetryPublisherEverest>(
+            [this](const std::string& topic, const nlohmann::json& data) { mqtt.publish(topic, data.dump()); },
+            config.telemetry_topic_prefix);
```
Pietfried: You could add a try catch to prevent issues when `mqtt.publish` or `data.dump()` throw.

Pietfried (commented):
lgtm, please see a few comments in line

modules/HardwareDrivers/PowerSupplies/Huawei_V100R023C10/connector_base/base.cpp
markxoe: Good point, should be fixed in 30d04805a75696ea7d577bacbb150df8fef423ba

modules/HardwareDrivers/PowerSupplies/Huawei_V100R023C10/Huawei_V100R023C10.cpp
markxoe: good point, fixed in 1e08be637412f68cc97c76a564c75cc664fa759f

modules/HardwareDrivers/PowerSupplies/Huawei_V100R023C10/telemetry_publisher_everest.cpp
markxoe: I think dropping some data is acceptable in this scenario. The underlying issue is that the PSU writes some registers in a batch, and because the registers are decoupled, each write triggers its own datapoint change. If I don't batch changes this would result in a burst of mqtt messages.
I could lower this sleep to 1ms though. Would that be acceptable from your side or would you prefer a different approach?

Pietfried: I see, if dropping some data in rare cases is acceptable I think this is fine. You could just extend your comment in the code basically with what you wrote in your comment here to explain the behavior

markxoe: I've added an comment: fb0d03e79132dabaec78c867c4769c5a44078284

PR #1649, opened 2025-12-15
DZG_GSH01: reject crc calculation if message is too small

hikinggrass:


modules/HardwareDrivers/PowerMeters/DZG_GSH01/slip_protocol.cpp
```diff
@@ -32,6 +35,9 @@ inline bool is_message_crc_correct(std::vector<std::uint8_t>& vec) {
 }
 
 inline void restore_special_characters(std::vector<std::uint8_t>& vec) {
+    if (vec.size() < 1) {
+        return;
+    }
```
Pietfried: Looks redundant because the for loop wouldn't execute for `size < 2`
```suggestion
```

hikinggrass: correct, removed the commit :)

PR #1652, opened 2025-12-15
libiso15118: use integrated libcbv2g during standalone builds

hikinggrass:
Add CMake code for using ExternalProject to use libcbv2g from everest-core when building libiso15118 standalone

note: in the future the everest-core hash used in the standalone build will be updated automatically when the library gets exported into another repository, for now it points to the import commit of libcbv2g


Pietfried (approved):
lgtm

PR #1670, opened 2025-12-16
Add command for setting SOC to the car simulator

nzbr:
## Describe your changes

Extends the car simulator with a command that may be used to set the SOC to a custom value before plugging it in.
The command can be used with `set_soc <soc>`, where the `soc` argument is parsed as a floating point number that must be in the interval [0,100].


mlitre:
@hikinggrass this looks good to me, but I wouldn't mind a second review here :)

PR #1671, opened 2025-12-17
libiso15118: Add -2 cable check messages

Dalachowsky:
## Describe your changes

Added messages related to cable check.
This PR also introduces EV/EVSE status structs and EVSEProcessing enum.

### DIFF
--- a/lib/everest/iso15118/test/exi/cb/iso2/cable_check.cpp
+++ b/lib/everest/iso15118/test/exi/cb/iso2/cable_check.cpp
@@ -1,0 +1,55 @@
+#include <catch2/catch_test_macros.hpp>
+
+#include <cstdint>
+#include <iso15118/message/d2/cable_check.hpp>
+#include <iso15118/message/d2/variant.hpp>
+
+#include "helper.hpp"
+
+#include <iomanip>
+#include <iostream>
+#include <vector>
+
+using namespace iso15118;
+
+SCENARIO("Ser/Deserialize d2 cable check messages") {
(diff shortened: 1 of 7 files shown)


lib/everest/iso15118/test/exi/cb/iso2/cable_check.cpp
```diff
@@ -0,0 +1,55 @@
...
+
+            const auto& msg = variant.get<d2::msg::CableCheckRequest>();
+            const auto& header = msg.header;
+
+            REQUIRE(header.session_id == std::array<uint8_t, 8>{0x02, 0xDB, 0x22, 0x07, 0x3B, 0x08, 0x4D, 0x2D});
+
+            // TODO(kd): Should EV/EVSEStatus also be tested here?
+            REQUIRE(msg.ev_status.ev_ready == true);
+            REQUIRE(msg.ev_status.ev_error_code == d2::msg::data_types::DC_EVErrorCode::NO_ERROR);
+            REQUIRE(msg.ev_status.ev_ress_soc == 0);
```
Dalachowsky: I'm wondering whether the EV/EVSE status structs shall always be tested in tests for specific messages? I've introduced `convert()` functions for those structs so these could be tested in a separate place.

Dalachowsky:
I have also noticed that for -20 cable check response gets tested for each value of `Processing` enum. Since enum conversion is not handled by libiso15118 I don't see a reason to introduce these tests here.
What I would introduce would be tests that check if optional fields are properly handled, since this logic is introduced in libiso15118.

PR #1673, opened 2025-12-18
feat(SerialCommHub):Add Timeout error in StatusCodeEnum and return this error in case of timeout to allow distinguish modules between application/functional errors and communication errors

florinmihut:
## Describe your changes
We need this change to be able to distinguish between application level errors and communication error.
This introduces Timeout return when the communication is lost.
I checked the modules using serial comm hub but all of them are comparing the result of operations against Success.
The only one I found needed a change is the IskraMeter (Rust). Not an expert in Rust, please check it.

### DIFF
--- a/modules/Misc/SerialCommHub/main/serial_communication_hubImpl.cpp
+++ b/modules/Misc/SerialCommHub/main/serial_communication_hubImpl.cpp
@@ -121,19 +121,21 @@ serial_communication_hubImpl::perform_modbus_request(uint8_t device_address, tin
                                       tiny_modbus::FunctionCode_to_string_with_hex(function), device_address,
                                       first_register_address, first_register_address, e.what());
 
-            if (retry_counter != 1)
+            if (retry_counter != 1) {
                 EVLOG_debug << logmsg;
-            else
+            } else {
                 EVLOG_warning << logmsg;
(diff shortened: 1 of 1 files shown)


modules/Misc/SerialCommHub/main/serial_communication_hubImpl.cpp
```diff
@@ -109,9 +110,21 @@ serial_communication_hubImpl::perform_modbus_request(uint8_t device_address, tin
...
+            // TimeoutException is a specific type of communication error
+            last_error_was_timeout = true;
+            auto logmsg = fmt::format("Modbus call {} for device id {} addr {}({:#06x}) failed: {}",
+                                      tiny_modbus::FunctionCode_to_string_with_hex(function), device_address,
+                                      first_register_address, first_register_address, e.what());
+
+            if (retry_counter != 1)
+                EVLOG_debug << logmsg;
+            else
+                EVLOG_warning << logmsg;
```
hikinggrass: Can you add {} here?

florinmihut: Sure, no problem, I added as well to the other existing catch part above.

PR #1680, opened 2025-12-19
Fix framework external mqtt message handling

Pietfried:
## Describe your changes
fix(framework): Add dedicated worker thread for external MQTT message handling
- Split external MQTT message processing into separate worker thread
- Prevents blocking of operation messages by external MQTT handlers
- Fix unsubscribe token to use correct external topic
- Add async API integration tests for system module request/reply commands

Without this change, AsyncAPI module commands deadlock, because the command handler subscribes to an external mqtt reply which is only processed after the command handler returns.

### DIFF
--- a/tests/async_api_tests/test_system.py
+++ b/tests/async_api_tests/test_system.py
@@ -1,8 +1,6 @@
 import pytest
 import pytest_asyncio
 import json
-import time
-import logging
 import asyncio
 from typing import Callable, Dict
 import paho.mqtt.client as mqtt


tests/async_api_tests/CMakeLists.txt
```diff
@@ -0,0 +1 @@
+add_subdirectory(config)
```
hikinggrass:
```suggestion
add_subdirectory(config)

```

tests/async_api_tests/config/CMakeLists.txt
```diff
@@ -0,0 +1,6 @@
+# install configs
+install(
+    DIRECTORY "."
+    DESTINATION "${CMAKE_INSTALL_SYSCONFDIR}/everest"
+    FILES_MATCHING PATTERN "*.yaml"
+)
```
hikinggrass:
```suggestion
)

```

hikinggrass (approved):

PR #1684, opened 2025-12-22
Various fixes for devcontainer script

catarial:
Following the documentation to setup the devcontainer in vscode results in several errors. The devrd script does not get copied when running `setupcontainer` and a warning is printed saying it's missing. When opening the devcontainer in vscode it fails to open because the `.env` file is missing. It needs to be created with the `./devrd env` command, but that can't be done since the devrd script isn't copied.

## Describe your changes

1. Get rid of template directory so all necessary devcontainer files are copied. The `devrd` script wasn't being copied when running setupcontainer.
2. Change links to reference everest-core instead of everest-dev-container.
3. Update vscode docs to make sure `./devrd env` is run. Opening the devcontainer in vscode fails if the .env file is missing.


cburandt:
Hi @catarial . Thanks for this contribution. We are closing this PR in favour of https://github.com/EVerest/everest-core/pull/1688, which is a big cleanup of many more related files but also fixes the issues addressed by this PR.

PR #1687, opened 2026-01-05
refactor(OCPP): Introduce DeviceModelInterface abstraction layer

Pietfried:
## Describe your changes
Extract DeviceModel into an abstract interface to enable dependency inversion and improve testability. Components now depend on DeviceModelInterface rather than the concrete implementation.

Changes:
- Create DeviceModelInterface with pure virtual methods
- Move template helpers and common types to interface header
- Make DeviceModel inherit from DeviceModelInterface
- Update all consumers (ChargePoint, ConnectivityManager, Evse, EvseManager, MonitoringUpdater, functional blocks) to use interface
- Move filter_criteria_monitor functions to utils.cpp
- Add explicit device_model.hpp includes in tests needing concrete impl

This maintains backward compatibility while enabling easier mocking and alternative implementations.

### DIFF
--- a/lib/everest/ocpp/include/ocpp/v2/device_model_interface.hpp
+++ b/lib/everest/ocpp/include/ocpp/v2/device_model_interface.hpp
@@ -1,0 +1,201 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Pionix GmbH and Contributors to EVerest
+
+#ifndef DEVICE_MODEL_INTERFACE_HPP
+#define DEVICE_MODEL_INTERFACE_HPP
+
+#include <optional>
+#include <string>
+#include <vector>
+
+#include <ocpp/v2/enums.hpp>
+#include <ocpp/v2/ocpp_types.hpp>
+
+namespace ocpp {
+namespace v2 {
+
+/// \brief Helper struct that holds values that don't have spec coverage
+struct VariableMonitoringMeta {
+    VariableMonitoring monitor;
+    VariableMonitorType type;
+    std::optional<std::string> reference_value;
+};
+
+/// \brief Helper struct that contains all monitors related to a variable that are of a periodic type
+struct VariableMonitoringPeriodic {
+    Component component;
+    Variable variable;
+    std::vector<VariableMonitoringMeta> monitors;
+};
+
+/// \brief Helper struct that combines VariableCharacteristics and VariableMonitoring
+struct VariableMetaData {
+    VariableCharacteristics characteristics;
+    std::unordered_map<std::int64_t, VariableMonitoringMeta> monitors;
+    std::optional<std::string> source;
+};
+
+using VariableMap = std::map<Variable, VariableMetaData>;
+using DeviceModelMap = std::map<Component, VariableMap>;
+
+class DeviceModelError : public std::exception {
+public:
+    [[nodiscard]] const char* what() const noexcept override {
+        return this->reason.c_str();
+    }
+    explicit DeviceModelError(std::string msg) : reason(std::move(msg)) {
+    }
+    explicit DeviceModelError(const char* msg) : reason(std::string(msg)) {
+    }
+
+private:
+    std::string reason;
+};
+
+/// \brief Pure abstract interface for device model access.
+class DeviceModelInterface {
+public:
+    virtual ~DeviceModelInterface() = default;
+
+    // ============================================================================
+    // Value getters
+    // ============================================================================
+
+    /// \brief Gets a variable attribute value
+    /// \param component_id The component
+    /// \param variable_id The variable
+    /// \param attribute_enum The attribute
+    /// \param value The value to get (as string)
+    /// \param allow_write_only If true, write-only variables can be accessed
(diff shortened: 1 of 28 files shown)


lib/everest/ocpp/include/ocpp/v2/device_model_interface.hpp
```diff
@@ -0,0 +1,364 @@
...
+using DeviceModelMap = std::map<Component, VariableMap>;
+
+class DeviceModelError : public std::exception {
+public:
+    [[nodiscard]] const char* what() const noexcept override {
+        return this->reason.c_str();
+    }
+    explicit DeviceModelError(std::string msg) : reason(std::move(msg)) {
+    }
+    explicit DeviceModelError(const char* msg) : reason(std::string(msg)) {
```
james-ctc, line 48: Not sure an extra object is needed.
The following should work fine.

```
explicit DeviceModelError(const char* msg) : reason(msg)
```

lib/everest/ocpp/include/ocpp/v2/device_model_interface.hpp
```diff
@@ -0,0 +1,364 @@
...
+        return std::stoi(value);
+    } else if constexpr (std::is_same_v<T, double>) {
+        return std::stod(value);
+    } else if constexpr (std::is_same_v<T, std::size_t>) {
+        const std::size_t res = std::stoul(value);
+        return res;
+    } else if constexpr (std::is_same_v<T, DateTime>) {
+        return DateTime(value);
+    } else if constexpr (std::is_same_v<T, bool>) {
+        return ocpp::conversions::string_to_bool(value);
```
james-ctc: I needs to be clear what the expectation of to_specific_type() is.

- should it be ensuring values are valid
- should it exception on invalid

at the moment it is a mixture.

e.g. 
"Hello" -> int raises an exception
"Hello" -> bool returns false

lib/everest/ocpp/include/ocpp/v2/device_model_interface.hpp
```diff
@@ -0,0 +1,364 @@
...
+                                        const AttributeEnum& attribute_enum = AttributeEnum::Actual) const {
+        std::string value;
+        const auto status = request_value_internal(component_id, variable_id, attribute_enum, value, true);
+
+        if (status != GetVariableStatusEnum::Accepted) {
+            return std::nullopt;
+        }
+
+        try {
+            return to_specific_type<T>(value);
```
james-ctc: see comment above about issues with string conversion to bool

lib/everest/ocpp/include/ocpp/v2/device_model_interface.hpp
james-ctc, line 1: in general I'm not keen on templates in interfaces. 
they could have issues with inheritance.

lib/everest/ocpp/include/ocpp/v2/device_model_interface.hpp
```diff
@@ -0,0 +1,364 @@
...
+    // ============================================================================
+    // Integrity check
+    // ============================================================================
+
+    /// \brief Check data integrity of the device model
+    /// \param evse_connector_structure The EVSE connector structure
+    virtual void check_integrity(const std::map<std::int32_t, std::int32_t>& evse_connector_structure) = 0;
+
+private:
+    virtual GetVariableStatusEnum request_value_internal(const Component& component_id, const Variable& variable_id,
```
james-ctc: it is unusual to have a private pure virtual function.

lib/everest/ocpp/include/ocpp/v2/device_model_interface.hpp
Pietfried: The templates are part of the existing DeviceModel interface that I tried to preserve. The templated functions of the interface aren't virtual. In my view using the non-virtual interface pattern was the best choice here to achieve backwards compatibility and polymorphism.

lib/everest/ocpp/include/ocpp/v2/device_model_interface.hpp
Pietfried: see comment above

lib/everest/ocpp/include/ocpp/v2/device_model_interface.hpp
Pietfried: Added a check if the string is actually a bool, throwing if not

lib/everest/ocpp/include/ocpp/v2/device_model_interface.hpp
Pietfried: see comment above

lib/everest/ocpp/include/ocpp/v2/device_model_interface.hpp
Pietfried: I seperated the class structure into:
* DeviceModelInterface being the pure virtual class
* DeviceModelBase containing the templates
* DeviceModelAbstract inheriting from DeviceModelInterface and DeviceModelBase. All consumers in v2 use DeviceModelAbstract

PR #1688, opened 2026-01-05
Chore/cleanup after dev env migration

andistorm:
## Describe your changes

Finish the  devcontainer feature.

This restructures the applications/* subdirectories:
* Move .devcontainer/ dir to root dir
* Move .prospector.yaml to root dir
* Move docker images in to applications/containers/
* Move dependency_manager to applications/
* Move everest_dev_tool to applications/
* Move everest-metadata.yaml to root dir
* Move devrd to applications/
* Update applications/README.md

The devcontainer feature is now working the the monorepo:
* Refactor devcontainer to work in monorepo
* Add initializeCommand to automatically create .env file for vscode

Backup of this branch with individual commits: [chore/cleanup-after-dev-env-migration_BACKUP_before_squash20260224](https://github.com/EVerest/everest-core/tree/chore/cleanup-after-dev-env-migration_BACKUP_before_squash20260224)

Documentation is prepared in another PR (https://github.com/EVerest/everest-core/pull/1693)


.devcontainer/general-devcontainer/devcontainer.json
```diff
@@ -1,10 +1,25 @@
...
+  "runServices": [ "mqtt-server", "ocpp-db", "steve", "mqtt-explorer" ],
 
-  "remoteUser": "docker",
+  // Workspace inside the container
   "workspaceFolder": "/workspace",
+
+  "initializeCommand": "./applications/devrd/devrd env",
+
+  // Prevent generation of broken override compose files
+  "overrideFeatureInstallCommand": false,
```
cburandt: Does this option exist?
```suggestion
  "overrideCommand": false,
```

CMakeLists.txt
```diff
@@ -195,7 +195,7 @@ add_subdirectory(config)
 add_subdirectory(doc)
 
 if(EVEREST_BUILD_APPLICATIONS)
-  add_subdirectory(applications)
+#   add_subdirectory(applications)
 endif()
```
cburandt, line 202: add_subdirectory(application/) unconditionally and use EVEREST_BUILD_APPLICATION inside application/CMakeLists.txt to include or not include optional subfolders (like pionix_chargebridge)?

.devcontainer/general-devcontainer/devcontainer.json
andistorm: You are right, I removed this option from the file

applications/dependency_manager/CMakeLists.txt
```diff
@@ -21,4 +21,4 @@ else()
     unset(EVEREST_DEPENDENCY_MANAGER CACHE)
     find_program(EVEREST_DEPENDENCY_MANAGER edm HINTS ${EV_ACTIVATE_PYTHON_VENV_PATH_TO_VENV}/bin REQUIRED)
     message(STATUS "Using edm from: ${EDM}")
-endif()
+endif()
```
andistorm:
```suggestion
endif()

```

applications/dependency_manager/CMakeLists.txt
andistorm, line 1: I would keep the CMakeLists.txt here out of `dependency_manager/` since it contains a python package and the CMake install magic doesn't fit in it's scope at least in my opinion. So suggestion would be to put it in the parent dir

andistorm (commented):
🐧 LGTM in general, left a view comments, Thanks @cburandt for cleaning up!

applications/dependency_manager/CMakeLists.txt
cburandt: Agree in principle, and yes it's not really nice, as it is. But the parent dir already has a CMakeLists.txt which is only conditionally included (via EVEREST_BUILD_APPLICATIONS). Thus simply adding the one from dependency_manager there, will not work. We could however evaluate the EVEREST_BUILD_APPLICATIONS option there and only then include the optional subdirs in applications/ (like pionix_chargebridge/).

andistorm: would make sense to me, but then this option should be renamed. Or we should add warnings_errors when this option is used in specific combinations with venv and edm to ensure the developer understands when edm is installed automatically and when not

cburandt: Tried to merge the applications/dependency_manager/CMakeLists.txt into applications/CMakeLists.txt, but this attempt failed, since these two need to be referenced via add_subdirectory from different locations from the top-level CMakeLists.txt.

cburandt (approved):
LGTM

Pietfried (approved):

hikinggrass (approved):

Issue #567, opened 2024-02-28
Auth (multiport): connector remains in plug_in_queue, although it is plugged out


FaHaGit:
**Version:**
2024.2.0

**Module config:**

```
    config_module:
      connection_timeout: 20
      prioritize_authorization_over_stopping_transaction: true
      selection_algorithm: PlugEvents

    connections:
      evse_manager:
        - module_id: connector_1
          implementation_id: evse
        - module_id: connector_2
          implementation_id: evse
```

**Describe the problem:**

In case when a plug is plugged-in and plugged-out without performing an authorization via RFID, the connector remains in the plug_in_queue although it has already been unplugged. As a result, the plug-selection-algorithm, bases on "PlugEvents", detects the unplugged connector and authorize it for charging when a RFID card is presented, without any plugged-in connectors.

**How to reproduce:**

1. In a setup for a Multi-port charger, plug in one of the connectors, without authorziation (Plug in before authoziation).
2. Plug the connector out, also without authoziation.
3. Swipe a RFID card. The authoziation will be assigned to the connector which was just plugged out.

**Expectation:**
The authoziation should not be assigned to any connector until one of them is plugged in.

**Describe your solution:**

Remove connector ID from list in case "SessionFinished" event has been received

**Additional context:**

--

barsnick:
Fixed in https://github.com/EVerest/everest-core/pull/568

PR #1689, opened 2026-01-05
refactor(OCPP): Move ConnectivityManager to common namespace

Pietfried:
## Describe your changes
Move ConnectivityManager from v2 to common namespace:
* Move ConnectivityManager to common namespace for reuse across OCPP versions
* Refactor constructor to accept logging and message callbacks via setters instead of constructor parameters
* Add ChargePoint constructor accepting shared ConnectivityManagerInterface for external injection
* Add validation that MessageLogging instance is provided to websocket at construction

### DIFF
--- a/lib/everest/ocpp/include/ocpp/v2/charge_point.hpp
+++ b/lib/everest/ocpp/include/ocpp/v2/charge_point.hpp
@@ -90,6 +90,28 @@ public:
     /// @{
 
     ///
+    /// \brief Shall be called when a websocket connection has been established in case the connectivity_handler is
+    /// provided exernally.
+    /// \param configuration_slot The network profile slot used for the connection.
+    /// \param network_connection_profile The network connection profile used for the connection.
+    virtual void on_websocket_connected(const int configuration_slot,
+                                        const NetworkConnectionProfile& network_connection_profile,
+                                        const OcppProtocolVersion ocpp_version) = 0;
+
+    ///
+    /// \brief Shall be called when a websocket connection has been disconnected in case the connectivity_handler is
+    /// provided externally.
+    /// \param configuration_slot The network profile slot used for the connection.
+    /// \param network_connection_profile The network connection profile used for the connection.
+    virtual void on_websocket_disconnected(const int configuration_slot,
+                                           const NetworkConnectionProfile& network_connection_profile) = 0;
+
+    /// \brief Shall be called when a websocket connection attempt has failed in case the connectivity_handler is
+    /// provided externally.
+    /// \param reason The reason why the connection failed.
+    virtual void on_websocket_connection_failed(ConnectionFailedReason reason) = 0;
+
+    ///
     /// \brief Can be called when a network is disconnected, for example when an ethernet cable is removed.
     ///
     /// This is introduced because the websocket can take several minutes to timeout when a network interface becomes
@@ -345,7 +367,7 @@ class ChargePoint : public ChargePointInterface, private ocpp::ChargingStationBa
 private:
     std::shared_ptr<DeviceModelAbstract> device_model;
     std::unique_ptr<EvseManager> evse_manager;
-    std::unique_ptr<ConnectivityManager> connectivity_manager;
+    std::shared_ptr<ConnectivityManagerInterface> connectivity_manager;
 
     std::unique_ptr<MessageDispatcherInterface<MessageType>> message_dispatcher;
 
@@ -373,13 +395,13 @@ private:
     fs::path share_path;
 
     // states
-    std::atomic<RegistrationStatusEnum> registration_status;
-    std::atomic<OcppProtocolVersion> ocpp_version =
-        OcppProtocolVersion::Unknown; // version that is currently in use, selected by CSMS in websocket handshake
-    std::atomic<UploadLogStatusEnum> upload_log_status;
+    std::atomic<RegistrationStatusEnum> registration_status{RegistrationStatusEnum::Rejected};
+    std::atomic<OcppProtocolVersion> ocpp_version{
+        OcppProtocolVersion::Unknown}; // version that is currently in use, selected by CSMS in websocket handshake
+    std::atomic<UploadLogStatusEnum> upload_log_status{UploadLogStatusEnum::Idle};
     std::atomic<std::int32_t> upload_log_status_id;
-    BootReasonEnum bootreason;
-    bool skip_invalid_csms_certificate_notifications;
+    BootReasonEnum bootreason{BootReasonEnum::PowerUp};
+    bool skip_invalid_csms_certificate_notifications{false};
 
     /// \brief Component responsible for maintaining and persisting the operational status of CS, EVSEs, and connectors.
     std::shared_ptr<ComponentStateManagerInterface> component_state_manager;
@@ -411,12 +433,7 @@ private:
     // internal helper functions
     void initialize(const std::map<std::int32_t, std::int32_t>& evse_connector_structure,
                     const std::string& message_log_path);
-    void websocket_connected_callback(const int configuration_slot,
-                                      const NetworkConnectionProfile& network_connection_profile,
-                                      const OcppProtocolVersion ocpp_version);
-    void websocket_disconnected_callback(const int configuration_slot,
-                                         const NetworkConnectionProfile& network_connection_profile);
-    void websocket_connection_failed(ConnectionFailedReason reason);
+    OcspUpdater make_ocsp_updater();
     void update_dm_availability_state(const std::int32_t evse_id, const std::int32_t connector_id,
                                       const ConnectorStatusEnum status);
 
@@ -461,6 +478,23 @@ public:
 
     /// @name Constructors for 2.0.1
     /// @{
+
+    /// \brief Construct a new ChargePoint object
+    /// \param evse_connector_structure Map that defines the structure of EVSE and connectors of the chargepoint. The
+    /// key represents the id of the EVSE and the value represents the number of connectors for this EVSE. The ids of
+    /// the EVSEs have to increment starting with 1.
+    /// \param device_model device model instance
+    /// \param database_handler database handler instance
+    /// \param evse_security Pointer to evse_security that manages security related operations
+    /// \param connectivity_manager connectivity manager instance
+    /// \param message_log_path Path to where logfiles are written to
+    /// \param callbacks Callbacks that will be registered for ChargePoint
+    ChargePoint(const std::map<int32_t, int32_t>& evse_connector_structure,
+                const std::shared_ptr<DeviceModelAbstract> device_model,
+                const std::shared_ptr<DatabaseHandler> database_handler,
+                const std::shared_ptr<EvseSecurity> evse_security,
+                const std::shared_ptr<ConnectivityManagerInterface> connectivity_manager,
+                const std::string& message_log_path, const Callbacks& callbacks);
 
     /// \brief Construct a new ChargePoint object
     /// \param evse_connector_structure Map that defines the structure of EVSE and connectors of the chargepoint. The
@@ -526,6 +560,12 @@ public:
     void connect_websocket(std::optional<std::int32_t> network_profile_slot = std::nullopt) override;
     void disconnect_websocket() override;
 
+    void on_websocket_connected(const int configuration_slot,
+                                const NetworkConnectionProfile& network_connection_profile,
+                                const OcppProtocolVersion ocpp_version) override;
+    void on_websocket_disconnected(const int configuration_slot,
+                                   const NetworkConnectionProfile& network_connection_profile) override;
+    void on_websocket_connection_failed(ConnectionFailedReason reason) override;
     void on_network_disconnected(OCPPInterfaceEnum ocpp_interface) override;
 
     void on_firmware_update_status_notification(std::int32_t request_id,
--- a/lib/everest/ocpp/include/ocpp/v2/charge_point.hpp
+++ b/lib/everest/ocpp/include/ocpp/v2/charge_point.hpp
@@ -506,10 +506,12 @@ public:
     /// \param message_log_path Path to where logfiles are written to
     /// \param evse_security Pointer to evse_security that manages security related operations
     /// \param callbacks Callbacks that will be registered for ChargePoint
+    /// \param share_path Path where utility files for OCPP are read and written to
     ChargePoint(const std::map<std::int32_t, std::int32_t>& evse_connector_structure,
                 std::shared_ptr<DeviceModelAbstract> device_model, std::shared_ptr<DatabaseHandler> database_handler,
                 std::shared_ptr<MessageQueue<v2::MessageType>> message_queue, const std::string& message_log_path,
-                const std::shared_ptr<EvseSecurity> evse_security, const Callbacks& callbacks);
+                const std::shared_ptr<EvseSecurity> evse_security, const Callbacks& callbacks,
+                const fs::path& share_path = {});
 
     /// \brief Construct a new ChargePoint object
     /// \param evse_connector_structure Map that defines the structure of EVSE and connectors of the chargepoint. The
--- a/lib/everest/ocpp/lib/ocpp/common/connectivity_manager.cpp
+++ b/lib/everest/ocpp/lib/ocpp/common/connectivity_manager.cpp
@@ -28,11 +28,10 @@ using SetNetworkProfileRequest = ocpp::v2::SetNetworkProfileRequest;
 namespace ControllerComponentVariables = ocpp::v2::ControllerComponentVariables;
 
 ConnectivityManager::ConnectivityManager(ocpp::v2::DeviceModelAbstract& device_model,
-                                         std::shared_ptr<EvseSecurity> evse_security,
-                                        std::shared_ptr<MessageLogging> logging, const fs::path& share_path,
-                                         const std::function<void(const std::string& message)>& message_callback) :
+                                         std::shared_ptr<EvseSecurity> evse_security, const fs::path& share_path) :
     device_model{device_model},
     evse_security{evse_security},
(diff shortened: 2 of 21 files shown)


Pietfried:
@james-ctc I just rebased this PR, it is now ready for review.

lib/everest/ocpp/include/ocpp/v2/charge_point.hpp
```diff
@@ -464,6 +481,22 @@ class ChargePoint : public ChargePointInterface, private ocpp::ChargingStationBa
...
+    /// key represents the id of the EVSE and the value represents the number of connectors for this EVSE. The ids of
+    /// the EVSEs have to increment starting with 1.
+    /// \param device_model device model instance
+    /// \param database_handler database handler instance
+    /// \param evse_security Pointer to evse_security that manages security related operations
+    /// \param connectivity_manager connectivity manager instance
+    /// \param message_log_path Path to where logfiles are written to
+    /// \param callbacks Callbacks that will be registered for ChargePoint
+    ChargePoint(const std::map<int32_t, int32_t>& evse_connector_structure,
+                std::shared_ptr<DeviceModelAbstract> device_model, std::shared_ptr<DatabaseHandler> database_handler,
```
james-ctc: The shared pointers should be passed as const reference:
```
    ChargePoint(const std::map<int32_t, int32_t>& evse_connector_structure,
                const std::shared_ptr<DeviceModelAbstract>& device_model,
                const std::shared_ptr<DatabaseHandler>& database_handler,
                const std::shared_ptr<EvseSecurity>& evse_security,
                const std::shared_ptr<ConnectivityManagerInterface>& connectivity_manager,
                const std::string& message_log_path, const Callbacks& callbacks);
```

lib/everest/ocpp/lib/ocpp/common/connectivity_manager.cpp
```diff
@@ -19,24 +19,37 @@ constexpr std::int32_t default_network_config_timeout_seconds = 60;
...
-                                         const std::function<void(const std::string& message)>& message_callback) :
+using NetworkConnectionProfile = ocpp::v2::NetworkConnectionProfile;
+using AttributeEnum = ocpp::v2::AttributeEnum;
+using DeviceModel = ocpp::v2::DeviceModel;
+using OCPPInterfaceEnum = ocpp::v2::OCPPInterfaceEnum;
+using SetNetworkProfileRequest = ocpp::v2::SetNetworkProfileRequest;
+namespace ControllerComponentVariables = ocpp::v2::ControllerComponentVariables;
+
+ConnectivityManager::ConnectivityManager(ocpp::v2::DeviceModelAbstract& device_model,
+                                         std::shared_ptr<EvseSecurity> evse_security) :
```
james-ctc: for consistency (and reduced stack usage) references should be used.
```
ConnectivityManager::ConnectivityManager(ocpp::v2::DeviceModelAbstract& device_model,
                                         const std::shared_ptr<EvseSecurity>& evse_security) :
```

lib/everest/ocpp/lib/ocpp/common/connectivity_manager.cpp
```diff
@@ -19,24 +19,37 @@ constexpr std::int32_t default_network_config_timeout_seconds = 60;
...
     websocket{nullptr},
-    message_callback{message_callback},
     wants_to_be_connected{false},
     active_network_configuration_priority{0},
     last_known_security_level{0},
     connected_ocpp_version{OcppProtocolVersion::Unknown} {
     cache_network_connection_profiles();
 }
 
+void ConnectivityManager::set_message_callback(const std::function<void(const std::string& message)>& callback) {
```
james-ctc, line 46: If these were in the header file then perhaps some optimisation could be used by the compiler.

lib/everest/ocpp/lib/ocpp/v2/charge_point.cpp
```diff
@@ -48,6 +48,25 @@ namespace v2 {
 const auto DEFAULT_MESSAGE_QUEUE_SIZE_THRESHOLD = 1000;
 
 ChargePoint::ChargePoint(const std::map<std::int32_t, std::int32_t>& evse_connector_structure,
+                         std::shared_ptr<DeviceModelAbstract> device_model,
+                         std::shared_ptr<DatabaseHandler> database_handler,
```
james-ctc: see hpp
```
ChargePoint::ChargePoint(const std::map<std::int32_t, std::int32_t>& evse_connector_structure,
                         const std::shared_ptr<DeviceModelAbstract>& device_model,
                         const std::shared_ptr<DatabaseHandler>& database_handler,
                         const std::shared_ptr<EvseSecurity>& evse_security,
                         const std::shared_ptr<ConnectivityManagerInterface>& connectivity_manager,
                         const std::string& message_log_path, const Callbacks& callbacks) :
```

lib/everest/ocpp/lib/ocpp/v2/charge_point.cpp
```diff
@@ -48,6 +48,25 @@ namespace v2 {
...
+                         const std::shared_ptr<EvseSecurity> evse_security,
+                         const std::shared_ptr<ConnectivityManagerInterface> connectivity_manager,
+                         const std::string& message_log_path, const Callbacks& callbacks) :
+    ocpp::ChargingStationBase(evse_security),
+    device_model(device_model),
+    database_handler(database_handler),
+    connectivity_manager(connectivity_manager),
+    registration_status(RegistrationStatusEnum::Rejected),
+    skip_invalid_csms_certificate_notifications(false),
```
james-ctc: having recently having to spend ages debugging constructors I recommend that member variables that are not dependent on constructor arguments are set to default values in the hpp file.
This is already done for some
e.g.
```
    std::atomic<OcppProtocolVersion> ocpp_version =
        OcppProtocolVersion::Unknown;
```

```
    std::atomic<OcppProtocolVersion> ocpp_version{OcppProtocolVersion::Unknown};
    std::atomic<RegistrationStatusEnum> registration_status{RegistrationStatusEnum::Rejected};
    bool skip_invalid_csms_certificate_notifications{false};
    std::atomic<UploadLogStatusEnum> upload_log_status{UploadLogStatusEnum::Idle};
    BootReasonEnum bootreason{BootReasonEnum::PowerUp};
```
There are two constructors and these initial values are the same for both. Being in the header ensures they remain the same. Better for code maintenance.

lib/everest/ocpp/lib/ocpp/v2/charge_point.cpp
```diff
@@ -180,6 +176,82 @@ void ChargePoint::disconnect_websocket() {
...
+                // B04.FR.02
+                // If offline period doesn't exceed offline threshold then send the status notification for all
+                // connectors that changed state
+                EVLOG_debug << "offline for less than offline threshold ";
+                this->component_state_manager->send_status_notification_changed_connectors();
+            }
+            this->security->init_certificate_expiration_check_timers(); // re-init as timers are stopped on disconnect
+        }
+    }
+    this->time_disconnected = std::chrono::time_point<std::chrono::steady_clock>();
```
james-ctc: the following comment would be useful
```
    // set time_disconnected to zero
```

lib/everest/ocpp/lib/ocpp/v2/charge_point.cpp
```diff
@@ -180,6 +176,82 @@ void ChargePoint::disconnect_websocket() {
...
+                                                                  ocpp_version);
+    }
+}
+
+void ChargePoint::on_websocket_disconnected(const int configuration_slot,
+                                            const NetworkConnectionProfile& network_connection_profile) {
+    this->message_queue->pause();
+
+    // check if offline threshold has been defined
+    if (this->device_model->get_value<int>(ControllerComponentVariables::OfflineThreshold) != 0) {
```
james-ctc: what should happen with negative values?

lib/everest/ocpp/lib/ocpp/v2/charge_point.cpp
```diff
@@ -180,6 +176,82 @@ void ChargePoint::disconnect_websocket() {
...
+
+    this->security->stop_certificate_expiration_check_timers();
+    if (this->callbacks.connection_state_changed_callback.has_value()) {
+        this->callbacks.connection_state_changed_callback.value()(false, configuration_slot, network_connection_profile,
+                                                                  this->ocpp_version);
+    }
+}
+
+void ChargePoint::on_websocket_connection_failed(ConnectionFailedReason reason) {
+    switch (reason) {
```
james-ctc: in general switch should include default: handling.
Even when all expected cases are handled the default: should log an error so it can provide useful information when analysing issues.
e.g. what if ConnectionFailedReason was uninitialised?

james-ctc (changes requested):

lib/everest/ocpp/lib/ocpp/common/connectivity_manager.cpp
Pietfried: This would requires a chain of changes that I would rather add in a separate PR

lib/everest/ocpp/lib/ocpp/v2/charge_point.cpp
Pietfried: I added a `minLimit` of 0 to the characteristics of offline threshold

lib/everest/ocpp/lib/ocpp/v2/charge_point.cpp
Pietfried: I prefer do not specify a `default` so the explicit handling of a new enum value is enforced. Compilation will fail with `Wswitch-enum` and `Werror` in any case.

I added handling after the switch clause

james-ctc: the compiler will still fail with an error even with the default clause.
So the handling should be part of the switch. 
(and break rather than return)

lib/everest/ocpp/include/ocpp/v2/charge_point.hpp
```diff
@@ -345,7 +367,7 @@ class ChargePoint : public ChargePointInterface, private ocpp::ChargingStationBa
 private:
     std::shared_ptr<DeviceModelAbstract> device_model;
     std::unique_ptr<EvseManager> evse_manager;
-    std::unique_ptr<ConnectivityManager> connectivity_manager;
+    std::shared_ptr<ConnectivityManagerInterface> connectivity_manager;
```
james-ctc, line 370: is the connectivity manager object used by others? if not then a unique_ptr could be used.

lib/everest/ocpp/lib/ocpp/v2/charge_point.cpp
```diff
@@ -425,6 +493,19 @@ void ChargePoint::on_ev_charging_needs(const NotifyEVChargingNeedsRequest& reque
 
 void ChargePoint::initialize(const std::map<std::int32_t, std::int32_t>& evse_connector_structure,
                              const std::string& message_log_path) {
+    if (!this->device_model) {
```
james-ctc: for consistency use the same as line #567
```
    if (this->connectivity_manager == nullptr) {
```

lib/everest/ocpp/lib/ocpp/v2/charge_point.cpp
```diff
@@ -425,6 +493,19 @@ void ChargePoint::on_ev_charging_needs(const NotifyEVChargingNeedsRequest& reque
...
+        EVLOG_AND_THROW(std::invalid_argument("Device model should not be null"));
+    }
+
+    // Make sure the received callback struct is completely filled early before we actually start running
+    if (!this->callbacks.all_callbacks_valid(this->device_model, evse_connector_structure)) {
+        EVLOG_AND_THROW(std::invalid_argument("All non-optional callbacks must be supplied"));
+    }
+
+    if (!this->database_handler) {
```
james-ctc: for consistency use the same as line #567
```
    if (this->connectivity_manager == nullptr) {
```

lib/everest/ocpp/lib/ocpp/v2/charge_point.cpp
```diff
@@ -480,23 +561,25 @@ void ChargePoint::initialize(const std::map<std::int32_t, std::int32_t>& evse_co
...
-        });
-    this->connectivity_manager->set_websocket_disconnected_callback(
-        [this](int configuration_slot, const NetworkConnectionProfile& network_connection_profile, auto) {
-            this->websocket_disconnected_callback(configuration_slot, network_connection_profile);
-        });
-    this->connectivity_manager->set_websocket_connection_failed_callback(
-        [this](ConnectionFailedReason reason) { this->websocket_connection_failed(reason); });
+    if (this->connectivity_manager == nullptr) {
+        // connectivity manager was not provided in constructor. Create a new one and set the message callbacks
+        this->connectivity_manager = std::make_shared<ConnectivityManager>(*this->device_model, this->evse_security);
```
james-ctc: It isn't obvious why connectivity_manager is not a unique_ptr

james-ctc (changes requested):

lib/everest/ocpp/include/ocpp/v2/charge_point.hpp
Pietfried: It is meant to be used by the consumer of the library to initiate the initial connection and then passed to the respective ChargePoint implementation.

lib/everest/ocpp/lib/ocpp/v2/charge_point.cpp
Pietfried: see comment above

lib/everest/ocpp/lib/ocpp/v2/charge_point.cpp
Pietfried: I am aware that compilation will fail with both approaches. My preference for omitting default is that it makes it slightly harder to be lazy. Without it, you must add an explicit case before the code compiles. With a default it's tempting to just add the new enum value and let it fall through to the default without thinking about the specific logic needed.

lib/everest/ocpp/include/ocpp/v2/charge_point.hpp
james-ctc: That makes sense, thank you for the explanation

lib/everest/ocpp/lib/ocpp/v2/charge_point.cpp
james-ctc: I still think
```
switch(x) {
case A:
   …
   break;
case B:
   …
   break;
}
```
is better than using return. 
As written an accidental change from return to break will have unintended consequences.
This is also the MISRA C recommendation. (use of break and requiring default).

james-ctc (dismissed):

lib/everest/ocpp/lib/ocpp/v2/charge_point.cpp
```diff
@@ -480,23 +565,25 @@ void ChargePoint::initialize(const std::map<std::int32_t, std::int32_t>& evse_co
...
+        this->connectivity_manager->set_websocket_connected_callback(
+            [this](int configuration_slot, const NetworkConnectionProfile& network_connection_profile,
+                   const OcppProtocolVersion ocpp_version) {
+                this->on_websocket_connected(configuration_slot, network_connection_profile, ocpp_version);
+            });
+        this->connectivity_manager->set_websocket_disconnected_callback(
+            [this](int configuration_slot, const NetworkConnectionProfile& network_connection_profile, auto) {
+                this->on_websocket_disconnected(configuration_slot, network_connection_profile);
+            });
+        this->connectivity_manager->set_websocket_connection_failed_callback(
```
mlitre, line 511: Shouldn't all of these set callbacks be outside of the if statement since they should always be set?

mlitre (changes requested):

lib/everest/ocpp/lib/ocpp/v2/charge_point.cpp
Pietfried: The idea is that if the connectivity_manager is provided externally, it also controls these callbacks.

This would allow the external application to check the ocpp protocol version and then either start/stop/control the OCPP1.6 or the OCPP2.x charge point instances

mlitre: Ok that's what I was thinking, but it is not reflected in the documentation so I was not sure if that was the actual behavior.

mlitre (dismissed):
Looking good!

lib/everest/ocpp/lib/ocpp/common/connectivity_manager.cpp
```diff
@@ -418,7 +430,7 @@ void ConnectivityManager::on_websocket_connected(OcppProtocolVersion protocol) {
 void ConnectivityManager::on_websocket_disconnected() {
     std::optional<NetworkConnectionProfile> network_connection_profile =
         this->get_network_connection_profile(this->get_active_network_configuration_slot());
-    if (this->time_disconnected.load().time_since_epoch() == 0s) {
+    if (this->time_disconnected.load().time_since_epoch() == std::chrono::seconds(0)) {
```
james-ctc, line 433: what is the issue with `0s`?  and using chrono_literals?

james-ctc (approved):

lib/everest/ocpp/lib/ocpp/common/connectivity_manager.cpp
Pietfried: std::chrono::seconds is also used in other places in this file so I found that more consistent.

mlitre (commented):
There is an issue with the PR where now the share_path for ConnecivityManager always defaults to empty, this means we lose the everest_version header info. I am not sure how to nicely add it back with these changes though. That is the only real issue I have at this time

mlitre (approved):

PR #1694, opened 2026-01-06
Fix CMake project-config when using out-of-tree modules in yocto

hikinggrass:
Add missing find_package calls to libraries integrated into everest-core

Add missing install for sqlite CollectMigrationFiles.cmake file


mhei (approved):
LGTM.
See also https://github.com/EVerest/meta-everest/pull/84 and https://github.com/EVerest/meta-everest/pull/85 for reference.

PR #1696, opened 2026-01-06
feat(MCS): add MCS related types and helpers

mhei:
## Describe your changes

This commit adds some missing pieces for MCS support. It was extracted from a working branch and should
contain only material which could already be merged.

A basic SIL test is added which uses the current SLAC simulation until a MCS specific module is available.

none

### DIFF
--- a/modules/EVSE/OCPP201/conversions.cpp
+++ b/modules/EVSE/OCPP201/conversions.cpp
@@ -1816,14 +1816,6 @@ std::vector<types::iso15118::EnergyTransferMode> to_everest_allowed_energy_trans
     value.reserve(allowed_energy_transfer_modes.size());
     for (const auto& mode : allowed_energy_transfer_modes) {
         value.push_back(to_everest_allowed_energy_transfer_mode(mode));
-        // revisit: at the moment, OCPP does not yet know about MCS types
-        // so in case of DC we allow also the corresponding MCS modes
-        if (mode == ocpp::v2::EnergyTransferModeEnum::DC) {
-            value.push_back(types::iso15118::EnergyTransferMode::MCS);
-        }
-        if (mode == ocpp::v2::EnergyTransferModeEnum::DC_BPT) {
-            value.push_back(types::iso15118::EnergyTransferMode::MCS_BPT);
-        }
     }
     return value;
 }
--- a/modules/EVSE/EvseManager/evse/evse_managerImpl.cpp
+++ b/modules/EVSE/EvseManager/evse/evse_managerImpl.cpp
@@ -448,25 +448,26 @@ evse_managerImpl::handle_update_allowed_energy_transfer_modes(
 
     filtered_energy_transfer_modes.reserve(allowed_energy_transfer_modes.size());
 
-    // TODO(mlitre): Add check for incompatible type(s), for now we just filter out
-    // MCS related stuff and only if a connector type was configured at all;
+    // TODO(mlitre): Add check for incompatible type(s), for now we just transform DC stuff
+    // in case of MCS and only if a connector type was configured at all;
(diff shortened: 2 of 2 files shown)


lib/everest/iso15118/src/iso15118/d20/state/service_discovery.cpp
```diff
@@ -29,6 +29,14 @@ convert_service_id_to_service_category(const std::uint16_t service_id) {
...
     case 7:
         return iso15118::message_20::datatypes::ServiceCategory::DC_ACDP_BPT;
+    case 8:
+        return iso15118::message_20::datatypes::ServiceCategory::MCS;
+    case 9:
+        return iso15118::message_20::datatypes::ServiceCategory::MCS_BPT;
+    case 10:
+        return iso15118::message_20::datatypes::ServiceCategory::AC_DER;
+    case 65:
+        return iso15118::message_20::datatypes::ServiceCategory::Internet;
```
SebaLukas: This function is only used to convert the ev supported energy services (line 90) and not the vas. So `Internet` should be deleted

modules/EVSE/OCPP201/conversions.cpp
```diff
@@ -1810,6 +1816,14 @@ std::vector<types::iso15118::EnergyTransferMode> to_everest_allowed_energy_trans
...
     for (const auto& mode : allowed_energy_transfer_modes) {
         value.push_back(to_everest_allowed_energy_transfer_mode(mode));
+        // revisit: at the moment, OCPP does not yet know about MCS types
+        // so in case of DC we allow also the corresponding MCS modes
+        if (mode == ocpp::v2::EnergyTransferModeEnum::DC) {
+            value.push_back(types::iso15118::EnergyTransferMode::MCS);
+        }
+        if (mode == ocpp::v2::EnergyTransferModeEnum::DC_BPT) {
+            value.push_back(types::iso15118::EnergyTransferMode::MCS_BPT);
+        }
```
SebaLukas: @mlitre is this okay?

config/config-sil-mcs.yaml
```diff
@@ -0,0 +1,178 @@
+active_modules:
+  iso15118_charger:
+    module: Evse15118D20
+    config_module:
+      device: auto
+      enable_tls_key_logging: true
```
SebaLukas: Not necessarily needed in the sil if you run wireshark parallel.

mhei: Ah yes, will drop in next round.

lib/everest/iso15118/src/iso15118/d20/state/service_discovery.cpp
mhei: Yeah, got it. 💡  Will remove.

modules/EVSE/OCPP201/conversions.cpp
mlitre: I would rather have this logic be elsewhere, since the modules receiving this info would already know if MCS is supported or not. Here in this case it is just be added any time we are doing DC

cburandt (changes requested):
I think docs/source/reference/EVerest_API/evse_manager_consumer_API.yaml is missing the cMCS connector variant in the ConnectorTypeEnum.

modules/EVSE/OCPP201/conversions.cpp
mhei: Yes, EvseManager::handle_update_allowed_energy_transfer_modes is only consumer at the moment and "knows" that we are doing MCS. I can change the my original "filtering" approach there to an "replace" approach for MCS.

mhei:
I've added two commits to address @cburandt and @mlitre feedback. And I've rebased to latest main branch.

PR #1373, opened 2025-09-19
add support for choosing a payment option

catarial:
## Describe your changes

Allow manually selecting the payment option for the car simulator as an argument to the 'iso_start_v2g_session` command.

These changes depend on my PR to the Josev library https://github.com/EVerest/ext-switchev-iso15118/pull/51

Todo:
- [X] Add option to automatically set the payment type
- [X] Update docs, configs, and tests

## Checklist before requesting a review
- [x] I have performed a self-review of my code
- [ ] I have made corresponding changes to the documentation
- [ ] I read the [contribution documentation](https://github.com/EVerest/EVerest/blob/main/CONTRIBUTING.md) and made sure that my changes meet its requirements

PR #1707, opened 2026-01-12
feat(EvManager): Option to select certain payment option and enforcing it

SebaLukas:
## Describe your changes
Now it is possible for the EV side in the SIL to select a certain payment option and enforce it
This PR is an extension for #1373.

Before the changes it was not possible to select the payment option on the ev side. Josev chose this option on its own.


modules/EV/EvManager/main/car_simulation.cpp
```diff
@@ -309,22 +309,44 @@ bool CarSimulation::iso_dc_power_on(const CmdArguments& arguments) {
 
 bool CarSimulation::iso_start_v2g_session(const CmdArguments& arguments, bool three_phases) {
     const auto& energy_mode = arguments[0];
-    const auto& departure_time = std::stoi(arguments[1]);
-    const auto& e_amount = std::stoi(arguments[2]);
+    const auto& payment_option = arguments[1];
```
mlitre, line 312: If this is a new argument let's add it after all the previous ones to avoid breaking current implementations

SebaLukas: It is a new argument, but I already consider this argument while I added the register command update to support more then one command register.
So basically it should not break the old and current implementation.

mlitre: Ok that's good for me then

PR #1711, opened 2026-01-13
feat(voltage_plausibility_monitor): Voltage plausibility check during DC charging to detect configuration of hardware inconsistencies

florinmihut:
## Describe your changes
Adds a voltage plausibility check during DC charging to detect inconsistencies between voltage measurements from multiple sources.

Monitors voltage from 4 sources:
- Power Supply (always available)
- Powermeter (optional)
- Isolation Monitor (optional)
- Over Voltage Monitor (optional)
Calculates standard deviation between available sources and compares it to a configurable threshold.
Raises a fault if the standard deviation value exceeds the threshold for a configurable duration, trying to prevent false positives from brief spikes.

Handles missing or stale data:
- Requires at least 2 valid sources to compute standard deviation
- Excludes measurements older than a configurable max age

Configuration:
voltage_plausibility_std_deviation_threshold_V: Maximum allowed standard deviation (default: 50.0 V)
voltage_plausibility_fault_duration_ms: Duration SD must exceed threshold before fault (default: 10000 ms)
voltage_plausibility_measurement_max_age_ms: Maximum age of measurements before considered stale (default: 3000 ms)

Implementation
New VoltagePlausibilityMonitor class following the same pattern as OverVoltageMonitor
New error type: evse_manager/VoltagePlausibilityFault (High severity)
This helps detect configuration or hardware issues by comparing voltage readings from independent sources.

### DIFF
--- a/modules/EVSE/EvseManager/CMakeLists.txt
+++ b/modules/EVSE/EvseManager/CMakeLists.txt
@@ -20,6 +20,7 @@ target_sources(${MODULE_NAME}
         backtrace.cpp
         PersistentStore.cpp
         over_voltage/OverVoltageMonitor.cpp
+        voltage_plausibility/VoltagePlausibilityMonitor.cpp
 )
 
 target_link_libraries(${MODULE_NAME}
--- a/modules/EVSE/EvseManager/EvseManager.cpp
+++ b/modules/EVSE/EvseManager/EvseManager.cpp
@@ -500,6 +500,16 @@ void EvseManager::ready() {
             });
 
         } else if (config.charge_mode == "DC") {
+            // Create voltage plausibility monitor for DC charging
+            voltage_plausibility_monitor = std::make_unique<VoltagePlausibilityMonitor>(
+                [this](const std::string& description) {
+                    if (this->error_handling) {
+                        this->error_handling->raise_voltage_plausibility_fault(description);
+                    }
+                },
+                config.voltage_plausibility_max_spread_threshold_V,
+                std::chrono::milliseconds(config.voltage_plausibility_fault_duration_ms));
+
             if (connector_type.has_value() and connector_type.value() == types::evse_manager::ConnectorTypeEnum::cMCS) {
                 transfer_modes.push_back(types::iso15118::EnergyTransferMode::MCS);
                 update_supported_energy_transfers(types::iso15118::EnergyTransferMode::MCS);
@@ -551,6 +561,10 @@ void EvseManager::ready() {
                     internal_over_voltage_monitor->reset();
                     internal_over_voltage_monitor->start_monitor();
                 }
+                if (voltage_plausibility_monitor) {
+                    voltage_plausibility_monitor->reset();
+                    voltage_plausibility_monitor->start_monitor();
+                }
             });
 
             r_hlc[0]->subscribe_current_demand_finished([this] {
@@ -558,6 +572,9 @@ void EvseManager::ready() {
                 sae_bidi_active = false;
                 if (not r_over_voltage_monitor.empty()) {
                     r_over_voltage_monitor[0]->call_stop();
+                }
+                if (voltage_plausibility_monitor) {
+                    voltage_plausibility_monitor->stop_monitor();
                 }
                 if (internal_over_voltage_monitor) {
                     internal_over_voltage_monitor->stop_monitor();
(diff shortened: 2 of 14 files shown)


modules/EVSE/OCPP201/error_handling.hpp
```diff
@@ -34,6 +34,7 @@ const std::unordered_map<std::string, std::string> MREC_ERROR_MAP = {
     {"evse_manager/MREC22ResistanceFault", "CX022"},
     {"evse_manager/MREC11CableCheckFault", "CX011"},
     {"evse_manager/MREC5OverVoltage", "CX005"},
+    {"evse_manager/VoltagePlausibilityFault", ""},
```
Pietfried: Since  this doesn't map to any MREC it is not required here

modules/EVSE/OCPP/error_mapping.hpp
```diff
@@ -36,6 +36,7 @@ const std::unordered_map<std::string, std::pair<ocpp::v16::ChargePointErrorCode,
     {"evse_manager/MREC22ResistanceFault", {ocpp::v16::ChargePointErrorCode::OtherError, "CX022"}},
     {"evse_manager/MREC11CableCheckFault", {ocpp::v16::ChargePointErrorCode::OtherError, "CX011"}},
     {"evse_manager/MREC5OverVoltage", {ocpp::v16::ChargePointErrorCode::OverVoltage, "CX005"}},
+    {"evse_manager/VoltagePlausibilityFault", {ocpp::v16::ChargePointErrorCode::OtherError, ""}},
```
Pietfried: same here

modules/EVSE/EvseManager/voltage_plausibility/VoltagePlausibilityMonitor.cpp
```diff
@@ -0,0 +1,246 @@
...
+
+void VoltagePlausibilityMonitor::update_over_voltage_monitor_voltage(double voltage_V) {
+    std::lock_guard<std::mutex> lock(data_mutex_);
+    over_voltage_monitor_sample_.voltage_V = voltage_V;
+    over_voltage_monitor_sample_.timestamp = std::chrono::steady_clock::now();
+    evaluate_voltages();
+}
+
+void VoltagePlausibilityMonitor::evaluate_voltages() {
+    // Check running state and fault latch (these are only modified from main thread, so no mutex needed here)
```
Pietfried: To me it looks like this is not correct, because `trigger_fault` is called from the timer thread. You can either make `running_` and `fault_latched_` atomic or protect them using the `data_mutex_`

modules/EVSE/EvseManager/EvseManager.cpp
```diff
@@ -292,6 +292,16 @@ void EvseManager::ready() {
     charger = std::make_unique<Charger>(bsp, error_handling, r_powermeter_billing(), store,
                                         hw_capabilities.connector_type, config.evse_id);
 
+    // Create voltage plausibility monitor for DC charging
+    voltage_plausibility_monitor = std::make_unique<VoltagePlausibilityMonitor>(
```
Pietfried: We only need the `voltage_plausibility_monitor` in the `DC` case

modules/EVSE/EvseManager/EvseManager.cpp
```diff
@@ -559,6 +573,9 @@ void EvseManager::ready() {
                 if (not r_over_voltage_monitor.empty()) {
                     r_over_voltage_monitor[0]->call_stop();
                 }
+                if (voltage_plausibility_monitor) {
+                    voltage_plausibility_monitor->stop_monitor();
```
Pietfried, line 577: Should we also `stop_monitor` on unplug? I am not sure if we currently always go through `current_demand_finish` e.g. in an error case

modules/EVSE/EvseManager/CMakeLists.txt
```diff
@@ -57,6 +57,7 @@ target_sources(${MODULE_NAME}
         "random_delay/uk_random_delayImpl.cpp"
         "dc_external_derate/dc_external_derateImpl.cpp"
         "over_voltage/OverVoltageMonitor.cpp"
+        "voltage_plausibility/VoltagePlausibilityMonitor.cpp"
```
Pietfried:
```
"dc_external_derate/dc_external_derateImpl.cpp"
"over_voltage/OverVoltageMonitor.cpp"
```
were added to the autogenerated part of this file. These should go to L21

PR #1713, opened 2026-01-13
fix(OCPP): improve message queue retry/timeout handling and connection resilience

marcemmers:
## Describe your changes

**Configurable websocket send timeout:**
The websocket layer had a hardcoded 1-second send timeout. When a transmission exceeded this, the message queue would pause and never resume until a new websocket connection was established. The timeout is now derived from the messageTimeout in the network connection profile (OCPP 2.x) or defaults to 10 seconds (OCPP 1.6), with a 1-second minimum enforced.

**Unified failed-send handling:**
Previously, a failed send_callback had its own inline retry/drop logic separate from the timeout/call-error path, leading to inconsistent behavior. Now the message is removed from the queue before sending, and failures are routed through handle_timeout_or_callerror uniformly. The queue is no longer paused on send failure — connection loss is detected via websocket ping/pong instead.

**OCPP 1.6 auto-reconnect:**
Added a `wants_to_be_connected` flag so the charge point automatically attempts to reconnect if the websocket stops connecting unexpectedly.

Minor fixes: Suppressed a spurious pong timeout warning when ping_interval_s is 0.

### DIFF
--- a/lib/everest/ocpp/include/ocpp/common/message_queue.hpp
+++ b/lib/everest/ocpp/include/ocpp/common/message_queue.hpp
@@ -3,6 +3,7 @@
 #ifndef OCPP_COMMON_MESSAGE_QUEUE_HPP
 #define OCPP_COMMON_MESSAGE_QUEUE_HPP
 
+#include <algorithm>
 #include <chrono>
 #include <condition_variable>
 #include <deque>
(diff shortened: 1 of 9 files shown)


Pietfried:
Hi @marcemmers , looks like some libocpp unit tests failed. Were the tests passing for you locally?

lib/everest/ocpp/include/ocpp/common/message_queue.hpp
james-ctc: Does this PR address this TODO?

lib/everest/ocpp/include/ocpp/common/message_queue.hpp
```diff
@@ -524,43 +525,27 @@ template <typename M> class MessageQueue {
...
-                        EVLOG_info << "The message in flight is not transaction related and will be dropped";
-                        if (queue_type == QueueType::Normal) {
-                            EnhancedMessage<M> enhanced_message;
-                            enhanced_message.offline = true;
-                            this->in_flight->promise.set_value(enhanced_message);
-                            this->normal_message_queue.pop_front();
-                        }
-                    }
-                    this->reset_in_flight();
+                    this->handle_timeout_or_callerror(std::nullopt);
```
james-ctc, line 650: A log would be useful here e.g.
```
EVLOG_error << "Could not send message, this is most likely because the charge point is offline.";
```

james-ctc, line 650: I'm not sure the logic here is correct.
- paused is no longer set so messages will slowly be removed from the queue
- the message has been removed from the queue hence it won't be retried
- hence: `this->in_flight->message_attempts += 1;` is never incremented again for the message

even if paused is set via a call to `pause()` a subsequent `resume()` can't attempt to resend the message since it is no longer in the queue and in_flight has been reset.
If in_flight hasn't been reset then another message send will never even be attempted.

It looks like the intent in the message queue is the following:
1. attempt to send a message
2. if send fails pause
3. when paused wait for resume
4. when resumed attempt resend (because message is still in the queue)

The change appears to be:
1. remove message from queue
2. attempt to send a message
3. if send fails reset_in_flight, which enables next message to be attempted

james-ctc (changes requested):

lib/everest/ocpp/include/ocpp/common/message_queue.hpp
Pietfried: The problem was that resume was never called in case the `send_callback` failed but the websocket didnt disconnect / reconnect. I will verify the implementation and potentially add some more tests.

Pietfried: I added additional test cases and verified the retry, timeout and error mechanism.

The change is :
1. remove message from the queue
2. attempt to send
3. if send fails, reuse handle_error_or_timeout mechanism to initiate message retry. This will re-add the message to the front of the queue and schedule the time to resend. Non-transactional messages may be send in between. Transactional messages maintain their ordering. 

If we don't apply the message retry mechanism and `send_callback` continously fails, we would never drop the message.

james-ctc (approved):
It looks to be fine.
But it is getting tricky to determine correct operation.
There are a few places where push_front() is used. Since retries involve removing and then adding back to the front on error it might be possible for some unexpected reordering to occur.
handle_timeout_or_callerror() is public and could be called at any time …

PR #1716, opened 2026-01-14
[OCPP] Fix incorrect composite schedule when duration is zero

james-ctc:
## Describe your changes
Updates to composite schedule to address issues identified in libocpp issue 1169

An incorrect composite schedule was generated when the duration was 0. The fallback vales were used rather than values from the active schedule.

Additional check added to ensure start time is not greater than end time.

https://github.com/EVerest/libocpp/issues/1169

### DIFF
--- a/lib/everest/ocpp/lib/ocpp/v16/profile.cpp
+++ b/lib/everest/ocpp/lib/ocpp/v16/profile.cpp
@@ -425,7 +425,7 @@ IntermediateProfile generate_profile_from_periods(std::vector<period_entry_t>& p
     IntermediateProfile combined{};
     DateTime current = now;
 
-    // run calculation at least once especially where current < end
+    // run calculation at least once especially where current >= end
     do {
         // find schedule to use for time: current
         DateTime earliest = end;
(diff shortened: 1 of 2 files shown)


lib/everest/ocpp/lib/ocpp/v16/profile.cpp
```diff
@@ -425,7 +425,8 @@ IntermediateProfile generate_profile_from_periods(std::vector<period_entry_t>& p
     IntermediateProfile combined{};
     DateTime current = now;
 
-    while (current < end) {
+    // run calculation at least once especially where current < end
```
Pietfried:
```suggestion
    // run calculation at least once, especially when current >= end
```

lib/everest/ocpp/lib/ocpp/v2/profile.cpp
```diff
@@ -408,7 +408,8 @@ IntermediateProfile generate_profile_from_periods(std::vector<period_entry_t>& p
     IntermediateProfile combined{};
     DateTime current = now;
 
-    while (current < end) {
+    // run calculation at least once especially where current < end
```
Pietfried:
```suggestion
    // run calculation at least once, especially when current >= end
```

Pietfried (approved):

PR #1765, opened 2026-01-21
Feature/ev response code

martinlukas84:
## Describe your changes
This PR publishes the response code via interface in Everest environment.
This is useful for process view or usage in other modules like API etc.


modules/EVSE/Evse15118D20/charger/ISO15118_chargerImpl.cpp
```diff
@@ -656,6 +656,135 @@ iso15118::session::feedback::Callbacks ISO15118_chargerImpl::create_callbacks()
         this->mod->p_charger->publish_ev_termination(termination_ctx);
     };
 
+    callbacks.response_code = [this](const iso15118::message_20::datatypes::ResponseCode& code) {
+        namespace rc = iso15118::message_20::datatypes;
+
+        switch (code) {
+        case rc::ResponseCode::OK:
+            this->mod->p_charger->publish_response_code("OK");
```
mhei: Instead of having the enum to string conversion here, I'd recommend to place it in a helper function in libiso15118 and rely on that helper here.

interfaces/ISO15118_charger.yaml
```diff
@@ -452,3 +452,7 @@ vars:
       The EV termination code as described in ISO15118-20.
     type: object
     $ref: /iso15118#/EvTermination
+  response_code:
+    description:
+      The response code send to EV
+    type: string
```
SebaLukas, line 457: I would like to have a ResponseCode type defined in the ISO15118.yaml type file and not a `string`. Then the other modules can use that as well.

modules/EVSE/Evse15118D20/charger/ISO15118_chargerImpl.cpp
SebaLukas: Yeah I agree with Michael

lib/everest/iso15118/src/iso15118/session/feedback.cpp
```diff
@@ -64,6 +64,10 @@ void Feedback::ev_termination(const std::string& ev_termination_code,
     call_if_available(callbacks.ev_termination, ev_termination_code, ev_termination_explanation);
 }
 
+void Feedback::response_code(const iso15118::message_20::datatypes::ResponseCode& code) const {
```
SebaLukas, line 67: Unit test(s) for response_code is/are missing

lib/everest/iso15118/src/iso15118/d20/state/ac_charge_loop.cpp
```diff
@@ -203,6 +203,7 @@ Result AC_ChargeLoop::feed(Event ev) {
         const auto res = handle_request(*req, m_ctx.session, false);
 
         m_ctx.respond(res);
+        m_ctx.feedback.response_code(res.response_code);
```
SebaLukas, line 206: Maybe it would be possible to call feedback.responce_code() not in every state. Maybe it is possible to call that function in the context respond function or in the Session::poll function

SebaLukas (changes requested):
See my comments.

interfaces/ISO15118_charger.yaml
martinlukas84: Ok... this new type is also from type string something like that?: 

ResponseCode:
    description: Response codes as defined in ISO15118-2 and ISO15118-20
    type: string
    enum:
      OK
      OK_CertificateExpiresSoon
      OK_NewSessionEstablished
      OK_OldSessionJoined

interfaces/ISO15118_charger.yaml
```diff
@@ -452,3 +452,8 @@ vars:
       The EV termination code as described in ISO15118-20.
     type: object
     $ref: /iso15118#/EvTermination
+  response_code:
+    description:
+      The response code send to EV
```
SebaLukas: One line or 
```
description: >-
    The response code [...]
```

types/iso15118.yaml
```diff
@@ -1857,3 +1857,47 @@ types:
       ev_termination_explanation:
         description: Contains the EV termination explanation
         type: string
+  ResponseCode:
```
SebaLukas: Add this to the other enums before the EVSEID type in line 159.

lib/everest/iso15118/src/iso15118/session/feedback.cpp
SebaLukas: Unit test is still missing

types/iso15118.yaml
```diff
@@ -1856,4 +1900,4 @@ types:
         type: string
       ev_termination_explanation:
         description: Contains the EV termination explanation
-        type: string
+        type: string
```
SebaLukas: Add again EOF

SebaLukas:
So this one can be closed because of #2690 ?

SebaLukas:
Closing this one because duplicate of #2693

PR #1770, opened 2026-01-26
Sending voltage/current to powersupply even the voltage is greater than 0

martinlukas84:
The standard ISO15518-20 [V2G20-2183] describes the provided EV target voltage and current in ChargeLoopReq from EV.

The voltage and current shall never send in the same message. In the current implementation, the voltage and current is only sent to powersupplyDC when voltage is greater than 0. But The voltage is only sent at the beginning of the chargeloop (and in other states). After that, only the current is sent peridically. This leads to the problem, that the current is never sent again during chargeloop. We can't found any limitation regarding voltage >= 0 in iso-2.  We assume that this change does not influence the existing implementation of DC.

### DIFF
--- a/modules/EVSE/EvseManager/EvseManager.cpp
+++ b/modules/EVSE/EvseManager/EvseManager.cpp
@@ -2587,6 +2587,19 @@ void EvseManager::process_dc_ev_target_voltage_current(const types::iso15118::Dc
         car_breaks_limit = true;
     }
 
+    // ISO15118-20 [V2G20-2183] allows EVs to only provide voltage or current,
+    // i.e. some EV implementations send target voltage as zero later.
+    // While this requirement is going to be dropped from the standard, we
+    // have to support older implementations by using a "cached" latest non-zero
+    // voltage the EV told us.
+    // EVs sending a current of zero has not yet been seen, so we don't care
+    // about this particular case here at the moment.
+    bool car_sent_zero_voltage{false};
+    if (almost_eq(clamped_voltage, 0.0) and ev_info.target_voltage.value_or(0.f) > 0.f) {
+        clamped_voltage = ev_info.target_voltage.value();
+        car_sent_zero_voltage = true;
+    }
+
     const auto actual_voltage =
         ev_info_snapshot.present_voltage.has_value() ? ev_info_snapshot.present_voltage.value() : clamped_voltage;
 
@@ -2611,6 +2624,9 @@ void EvseManager::process_dc_ev_target_voltage_current(const types::iso15118::Dc
         }
         if (car_breaks_limit) {
             EVLOG_warning << "EV ignores new EVSE max limits. Setting target current to new EVSE max limits";
+        }
+        if (car_sent_zero_voltage) {
+            EVLOG_warning << "EV did not provide a recent voltage. Re-using the previously communicated value again.";
         }
 
         {


mhei (commented):
I'm not sure... when lastest_target_voltage cannot be negative, then the check can be removed completely?

kwoyhope:
I'm not sure, but EvseManager::powersupply_DC_set() function checks '_voltage' parameter(='latest_target_voltage') and seems that it may throw "DC voltage/current out of limits requested: Voltage {:.2f} Current {:.2f}." -- IMHO, is this PR enough for your intention?

SebaLukas:
I cannot accept the current solution. With these changes, a voltage of 0V would be set at the power supply, which would be quite fatal for the further charging process.

You would now have to think about how to implement the case for D20 Scheduled Mode if the car only sends TargetCurrent > 0A or TargetVoltage > 0V and the other values with 0A/V. Should the last target values be used instead of 0A/0V, for example?
For D2/DIN, the car always sends both target values above 0A/0V when it wants to charge. Therefore, this would also have to be synchronized with D20 Scheduled Mode here.

mhei:
It will take some time until we can work on this again.

mhei:
@SebaLukas , @kwoyhope : I've updated this PR. We tested this on our side and it works in our MCS use-case. If you have some time, feel free to take another look.

SebaLukas (changes requested):
Looking better but for me but this is also not ideal. `apply_new_target_voltage_current` is also used in DIN, ISO-2, and D20 dynamic modes. Have you also tested these changes with the above standards?

I would rather see the changes in the `subscribe_dc_ev_target_voltage_current()` function: https://github.com/EVerest/everest-core/blob/main/modules/EVSE/EvseManager/EvseManager.cpp#L704. Then you could also save yourself `latest_non_zero_target_voltage`.
Here, it would also be necessary to check again what happens when a `TargetVoltage` or `TargetCurrent` of 0V/A is sent.

modules/EVSE/EvseManager/EvseManager.cpp
```diff
@@ -2563,6 +2563,19 @@ void EvseManager::process_dc_ev_target_voltage_current(const types::iso15118::Dc
...
+    // While this requirement is going to be dropped from the standard, we
+    // have to support older implementations by using a "cached" latest non-zero
+    // voltage the EV told us.
+    // EVs sending a current of zero has not yet been seen, so we don't care
+    // about this particular case here at the moment.
+    bool car_v2g20_2183{false};
+    if (almost_eq(clamped_voltage, 0.0) and ev_info.target_voltage.value_or(0.f) > 0.f) {
+        clamped_voltage = ev_info.target_voltage.value();
+        car_v2g20_2183 = true;
+    }
```
SebaLukas, line 2601: One could argue that this check should only be activated when a D20 session is actually active. However, this check also makes sense for D2 and DIN.

modules/EVSE/EvseManager/EvseManager.cpp
```diff
@@ -2563,6 +2563,19 @@ void EvseManager::process_dc_ev_target_voltage_current(const types::iso15118::Dc
...
     }
 
+    // ISO15118-20 [V2G20-2183] allows EVs to only provide voltage or current,
+    // i.e. some EV implementations send target voltage as zero later.
+    // While this requirement is going to be dropped from the standard, we
+    // have to support older implementations by using a "cached" latest non-zero
+    // voltage the EV told us.
+    // EVs sending a current of zero has not yet been seen, so we don't care
+    // about this particular case here at the moment.
+    bool car_v2g20_2183{false};
```
SebaLukas: Maybe choose another variable name? Kind of cryptic

modules/EVSE/EvseManager/EvseManager.cpp
```diff
@@ -2588,6 +2601,9 @@ void EvseManager::process_dc_ev_target_voltage_current(const types::iso15118::Dc
         if (car_breaks_limit) {
             EVLOG_warning << "EV ignores new EVSE max limits. Setting target current to new EVSE max limits";
         }
+        if (car_v2g20_2183) {
+            EVLOG_warning << "EV relies on ISO15118-20 [V2G20-2183], applied fixup.";
```
SebaLukas: It may cause confusion if this warning is also printed on DIN or ISO-2. I don't expect this to happen with DIN or ISO-2, but I would remove the D20 part and prefer a more general text.

SebaLukas (commented):
I like the changes 👍 
Please look into my comments

mhei:
Thanks for your feedback. I tried to find a better flag name and adapted the message. While at, rebased to latest main.

SebaLukas (approved):

SebaLukas:
Please check the DCO errors. I thinkl your signed-off message is not correct.

PR #1780, opened 2026-01-30
Bugfix: Fix BSP enable handling

FaHaGit:
## Describe your changes

Until now, the evse_board_support interface description did not document how a BSP should behave when an enable “false” command is set via the interface. The correct behavior was requested in the [Zulip chat](https://lfenergy.zulipchat.com/#narrow/channel/417675-EVerest-General/topic/evse_board_support.20.22enable.22.20cmd/with/570601793) and has now been corrected. The PhyVersoBSP module did not implement the enable correctly until now. This has also been fixed in this PR.

--


modules/HardwareDrivers/EVSE/PhyVersoBSP/connector_1/evse_board_supportImpl.cpp
Pietfried, line 1: `last_pwm_raw` is cached but unconditionally applied to the HW using `set_pwm`. Is that intentional? If the port was previously disabled with `handle_enable(false)`, should `set_pwm` calls be suppressed until the re-enable?

FaHaGit: Hi Piet, that's a good question. Can you clarify the correct behavior internally? Until now, it has been implemented unconditionally. If you want, I can also adjust the description for pmwOff(), pwmOn(), and pwmF() in case enable() was previously sent with "false". So, as it is currently implemented, we have also implemented it in our BSP implementations.

Pietfried: We think it's best to cache the PWM in the EVerest driver and only apply it once `enable()` is called. This keeps this complexity out of the MCU.

In your current implementation `last_pwm_raw` is cached but unconditionally applied to the HW. `pmwOff`, `pwmOn` and `pwmF` should contain a check if we're disabled and just cache instead of applying the PWM directly. Once enabled again, we can send the cached `last_pwm_raw` (this is already implemented)

PR #1782, opened 2026-01-30
[OCPP 1.6] setting JSON configuration to include schema validation

james-ctc:
## Describe your changes

fix: OCPP 1.6 set could accept invalid values that would fail schema validation

Once an invalid value is allowed then EVerest won't start due to invalid configuration. This fix checks that the new value will pass schema validation before allowing the set to be actioned.

### DIFF
--- a/lib/everest/ocpp/lib/ocpp/v16/charge_point_configuration.cpp
+++ b/lib/everest/ocpp/lib/ocpp/v16/charge_point_configuration.cpp
@@ -313,12 +313,12 @@ void ChargePointConfiguration::setChargepointInformationProperty(json& user_conf
 }
 
 bool ChargePointConfiguration::validate(const std::string_view& schema_file, const json& object) {
-    fs::path schema_path = ocpp_main_path / "profile_schemas" / schema_file;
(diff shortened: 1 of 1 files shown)


lib/everest/ocpp/lib/ocpp/v16/charge_point_configuration.cpp
```diff
@@ -312,6 +312,25 @@ void ChargePointConfiguration::setChargepointInformationProperty(json& user_conf
     }
 }
 
+bool ChargePointConfiguration::validate(const std::string_view& schema_file, const json& object) {
+    fs::path schema_path = ocpp_main_path / "profile_schemas" / schema_file;
+    std::ifstream ifs(schema_path);
+    auto schema_json = json::parse(ifs);
```
Pietfried: This could be moved inside the try block to avoid issues in case the schema file doesn't exist

james-ctc: updated

Pietfried (approved):

PR #1784, opened 2026-02-02
libiso15118: Add response msg interval control

FaHaGit:
## Describe your changes

There are EV implementations that can configure request messages within a few milliseconds, which can currently trigger a flood of internal communication. To avoid performance issues, we should therefore wait a certain amount of time after receiving a request message before sending the next response message. The delay of maximum 100 ms should not cause any problems, as the most critical time of the ISO 15118-20 (V2G_SECC_Msg_Performance_Time of the charging loop) is 250 ms.

Already successfully tested in the SIL environment (with a higher delay, because the josev lib based implementation needs ~500ms to configure a request message).

Changes:
- introduce MIN_RESPONSE_INTERVAL_MS to manage time between responses
- add last_response_tx_time to track last response timestamp
- implement delay to ensure minimum interval between response messages

- [X ] I have performed a self-review of my code
- [X ] I read the [contribution documentation](https://github.com/EVerest/EVerest/blob/main/CONTRIBUTING.md) and made sure that my changes meet its requirements

### DIFF
--- a/lib/everest/iso15118/src/iso15118/session/iso.cpp
+++ b/lib/everest/iso15118/src/iso15118/session/iso.cpp
@@ -16,6 +16,7 @@
 namespace iso15118 {
 
 static constexpr auto SESSION_IDLE_TIMEOUT_MS = 5000;
+static constexpr auto MIN_RESPONSE_INTERVAL_MS = 100; // minimum time between two response messages
 
 static void log_sdp_packet(const iso15118::io::SdpPacket& sdp) {
     static constexpr auto ESCAPED_BYTE_CHAR_COUNT = 4;
@@ -140,10 +141,11 @@ void Session::push_control_event(const d20::ControlEvent& event) {
 
 TimePoint const& Session::poll() {
     const auto now = get_current_time_point();
+    // FIXME (aw): proper timeout handling!
+    next_session_event = offset_time_point_by_ms(now, SESSION_IDLE_TIMEOUT_MS);
 
     if (not state.connected) {
         // nothing happened so far, just return
-        next_session_event = offset_time_point_by_ms(now, SESSION_IDLE_TIMEOUT_MS);
         return next_session_event;
     }
 
@@ -223,19 +225,35 @@ TimePoint const& Session::poll() {
         // FIXME(sl): check result!
     }
 
-    const auto [got_response, payload_size, payload_type, response_type] = message_exchange.check_and_clear_response();
+    if (message_exchange.has_response()) {
+        const auto [payload_type, response_type] = message_exchange.peek_response_meta();
+        // Before we send back the response message, we check the time between two response messages
+        // sent out. If this is less than MIN_RESPONSE_INTERVAL_MS, we delay the response message to
+        // avoid potential performance issues.
+        if (not response_send_after.has_value()) {
+            response_send_after = now;
+            if (last_response_tx_time.has_value()) {
+                const auto elapsed = std::chrono::duration_cast<std::chrono::milliseconds>(
+                    get_current_time_point() - last_response_tx_time.value());
+                if (elapsed < std::chrono::milliseconds(MIN_RESPONSE_INTERVAL_MS)) {
+                    response_send_after =
+                        offset_time_point_by_ms(last_response_tx_time.value(), MIN_RESPONSE_INTERVAL_MS);
+                }
+            }
+        }
 
-    if (got_response) {
-        const auto response_size = setup_response_header(response_buffer, payload_type, payload_size);
-        connection->write(response_buffer, response_size);
-
-        timeouts.start_timeout(d20::TimeoutType::SEQUENCE, d20::TIMEOUT_SEQUENCE);
-
-        // FIXME (aw): this is hacky ...
-        log.exi(static_cast<uint16_t>(payload_type), response_buffer + io::SdpPacket::V2GTP_HEADER_SIZE, payload_size,
-                session::logging::ExiMessageDirection::TO_EV);
-
-        ctx.feedback.v2g_message(response_type);
+        // Send the response as soon as the response interval is reached
+        if (response_send_after.has_value() && now < response_send_after.value()) {
+            next_session_event = response_send_after.value();
+        } else {
+            response_send_after.reset();
+            // FIXME(fh): Currently, the response is still generated when the request is received.
+            // This may have changed after the delay. Ideally, the processing of the request message
+            // and the generated of the response message should be decoupled from each other.
+            send_response(payload_type, response_type);
+        }
+    } else {
+        response_send_after.reset();
     }
 
     if (ctx.session_stopped or ctx.session_paused) {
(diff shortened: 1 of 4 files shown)


SebaLukas (changes requested):
In general I like the PR. Wait for some time before sending the response is a good idea.

My problem with this implementation is that it blocks the whole process for 100ms.
That could be really a problem in some cases.

I need to think how to fix this problem not blocking the whole process.

FaHaGit:
@SebaLukas what is your opinion on how the delay should be implemented? Would you prefer it to be implemented as it currently is, with a fixed response interval (i.e., keeping the time between two res messages constant), or to wait a fixed amount of time after receiving a request message? The first approach has been implemented here. I think that's quite good because we don't necessarily delay, but only when the EV generates request messages quickly. Or do you think a delay after a response would always be useful in order to have the chance to react to the content of the request message? I can also imagine a hybrid form: max 100ms interval + constant delay (e.g. 50ms) after a request.

SebaLukas:
I would prefer the second option, as a delay here does not block the entire process.

I would start a separate timer after sending the response and check this timer after receiving the request to see whether the timer is over 100 ms or not.
If the car sends the request after the response very fast in less than 100 ms, the session poll function should return early and check later whether the timer is over 100 ms. Only then is the response created via the state machine.
The advantage of this is that various events can also be received by EVerest during the waiting time. And the events can also be reflected in the response.

SebaLukas:
> The problem is that if the EV itself needs 100 ms to send the request message, for example, then the delay timer expires immediately and we generate the response directly without first communicating the request content internally.

I think this is actually okay. Do you have an example of what could be problematic if the state machine responds directly to the req message? 
The state machine is handling the req message and react depending on the req. And right now this behaivour is happing in libiso15118 and EvseV2G.

SebaLukas:
Okay :+1: Lets see if we will have some problems in the future then we can add this extra 50ms. But for now I think we dont need them.

Pietfried:
The problem should be mitigated with https://github.com/EVerest/everest-core/commit/16bc99ba54b7c253d6d0479c9ace5f77d082a17b . @FaHaGit Can you please verify if we still need this suggested change?

FaHaGit:
> The problem should be mitigated with [16bc99b](https://github.com/EVerest/everest-core/commit/16bc99ba54b7c253d6d0479c9ace5f77d082a17b) . @FaHaGit Can you please verify if we still need this suggested change?

I believe that sooner or later we will need such a mechanism. At least for hardware platforms that do not have much performance. At the time, we had to implement this delay for request messages into the EvseV2G module because EV implementations with a request interval of <30ms overloaded the entire internal MQTT communication channel. The stability of the entire system should not depend on external factors (such as the request interval).

lib/everest/iso15118/src/iso15118/session/iso.cpp
```diff
@@ -250,11 +268,30 @@ TimePoint const& Session::poll() {
...
 }
 
+void Session::send_response(io::v2gtp::PayloadType payload_type, message_20::Type response_type) {
+    const auto [got_response, payload_size, stored_payload_type, stored_response_type] =
+        message_exchange.check_and_clear_response();
+    if (not got_response) {
+        return;
+    }
+    payload_type = stored_payload_type;
+    response_type = stored_response_type;
```
SebaLukas: I dont get quite these lines

lib/everest/iso15118/src/iso15118/session/iso.cpp
```diff
@@ -223,19 +225,35 @@ TimePoint const& Session::poll() {
...
-        const auto response_size = setup_response_header(response_buffer, payload_type, payload_size);
-        connection->write(response_buffer, response_size);
-
-        timeouts.start_timeout(d20::TimeoutType::SEQUENCE, d20::TIMEOUT_SEQUENCE);
-
-        // FIXME (aw): this is hacky ...
-        log.exi(static_cast<uint16_t>(payload_type), response_buffer + io::SdpPacket::V2GTP_HEADER_SIZE, payload_size,
-                session::logging::ExiMessageDirection::TO_EV);
+    if (message_exchange.has_response()) {
+        const auto [payload_type, response_type] = message_exchange.peek_response_meta();
```
SebaLukas: Why do you need the `payload_type` and `response_type` already here? You are passing them into `send_response()` but there are overwritten?

SebaLukas (commented):
Sorry for the late review. Definitely better but one/two things we need to discuss before merging

lib/everest/iso15118/src/iso15118/session/iso.cpp
FaHaGit: accidentally left in there while debugging

FaHaGit:
I've revised the PR. We've been testing it internally for a while now, and it turned out there was a bug in the timeout calculation, as well as another issue where the `close` method was being called even though a response was still pending.

lib/everest/iso15118/src/iso15118/session/iso.cpp
```diff
@@ -140,15 +141,16 @@ void Session::push_control_event(const d20::ControlEvent& event) {
...
 
     if (not state.connected) {
         // nothing happened so far, just return
-        next_session_event = offset_time_point_by_ms(now, SESSION_IDLE_TIMEOUT_MS);
         return next_session_event;
     }
 
     // check for new data to read
-    if (state.new_data) {
+    if (state.new_data and not message_exchange.has_response()) {
```
SebaLukas: That really shouldn't happen, should it? That the EV sends another request message without receiving a response and does not wait for the timeout

lib/everest/iso15118/src/iso15118/session/iso.cpp
```diff
@@ -202,7 +204,7 @@ TimePoint const& Session::poll() {
     }
 
     // check for complete sdp packet
-    if (packet.is_complete()) {
+    if (packet.is_complete() and not message_exchange.has_response()) {
```
SebaLukas: Same as above

SebaLukas (dismissed):
Two comments but overall looking good!
In the next weeks I will have another look into this again. But I think we should merge this PR before that

SebaLukas:
Please squash your commits into one commit and modify the message before merging :)

lib/everest/iso15118/src/iso15118/session/iso.cpp
FaHaGit: It simply ensures that no new request is processed until the response has been sent. The EV shouldn't send out another request, thats right. But it makes the system more robust in case it does..

FaHaGit: I'm fine with removing the check if you think that's better

FaHaGit: I can only imagine a scenario where the EV might send a SessionStopReq even though it hasn't yet received a response to the last request in the event of an emergency shutdown.

SebaLukas: If the Ev sends another request before getting the response then there is something wrong. And this case is kind of already handled in the `MessageExchange::set_request()` throwing an exception.
I can not think of one scenario the Ev sends another req in a not failure state.
So please remove both `message_exchange.has_response()` calls :+1:

SebaLukas (approved):

PR #1789, opened 2026-02-04
Enumerate evse managers correctly in API module

### LINKED ISSUE #1710: Everest module API: evse_id is not initialized from the mapping
### Describe the bug

`API::init` enumerates evse managers based on the order these are listed in the connections: 

```
int evse_id = 1;
...
evse_id++
```

This is quite error prone, since the IDs should be assigned to the managers. We have a configuration with two managers; although the mappings were set up correctly (AC with id = 1, DC with id = 2), the links to the API modules were done in the opposite order, i.e. first we connected DC, then AC. Either we should mention this in the documentation (I could not find any reference) or the approach should be changed.

### EVerest Domain

Utility

### Affected EVerest Module

https://github.com/EVerest/everest-core/blob/main/modules/API/API/API.cpp

### To Reproduce

_No response_

### Anything else?

_No response_

d-u-d-e:
## Describe your changes
Fix: `evse_id` should not be defined based on the order the evse managers are listed among the API connections.
Moved all the code requiring the `evse_id` parameter from the `init` function to the `ready` function where it's safe to call `call_get_evse` on the evse managers. 

Refactor: performed a couple of suggested changes from clang tidy

Fix #1710 

### DIFF
--- a/modules/API/API/API.cpp
+++ b/modules/API/API/API.cpp
@@ -26,11 +26,8 @@ SessionInfo::SessionInfo() :
 }
 
 bool SessionInfo::is_state_charging(const SessionInfo::State current_state) {
-    if (current_state == State::AuthRequired || current_state == State::Charging ||
-        current_state == State::ChargingPausedEV || current_state == State::ChargingPausedEVSE) {
-        return true;
-    }
-    return false;
+    return current_state == State::AuthRequired || current_state == State::Charging ||
+           current_state == State::ChargingPausedEV || current_state == State::ChargingPausedEVSE;
 }
 
 void SessionInfo::reset() {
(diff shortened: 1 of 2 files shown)


modules/API/API/API.cpp
```diff
@@ -757,12 +687,73 @@ void API::ready() {
         this->charger_information = r_charger_information.at(0)->call_get_charger_information();
     }
 
+    // The following API commands require the EVSE managers to be initialized
```
Pietfried:
```suggestion
    this->evse_manager_check.wait_ready();

    // The following API commands require the EVSE managers to be initialized
```

Pietfried (approved):
lgtm, please rebase and squash your commits into one

PR #1791, opened 2026-02-04
EvseManager: set three phases correctly

jalbert514:
Currently the three phases boolean is always set to true, even for BSPs that only support one phase. This leads to an incorrect pp_amp. The changes here try to fix that.

### DIFF
--- a/modules/EVSE/EvseManager/EvseManager.cpp
+++ b/modules/EVSE/EvseManager/EvseManager.cpp
@@ -235,7 +235,22 @@ void EvseManager::init() {
 
         signalNrOfPhasesAvailable(ac_nr_phases_active);
 
-        bsp->set_three_phases(c.max_phase_count_import);
+        const auto get_max_phases = [](int max_phase_count) -> AcPhases {
+            switch (max_phase_count) {
+            case 1:
+                return AcPhases::SinglePhase;
+            case 2:
+                return AcPhases::TwoPhases;
+            case 3:
+                return AcPhases::ThreePhases;
+
+            default:
+                EVLOG_warning << "Invalid max_phase_count of " << max_phase_count << ". Falling back to single phase";
+            }
+            return AcPhases::SinglePhase;
+        };
+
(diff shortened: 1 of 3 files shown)


modules/EVSE/EvseManager/EvseManager.cpp
```diff
@@ -235,7 +235,7 @@ void EvseManager::init() {
 
         signalNrOfPhasesAvailable(ac_nr_phases_active);
 
-        bsp->set_three_phases(c.max_phase_count_import);
+        bsp->set_three_phases(c.max_phase_count_import != 1);
```
Pietfried:
```suggestion
        bsp->set_three_phases(c.max_phase_count_import == 3);
```

Pietfried (changes requested):
Thanks for the contribution! This would even be more appropriate.

```cpp
bsp->set_three_phases(c.max_phase_count_import == 3);
```

modules/EVSE/EvseManager/EvseManager.cpp
jalbert514: thanks for the feedback, but is the mentioned change correct? that would allow for a maximum current of 70A inside this (modules/EVSE/EvseManager/IECStateMachine.cpp) if the max phase count is 2

or is this never the case, since there are only the 3 or 1 phase options? not really sure about this

Pietfried: Good point, this should be adressed as well! Per IEC61851 70A should only be allowed for single phase charging. Here's a suggestion of how this could be implemented. What do you think about this? Feel free to cherry-pick: https://github.com/EVerest/everest-core/commit/c76a9bc25560e7a968bf26ea3f659d653f8e129e

jalbert514: thanks for the suggestion, added your changes and rebased again

Pietfried (approved):

PR #1795, opened 2026-02-05
feat(util): Added new fixed vector util

mlitre:
This new feature provides some of the "dynamic" functionality of vectors to an std array.

### DIFF
--- a/lib/everest/util/include/everest/util/vector/fixed_vector.hpp
+++ b/lib/everest/util/include/everest/util/vector/fixed_vector.hpp
@@ -1,0 +1,747 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2020 - 2026 Pionix GmbH and Contributors to EVerest
+
+/**
+ * @file fixed_vector.hpp
+ * @brief Provides a std::vector-like container with fixed capacity, avoiding dynamic memory allocation.
+ *
+ * The fixed_vector is a sequence container that encapsulates a fixed-size array. It provides an interface
+ * similar to std::vector but does not allocate memory on the heap. Its capacity is determined at compile time
+ * by the template parameter N. This makes it suitable for real-time and embedded applications where
+ * dynamic memory allocation is disallowed or undesirable.
+ */
+
+#pragma once
+
+#include <array>
+#include <cstddef>
+#include <initializer_list>
+#include <iterator>
+#include <memory>
+#include <stdexcept>
+#include <type_traits>
+#include <utility>
+#include <vector>
+
+namespace everest::lib::util {
+
+/**
+ * @brief A container with std::vector-like interface but with fixed capacity.
+ *
+ * This class mimics the behavior of std::vector but stores its elements in a fixed-size internal buffer,
+ * avoiding heap allocations. The capacity is specified at compile time.
+ * If the number of elements exceeds the capacity, it results in an exception (`std::length_error`).
+ *
+ * @tparam T The type of elements.
+ * @tparam N The maximum number of elements the vector can hold (its capacity).
+ */
+template <typename T, std::size_t N> class fixed_vector {
+public:
+    //- Member types
+    using value_type = T;
+    using size_type = std::size_t;
+    using difference_type = std::ptrdiff_t;
+    using reference = value_type&;
+    using const_reference = const value_type&;
+    using pointer = value_type*;
+    using const_pointer = const value_type*;
+    using iterator = pointer;
+    using const_iterator = const_pointer;
+    using reverse_iterator = std::reverse_iterator<iterator>;
+    using const_reverse_iterator = std::reverse_iterator<const_iterator>;
+
+    /**
+     * @brief Constructs an empty fixed_vector.
+     */
+    constexpr fixed_vector() noexcept = default;
+
+    /**
+     * @brief Destroys the fixed_vector, calling destructors for all contained elements.
+     */
+    ~fixed_vector() {
+        clear();
+    }
+
+    /**
+     * @brief Copy constructor. Constructs the vector with a copy of the contents of other.
+     * @param other another fixed_vector object to be used as source to initialize the elements of the container with.
+     */
+    fixed_vector(const fixed_vector& other) {
+        copy_construct_from(other);
+    }
+
+    /**
+     * @brief Move constructor. Constructs the vector with the contents of other using move semantics.
+     * @details This constructor has two behaviors depending on `T`:
+     *          1. If `T` is `nothrow_move_constructible`, this operation is `noexcept` and highly efficient.
+     *          2. If `T`'s move constructor can throw, this provides a strong guarantee: if a failure occurs, the
+     *             source `other` is restored to its original state. However, the exception is **swallowed**, and
+     * `*this` will be empty. The caller must be prepared to handle an empty vector even if `other` was not.
+     * @param other another fixed_vector object to be used as source.
+     */
+    fixed_vector(fixed_vector&& other) noexcept(std::is_nothrow_move_constructible_v<T>) {
+        move_construct_from(std::move(other));
+    }
+
+    /**
+     * @brief Constructs the vector with the contents of the initializer list.
+     * @param init initializer list to initialize the elements of the container with.
+     * @throws std::length_error if the size of the initializer list is greater than the capacity of the vector.
+     */
+    fixed_vector(std::initializer_list<T> init) {
+        if (init.size() > N) {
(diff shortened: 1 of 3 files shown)


lib/everest/util/include/everest/util/vector/fixed_vector.hpp
```diff
@@ -0,0 +1,447 @@
...
+    /**
+     * @brief Move constructor. Constructs the vector with the contents of other using move semantics.
+     * @details If `T` is not move-constructible, the new vector will be empty and `other` will be unmodified.
+     * After the move, `other` is left in a valid but empty state if `T` is move-constructible.
+     * @param other another fixed_vector object to be used as source.
+     */
+    fixed_vector(fixed_vector&& other) noexcept {
+        if constexpr (std::is_move_constructible_v<T>) {
+            for (auto& elem : other) {
+                emplace_back(std::move(elem));
```
cburandt: Looks like this is a valid warning from codacy here (can throw from within emplace_back despite being declared noexcept)

lib/everest/util/tests/vector/fixed_vector_tests.cpp
```diff
@@ -0,0 +1,330 @@
...
+    ThrowsOnCopy::should_throw = true;
+    try {
+        vec.push_back(ThrowsOnCopy{});
+    } catch (const std::runtime_error&) {
+        // exception caught
+    }
+
+    EXPECT_EQ(vec.size(), 2); // Strong exception guarantee: state is unchanged
+    ThrowsOnCopy::should_throw = false;
+}
```
cburandt, line 381: None of these TESTs is using pop_back(). Add one test which calls pop_back() beyond emptying the fixed_vector.

lib/everest/util/tests/vector/fixed_vector_tests.cpp
```diff
@@ -0,0 +1,330 @@
...
+    EXPECT_EQ(vec.size(), 2);
+    EXPECT_EQ(it, vec.end());
+    EXPECT_EQ(vec[0], 1);
+    EXPECT_EQ(vec[1], 2);
+
+    // Erase everything
+    it = vec.erase(vec.begin(), vec.end());
+    EXPECT_TRUE(vec.empty());
+    EXPECT_EQ(it, vec.end());
+}
```
cburandt, line 353: Maybe add test of an invalid range (first>last).

lib/everest/util/include/everest/util/vector/fixed_vector.hpp
```diff
@@ -0,0 +1,447 @@
...
+        clear();
+    }
+
+    /**
+     * @brief Copy constructor. Constructs the vector with a copy of the contents of other.
+     * @param other another fixed_vector object to be used as source to initialize the elements of the container with.
+     */
+    fixed_vector(const fixed_vector& other) {
+        for (const auto& elem : other) {
+            push_back(elem);
```
cburandt: Assuming T's constructor can throw: What if push_back fails for some random element?

lib/everest/util/include/everest/util/vector/fixed_vector.hpp
```diff
@@ -0,0 +1,447 @@
...
+        }
+    }
+
+    /**
+     * @brief Move constructor. Constructs the vector with the contents of other using move semantics.
+     * @details If `T` is not move-constructible, the new vector will be empty and `other` will be unmodified.
+     * After the move, `other` is left in a valid but empty state if `T` is move-constructible.
```
cburandt: Just doing nothing if sth. is not move-constructible may be dangerous. Somebody might bang his head against the wall before he finds the cause of the problem. Could this be made to fail at compile-time when used with non-movable types?

cburandt (changes requested):
This code looks like you spent quite some thought on it. Find a few comments below that may be worth to be taken care of. The most important one concerns the move-constructible criteria you apply in some function.

lib/everest/util/include/everest/util/vector/fixed_vector.hpp
```diff
@@ -0,0 +1,447 @@
...
+            clear();
+            if constexpr (std::is_move_constructible_v<T>) {
+                for (auto& elem : other) {
+                    emplace_back(std::move(elem));
+                }
+                other.clear();
+            }
+        }
+        return *this;
+    }
```
djchhp:
```suggestion
    template < class U = T, std::enable_if_t<std::is_move_constructible_v<U>>* = nullptr>
    fixed_vector& operator=(fixed_vector&& other) noexcept {
        if (this != &other) {
            clear();
            for (auto& elem : other) {
                emplace_back(std::move(elem));
            }
            other.clear();
        }
        return *this;
    }

```

This results in compile time failure for non move_constructible classes, like

```
 class no_move{
    public:
        no_move(no_move&&) = delete;
        no_move() = default;
    };

    fixed_vector<no_move, 6> no_move_vec;
    fixed_vector<no_move, 6> no_move_vec_b;
   //// This line won't compile (where for e.g. std::string it would)
    no_move_vec = std::move(no_move_vec_b);

```

mlitre:
@cburandt all of your comments should have been addressed if there are still any issues feel free to bring them up :)

lib/everest/util/tests/vector/fixed_vector_tests.cpp
```diff
@@ -0,0 +1,586 @@
...
+
+    EXPECT_FALSE(vec1 == vec4);
+    EXPECT_TRUE(vec1 != vec4);
+
+    EXPECT_TRUE(empty1 == empty2);
+    EXPECT_FALSE(empty1 != empty2);
+
+    EXPECT_FALSE(vec1 == empty1);
+    EXPECT_TRUE(vec1 != empty1);
+}
```
cburandt, line 586:
```suggestion
}

struct ThrowsAt_Nth_Move {
    ThrowsAt_Nth_Move() = default;
    ThrowsAt_Nth_Move(const ThrowsAt_Nth_Move&) = default;
    ThrowsAt_Nth_Move& operator=(const ThrowsAt_Nth_Move&) = default;
    ThrowsAt_Nth_Move(const ThrowsAt_Nth_Move&&) {
        move_countdown -= 1;
        if (move_countdown == -1) {
            throw std::runtime_error("Throwing on move construction");
        }
    }
    ThrowsAt_Nth_Move& operator=(const ThrowsAt_Nth_Move&&) {
        move_countdown -= 1;
        if (move_countdown == -1) {
            throw std::runtime_error("Throwing on move construction");
        }
        return *this;
    }
    static int move_countdown;
};
int ThrowsAt_Nth_Move::move_countdown = -1;

TEST(FixedVectorTest, FailToMoveConstruct) {
    ThrowsAt_Nth_Move::move_countdown = 2;
    fixed_vector<ThrowsAt_Nth_Move, 5> vec;
    vec.emplace_back();
    vec.emplace_back();
    vec.emplace_back();
    vec.emplace_back();

    auto moved_to = fixed_vector<ThrowsAt_Nth_Move, 5>(std::move(vec));

    EXPECT_EQ(moved_to.size(), 0);
    EXPECT_EQ(vec.size(), 4);
}

TEST(FixedVectorTest, FailToMoveAssign) {
    ThrowsAt_Nth_Move::move_countdown = 2;
    fixed_vector<ThrowsAt_Nth_Move, 5> vec;
    vec.emplace_back();
    vec.emplace_back();
    vec.emplace_back();
    vec.emplace_back();

    fixed_vector<ThrowsAt_Nth_Move, 5> moved_to;

    EXPECT_EQ(moved_to.size(), 0);

    moved_to = std::move(vec);

    EXPECT_EQ(moved_to.size(), 0);
    EXPECT_EQ(vec.size(), 4);
}
```

This tests for strong exception-safety for the move construction/assignment.

mlitre:
> > With the tests you have added, I see exactly the issue you are pointing out. I would agree we should succeed or fail completely so I'll implement a mechanism to move back all of the elements that were already moved, and as a best effort we would assume when moving back that there will not be any issue.
> 
> Reading some bits about the STL in the meantime, I have to assume that the tests may be a bit too cruel - using a std::vector will fail these tests as well ... But maybe one can simply require the contained type to provide noexcept move-constructors? (what STL implementations do: require the move constructors to be declared noexcept, or otherwise fallback to the copy-constructors, or simply throw themselves if there is no copy-constructor to fallback to)

Yes I also saw in the end that we are technically giving way more guarantees than even the standard STL containers do. So I am pretty happy with where the code stands at the moment. Let me know if you see any new issues. I will probably rebase and squash all the commits into one.

SebaLukas (approved):
Wow! I really like the implementation 👍

PR #1799, opened 2026-02-08
framework: replace mqtt-c with mosquitto

hikinggrass:


lib/everest/framework/lib/mqtt_abstraction_impl.cpp
```diff
@@ -325,57 +323,16 @@ json MQTTAbstractionImpl::get(const MQTTRequest& request) {
...
-
-                        return;
-                    }
-                } else {
-                    // probably we got hit by a signal, nothing to do, just reloop
-                }
-            }
+            // Create event handler and register clients
+            everest::lib::io::event::fd_event_handler ev_handler;
+            ev_handler.register_event_handler(&*this->mqtt_client);
```
djchhp: isn't that just 
`mqtt_client.get()`

lib/everest/framework/lib/mqtt_abstraction_impl.cpp
```diff
@@ -190,25 +186,28 @@ void MQTTAbstractionImpl::subscribe(const std::string& topic, QOS qos) {
...
 
     this->subscribed_topics.insert(topic);
 
-    mqtt_subscribe(&this->mqtt_client, topic.c_str(), max_qos_level);
-    notify_write_data();
+    const auto result = this->mqtt_client->subscribe(
+        topic,
+        [this](everest::lib::io::mqtt::mosquitto_cpp& client,
+               everest::lib::io::mqtt::mosquitto_cpp::message const& message) {
+            this->message_queue.add(std::unique_ptr<Message>(new Message{message.topic, message.payload}));
```
djchhp: I suggest `std::make_unique`

dorezyuk:
I think you will need this https://github.com/bazelbuild/bazel-central-registry/pull/7746 to make it work with bazel

lib/everest/framework/lib/mqtt_abstraction_impl.cpp
hikinggrass: fixed

lib/everest/framework/lib/mqtt_abstraction_impl.cpp
hikinggrass: added a ctor to Message to be able to use make_unique here

hikinggrass:
> I think you will need this [bazelbuild/bazel-central-registry#7746](https://github.com/bazelbuild/bazel-central-registry/pull/7746) to make it work with bazel

this needed a patch to work, but should build now

lib/everest/framework/lib/error/error_manager_req.cpp
```diff
@@ -74,6 +74,17 @@ void ErrorManagerReq::on_error_raised(const Error& error) {
...
+        database->get_errors({ErrorFilter(TypeFilter(error.type)), ErrorFilter(SubTypeFilter(error.sub_type)),
+                              ErrorFilter(OriginFilter(error.origin))});
+    if (!errors.empty()) {
+        std::stringstream ss;
+        ss << "Error of type '" << error.type << "' and sub type '" << error.sub_type
+           << "' is already raised, ignoring new error";
+        ss << std::endl << "Error object: " << nlohmann::json(error).dump(2);
+        EVLOG_error << ss.str();
+        return;
+    }
```
andistorm, line 84: Looks mostly good to me. I didn't get in which cases this happens, but if this happen quite regularily,  we probably should think about changing/fixing the subscription mechanism instead and/or don't log it as error since this can be quite verbose if this happens often

third-party/bazel/BUILD.mosquitto.bazel
dorezyuk, line 1: can't we not just apply the patch to the file? It seems its mostly copy and pasted

hikinggrass: it drops ~90% of the file content and all the 2.1 dependencies to cjson etc. but feel free to change it into a patch :)

lib/everest/framework/lib/mqtt_abstraction_impl.cpp
```diff
@@ -174,14 +169,14 @@ void MQTTAbstractionImpl::publish(const std::string& topic, const std::string& d
         return;
     }
 
-    const MQTTErrors error = mqtt_publish(&this->mqtt_client, topic.c_str(), data.c_str(), data.size(), publish_flags);
-    if (error != MQTT_OK) {
-        EVLOG_error << fmt::format("MQTT Error {}", mqtt_error_str(error));
+    const auto error =
+        this->mqtt_client->publish(topic, data, mqtt_qos, retain, everest::lib::io::mqtt::PropertiesBase{});
```
djchhp: I expect this works, since the mosquitto client in lib::io calles `moquitto_threaded_set(true)` and if so, this is likely the most efficient way. The design concept of lib::io would be to pass the data on via an event to be handled in the event_loop, like it is done later for the `subscribe`

lib/everest/io/src/mqtt/mosquitto_cpp.cpp
```diff
@@ -372,8 +372,11 @@ ErrorCode mosquitto_cpp::connect(const std::string_view& unix_domain_socket, std
 
 ErrorCode mosquitto_cpp::connect_impl(const std::string_view& bind_address, const std::string_view& host,
                                       std::uint16_t port, std::uint16_t keepalive_seconds) {
-    const char* bind_to = bind_address.empty() ? nullptr : bind_address.data();
-    return convertEC(mosquitto_connect_bind_async(client.get(), host.data(), port, keepalive_seconds, bind_to));
+    if (bind_address.empty()) {
+        return convertEC(mosquitto_connect_async(client.get(), host.data(), port, keepalive_seconds));
+    }
+    return convertEC(
+        mosquitto_connect_bind_async(client.get(), host.data(), port, keepalive_seconds, bind_address.data()));
```
djchhp: If not strictly necessary, please revert, as the original version is well tested.

djchhp (changes requested):
Looking good in general. Please check mosquitto_cpp.cpp changes could be reverted.

lib/everest/io/src/mqtt/mosquitto_cpp.cpp
hikinggrass: reverted

lib/everest/io/src/can/socket_can_handler.cpp
```diff
@@ -58,7 +67,11 @@ int socket_can_handler::rx(uint32_t& can_id, uint8_t& len8_dlc, can_payload& pay
             payload.clear();
             payload.assign(frame.data, frame.data + frame.can_dlc);
             can_id = frame.can_id;
+            #ifdef CAN_MAX_RAW_DLC
             len8_dlc = frame.len8_dlc;
+            #else
+            // FIXME: do we need this information?
```
djchhp: `len8_dlc = 0` is correct  whenever `len <= 8 `. 
In a raw can frame the length value could can physically be up to 15, even though only only a length up to 8 bytes is possible. len8_dlc carries that raw value in case len > 8. If the kernel does not support forwarding the raw value, we set it to zero. The extra information is already lost anyway. Depending on the driver it is quite likely any such would be dropped.

lib/everest/io/src/can/socket_can_handler.cpp
```diff
@@ -58,7 +67,11 @@ int socket_can_handler::rx(uint32_t& can_id, uint8_t& len8_dlc, can_payload& pay
             payload.clear();
             payload.assign(frame.data, frame.data + frame.can_dlc);
```
djchhp, line 62: can_dlc is deprecated. It should be replaced by len

lib/everest/io/src/socket/socket.cpp
```diff
@@ -27,6 +27,10 @@
 #include <everest/io/event/unique_fd.hpp>
 #include <everest/io/socket/socket.hpp>
 
+#ifndef PACKET_IGNORE_OUTGOING
+#define PACKET_IGNORE_OUTGOING 11
```
djchhp: That turns a compile time error into a runtime error. The correct approach would be to fallback to Berkeley Packet Filter, which is a bit painful. 
I suggest to at least screaming out loudly a compiler warning, that by default results in a compile failure, but can be deactivated. An other option would be to define via compiler command line manually.

lib/everest/io/src/tun_tap/tap_handler.cpp
```diff
@@ -3,6 +3,8 @@
 #include <cstring>
 #include <everest/io/tun_tap/tap_handler.hpp>
 #include <fcntl.h>
+#include <sys/socket.h>
+#include <sys/types.h>
```
djchhp, line 7: Why is that necessary?

djchhp (changes requested):

lib/everest/io/src/tun_tap/tap_handler.cpp
hikinggrass: Building with musl complained about incomplete types when including if.h without these

lib/everest/framework/include/utils/types.hpp
```diff
@@ -167,6 +167,7 @@ NLOHMANN_JSON_NAMESPACE_END
 
 namespace Everest {
 inline constexpr int mqtt_get_config_timeout_ms = 5000;
+inline constexpr std::size_t mqtt_get_config_retries = 1;
```
dorezyuk, line 164: maybe move that to where you need it

lib/everest/framework/lib/mqtt_abstraction_impl.cpp
```diff
@@ -21,72 +23,51 @@
...
+#include <everest/io/mqtt/mosquitto_cpp.hpp>
+#include <everest/io/mqtt/mqtt_client.hpp>
 #include <everest/logging.hpp>
 
 #include <utils/mqtt_abstraction_impl.hpp>
 
 namespace Everest {
 const auto mqtt_keep_alive = 600;
-const auto mqtt_get_timeout_ms = 5000; ///< Timeout for MQTT get in milliseconds
+const auto mqtt_get_timeout_ms = 5000;       ///< Timeout for MQTT get in milliseconds
```
dorezyuk: constexpr?

lib/everest/framework/lib/mqtt_abstraction_impl.cpp
```diff
@@ -21,72 +23,51 @@
...
-    this->mqtt_client.publish_response_callback_state = &this->message_queue;
-
-    this->disconnect_event_fd = eventfd(0, 0);
-    if (this->disconnect_event_fd == -1) {
-        throw EverestInternalError("Could not setup eventfd for disconnect event");
+    if (mqtt_settings.uses_socket()) {
+        this->mqtt_server_socket_path = mqtt_settings.broker_socket_path;
+    } else {
+        this->mqtt_server_address = mqtt_settings.broker_host;
+        this->mqtt_server_port = std::to_string(mqtt_settings.broker_port);
```
dorezyuk: i think you could change the mqtt_server_port type to int

dorezyuk: you could make this uint16_t

lib/everest/framework/lib/mqtt_abstraction_impl.cpp
dorezyuk: maybe just {}

lib/everest/framework/lib/mqtt_abstraction_impl.cpp
```diff
@@ -174,14 +169,14 @@ void MQTTAbstractionImpl::publish(const std::string& topic, const std::string& d
         return;
     }
 
-    const MQTTErrors error = mqtt_publish(&this->mqtt_client, topic.c_str(), data.c_str(), data.size(), publish_flags);
-    if (error != MQTT_OK) {
-        EVLOG_error << fmt::format("MQTT Error {}", mqtt_error_str(error));
+    const auto error =
+        this->mqtt_client->publish(topic, data, mqtt_qos, retain, everest::lib::io::mqtt::PropertiesBase{});
+    if (error != everest::lib::io::mqtt::ErrorCode::Success) {
+        EVLOG_error << fmt::format("MQTT Error");
```
dorezyuk: you don't need to format that

lib/everest/framework/lib/mqtt_abstraction_impl.cpp
```diff
@@ -194,25 +189,30 @@ void MQTTAbstractionImpl::subscribe(const std::string& topic, QOS qos) {
     BOOST_LOG_FUNCTION();
     const std::lock_guard<std::mutex> lock(topics_mutex);
 
-    auto max_qos_level = 0;
+    auto max_qos_level = everest::lib::io::mqtt::mqtt_client::QoS::at_most_once;
     switch (qos) {
```
dorezyuk: can we make this into a converter

third-party/bazel/BUILD.mosquitto.bazel
```diff
@@ -0,0 +1,137 @@
...
+        "//conditions:default": ["WITH_THREADING"],
+    }) + select({
+        "//conditions:default": [],
+    }),
+    visibility = ["//visibility:public"],
+    deps = [
+        ":bundled_deps",
+        ":config_header",
+        ":mosquitto_common",
+        "@openssl//:ssl",
```
dorezyuk: you have to say `//third-party/bazel/openssl:ssl`

lib/everest/io/src/can/socket_can_handler.cpp
hikinggrass: done

lib/everest/framework/everestjs/everestjs.cpp
```diff
@@ -592,7 +592,7 @@ static Napi::Value boot_module(const Napi::CallbackInfo& info) {
         if (mqtt_broker_socket_path.empty()) {
             auto mqtt_broker_port = Everest::defaults::MQTT_BROKER_PORT;
             try {
-                mqtt_broker_port = std::stoi(mqtt_server_port);
+                mqtt_broker_port = std::stoul(mqtt_server_port);
```
james-ctc, line 595: just curious why stoul() rather than stoi() since the port number is 16 bits?

lib/everest/framework/lib/mqtt_abstraction_impl.cpp
```diff
@@ -21,72 +23,72 @@
...
-    if (this->disconnect_event_fd == -1) {
-        throw EverestInternalError("Could not setup eventfd for disconnect event");
-    }
+    this->mqtt_client = std::make_unique<everest::lib::io::mqtt::mqtt_client>(mqtt_reconnect_timeout_ms);
+    this->mqtt_client->set_error_handler([&](int error, std::string const& msg) {
+        if (error) {
+            EVLOG_error << fmt::format("MQTT error: {}", msg);
+        }
+    });
+    this->mqtt_client->set_callback_connect([this](auto& mqtt, auto, auto, auto const&) { this->on_mqtt_connect(); });
```
james-ctc, line 86: MQTT v5 brokers can set a property that could lower the keep alive setting for the connection.
To check this on_mqtt_connect() would need access to the properties argument which is available with mosquitto_connect_v5_callback_set();

lib/everest/framework/everestjs/everestjs.cpp
hikinggrass: I thought `stoul` at least returns an unsigned value

djchhp (dismissed):
Looks good to me

lib/everest/io/src/can/socket_can_handler.cpp
hikinggrass: dropped this from the PR completely

lib/everest/io/src/socket/socket.cpp
hikinggrass: dropped this from the PR completely

lib/everest/framework/lib/mqtt_abstraction_impl.cpp
```diff
@@ -194,25 +188,19 @@ void MQTTAbstractionImpl::subscribe(const std::string& topic, QOS qos) {
...
-        break;
-    }
+    auto max_qos_level = to_io_qos(qos, everest::lib::io::mqtt::mqtt_client::QoS::at_most_once);
 
     this->subscribed_topics.insert(topic);
 
-    mqtt_subscribe(&this->mqtt_client, topic.c_str(), max_qos_level);
-    notify_write_data();
+    this->ev_handler.add_action([this, topic, max_qos_level]() {
+        const auto result = this->mqtt_client->subscribe(
```
Pietfried, line 196: nit: We should log on failure

lib/everest/framework/lib/mqtt_abstraction_impl.cpp
```diff
@@ -21,72 +23,71 @@
...
     this->mqtt_is_connected = false;
 
-    this->mqtt_client.publish_response_callback_state = &this->message_queue;
-
-    this->disconnect_event_fd = eventfd(0, 0);
-    if (this->disconnect_event_fd == -1) {
-        throw EverestInternalError("Could not setup eventfd for disconnect event");
-    }
+    this->mqtt_client = std::make_unique<everest::lib::io::mqtt::mqtt_client>(mqtt_reconnect_timeout_ms);
+    this->mqtt_client->set_error_handler([&](int error, std::string const& msg) {
```
Pietfried: should also be fine

```suggestion
    this->mqtt_client->set_error_handler([](int error, std::string const& msg) {
```

Pietfried (dismissed):

Pietfried (dismissed):

lib/everest/framework/lib/error/error_manager_req.cpp
hikinggrass: removed the verbose error message and added a FIXME comment to look into this issue in the future

lib/everest/framework/lib/mqtt_abstraction_impl.cpp
hikinggrass: fixed

lib/everest/framework/lib/mqtt_abstraction_impl.cpp
hikinggrass: This should already be reported via the error handler, so we should not log here twice

Pietfried (approved):

PR #1802, opened 2026-02-10
Fix(EvseV2G): PnC enabled check did not work correctly in ServiceDiscovery state

SebaLukas:
## Describe your changes
Replacing the c array `payment_option_list` with std::vector. With a std::vector it is easier to check if certain payment options are supported or not.

In the ServiceDiscovery state this warning log message was falsly printed:
```PnC is not allowed without TLS-communication. Correcting value to '1' (ExternalPayment)```
This fix changes that.

### DIFF
--- a/modules/EVSE/EvseV2G/charger/ISO15118_chargerImpl.cpp
+++ b/modules/EVSE/EvseV2G/charger/ISO15118_chargerImpl.cpp
@@ -5,6 +5,7 @@
 #include "log.hpp"
 #include "tools.hpp"
 #include "v2g_ctx.hpp"
+#include <algorithm>
 #include <string.h>
 #include <string_view>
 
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: modules/EVSE/EvseV2G/v2g_ctx.cpp [hunk 1]
```diff
@@ -193,8 +193,8 @@ void v2g_ctx_init_charging_values(struct v2g_context* const ctx) {
         ctx->evse_v2g_data.rcd = (int)0; // 0 if RCD has not detected an error
         ctx->contactor_is_closed = false;
 
-        ctx->evse_v2g_data.payment_option_list[0] = iso2_paymentOptionType_ExternalPayment;
-        ctx->evse_v2g_data.payment_option_list_len = (uint8_t)1; // One option must be set
+        ctx->evse_v2g_data.payment_option_list.reserve(iso2_paymentOptionType_2_ARRAY_SIZE);
+        ctx->evse_v2g_data.payment_option_list = {iso2_paymentOptionType_ExternalPayment};
```
Pietfried: The call to `reserve` has no effect if an initializer list is used afterwards
```suggestion
        ctx->evse_v2g_data.payment_option_list.push_back(iso2_paymentOptionType_ExternalPayment);
```

modules/EVSE/EvseV2G/v2g_ctx.cpp
```diff
@@ -261,8 +261,7 @@ void v2g_ctx_init_charging_values(struct v2g_context* const ctx) {
     if (ctx->hlc_pause_active != true) {
         ctx->session.iso_selected_payment_option = iso2_paymentOptionType_ExternalPayment;
     } else {
-        ctx->evse_v2g_data.payment_option_list[0] = ctx->session.iso_selected_payment_option;
-        ctx->evse_v2g_data.payment_option_list_len = (uint8_t)1; // One option must be set
+        ctx->evse_v2g_data.payment_option_list = {ctx->session.iso_selected_payment_option};
```
Pietfried: Same as above applies here:
```suggestion
        ctx->evse_v2g_data.payment_option_list.clear();
        ctx->evse_v2g_data.payment_option_list.push_back(ctx->session.iso_selected_payment_option);
```

modules/EVSE/EvseV2G/charger/ISO15118_chargerImpl.cpp
```diff
@@ -181,34 +182,25 @@ void ISO15118_chargerImpl::handle_set_charging_parameters(types::iso15118::Setup
...
-                dlog(DLOG_LEVEL_WARNING, "Unable to configure PaymentOption %s",
-                     types::iso15118::payment_option_to_string(option).c_str());
+    if (not v2g_ctx->hlc_pause_active) {
+        if (payment_options.size() > 0) {
+            v2g_ctx->evse_v2g_data.payment_option_list.clear();
+            const auto max_payment_options =
+                std::min(static_cast<size_t>(iso2_paymentOptionType_2_ARRAY_SIZE), payment_options.size());
+            for (size_t i = 0; i < max_payment_options; i++) {
+                v2g_ctx->evse_v2g_data.payment_option_list.push_back(
+                    static_cast<iso2_paymentOptionType>(payment_options.at(i)));
```
Pietfried: I would avoid a static cast here because it can easily break if e.g. the order in the definition of the enum values changes. I would recommend to keep the comparisons.

modules/EVSE/EvseV2G/charger/ISO15118_chargerImpl.cpp
```diff
@@ -181,34 +182,25 @@ void ISO15118_chargerImpl::handle_set_charging_parameters(types::iso15118::Setup
...
             }
-        }
 
-        if (v2g_ctx->evse_v2g_data.payment_option_list_len == 0) {
+        } else {
             dlog(DLOG_LEVEL_ERROR, "No valid PaymentOptions configured, falling back to ExternalPayment");
-            v2g_ctx->evse_v2g_data.payment_option_list[0] = iso2_paymentOptionType_ExternalPayment;
-            v2g_ctx->evse_v2g_data.payment_option_list_len = 1;
+            v2g_ctx->evse_v2g_data.payment_option_list = {iso2_paymentOptionType_ExternalPayment};
```
Pietfried: same with initializer list. Use push_back instead
```suggestion
            v2g_ctx->evse_v2g_data.payment_option_list.push_back(iso2_paymentOptionType_ExternalPayment);
```

modules/EVSE/EvseV2G/iso_server.cpp
```diff
@@ -812,8 +813,8 @@ static enum v2g_event handle_iso_payment_service_selection(struct v2g_connection
      * this also covers the case that the peer sends any invalid/unknown payment option
      * in the message; if we are not happy -> bail out
      */
-    for (idx = 0; idx < conn->ctx->evse_v2g_data.payment_option_list_len; idx++) {
-        if (conn->ctx->evse_v2g_data.payment_option_list[idx] == req->SelectedPaymentOption) {
+    for (idx = 0; idx < conn->ctx->evse_v2g_data.payment_option_list.size(); idx++) {
+        if (conn->ctx->evse_v2g_data.payment_option_list.at(idx) == req->SelectedPaymentOption) {
             list_element_found = true;
             conn->ctx->p_charger->publish_selected_payment_option(
                 static_cast<types::iso15118::PaymentOption>(req->SelectedPaymentOption));
```
Pietfried: The static_cast should also be avoided here. Maybe a conversion function makes sense since its needed in multiple places

Pietfried (changes requested):

PR #1815, opened 2026-02-12
fix: resolve data races in OCPP and OCPP201 modules

Pietfried:
## Describe your changes

Replace unprotected ``std::map`` members with ``everest::lib::util::monitor`` for thread-safe access:
* evse_soc_map (both modules): concurrent reads from powermeter callbacks and writes from ev_info/session_finished callbacks
* evse_evcc_id (OCPP201): concurrent reads/writes across session event processing and ev_info callbacks
* evse_ready_map (both modules): migrated from manual mutex/condition_variable to monitor, removing
* evse_ready_mutex, evse_ready_cv, and all_evse_ready()

Additionally:
* event_queue (OCPP201): add missing session_event_mutex lock in firmware/log status callbacks
* event_queue (OCPP): add missing event_mutex lock in subscribe_powermeter_public_key_ocmf callback

### DIFF
--- a/modules/EVSE/OCPP/OCPP.cpp
+++ b/modules/EVSE/OCPP/OCPP.cpp
@@ -278,7 +278,7 @@ void OCPP::process_session_event(int32_t evse_id, const types::evse_manager::Ses
         EVLOG_debug << "Connector#" << ocpp_connector_id << ": "
                     << "Received SessionFinished";
         // ev side disconnect
-        this->evse_soc_map[evse_id].reset();
+        this->evse_soc_map.handle()->at(evse_id).reset();
         this->charge_point->on_session_stopped(ocpp_connector_id, session_event.uuid);
     } else if (session_event.event == types::evse_manager::SessionEventEnum::ReservationStart) {
         this->charge_point->on_reservation_start(ocpp_connector_id);
@@ -295,9 +295,10 @@ void OCPP::init_evse_subscriptions() {
         evse->subscribe_powermeter([this, evse_id](types::powermeter::Powermeter powermeter) {
             ocpp::Measurement measurement;
             measurement.power_meter = conversions::to_ocpp_power_meter(powermeter);
-            if (this->evse_soc_map[evse_id].has_value()) {
+            auto evse_soc_map_handle = this->evse_soc_map.handle();
+            if (evse_soc_map_handle->at(evse_id).has_value()) {
                 // soc is present, so add this to the measurement
-                measurement.soc_Percent = ocpp::StateOfCharge{this->evse_soc_map[evse_id].value()};
+                measurement.soc_Percent = ocpp::StateOfCharge{evse_soc_map_handle->at(evse_id).value()};
             }
             if (powermeter.temperatures.has_value()) {
                 measurement.temperature_C = conversions::to_ocpp_temperatures(powermeter.temperatures.value());
@@ -307,7 +308,7 @@ void OCPP::init_evse_subscriptions() {
 
         evse->subscribe_ev_info([this, evse_id](const types::evse_manager::EVInfo& ev_info) {
             if (ev_info.soc.has_value()) {
-                this->evse_soc_map[evse_id] = ev_info.soc.value();
+                this->evse_soc_map.handle()->at(evse_id) = ev_info.soc.value();
             }
         });
 
@@ -346,6 +347,7 @@ void OCPP::init_evse_subscriptions() {
         });
 
         evse->subscribe_powermeter_public_key_ocmf([this, evse_id](std::string public_key_ocmf) {
+            std::lock_guard<std::mutex> lg(this->event_mutex);
             if (!this->started) {
                 this->event_queue.emplace(evse_id, PowermeterPublicKey{public_key_ocmf});
                 return;
@@ -401,10 +403,17 @@ void OCPP::init_evse_connector_map() {
 }
 
 void OCPP::init_evse_maps() {
-    std::lock_guard<std::mutex> lk(this->evse_ready_mutex);
-    for (size_t evse_id = 1; evse_id <= this->r_evse_manager.size(); evse_id++) {
-        this->evse_ready_map[evse_id] = false;
-        this->evse_soc_map[evse_id] = std::nullopt;
+    {
+        auto ready_handle = this->evse_ready_map.handle();
+        for (size_t evse_id = 1; evse_id <= this->r_evse_manager.size(); evse_id++) {
+            (*ready_handle)[evse_id] = false;
+        }
+    }
+    {
+        auto soc_handle = this->evse_soc_map.handle();
+        for (size_t evse_id = 1; evse_id <= this->r_evse_manager.size(); evse_id++) {
+            (*soc_handle)[evse_id] = std::nullopt;
+        }
     }
 }
 
@@ -452,16 +461,6 @@ void OCPP::handle_config_key(const ocpp::v16::KeyValue& kv) {
         pnc_config.central_contract_validation_allowed = ocpp::conversions::string_to_bool(kv.value.value());
         set_pnc_config(pnc_config);
     }
-}
-
-bool OCPP::all_evse_ready() {
-    for (auto const& [evse, ready] : this->evse_ready_map) {
-        if (!ready) {
-            return false;
-        }
-    }
-    EVLOG_info << "All EVSE ready. Starting OCPP1.6 service";
-    return true;
 }
 
 ocpp::v16::ChargingRateUnit get_unit_or_default(const std::string& unit_string) {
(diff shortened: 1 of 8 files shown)


modules/EVSE/OCPP/OCPP.hpp
```diff
@@ -178,12 +179,9 @@ class OCPP : public Everest::ModuleBase {
                                                          // EVerests evse and connector id
     std::map<int32_t, int32_t> connector_evse_index_map; // provides access to r_evse_manager index by
                                                          // using OCPP connector id
-    std::map<int32_t, bool> evse_ready_map;
-    std::map<int32_t, std::optional<float>> evse_soc_map;
+    everest::lib::util::monitor<std::map<int32_t, bool>> evse_ready_map;
+    everest::lib::util::monitor<std::map<int32_t, std::optional<float>>> evse_soc_map;
```
djchhp, line 183: Just an observation:
It looks like `std::map<X, std::optional<Y>> data`  is redundant. Is there any information derived from the key existing in the map without a value?

modules/EVSE/OCPP/OCPP.cpp
```diff
@@ -307,7 +308,7 @@ void OCPP::init_evse_subscriptions() {
 
         evse->subscribe_ev_info([this, evse_id](const types::evse_manager::EVInfo& ev_info) {
             if (ev_info.soc.has_value()) {
-                this->evse_soc_map[evse_id] = ev_info.soc.value();
+                this->evse_soc_map.handle()->at(evse_id) = ev_info.soc.value();
```
djchhp, line 311: A matter of preferred style maybe. I find it easier to read and to understand if the different actions are perfromed on different lines
```
auto ref = evse_soc_map.handle();       // get the handle
ref->at(evse_id) = ev_indo.soc.value()  // assing the value.
```

modules/EVSE/OCPP/OCPP.cpp
```diff
@@ -401,10 +403,9 @@ void OCPP::init_evse_connector_map() {
 }
 
 void OCPP::init_evse_maps() {
-    std::lock_guard<std::mutex> lk(this->evse_ready_mutex);
     for (size_t evse_id = 1; evse_id <= this->r_evse_manager.size(); evse_id++) {
-        this->evse_ready_map[evse_id] = false;
-        this->evse_soc_map[evse_id] = std::nullopt;
+        (*this->evse_ready_map.handle())[evse_id] = false;
+        (*this->evse_soc_map.handle())[evse_id] = std::nullopt;
```
djchhp: This is very likely not correct.

The call to handle returns an object, that grants mutually exclusive access to the protected data. That access is bound to the lifetime of the object. What happens here, is that a temporary object is created and destructed immediately after usage, effectively locking and unlocking mutexed on every iteration.

The original code protected data access not per iteration, but for the whole loop.

Also beware of the danger of deadlock: You try to lock a mutex while holding another one. If that happens somewhere else but in reverse order some unfortunate timing will result in severe problems.

So what can be done?

If the locking per iteration is intended, I think the code is ok.

If both maps need to be locked at the same time, the only viable solution is to pack them in one struct and protect that struct with a single monitor. (btw. this problem did not arise because of using monitor, it would be the same for a simple mutex)

If both maps do not need to stay in sync, it would be possible to one after the other

```
{ 
   auto ref = this->evse_ready_map.handle();
   auto& obj = *ref;
   for (size_t evse_id = 1; evse_id <= this->r_evse_manager.size(); evse_id++) {
           obj[evse_id] = false;
   }
} // the first monitor is released here

{ 
   auto ref = this->evse_soc_map.handle();
   auto& obj = *ref;
   for (size_t evse_id = 1; evse_id <= this->r_evse_manager.size(); evse_id++) {
           obj[evse_id] = std::nullopt;
   }
}
```

modules/EVSE/OCPP201/OCPP201.cpp
```diff
@@ -171,14 +171,13 @@ int32_t get_connector_id_from_error(const Everest::error::Error& error) {
...
     for (size_t evse_id = 1; evse_id <= this->r_evse_manager.size(); evse_id++) {
-        this->evse_ready_map[evse_id] = false;
-        this->evse_soc_map[evse_id] = std::nullopt;
+        (*this->evse_ready_map.handle())[evse_id] = false;
+        (*this->evse_soc_map.handle())[evse_id] = std::nullopt;
         this->evse_hardware_capabilities_map[evse_id] = types::evse_board_support::HardwareCapabilities{};
         this->evse_supported_energy_transfer_modes[evse_id] = {};
         this->evse_service_renegotiation_supported[evse_id] = false;
-        this->evse_evcc_id[evse_id] = "";
+        (*this->evse_evcc_id.handle())[evse_id] = "";
```
djchhp: A similar observation to above.
- locking on every iteration
- data consistency not guaranteed due to locks released by temporary objects

What is protected here? If it is the consistent data for the different maps, this needs to go into one struct with one monitor around it.

modules/EVSE/OCPP/OCPP.cpp
```diff
@@ -346,6 +347,7 @@ void OCPP::init_evse_subscriptions() {
         });
 
         evse->subscribe_powermeter_public_key_ocmf([this, evse_id](std::string public_key_ocmf) {
+            std::lock_guard<std::mutex> lg(this->event_mutex);
```
djchhp, line 350: Could be a monitor on event_queue

modules/EVSE/OCPP201/OCPP201.cpp
```diff
@@ -415,6 +404,7 @@ void OCPP201::init() {
             this->charge_point->on_firmware_update_status_notification(
                 status.request_id, conversions::to_ocpp_firmware_status_enum(status.firmware_update_status));
         } else {
+            std::scoped_lock lock(this->session_event_mutex);
```
djchhp, line 423: could be a monitor on event_queue

modules/EVSE/OCPP201/OCPP201.cpp
```diff
@@ -424,6 +414,7 @@ void OCPP201::init() {
             this->charge_point->on_log_status_notification(
                 conversions::to_ocpp_upload_logs_status_enum(status.log_status), status.request_id);
         } else {
+            std::scoped_lock lock(this->session_event_mutex);
```
djchhp, line 433: could be a monitor on event_queue

djchhp (changes requested):

modules/EVSE/OCPP/OCPP.hpp
Pietfried: A key existing in the map without a value would indicate that there is no SoC for this EVSE because the SoC is not known or no session is active

modules/EVSE/OCPP/OCPP.cpp
Pietfried: I updated it using the approach you suggested. The two maps don't need to stay in sync. They serve independent purposes (startup ready signal and SoC tracking).

modules/EVSE/OCPP201/OCPP201.cpp
Pietfried: see comment above

modules/EVSE/OCPP/OCPP.cpp
```diff
@@ -403,9 +403,17 @@ void OCPP::init_evse_connector_map() {
 }
 
 void OCPP::init_evse_maps() {
-    for (size_t evse_id = 1; evse_id <= this->r_evse_manager.size(); evse_id++) {
-        (*this->evse_ready_map.handle())[evse_id] = false;
-        (*this->evse_soc_map.handle())[evse_id] = std::nullopt;
+    {
+        auto ready_handle = this->evse_ready_map.handle();
+        for (size_t evse_id = 1; evse_id <= this->r_evse_manager.size(); evse_id++) {
```
djchhp, line 408: use std::size_t

modules/EVSE/OCPP/OCPP.cpp
```diff
@@ -403,9 +403,17 @@ void OCPP::init_evse_connector_map() {
...
+            (*ready_handle)[evse_id] = false;
+        }
+    }
+    {
+        auto soc_handle = this->evse_soc_map.handle();
+        for (size_t evse_id = 1; evse_id <= this->r_evse_manager.size(); evse_id++) {
```
djchhp, line 414: use std::size_t

modules/EVSE/OCPP201/OCPP201.cpp
```diff
@@ -171,13 +171,29 @@ int32_t get_connector_id_from_error(const Everest::error::Error& error) {
 }
 
 void OCPP201::init_evse_maps() {
+    {
+        auto ready_handle = this->evse_ready_map.handle();
+        for (size_t evse_id = 1; evse_id <= this->r_evse_manager.size(); evse_id++) {
```
djchhp, line 176: use std::size_t

modules/EVSE/OCPP201/OCPP201.cpp
```diff
@@ -171,13 +171,29 @@ int32_t get_connector_id_from_error(const Everest::error::Error& error) {
...
+            (*ready_handle)[evse_id] = false;
+        }
+    }
+    {
+        auto soc_handle = this->evse_soc_map.handle();
+        for (size_t evse_id = 1; evse_id <= this->r_evse_manager.size(); evse_id++) {
```
djchhp, line 182: use std::size_t

modules/EVSE/OCPP201/OCPP201.cpp
```diff
@@ -171,13 +171,29 @@ int32_t get_connector_id_from_error(const Everest::error::Error& error) {
...
+            (*soc_handle)[evse_id] = std::nullopt;
+        }
+    }
+    {
+        auto evse_evcc_id_handle = this->evse_evcc_id.handle();
+        for (size_t evse_id = 1; evse_id <= this->r_evse_manager.size(); evse_id++) {
```
djchhp, line 188: use std::size_t

modules/EVSE/OCPP201/OCPP201.cpp
```diff
@@ -171,13 +171,29 @@ int32_t get_connector_id_from_error(const Everest::error::Error& error) {
...
+            (*evse_evcc_id_handle)[evse_id] = "";
+        }
+    }
+
     for (size_t evse_id = 1; evse_id <= this->r_evse_manager.size(); evse_id++) {
```
djchhp, line 193: use std::size_t

modules/EVSE/OCPP/OCPP.hpp
```diff
@@ -178,12 +179,9 @@ class OCPP : public Everest::ModuleBase {
                                                          // EVerests evse and connector id
     std::map<int32_t, int32_t> connector_evse_index_map; // provides access to r_evse_manager index by
                                                          // using OCPP connector id
-    std::map<int32_t, bool> evse_ready_map;
-    std::map<int32_t, std::optional<float>> evse_soc_map;
+    everest::lib::util::monitor<std::map<int32_t, bool>> evse_ready_map;
```
djchhp, line 182: use std::int32_t

modules/EVSE/OCPP201/OCPP201.hpp
```diff
@@ -139,23 +140,20 @@ class OCPP201 : public Everest::ModuleBase {
     std::string source_ext_limit;
 
     // key represents evse_id, value indicates if ready
-    std::map<int32_t, bool> evse_ready_map;
-    std::map<int32_t, std::optional<float>> evse_soc_map;
+    everest::lib::util::monitor<std::map<int32_t, bool>> evse_ready_map;
```
djchhp, line 143: use std::int32_t

modules/EVSE/OCPP201/OCPP201.hpp
```diff
@@ -139,23 +140,20 @@ class OCPP201 : public Everest::ModuleBase {
...
+    everest::lib::util::monitor<std::map<int32_t, std::optional<float>>> evse_soc_map;
```
djchhp, line 144: use std::int32_t

modules/EVSE/OCPP/OCPP.cpp
djchhp: will be done in follow up PR

PR #1816, opened 2026-02-12
feat(framework): Add thread pool to message handler to avoid backed up mqtt queues

### LINKED ISSUE #1800: Timing issues with new framework MQTT handler
### Describe the bug

We have been encountering some issues during ISO 15118-20 charging, and have potentially tracked them down to the framework's MQTT handling.

The basic symptom is that we start getting "Power budget expired, falling back to 0." log entries early during the charge loop, leading to charging being interrupted. (Seen only in ISO 15118-20, scheduled mode, recently.)

During the investigation, we noticed that some processes were getting delays when waiting for the results of an interface cmd call. We added logging into the framework to notify us of all subscribe handlers and cmd calls which took more than a certain amount of time. It showed that, in regular operation, as seen in the SIL (on a powerful CPU), some calls can take around 130ms to be processed.

In the mentioned critical situation, on a target CPU, we observe that some of these processing times start increasing, even well above one second, and probably lead to a chain of delays, finally resulting in the named power budget loss.

Reducing the pressure on Evse15118D20 by "slowing down" the charge cycle loop (PR https://github.com/EVerest/everest-core/pull/1784) does seem to vastly reduce the probability of the occurrence of this issue.

We managed to provoke the scenario in the SIL with two different methods:
* adding an artificial 250ms delay in `subscribe_voltage_current()` in EvseManager
* adding additional load to Evse15118D20 by scripted MQTT publishing of `update_isolation_status` to the ISO15118_charger interface.

As pointed out to us in the recent Car Communication Working Group call, MQTT handling was reworked in everest-framework commit https://github.com/EVerest/everest-framework/commit/906e4b6c783a3bf420ffba3d2751647cfc479f59 (from PR https://github.com/EVerest/everest-framework/pull/292), included from release v0.25.0.

We reverted this commit (albeit within everest-core, since everest-framework has been migrated there), to find our problems resolved.

We are looking for useful ideas to solve this issue. Does the MQTT rework need to be revisited?

### EVerest Domain

Other

### Affected EVerest Module

_No response_

### To Reproduce

_No response_

### Anything else?

_No response_

mlitre:
A thread pool for handling the operation worker was added so that
commands and vars don't accumulate in the queue and cause delays which
can lead to timeouts in certain parts of the code.

### DIFF
--- a/lib/everest/framework/include/utils/message_handler.hpp
+++ b/lib/everest/framework/include/utils/message_handler.hpp
@@ -5,15 +5,17 @@
 
 #include <atomic>
 #include <condition_variable>
-#include <functional>
 #include <map>
 #include <memory>
 #include <mutex>
 #include <queue>
 #include <string>
 #include <thread>
+#include <unordered_map>
+#include <unordered_set>
 #include <vector>
 
+#include <everest/util/async/thread_pool_scaling.hpp>
 #include <utils/message_queue.hpp>
 #include <utils/types.hpp>
 
@@ -21,6 +23,10 @@ using MqttTopic = std::string;
 using CmdId = std::string;
 
 namespace Everest {
+
+constexpr int THREAD_POOL_SCALING_LATENCY_THRESHOLD_MS = 50;
+constexpr int THREAD_POOL_SCALING_LATENCY_THREAD_IDLE_TIMEOUT_S = 2;
+constexpr int THREAD_POOL_SCALING_MIN_THREAD_COUNT = 1;
 
 /// \brief Handles message dispatching and thread-safe queuing of different message types. This class uses two separate
 /// threads and message queues: one for operation messages (vars, cmds, errors, GetConfig, ModuleReady) and one for
@@ -40,9 +46,13 @@ public:
     void register_handler(const std::string& topic, std::shared_ptr<TypedHandler> handler);
 
 private:
-    void run_operation_message_worker();
+    void run_operation_dispatcher();
     void run_result_message_worker();
     void run_external_mqtt_worker();
+
+    void dispatch_operation_message(ParsedMessage&& message);
+    void schedule_operation_message(ParsedMessage&& message);
+    void on_operation_message_done(const std::string& topic);
 
     void handle_operation_message(const std::string& topic, const json& payload);
     void handle_result_message(const std::string& topic, const json& payload);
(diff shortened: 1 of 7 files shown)


Pietfried (changes requested):
lgtm! The bazel build is failing. You need to add the everest::util lib to https://github.com/EVerest/everest-core/blob/feat/thread-pool-message-handler/lib/BUILD.bazel and https://github.com/EVerest/everest-core/blob/feat/thread-pool-message-handler/lib/everest/framework/BUILD.bazel#L110

hikinggrass (approved):

Pietfried (dismissed):

hikinggrass (changes requested):
Blocked until a solution to the issue reported by @Pietfried is implemented

lib/everest/framework/lib/message_handler.cpp
```diff
@@ -145,9 +152,57 @@ void MessageHandler::run_operation_message_worker() {
...
+}
+
+void MessageHandler::schedule_operation_message(ParsedMessage&& message) {
+    auto operation = [this, message = std::move(message)]() {
         handle_operation_message(message.topic, message.data);
+        on_operation_message_done(message.topic);
+    };
+
+    if (operation_thread_pool) {
+        operation_thread_pool->run(std::move(operation));
```
djchhp, line 192: This is currently broken in main

djchhp (changes requested):

lib/everest/framework/include/utils/message_handler.hpp
```diff
@@ -69,20 +79,30 @@ class MessageHandler {
...
     std::thread external_mqtt_worker_thread; // processes external MQTT messages
     std::thread ready_thread;                // runs the modules ready function
 
     // Queues and sync primitives
+    std::unique_ptr<everest::lib::util::thread_pool_scaling<
+        everest::lib::util::LatencyScaling<THREAD_POOL_SCALING_LATENCY_THRESHOLD_MS>>>
+        operation_thread_pool{std::make_unique<everest::lib::util::thread_pool_scaling<
+            everest::lib::util::LatencyScaling<THREAD_POOL_SCALING_LATENCY_THRESHOLD_MS>>>(
+            THREAD_POOL_SCALING_MIN_THREAD_COUNT, std::thread::hardware_concurrency(),
+            std::chrono::seconds(THREAD_POOL_SCALING_LATENCY_THREAD_IDLE_TIMEOUT_S))};
```
djchhp, line 94: That is quite some declaration, almost impossible to read. What about something like this:

```
using latency_scaling = everest::lib::util::LatencyScaling<THREAD_POOL_SCALING_LATENCY_THRESHOLD_MS>;
using thread_pool = everest::lib::util::thread_pool_scaling<latency_scaling>;

// I suppose you need to keep the unique_ptr to be able to kill the pool later in a controlled way
std::unique_ptr<thread_pool> operation_thread_pool;

// moving the initilization to the implemetation 
// in message_handler.cpp
MessageHandler::MessageHandler()
: operation_thread_pool(std::make_unique<thread_pool>(THREAD_POOL_SCALING_MIN_THREAD_COUNT, std::thread::hardware_concurrenty(), std::chrono::seconds(THREAD_POOL_SCLING_LATENCY_THREAD_IDLE_TIMEOUT_S))
{
...
}
```

djchhp (approved):
Looking good.

Pietfried (approved):

hikinggrass (approved):

PR #1833, opened 2026-02-17
Bazel: make it a monorepo

dorezyuk:


lib/everest/framework/src/system_unix.cpp
```diff
@@ -69,7 +69,14 @@ GetPasswdEntryResult get_passwd_entry(const std::string& user_name) {
 } // namespace
 
 bool keep_caps() {
+    // cap_set_secbits was added in libcap 2.30.
+    // LIBCAP_MAJOR/LIBCAP_MINOR are defined in libcap >= 2.33.
```
dorezyuk: actually this is defined since 2.64, so this might be a poor idea - but at least on kirkstone we have libcap 2.66

@hikinggrass do you think this is reasonable to assume for most users?

hikinggrass: looking at what's officially supported (yocto >= kirkstone and ubuntu 24.04) that should be fine. for older systems (like debian11) you could just add a #if defined(LIBCAP_MAJOR) with a comment

PR #1838, opened 2026-02-19
Fix(EvseManager): Raising a cable check fault should not happen if a cancel_transaction (DeAuthorized) is triggered

SebaLukas:
## Describe your changes
If the cancel transaction reason is DeAuthorize, then no cable check error is raised. In other cases the error will be raised

During cable check a fault was raised if the cable check was interrupted due a start transaction (blocked id tag). EvseManager should not go into error mode.

### DIFF
--- a/modules/EVSE/EvseManager/EvseManager.cpp
+++ b/modules/EVSE/EvseManager/EvseManager.cpp
@@ -2517,8 +2517,16 @@ void EvseManager::fail_cable_check(const std::string& reason) {
         }
         r_hlc[0]->call_cable_check_finished(false);
     }
-    // Raising the cable check error also causes the HLC stack to get notified
-    this->error_handling->raise_cable_check_fault(reason);
+
+    // Raising a cable check fault should not happen if a cancel_transaction (DeAuthorized) is triggered
+    // during cable check
+    const auto last_stop_transaction_reason = charger->get_last_stop_transaction_reason();
+    if (not last_stop_transaction_reason.has_value() or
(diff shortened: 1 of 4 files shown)


modules/EVSE/EvseManager/EvseManager.cpp
```diff
@@ -2517,8 +2517,16 @@ void EvseManager::fail_cable_check(const std::string& reason) {
...
-    this->error_handling->raise_cable_check_fault(reason);
+
+    // Raising a cable check fault should not happen if a cancel_transaction (DeAuthorized) is triggered
+    // during cable check
+    if (not charger->get_last_stop_transaction_reason().has_value() or
+        (charger->get_last_stop_transaction_reason().has_value() and
+         charger->get_last_stop_transaction_reason().value() !=
+             types::evse_manager::StopTransactionReason::DeAuthorized)) {
+        // Raising the cable check error also causes the HLC stack to get notified
+        this->error_handling->raise_cable_check_fault(reason);
```
mlitre: Probably nicer to just copy the value once to avoid locking and unlocking the mutex.
If there is no stop reason maybe we should raise the error?

SebaLukas: Copying is a good point. I will change that.

> If there is no stop reason maybe we should raise the error?

This should be covered with the first `not [...].has_value()` condition?

mlitre: Ah right I got confused while reading the conditions

PR #1852, opened 2026-02-20
[OCPP] Add support for OCPP 2.1 Q11 and Q12

JB-CompleoCS:
## Describe your changes

Adds support for OCPP 2.1 use cases Q11 and Q12 to libocpp.

### DIFF
--- a/lib/everest/ocpp/doc/v2/ocpp_2x_status.md
+++ b/lib/everest/ocpp/doc/v2/ocpp_2x_status.md
@@ -2705,6 +2705,18 @@ CSMS setpoint and Dynamic charging profiles from K01. There are no specific requ
 | Q10.FR.04 |        |        |
 | Q10.FR.05 |        |        |
 
+## Bidirectional Power Transfer - Going offline during V2X operation
+
+| ID  | Status | Remark |
+| --- | ------ | ------ |
+| Q11 | ✅      |Use Case doesn't contain any functional requirement|
+
+## Bidirectional Power Transfer - Resuming a V2X operation after an offline period
+
+| ID  | Status | Remark |
+| --- | ------ | ------ |
+| Q12 | ✅      | Use Case doesn't contain any functional requirement|
+
 ## DER Control - Starting a V2X session with DER control in EVSE (New in OCPP 2.1)
 
 | ID        | Status | Remark |
--- a/lib/everest/ocpp/include/ocpp/v2/connectivity_manager.hpp
+++ b/lib/everest/ocpp/include/ocpp/v2/connectivity_manager.hpp
@@ -7,9 +7,12 @@
 #include <ocpp/v2/messages/SetNetworkProfile.hpp>
 #include <ocpp/v2/ocpp_types.hpp>
 
+#include <atomic>
+#include <chrono>
 #include <functional>
 #include <future>
 #include <optional>
+
 namespace ocpp {
 namespace v2 {
 
@@ -83,6 +86,11 @@ public:
     /// \return True is the websocket is connected, else false
     ///
     virtual bool is_websocket_connected() = 0;
+
+    /// @brief Get the time the websocket has been disconnected.
+    /// @return The time the websocket has been disconnected
+    ///
+    virtual std::chrono::time_point<std::chrono::steady_clock> get_time_disconnected() const = 0;
 
     /// \brief Connect to the websocket
     /// \param configuration_slot Optional the network_profile_slot to connect to. std::nullopt will select the slot
@@ -168,6 +176,7 @@ public:
     std::optional<std::int32_t> get_priority_from_configuration_slot(const int configuration_slot) const override;
     const std::vector<int>& get_network_connection_slots() const override;
     bool is_websocket_connected() override;
+    std::chrono::time_point<std::chrono::steady_clock> get_time_disconnected() const override;
     void connect(std::optional<std::int32_t> network_profile_slot = std::nullopt) override;
     void disconnect() override;
     bool send_to_websocket(const std::string& message) override;
@@ -176,6 +185,8 @@ public:
     void confirm_successful_connection() override;
 
 private:
+    std::atomic<std::chrono::time_point<std::chrono::steady_clock>> time_disconnected{};
+
     /// \brief Initializes the websocket and tries to connect
     ///
     void try_connect_websocket();
--- a/lib/everest/ocpp/lib/ocpp/v2/charge_point.cpp
+++ b/lib/everest/ocpp/lib/ocpp/v2/charge_point.cpp
@@ -1092,10 +1092,13 @@ void ChargePoint::websocket_connected_callback(const int configuration_slot,
     if (this->registration_status == RegistrationStatusEnum::Accepted) {
         this->connectivity_manager->confirm_successful_connection();
 
-        if (this->time_disconnected.time_since_epoch() != 0s) {
+        // check if we are disconnected and offline theshold has been defined
+        if (const auto time_disconnected = this->connectivity_manager->get_time_disconnected();
+            time_disconnected.time_since_epoch() != 0s &&
+            this->device_model->get_value<int>(ControllerComponentVariables::OfflineThreshold) != 0) {
             // handle offline threshold
             //  Get the current time point using steady_clock
-            auto offline_duration = std::chrono::steady_clock::now() - this->time_disconnected;
+            const auto offline_duration = std::chrono::steady_clock::now() - time_disconnected;
 
             // B04.FR.01
             // If offline period exceeds offline threshold then send the status notification for all connectors
@@ -1113,7 +1116,6 @@ void ChargePoint::websocket_connected_callback(const int configuration_slot,
             this->security->init_certificate_expiration_check_timers(); // re-init as timers are stopped on disconnect
         }
     }
-    this->time_disconnected = std::chrono::time_point<std::chrono::steady_clock>();
 
     // We have a connection again so next time it fails we should send the notification again
     this->skip_invalid_csms_certificate_notifications = false;
@@ -1127,12 +1129,6 @@ void ChargePoint::websocket_connected_callback(const int configuration_slot,
 void ChargePoint::websocket_disconnected_callback(const int configuration_slot,
                                                   const NetworkConnectionProfile& network_connection_profile) {
     this->message_queue->pause();
-
-    // check if offline threshold has been defined
-    if (this->device_model->get_value<int>(ControllerComponentVariables::OfflineThreshold) != 0) {
-        // Get the current time point using steady_clock
-        this->time_disconnected = std::chrono::steady_clock::now();
-    }
 
     this->security->stop_certificate_expiration_check_timers();
     if (this->callbacks.connection_state_changed_callback.has_value()) {
--- a/lib/everest/ocpp/lib/ocpp/v2/connectivity_manager.cpp
+++ b/lib/everest/ocpp/lib/ocpp/v2/connectivity_manager.cpp
@@ -21,6 +21,8 @@ constexpr std::int32_t default_network_config_timeout_seconds = 60;
 namespace ocpp {
 namespace v2 {
 
+using namespace std::chrono_literals;
+
 ConnectivityManager::ConnectivityManager(DeviceModelAbstract& device_model, std::shared_ptr<EvseSecurity> evse_security,
                                          std::shared_ptr<MessageLogging> logging, const fs::path& share_path,
                                          const std::function<void(const std::string& message)>& message_callback) :
@@ -120,6 +122,10 @@ const std::vector<int>& ConnectivityManager::get_network_connection_slots() cons
 
 bool ConnectivityManager::is_websocket_connected() {
     return this->websocket != nullptr && this->websocket->is_connected();
+}
+
+std::chrono::time_point<std::chrono::steady_clock> ConnectivityManager::get_time_disconnected() const {
+    return this->time_disconnected;
 }
 
 void ConnectivityManager::connect(std::optional<std::int32_t> network_profile_slot) {
@@ -406,11 +412,15 @@ void ConnectivityManager::on_websocket_connected(OcppProtocolVersion protocol) {
         this->websocket_connected_callback.value()(actual_configuration_slot, network_connection_profile.value(),
                                                    this->connected_ocpp_version);
     }
+    this->time_disconnected = std::chrono::time_point<std::chrono::steady_clock>();
 }
 
 void ConnectivityManager::on_websocket_disconnected() {
     std::optional<NetworkConnectionProfile> network_connection_profile =
         this->get_network_connection_profile(this->get_active_network_configuration_slot());
+    if (this->time_disconnected.load().time_since_epoch() == 0s) {
+        this->time_disconnected = std::chrono::steady_clock::now();
+    }
 
     if (this->websocket_disconnected_callback.has_value() and network_connection_profile.has_value()) {
         this->websocket_disconnected_callback.value()(this->get_active_network_configuration_slot(),
--- a/lib/everest/ocpp/lib/ocpp/v2/functional_blocks/smart_charging.cpp
+++ b/lib/everest/ocpp/lib/ocpp/v2/functional_blocks/smart_charging.cpp
@@ -25,6 +25,7 @@
 const std::int32_t STATION_WIDE_ID = 0;
 
 using namespace std::chrono;
+using namespace std::chrono_literals;
 
 namespace ocpp::v2 {
 namespace {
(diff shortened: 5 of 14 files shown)


lib/everest/ocpp/lib/ocpp/v2/charge_point.cpp
```diff
@@ -1092,10 +1092,11 @@ void ChargePoint::websocket_connected_callback(const int configuration_slot,
     if (this->registration_status == RegistrationStatusEnum::Accepted) {
         this->connectivity_manager->confirm_successful_connection();
 
-        if (this->time_disconnected.time_since_epoch() != 0s) {
+        if (const auto time_disconnected = this->connectivity_manager->get_time_disconnected();
+            time_disconnected.has_value()) {
```
mlitre: If statement condition has changed and is most likely the reason the tests are failing

lib/everest/ocpp/lib/ocpp/v2/charge_point.cpp
```diff
@@ -1092,10 +1092,11 @@ void ChargePoint::websocket_connected_callback(const int configuration_slot,
...
             // handle offline threshold
             //  Get the current time point using steady_clock
-            auto offline_duration = std::chrono::steady_clock::now() - this->time_disconnected;
+            const auto offline_duration = std::chrono::steady_clock::now() - *time_disconnected;
```
mlitre:
```suggestion
            const auto offline_duration = std::chrono::steady_clock::now() - time_disconnected.value();
```

lib/everest/ocpp/lib/ocpp/v2/functional_blocks/smart_charging.cpp
```diff
@@ -339,6 +339,28 @@ ProfileValidationResultEnum SmartCharging::verify_rate_limit(const ChargingProfi
...
+    }
+
+    // Absent maxOfflineDuration means the profile is valid independent of the offline time
+    if (!profile.maxOfflineDuration.has_value()) {
+        return {true, false};
+    }
+
+    // Not being offline for long enough means profile is valid
+    if (std::chrono::duration_cast<std::chrono::seconds>(std::chrono::steady_clock::now() - *time_disconnected)
+            .count() <= *profile.maxOfflineDuration) {
```
mlitre:
```suggestion
    if (std::chrono::duration_cast<std::chrono::seconds>(std::chrono::steady_clock::now() - time_disconnected.value())
            .count() <= profile.maxOfflineDuration.value()) {
```

lib/everest/ocpp/lib/ocpp/v2/functional_blocks/smart_charging.cpp
```diff
@@ -339,6 +339,28 @@ ProfileValidationResultEnum SmartCharging::verify_rate_limit(const ChargingProfi
...
+        return {true, false};
+    }
+
+    // Profile must be cleared when we are offline for too long and invalidAfterOfflineDuration is set
+    return {false, profile.invalidAfterOfflineDuration.value_or(false)};
```
mlitre, line 362:
```suggestion
    return {false, true};
```

lib/everest/ocpp/include/ocpp/v2/connectivity_manager.hpp
```diff
@@ -168,6 +175,7 @@ class ConnectivityManager : public ConnectivityManagerInterface {
     std::optional<std::int32_t> get_priority_from_configuration_slot(const int configuration_slot) const override;
     const std::vector<int>& get_network_connection_slots() const override;
     bool is_websocket_connected() override;
+    std::optional<std::chrono::time_point<std::chrono::steady_clock>> get_time_disconnected() const;
```
mlitre: Why make this optional? It was not before so I don't think it should be. I think this also affects the construction of the time disconnected time point

mlitre:
```suggestion
    std::optional<std::chrono::time_point<std::chrono::steady_clock>> get_time_disconnected() const override;
```

lib/everest/ocpp/lib/ocpp/v2/functional_blocks/smart_charging.cpp
```diff
@@ -1397,6 +1419,17 @@ SmartCharging::get_valid_profiles_for_evse(std::int32_t evse_id,
...
+        // Q11
+        if (const auto [valid, clear] = this->validate_profile_with_offline_time(profile); !valid) {
+            if (clear) {
+                // Q12
+                EVLOG_debug << "Clearing profile with ID: " << profile.id
+                            << ", because of it is invalid after offline duration";
+                this->context.database_handler.clear_charging_profiles_matching_criteria(profile.id, std::nullopt);
+            }
+            continue;
+        }
```
mlitre, line 1432: Do we really need an std pair for valid and clear? For this use case we could just have a singular bool for clearing the profile. If you don't have to clear it then it is always valid and does not need special handling. If it is invalid because it is too old then you clear it.

mlitre (changes requested):

lib/everest/ocpp/lib/ocpp/v2/functional_blocks/smart_charging.cpp
JB-CompleoCS: OCPP 2.1 use case Q11 says that a profile that has 'maxOfflineDuration' set has to be invalidated as soon as the offline duration is reached. OCPP 2.1 use case Q12 says that a profile that has 'invalidAfterOfflineDuration' set will not get valid again after getting online again. However, if 'invalidAfterOfflineDuration' is not set, the profile will be valid again as soon as we get online again. So in this case we cannot clear the profile, because we still need it later on

lib/everest/ocpp/include/ocpp/v2/connectivity_manager.hpp
JB-CompleoCS: Before this change a value of `0s` was used for 'time disconnected not set'. To me this is a use case for an optional where `std::nullopt` means 'time disconnected not set'. But if you like it more the way it was before, I can revert that change

lib/everest/ocpp/lib/ocpp/v2/functional_blocks/smart_charging.cpp
JB-CompleoCS: The profile shall not be cleared in every case when 'maxOfflineDuration' is reached. Only when 'invalidAfterOfflineDuration' is set

lib/everest/ocpp/include/ocpp/v2/connectivity_manager.hpp
mlitre: You are right. However, as it is implemented now this "feature" is not used anymore as we always set the disconnected time.
I think a good way to move forward would be to move this "not set" feature to where the old uses of time_disconnected were done and always update time_disconnected since it is now always needed by the new invalid offline profile functionality.

lib/everest/ocpp/lib/ocpp/v2/functional_blocks/smart_charging.cpp
mlitre: Got it, this makes sense to keep as is then

lib/everest/ocpp/lib/ocpp/v2/functional_blocks/smart_charging.cpp
```diff
@@ -1397,6 +1419,17 @@ SmartCharging::get_valid_profiles_for_evse(std::int32_t evse_id,
 
     auto evse_profiles = this->context.database_handler.get_charging_profiles_for_evse(evse_id);
     for (auto profile : evse_profiles) {
+        // Q11
+        if (const auto [valid, clear] = this->validate_profile_with_offline_time(profile); !valid) {
+            if (clear) {
+                // Q12
+                EVLOG_debug << "Clearing profile with ID: " << profile.id
+                            << ", because of it is invalid after offline duration";
```
Pietfried:
```suggestion
                            << ", because it is invalid after offline duration";
```

lib/everest/ocpp/lib/ocpp/v2/connectivity_manager.cpp
```diff
@@ -401,6 +405,7 @@ void ConnectivityManager::on_websocket_connected(OcppProtocolVersion protocol) {
     const int actual_configuration_slot = get_active_network_configuration_slot();
     std::optional<NetworkConnectionProfile> network_connection_profile =
         this->get_network_connection_profile(actual_configuration_slot);
+    this->time_disconnected = std::nullopt;
```
Pietfried: `time_disconnected` is cleared before the `websocket_connected_callback` reads it.

The `OfflineThreshold` handling in the `charge_point.cpp` will never execute because `time_disconnected` is always `nullopt`.

It may be solved by moving this below the execution of the callback.

Pietfried (changes requested):
Please also update the functional requirements in https://github.com/EVerest/everest-core/blob/main/lib/everest/ocpp/doc/v2/ocpp_2x_status.md

lib/everest/ocpp/lib/ocpp/v2/connectivity_manager.cpp
JB-CompleoCS: You're right, thank you. Noticed that yesterday, too. I'm going to move it below

lib/everest/ocpp/include/ocpp/v2/connectivity_manager.hpp
JB-CompleoCS: The idea was that `get_time_disconnected` returns `std::nullopt`, if we are not disconnected at all. And if we are disconnected, it returns the corresponding time point. The same way `time_disconnected` had `time_since_epoch() == 0s` before when we were not disconnected.
The new invalid offline profile functionality treats it exactly this way.

lib/everest/ocpp/doc/v2/ocpp_2x_status.md
```diff
@@ -2705,6 +2705,18 @@ CSMS setpoint and Dynamic charging profiles from K01. There are no specific requ
...
+
+| ID  | Status | Remark |
+| --- | ------ | ------ |
+| Q11 | ✅      |        |
+
+## Bidirectional Power Transfer - Resuming a V2X operation after an offline period
+
+| ID  | Status | Remark |
+| --- | ------ | ------ |
+| Q12 | ✅      |        |
```
Pietfried: nit:
```suggestion
| Q12 | ✅      | Use Case doesn't contain any functional requirement|
```

lib/everest/ocpp/doc/v2/ocpp_2x_status.md
```diff
@@ -2705,6 +2705,18 @@ CSMS setpoint and Dynamic charging profiles from K01. There are no specific requ
 | Q10.FR.04 |        |        |
 | Q10.FR.05 |        |        |
 
+## Bidirectional Power Transfer - Going offline during V2X operation
+
+| ID  | Status | Remark |
+| --- | ------ | ------ |
+| Q11 | ✅      |        |
```
Pietfried: nit:
```suggestion
| Q11 | ✅      |Use Case doesn't contain any functional requirement|
```

Pietfried (dismissed):
lgtm

Pietfried (dismissed):

lib/everest/ocpp/include/ocpp/v2/connectivity_manager.hpp
```diff
@@ -168,6 +175,7 @@ class ConnectivityManager : public ConnectivityManagerInterface {
     std::optional<std::int32_t> get_priority_from_configuration_slot(const int configuration_slot) const override;
     const std::vector<int>& get_network_connection_slots() const override;
     bool is_websocket_connected() override;
+    std::optional<std::chrono::time_point<std::chrono::steady_clock>> get_time_disconnected() const override;
```
mlitre: We should wrap this in a monitor as it can now be used from multiple threads

lib/everest/ocpp/lib/ocpp/v2/charge_point.cpp
```diff
@@ -1128,12 +1128,6 @@ void ChargePoint::websocket_disconnected_callback(const int configuration_slot,
                                                   const NetworkConnectionProfile& network_connection_profile) {
     this->message_queue->pause();
 
-    // check if offline threshold has been defined
-    if (this->device_model->get_value<int>(ControllerComponentVariables::OfflineThreshold) != 0) {
-        // Get the current time point using steady_clock
-        this->time_disconnected = std::chrono::steady_clock::now();
-    }
```
mlitre, line 1135: There is a regression on this functionality as we now always set the time disconnected. We should most likely remove the optional from time_disconnected and a check on this variable when we do the reconnection check. See other comment

lib/everest/ocpp/lib/ocpp/v2/charge_point.cpp
```diff
@@ -1092,10 +1092,11 @@ void ChargePoint::websocket_connected_callback(const int configuration_slot,
     if (this->registration_status == RegistrationStatusEnum::Accepted) {
         this->connectivity_manager->confirm_successful_connection();
 
-        if (this->time_disconnected.time_since_epoch() != 0s) {
+        if (const auto time_disconnected = this->connectivity_manager->get_time_disconnected();
```
mlitre, line 1096: We should use the check on the variable here since we always set time disconnected

lib/everest/ocpp/lib/ocpp/v2/functional_blocks/smart_charging.cpp
```diff
@@ -1397,6 +1419,17 @@ SmartCharging::get_valid_profiles_for_evse(std::int32_t evse_id,
...
     auto evse_profiles = this->context.database_handler.get_charging_profiles_for_evse(evse_id);
     for (auto profile : evse_profiles) {
+        // Q11
+        if (const auto [valid, clear] = this->validate_profile_with_offline_time(profile); !valid) {
+            if (clear) {
+                // Q12
+                EVLOG_debug << "Clearing profile with ID: " << profile.id
+                            << ", because it is invalid after offline duration";
+                this->context.database_handler.clear_charging_profiles_matching_criteria(profile.id, std::nullopt);
+            }
```
mlitre, line 1430: @Pietfried I am wondering if it is not better to clear these profiles at a different time so that we properly separate concerns for this function? In the end we want to only get valid profiles here, deleting things from the DB might create unnecessary load. We should probably do some sort of regular cleanup

mlitre (changes requested):

lib/everest/ocpp/lib/ocpp/v2/functional_blocks/smart_charging.cpp
Pietfried: I see your point, but I actually don't see good alternative that doesn't add much overhead. I'd therefore be fine with keeping it as is.

mlitre:
Hello @JB-CompleoCS did you see my latest comments?

lib/everest/ocpp/lib/ocpp/v2/charge_point.cpp
JB-CompleoCS: Sorry, haven't had time to work on this. Yes, you're right, thank you. I will remove the optional and check on the variable again

lib/everest/ocpp/lib/ocpp/v2/charge_point.cpp
JB-CompleoCS: Added the check back

lib/everest/ocpp/include/ocpp/v2/connectivity_manager.hpp
JB-CompleoCS: I used an `atomic` to make it thread-safe. Is that ok for you? Or what did you mean by wrapping it in a monitor? Is it still necessary?

lib/everest/ocpp/lib/ocpp/v2/connectivity_manager.cpp
```diff
@@ -406,11 +412,15 @@ void ConnectivityManager::on_websocket_connected(OcppProtocolVersion protocol) {
...
     }
+    this->time_disconnected = std::chrono::time_point<std::chrono::steady_clock>();
 }
 
 void ConnectivityManager::on_websocket_disconnected() {
     std::optional<NetworkConnectionProfile> network_connection_profile =
         this->get_network_connection_profile(this->get_active_network_configuration_slot());
+    if (this->time_disconnected.load().time_since_epoch() == 0s) {
+        this->time_disconnected = std::chrono::steady_clock::now();
+    }
```
mlitre, line 423: We do not need this if statement anymore we can just always set the time disconnected

mlitre (changes requested):

lib/everest/ocpp/include/ocpp/v2/connectivity_manager.hpp
mlitre: atomic is fine, we created a helper that handles all of the locking for you (the monitor), it is located here: https://github.com/EVerest/EVerest/blob/main/lib/everest/util/include/everest/util/async/monitor.hpp

lib/everest/ocpp/lib/ocpp/v2/connectivity_manager.cpp
JB-CompleoCS: When I tested it, the disconnected callback of the websocket was called repeatedly, I guess everytime it tried to reconnect. And I wanted to store the time the disconnect happened initially. Has that changed since?

mlitre: That probably has not changed, then your implementation is correct then.

mlitre (dismissed):
@JB-CompleoCS Could you squash all of your commits into one please? Except for that it's all good from me! Thanks for your contribution!

Pietfried (approved):

mlitre (approved):

PR #1857, opened 2026-02-23
EvseV2G: add d-link error handling

FaHaGit:
## Describe your changes

This change is necessary to notify when the connection is unexpectedly closed (for example, due to a failed response code). Otherwise, it is not possible to reinitialize a charging session at the EvseManager level in case of an error (as described in the ISO15118-3), especially if a switch to basic PWM charging is necessary in the AC case.

Changes:
- change default value of conn member d_link_action to D_LINK_ACTION_ERROR
- introduce D_LINK_ACTION_ERROR to manage dLink errors
- trigger dLink error when iso session stop validation fails
- handle dLink errors by setting appropriate state and actions

--

### DIFF
--- a/modules/EVSE/EvseV2G/iso_server.cpp
+++ b/modules/EVSE/EvseV2G/iso_server.cpp
@@ -2155,7 +2155,7 @@ static enum v2g_event handle_iso_session_stop(struct v2g_connection* conn) {
         conn->d_link_action = dLinkAction::D_LINK_ACTION_ERROR;
         conn->ctx->hlc_pause_active = false;
         /* Set next expected req msg */
-        conn->ctx->state = (int)iso_dc_state_id::WAIT_FOR_TERMINATED_SESSION;
+        conn->ctx->state = static_cast<int>(iso_dc_state_id::WAIT_FOR_TERMINATED_SESSION);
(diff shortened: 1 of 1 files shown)


modules/EVSE/EvseV2G/iso_server.cpp
```diff
@@ -2148,7 +2148,16 @@ static enum v2g_event handle_iso_session_stop(struct v2g_connection* conn) {
...
     /* Check the current response code and check if no external error has occurred */
-    iso_validate_response_code(&res->ResponseCode, conn);
+    auto event = iso_validate_response_code(&res->ResponseCode, conn);
+
+    /* In case of an error, ignore payload of the request message and trigger a d-link error */
+    if (event != V2G_EVENT_NO_EVENT) {
+        conn->d_link_action = dLinkAction::D_LINK_ACTION_ERROR;
+        conn->ctx->hlc_pause_active = false;
+        /* Set next expected req msg */
+        conn->ctx->state = (int)iso_dc_state_id::WAIT_FOR_TERMINATED_SESSION;
```
SebaLukas: Instead if c-style casting, use `static_cast<int>()`

PR #1860, opened 2026-02-24
Fix(EvseV2G): Each iso15118_charger cmd handler function now checks whether v2g_ctx has been set

SebaLukas:
## Describe your changes
1. Adding check if v2g_ctx is set
2. Adding generic error to ISO15118_charger interface

Before the check it could happen that the v2g_ctx was freed but the module continued running. If a cmd handler is called in this case then the process could crash.

### DIFF
--- a/modules/EVSE/EvseV2G/EvseV2G.cpp
+++ b/modules/EVSE/EvseV2G/EvseV2G.cpp
@@ -116,7 +116,8 @@ void EvseV2G::ready() {
     return;
 
 err_out:
-    v2g_ctx_free(v2g_ctx);
+    p_charger->raise_error(p_charger->error_factory->create_error("generic/CommunicationFault", "",
+                                                                  "Could not initialise EvseV2G module"));
 }
 
 EvseV2G::~EvseV2G() {
--- a/interfaces/ISO15118_charger.yaml
+++ b/interfaces/ISO15118_charger.yaml
@@ -452,3 +452,5 @@ vars:
       The EV termination code as described in ISO15118-20.
     type: object
     $ref: /iso15118#/EvTermination
+errors:
+  - reference: /errors/generic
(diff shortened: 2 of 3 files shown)


modules/EVSE/EvseV2G/EvseV2G.cpp
```diff
@@ -116,7 +116,8 @@ void EvseV2G::ready() {
     return;
 
 err_out:
-    v2g_ctx_free(v2g_ctx);
+    p_charger->raise_error(p_charger->error_factory->create_error("generic/CommunicationFault", "",
```
SebaLukas, line 119: @corneliusclaussen Is this enough or do I have to handle/subscribe this error in the EvseManager as well? If this error is raised then the EvseV2G or the whole charging station should be restarted.

modules/EVSE/EvseV2G/EvseV2G.cpp
```diff
@@ -116,7 +116,8 @@ void EvseV2G::ready() {
     return;
 
 err_out:
-    v2g_ctx_free(v2g_ctx);
```
Pietfried, line 119: Is it intended to remove the `v2g_ctx_free(v2g_ctx);` statement? Shouldnt we raise the error in addition to freeing the context?

modules/EVSE/EvseV2G/EvseV2G.cpp
Pietfried: If the module can not recover on its own shoud we just throw here? The EvseManager cant do much about it at the moment. If its AC, it could try to disable HLC, but for DC charging would not be possible

modules/EVSE/EvseV2G/EvseV2G.cpp
SebaLukas: Yes :+1:
The original problem was freeing the ctx in the event of an error during the ready or init(). In this case the ctx pointed to zero, but the module continued to run. This is fatal if, for example, the EvseManager calls a cmd from the EvseV2G module.
That's why I thought that every cmd handler should check whether the ctx is not zero and the ctx should only be freed when the module is also terminated.

modules/EVSE/EvseV2G/EvseV2G.cpp
SebaLukas: In my understanding if the EvseV2G module is at this point, a hard reset of the charging station is necessary. So raising an error should be okay and the next steps would have to be discussed

PR #1862, opened 2026-02-24
Fix(EvseManager): Introduce monitor for hw_capabilites and supported_energy_transfers

Pietfried:
fix(EvseManager,EvseV2G): Make supported_energy_transfer modes and hw_capabilites thread safe

* Added monitor utility to EvseManager to protect supported_energy_transfers and hw_capabilities
* Fix issue that occured at encoding ServiceDiscoveryRes message that occured when call_update_energy_transfer_modes was called with an empty vector. New helper get_supported_ac_energy_transfers() ensures that at least AC_single_phase_core is provided
* For hw_capabilities concurrent read and write operations could occur. The access is now protected by the monitor utility

### DIFF
--- a/modules/EVSE/EvseManager/EvseManager.cpp
+++ b/modules/EVSE/EvseManager/EvseManager.cpp
@@ -110,6 +110,30 @@ get_dc_external_derate(std::optional<float> present_voltage,
     return d;
 }
 
+std::vector<types::iso15118::EnergyTransferMode>
+get_supported_ac_energy_transfers(const Conf& config, const types::evse_board_support::HardwareCapabilities& caps) {
+    std::vector<types::iso15118::EnergyTransferMode> energy_transfers;
+
+    const auto min_phases = std::clamp(caps.min_phase_count_import, 1, 3);
+    const auto max_phases = std::clamp(caps.max_phase_count_import, min_phases, 3);
+
+    for (const auto& [count, mode] : {
+             std::pair{1, types::iso15118::EnergyTransferMode::AC_single_phase_core},
+             std::pair{2, types::iso15118::EnergyTransferMode::AC_two_phase},
+             std::pair{3, types::iso15118::EnergyTransferMode::AC_three_phase_core},
+         }) {
+        if (count >= min_phases and count <= max_phases) {
+            energy_transfers.push_back(mode);
+        }
+    }
+
+    if (config.supported_iso_ac_bpt and caps.max_current_A_export > 0 and caps.max_phase_count_export >= 1) {
+        energy_transfers.push_back(types::iso15118::EnergyTransferMode::AC_BPT);
+    }
+
+    return energy_transfers;
+}
+
 } // namespace
 
 void EvseManager::init() {
@@ -140,6 +164,24 @@ void EvseManager::init() {
         session_log.enable();
     }
     session_log.xmlOutput(config.session_logging_xml);
+
+    // default initialize hardware capabilities with safe values until actual capabilities are received.
+    // This is important to have a defined behavior of the charger even if the bsp does not provide
+    // capabilities or is slow in doing so.
+    {
+        auto hw_capabilities_handle = hw_capabilities.handle();
+        hw_capabilities_handle->max_current_A_import = 0;
+        hw_capabilities_handle->min_current_A_import = 0;
+        hw_capabilities_handle->max_phase_count_import = 1;
+        hw_capabilities_handle->min_phase_count_import = 0;
+        hw_capabilities_handle->max_current_A_export = 0;
+        hw_capabilities_handle->min_current_A_export = 0;
+        hw_capabilities_handle->max_phase_count_export = 0;
+        hw_capabilities_handle->min_phase_count_export = 0;
+        hw_capabilities_handle->supports_changing_phases_during_charging = false;
+        hw_capabilities_handle->supports_cp_state_E = false;
+        hw_capabilities_handle->connector_type = types::evse_board_support::Connector_type::IEC62196Type2Cable;
+    }
 
     invoke_init(*p_evse);
     invoke_init(*p_energy_grid);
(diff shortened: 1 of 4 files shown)


modules/EVSE/EvseManager/EvseManager.cpp
```diff
@@ -110,6 +110,30 @@ get_dc_external_derate(std::optional<float> present_voltage,
     return d;
 }
 
+std::vector<types::iso15118::EnergyTransferMode>
+get_supported_ac_energy_transfers(const Conf& config, const types::evse_board_support::HardwareCapabilities& caps) {
+    std::vector<types::iso15118::EnergyTransferMode> energy_transfers;
+
+    const auto min_phases = std::max(1, std::min(3, caps.min_phase_count_import));
+    const auto max_phases = std::max(min_phases, std::min(3, caps.max_phase_count_import));
```
djchhp:
```suggestion
    const auto min_phases = std::clamp(caps.min_phase_count_import, 1, 3);
    const auto max_phases = std::clamp(caps.max_phase_count_import, min_phases, 3);
```

Prefer using std::clamp instead of replicating it.

modules/EVSE/EvseManager/EvseManager.cpp
```diff
@@ -110,6 +110,30 @@ get_dc_external_derate(std::optional<float> present_voltage,
...
+
+    if (min_phases <= 1 and max_phases >= 1) {
+        energy_transfers.push_back(types::iso15118::EnergyTransferMode::AC_single_phase_core);
+    }
+    if (min_phases <= 2 and max_phases >= 2) {
+        energy_transfers.push_back(types::iso15118::EnergyTransferMode::AC_two_phase);
+    }
+    if (min_phases <= 3 and max_phases >= 3) {
+        energy_transfers.push_back(types::iso15118::EnergyTransferMode::AC_three_phase_core);
+    }
```
djchhp: You might consider accumulating the data in one place, and the logic in another, e.g. like this

```suggestion
    for (const auto& [count, mode] : {
             std::pair{1, types::iso15118::EnergyTransferMode::AC_single_phase_core},
             std::pair{2, types::iso15118::EnergyTransferMode::AC_two_phase},
             std::pair{3, types::iso15118::EnergyTransferMode::AC_three_phase_core},
         }) {
        if (count >= min_phases and count <= max_phases) {
            energy_transfers.push_back(mode);
        }
    }

```

modules/EVSE/EvseManager/EvseManager.cpp
```diff
@@ -397,34 +421,15 @@ void EvseManager::ready() {
...
-            }
-
-            update_supported_energy_transfers(types::iso15118::EnergyTransferMode::AC_three_phase_core);
-
-            if (config.supported_iso_ac_bpt and hw_capabilities.max_current_A_export > 0 and
-                hw_capabilities.max_phase_count_export >= 1) {
-                transfer_modes.push_back({types::iso15118::EnergyTransferMode::AC_BPT});
-                update_supported_energy_transfers(types::iso15118::EnergyTransferMode::AC_BPT);
-            }
+            const auto hw_caps = *hw_capabilities.handle();
```
djchhp, line 449: This shadows the earlier copy in line 314. The lock should be avoided here.

```suggestion
```

modules/EVSE/EvseManager/EvseManager.cpp
```diff
@@ -1811,61 +1812,83 @@ void EvseManager::set_contract_certificate_installation_enabled(const bool value
...
+    }
+
+    p_evse->publish_supported_energy_transfer_modes(transfer_modes);
+
+    if (hlc_enabled and not r_hlc.empty()) {
+        r_hlc[0]->call_update_energy_transfer_modes(transfer_modes);
     }
+}
+
+bool EvseManager::update_supported_energy_transfers(
```
djchhp, line 1849: This does not try to remove items from the global supported_energy_transfers. Is that correct?

modules/EVSE/EvseManager/EvseManager.cpp
```diff
@@ -1811,61 +1812,83 @@ void EvseManager::set_contract_certificate_installation_enabled(const bool value
...
+    const std::vector<types::iso15118::EnergyTransferMode>& energy_transfers) {
+    auto handle = this->supported_energy_transfers.handle();
+    bool was_updated = false;
+
+    for (const auto& energy_transfer : energy_transfers) {
+        const bool exists = std::find(handle->begin(), handle->end(), energy_transfer) != handle->end();
```
djchhp: This screams for a template in the util lib

modules/EVSE/EvseManager/EvseManager.cpp
```diff
@@ -1811,61 +1812,83 @@ void EvseManager::set_contract_certificate_installation_enabled(const bool value
...
+
     return was_updated;
 }
 
+bool EvseManager::update_supported_energy_transfers(const types::iso15118::EnergyTransferMode& energy_transfer) {
+    return update_supported_energy_transfers(std::vector<types::iso15118::EnergyTransferMode>{energy_transfer});
+}
+
 void EvseManager::update_hlc_ac_parameters() {
     std::scoped_lock lock(hlc_ac_parameters_mutex);
```
djchhp: danger of deadlock, locking two things at once

modules/EVSE/EvseManager/EvseManager.cpp
```diff
@@ -2487,8 +2510,8 @@ types::energy::ExternalLimits EvseManager::get_local_energy_limits() {
     // external limits are empty
     if (external_local_energy_limits.schedule_import.empty() and external_local_energy_limits.schedule_export.empty()) {
         if (config.charge_mode == "AC") {
-            update_max_current_limit(active_local_limits, get_hw_capabilities().max_current_A_import,
-                                     get_hw_capabilities().max_current_A_export);
+            const auto hw_caps = *hw_capabilities.handle();
+            update_max_current_limit(active_local_limits, hw_caps.max_current_A_import, hw_caps.max_current_A_export);
```
djchhp: No need to copy the full struct. Consider
```
auto handle = hw_capabilities.handle();
update_max_current_limit(active_local_limits, handle->max_current_A_import, handle->max_current_A_export);
```

modules/EVSE/EvseManager/EvseManager.cpp
Pietfried: The existing implementation wasn't doing it, although it should be considered to remove supported transfer modes when capabilities change. I'll look into it.

modules/EVSE/EvseManager/EvseManager.cpp
Pietfried: Changed the order so that hw_caps mutex is acquired and released before hlc_ac_parameters_mutex is locked.

djchhp (approved):
lgtm

SebaLukas (approved):
Looking good!

PR #1866, opened 2026-02-25
EvseV2G, Evse15118D20, IsoMux: Add configurable supported app protocols

FaHaGit:
## Describe your changes

This PR introduces configurable handling of supported app protocols across EvseV2G, Evse15118D20, and IsoMux, and aligns how protocol configuration is logged and published.

Note: This command is intended exclusively for Fake DC, primarily to disable specific supported protocols in case of incompatibilities between EVCC and SECC.

The interface extensions are not currently in use. The feature can currently only be tested via internal MQTT messages such as:

mosquitto_pub -h localhost   -t ‘everest/modules/iso_mux/impl/charger/cmd/update_supported_app_protocols’   -m '{“data”:{‘args’:{“supported_app_protocols”: {“app_protocols”:[0,1,2]}},“id”:“11111111-2222-3333-4444-555555555555”,“origin”:“cli”},‘msg_type’:“Cmd”}'

--


FaHaGit:
Updated the command description to clarify that this command is only intended for fake DC.

types/iso15118.yaml
```diff
@@ -376,6 +376,30 @@ types:
           (approx. 80% SOC) is complete"
         type: string
         format: date-time
+  SupportedAppProtocol:
+    description: Protocols supported by the SECC
+    type: string
+    enum:
+      - DIN70121
+      - ISO15118d2
+      - ISO15118d20
```
SebaLukas: `D` instead of `d`

modules/EVSE/Evse15118D20/charger/ISO15118_chargerImpl.cpp
```diff
@@ -169,6 +169,10 @@ void ISO15118_chargerImpl::init() {
 }
 
 void ISO15118_chargerImpl::ready() {
+    types::iso15118::SupportedAppProtocols supp_app_protocols_secc;
+    supp_app_protocols_secc.app_protocols.push_back(types::iso15118::SupportedAppProtocol::ISO15118d20);
+    publish_supported_app_protocols_secc(supp_app_protocols_secc);
```
SebaLukas: This is fine but you can always use one line:
`publish_supported_app_protocols_secc({types::iso15118::SupportedAppProtocol::ISO15118d20});`

modules/EVSE/Evse15118D20/charger/ISO15118_chargerImpl.cpp
```diff
@@ -842,6 +846,41 @@ void ISO15118_chargerImpl::handle_no_energy_pause_charging(types::iso15118::NoEn
     // your code for cmd no_energy_pause_charging goes here
 }
 
+bool ISO15118_chargerImpl::handle_update_supported_app_protocols(
+    types::iso15118::SupportedAppProtocols& supported_app_protocols) {
+    if (supported_app_protocols.app_protocols.empty()) {
+        EVLOG_warning << "No supported app protocols configured";
+        return false;
+    } else {
```
SebaLukas: You dont need the `else` because of the `return`

modules/EVSE/EvseV2G/charger/ISO15118_chargerImpl.cpp
```diff
@@ -362,64 +364,116 @@ void ISO15118_chargerImpl::handle_no_energy_pause_charging(types::iso15118::NoEn
...
 
-    uint16_t& energyArrayLen =
-        (v2g_ctx->evse_v2g_data.charge_service.SupportedEnergyTransferMode.EnergyTransferMode.arrayLen);
-    iso2_EnergyTransferModeType* energyArray =
-        v2g_ctx->evse_v2g_data.charge_service.SupportedEnergyTransferMode.EnergyTransferMode.array;
-    energyArrayLen = 0;
+    if (supported_app_protocols.app_protocols.empty()) {
+        dlog(DLOG_LEVEL_WARNING, "No supported app protocols configured");
+        return true;
+    } else {
```
SebaLukas: See above.

modules/EVSE/IsoMux/charger/ISO15118_chargerImpl.cpp
```diff
@@ -500,6 +511,28 @@ void ISO15118_chargerImpl::init() {
...
+            changed = true;
+        }
+        protocol_sources[protocol] = source;
+    }
+
+    if (!changed) {
+        return std::nullopt;
+    }
+
+    return supported_app_protocols;
```
SebaLukas: Could be handled via `return (changed) ? supported_app_protocols : std::nullopt;`

modules/EVSE/IsoMux/charger/ISO15118_chargerImpl.cpp
```diff
@@ -594,6 +627,63 @@ void ISO15118_chargerImpl::handle_no_energy_pause_charging(types::iso15118::NoEn
...
 
+bool ISO15118_chargerImpl::handle_update_supported_app_protocols(
+    types::iso15118::SupportedAppProtocols& supported_app_protocols) {
+    bool iso20_configured = false;
+
+    if (supported_app_protocols.app_protocols.empty()) {
+        v2g_ctx->iso20_proxy_enabled = false;
+        EVLOG_info << "No protocols configured, disabling ISO-20 proxy routing";
+        // Forward empty updates to both implementations
+        const bool ok_iso2 = mod->r_iso2->call_update_supported_app_protocols(supported_app_protocols);
```
SebaLukas: Why not calling `r_iso20` as well? Right now only the d2 module will get a empty update

modules/EVSE/IsoMux/charger/ISO15118_chargerImpl.cpp
```diff
@@ -594,6 +627,63 @@ void ISO15118_chargerImpl::handle_no_energy_pause_charging(types::iso15118::NoEn
...
+    /* update supported protocols */
+    bool ok = true;
+    if (!iso2_payload.app_protocols.empty()) {
+        ok = mod->r_iso2->call_update_supported_app_protocols(iso2_payload) && ok;
+    }
+    if (!iso20_payload.app_protocols.empty()) {
+        ok = mod->r_iso20->call_update_supported_app_protocols(iso20_payload) && ok;
+    }
+
+    /* route to ISO-20 proxy only when ISO15118-20 is explicitly configured. */
```
SebaLukas, line 690: One could argue that D2 should only be enabled if a D2 proxy such as D20 is actually active.
Either one assumes that a D2 and D20 module is connected to the IsoMux an no check is done or one must check every combination. D2 as well.

SebaLukas (changes requested):
See my comments

modules/EVSE/IsoMux/charger/ISO15118_chargerImpl.cpp
FaHaGit: The D20 implementation currently does not allow for an empty list and returns false when call_update_supported_app_protocols is called.  That's also why I made it possible to disable the d20 proxy (see my comment below).
But I agree the comment is currently confusing. I can add a TODO to include this here if this logic is also required for the D20.

modules/EVSE/IsoMux/charger/ISO15118_chargerImpl.cpp
FaHaGit: To be honest, this is currently more of a workaround because the Evse15118D20 module doesn't currently support disabling all protocols. If the Evse15118D20 also had the capability to disable all protocols, there would be no need to distinguish between the proxies here. But since the library currently only supports D20, I thought the simplest solution for now would be to simply forward SupportedApp requests to the D2 module as soon as D20 has been disabled.

FaHaGit: Otherwise, for the sake of completeness, we'll have to implement it in the Evse15118D20 implementation as well. What do you think?

FaHaGit (commented):
I have updated the PR and wrote some comments. I would be happy if you would review your points again.

SebaLukas:
Please revert the InfyPower commits here. This is not part of this PR

FaHaGit:
> Please revert the InfyPower commits here. This is not part of this PR

Sorry, I accidentally rebased from the wrong branch... I've fixed it.

modules/EVSE/Evse15118D20/charger/ISO15118_chargerImpl.cpp
```diff
@@ -169,6 +169,9 @@ void ISO15118_chargerImpl::init() {
 }
 
 void ISO15118_chargerImpl::ready() {
+    publish_supported_app_protocols_secc(
+        types::iso15118::SupportedAppProtocols{.app_protocols = {types::iso15118::SupportedAppProtocol::ISO15118D20}});
```
SebaLukas: This is not a C++17 feature. So please dont use `.app_protocols`

modules/EVSE/IsoMux/charger/ISO15118_chargerImpl.cpp
```diff
@@ -605,6 +634,68 @@ void ISO15118_chargerImpl::handle_no_energy_pause_charging(types::iso15118::NoEn
     }
 }
 
+bool ISO15118_chargerImpl::handle_update_supported_app_protocols(
+    types::iso15118::SupportedAppProtocols& supported_app_protocols) {
```
SebaLukas: Please use another name for this parameter. Its the same name as the class member

modules/EVSE/IsoMux/charger/ISO15118_chargerImpl.cpp
```diff
@@ -511,6 +522,24 @@ void ISO15118_chargerImpl::init() {
...
+    for (const auto protocol : value.app_protocols) {
+        if (std::find(supported_app_protocols.app_protocols.begin(), supported_app_protocols.app_protocols.end(),
+                      protocol) == supported_app_protocols.app_protocols.end()) {
+            supported_app_protocols.app_protocols.push_back(protocol);
+            changed = true;
+        }
+        protocol_sources[protocol] = source;
+    }
+
+    return changed ? std::optional<types::iso15118::SupportedAppProtocols>{supported_app_protocols} : std::nullopt;
```
SebaLukas: You can also use `std::make_optional<>()`

modules/EVSE/IsoMux/charger/ISO15118_chargerImpl.cpp
SebaLukas: Okay then its fine. Lets see if it is necessary in the future

modules/EVSE/IsoMux/charger/ISO15118_chargerImpl.cpp
SebaLukas: Okay then its fine.

modules/EVSE/Evse15118D20/charger/ISO15118_chargerImpl.cpp
FaHaGit: you are right. I just tried to refactore it to one line as suggested. I found now a compiling one liner 👌

FaHaGit:
Thank you very much for the review @SebaLukas! I hope everything is okay now.

SebaLukas (approved):
Looking good!
Please squash your commits into one with a fitting commit message :)

SebaLukas:
This PR needs some additional approval because of the csv file. Maybe @Pietfried, @hikinggrass or @cburandt

cburandt (approved):

PR #1869, opened 2026-02-25
Fix/api test file autogen error messages

cburandt:
## Describe your changes
Improves the error message during autogenerating tests for the EVerest_API-types.

Adding a new type to the EVerestAPI requires changing of/creating a new lib/everest/everest_api_types/include/.../API.hpp.
Unit-test for (de-)serialization of these tests are auto-generated during the build.
Auto-generation fails for unsupported base-types. The resulting error-message did not point to the actual problem.

The change allows the auto-generator to name the data-structure that caused the problem and clearly states which type cannot be used (e.g. use of an int instead of int32_t).

### DIFF
--- a/lib/everest/everest_api_types/tests/test_file_autogenerator/valuegenerator.py
+++ b/lib/everest/everest_api_types/tests/test_file_autogenerator/valuegenerator.py
@@ -106,6 +106,12 @@ def get_vector_variable_name(variable_name_suffix=""):
 class ValueGenerator:
     manual_generator = ManualGenerator()
     base_types = ["int32_t", "int64_t", "float", "std::string", "bool"]
+    unsupported_base_types = ["void",
+                              "char", "signed char", "unsigned char", "wchar_t", "char16_t", "char32_t", "char8_t",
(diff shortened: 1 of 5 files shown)


hikinggrass:
What's the state of this PR @cburandt @andistorm ?

lib/everest/everest_api_types/tests/test_file_autogenerator/valuegenerator.py
```diff
@@ -216,6 +222,8 @@ def generate_corresponding_field_test(self, field_name, field_type, namespace, i
                                                 field_type, namespace, is_optional)
 
     def generate_corresponding_test(self, original_object, result_object, field_type, namespace, is_optional):
+        if field_type in self.unsupported_base_types:
+            raise TypeError(f"Unsupported type {field_type!r}, supported base types are: {self.base_types}")
```
andistorm: Why not checking the supported types instead?

andistorm (changes requested):
Looks good in general. This PR needs an update.

lib/everest/everest_api_types/tests/test_file_autogenerator/valuegenerator.py
cburandt: The reason is, that field_type can also be a custom type so neither a supported_base_type nor a unsupported_base_type.

cburandt: Removed the hard to maintain unsupported_base_types set. Now it checks if a type had been declared by the user (that is in the API.hpp files) or is one of the few supported_base_types.

PR #1870, opened 2026-02-25
EvseManager: Don't lock without auth

dorezyuk:
## Describe your changes

We want to avoid locking the lock if the user is not authorized. This feature is gated behind the `unlock_when_deauthorized` config flag and is disabled by default. It also removes the `lock_connector_in_state_b` config which aimed to achieve similar but without much impact. 

The change adds transparently a new type `SignalingBool` which extends the `atomic_bool` and replaces the `flag_authorized` so we can signal it to our IEC state machine. In there we add the `authorized` flag so we can react on the authoziation

### DIFF
--- a/modules/EVSE/EvseManager/IECStateMachine.cpp
+++ b/modules/EVSE/EvseManager/IECStateMachine.cpp
@@ -71,9 +71,8 @@ const std::string cpevent_to_string(CPEvent e) {
     throw std::out_of_range("No known string conversion for provided enum of type CPEvent");
 }
 
-IECStateMachine::IECStateMachine(const std::unique_ptr<evse_board_supportIntf>& r_bsp_,
-                                 bool lock_connector_in_state_b_) :
-    r_bsp(r_bsp_), lock_connector_in_state_b(lock_connector_in_state_b_) {
+IECStateMachine::IECStateMachine(const std::unique_ptr<evse_board_supportIntf>& r_bsp_, bool use_authorized_) :
+    r_bsp(r_bsp_), use_authorized(use_authorized_), authorized(!use_authorized_) {
     // feed the state machine whenever the timer expires
     timeout_state_c1.signal_reached.connect([this]() { feed_state_machine(std::nullopt); });
     timeout_unlock_state_F.signal_reached.connect([this]() { feed_state_machine(std::nullopt); });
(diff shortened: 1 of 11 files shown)


modules/EVSE/EvseManager/IECStateMachine.cpp
```diff
@@ -71,9 +71,8 @@ const std::string cpevent_to_string(CPEvent e) {
     throw std::out_of_range("No known string conversion for provided enum of type CPEvent");
 }
 
-IECStateMachine::IECStateMachine(const std::unique_ptr<evse_board_supportIntf>& r_bsp_,
-                                 bool lock_connector_in_state_b_) :
-    r_bsp(r_bsp_), lock_connector_in_state_b(lock_connector_in_state_b_) {
+IECStateMachine::IECStateMachine(const std::unique_ptr<evse_board_supportIntf>& r_bsp_, bool use_authorized_) :
+    r_bsp(r_bsp_), use_authorized(use_authorized_), authorized(!use_authorized_) {
```
dorezyuk: if `use_authorized_` is false we set the `authorized` flag to always be true and basically disable its impact

dorezyuk:
@Pietfried @hikinggrass We tested it and it should work for us. I think we can proceed here

modules/EVSE/EvseManager/manifest.yaml
```diff
@@ -297,8 +297,16 @@ config:
...
       Indicates if the connector lock should be locked in state B. If set to false, connector will remain unlocked in CP state B
       and this violates IEC61851-1:2019 D.6.5 Table D.9 line 4 and should not be used in public environments!
+      This feature is deprecated.
     type: boolean
     default: true
+  unlock_when_deauthorized:
+    description: >-
+      Indicates if the connector lock should be locked in state B/C regardless of if a user authorized. If set to true,
+      connector will remain unlocked in CP state B/C if the relays are open and no user is authorized. This violates
+      violates IEC61851-1:2019 D.6.5 Table D.9 line 4 and should not be used in public environments!
```
Pietfried: nit
```suggestion
      IEC61851-1:2019 D.6.5 Table D.9 line 4 and should not be used in public environments!
```

modules/EVSE/EvseManager/scoped_lock_timeout.hpp
```diff
@@ -61,6 +61,7 @@ enum class MutexDescription {
     IEC_set_cp_state_F,
     IEC_allow_power_on,
     IEC_force_unlock,
+    IEC_set_authorized,
```
Pietfried: afaics its never used. Remove?

Pietfried (dismissed):

Pietfried (approved):

PR #1873, opened 2026-02-26
feat(PacketSniffer): enhance session logging path with timestamp and session ID, if config path is used

ManuelJanocha:
## Describe your changes

This PR implements timestamp generation for pcap filenames in the PacketSniffer module. The code will be active if session_logging_path is set:

Example pcap filename: `2026-02-26_06-14-35+0100_7c2a986a-3d37-4cca-b1ca-c7ea48af10b0.pcap`

N/A

### DIFF
--- a/modules/Misc/PacketSniffer/PacketSniffer.cpp
+++ b/modules/Misc/PacketSniffer/PacketSniffer.cpp
@@ -71,10 +71,8 @@ void PacketSniffer::ready() {
 void PacketSniffer::capture(const std::string& logpath, const std::string& session_id) {
     already_started = true;
 
-    std::string fn;
-    if (config.session_logging_path.empty()) {
-        fn = fmt::format("{}/ethernet-traffic.pcap", logpath);
-    } else {
+    std::string fn = fmt::format("{}/ethernet-traffic.pcap", logpath);
+    if (not config.session_logging_path.empty()) {
         const auto now = std::chrono::system_clock::now();
         const auto time_t_now = std::chrono::system_clock::to_time_t(now);
         std::tm local_tm{};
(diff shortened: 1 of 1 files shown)


modules/Misc/PacketSniffer/PacketSniffer.cpp
```diff
@@ -68,7 +70,18 @@ void PacketSniffer::ready() {
 
 void PacketSniffer::capture(const std::string& logpath, const std::string& session_id) {
     already_started = true;
-    const std::string fn = fmt::format("{}/ethernet-traffic.pcap", logpath);
+
+    std::string fn;
+    if (config.session_logging_path.empty()) {
+        fn = fmt::format("{}/ethernet-traffic.pcap", logpath);
+    } else {
```
SebaLukas: An alternative would be:
```cpp
std::string fn = fmt::format("{}/ethernet-traffic.pcap", logpath);
if (not config.session_logging_path.empty()) {
    const auto now = std::chrono::system_clock::now();
    [...]
}
```

modules/Misc/PacketSniffer/PacketSniffer.cpp
```diff
@@ -68,7 +70,16 @@ void PacketSniffer::ready() {
 
 void PacketSniffer::capture(const std::string& logpath, const std::string& session_id) {
     already_started = true;
-    const std::string fn = fmt::format("{}/ethernet-traffic.pcap", logpath);
+
+    std::string fn = fmt::format("{}/ethernet-traffic.pcap", logpath);
+    if (not config.session_logging_path.empty()) {
```
SebaLukas, line 75: I am asking myself right now why not add timestamp and session_id for every pcap file as default. Not only when `session_logging_path` is set

ManuelJanocha: By default, the session start event creates a new folder that includes the session ID. I am not sure if a timestamp is also part of the folder name. Therefore, it is acceptable to use the same file name within these folders.

SebaLukas (approved):

PR #1893, opened 2026-02-27
fix(EvseManager): Apply EVSE limits on DC target values if EV doesnt update its target values

Pietfried:
## Describe your changes
The ISO15118 stack / EvseV2G only publishes EV target voltage/current on change, so updated EVSE limits from energy management or BSP caps were never enforced when the EV kept requesting the same values.

Store the raw EV targets as atomics and extract the clamping logic into process_dc_ev_target_voltage_current(). The Charger state machine now now calls process_dc_ev_target_voltage_current via signal_dc_enforce_target_limits to re-clamp against current EVSE limits, in case apply_new_target_voltage_current has no been called for 5 seconds (`DC_ENFORCE_TARGET_LIMITS_INTERVAL_MS`)

TODO: verify thread safety because target limits are now also controled by charger state machine thread

### DIFF
--- a/modules/EVSE/EvseManager/EvseManager.cpp
+++ b/modules/EVSE/EvseManager/EvseManager.cpp
@@ -2290,6 +2290,10 @@ void EvseManager::powersupply_DC_on() {
 // input voltage/current is what the evse/car would like to set.
 // if it is more then what the energymanager gave us, we can limit it here.
 bool EvseManager::powersupply_DC_set(double _voltage, double _current) {
+    if (last_power_supply_voltage == _voltage and last_power_supply_current == _current) {
+        return true;
+    }
+
     double voltage = _voltage;
     double current = _current;
     static bool last_is_actually_exporting_to_grid{false};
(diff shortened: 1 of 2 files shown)


corneliusclaussen (dismissed):

modules/EVSE/EvseManager/Charger.hpp
```diff
@@ -398,12 +400,15 @@ class Charger {
 
         std::chrono::time_point<std::chrono::steady_clock> fatal_error_became_active;
         bool fatal_error_timer_running{false};
+
```
SebaLukas, line 403: Is the new line intentional?

modules/EVSE/EvseManager/EvseManager.cpp
```diff
@@ -2574,10 +2526,75 @@ types::evse_manager::EVInfo EvseManager::get_ev_info() {
...
+    }
+
+    bool target_changed = false;
+    if (clamped_voltage not_eq latest_target_voltage or clamped_current not_eq latest_target_current) {
+        latest_target_voltage = clamped_voltage;
+        latest_target_current = clamped_current;
+        target_changed = true;
+    }
+
+    apply_new_target_voltage_current();
```
SebaLukas, line 2582: So every time the ev requieres a new voltage and/or current, the target voltage and current will be applied to the power supply.
I get why we want that. But hopefully, this will not generate too much traffic on the MQTT bus. It may be the case that the power supply receives a new target every 100 ms.

SebaLukas (dismissed):

modules/EVSE/EvseManager/EvseManager.cpp
Pietfried: I adressed this concern with https://github.com/EVerest/everest-core/pull/1893/changes/aa2a9b5ea26c211888927f400f046fee23ef8ee3

PR #1895, opened 2026-02-27
Bazel: Add aarch64 and armv7 cross compilation toolchains

dorezyuk:
## Describe your changes

Adds 4 cross compilation toolchains

- aarch64-linux-gnu: Here glibc is picked to be 2.35 which is a compatible version with yocto kirkstone. 
- aarch64-linux-musl: Musl version - everything is liked into the binaries
- armv7-linux-gnueabihf: As as for aarch64-linux-gnu - sysroot uses glibc 2.35.
- armv7-linux-musleabihf: As as armv7-linux-gnueabihf but with musl as libc

Some notes: the bcr version of openssl did not support armv7 musl cross compilation, so I've added a switch/select and used the foreign_rules_cc as fallback


SirVer (dismissed):
Looks pretty similar to what we do internally now, So i guess that is good.

hikinggrass (dismissed):
looks like it needs a rebase though

hikinggrass (approved):

andistorm (approved):
LGTM 🐧

PR #1896, opened 2026-02-27
EvseSlac: Make TT_EVSE_SLAC_INIT_MS configurable

FaHaGit:
## Describe your changes

This PR introduces a configurable TT_EVSE_SLAC_INIT_MS in EvseSlac to improve Fake DC behavior.

Not all EV AC implementations support HLC and respond after applying 5% duty cycle. To ensure robust interoperability, a fast fallback to AC basic charging is important when no SLAC request is received.
The current default timeout of 40 seconds is too high for fallback scenarios. Therefore, this PR makes the SLAC init timeout configurable so EVSE's can use a shorter, appropriate fallback time.

### DIFF
--- a/modules/EVSE/EvseSlac/manifest.yaml
+++ b/modules/EVSE/EvseSlac/manifest.yaml
@@ -82,6 +82,14 @@ provides:
           On some hardware platforms with multiple EvseSlac modules loaded this can be necessary to ensure that the initial query of the device information does not happen at the same time.
         type: integer
         default: 0
+      slac_init_timeout_ms:
+        description: >-
+          Timeout for CM_SLAC_PARM.REQ. Values between 10000 and 20000 can be set at your own risk, as they
+          violate ISO15118-3. Experience has shown that this should not cause any problems.
+        type: integer
+        minimum: 10000
+        maximum: 50000
+        default: 40000
 metadata:
   base_license: https://directory.fsf.org/wiki/License:BSD-3-Clause-Clear
   license: https://opensource.org/licenses/Apache-2.0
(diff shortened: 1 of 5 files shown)


modules/EVSE/EvseSlac/manifest.yaml
```diff
@@ -82,6 +82,12 @@ provides:
           On some hardware platforms with multiple EvseSlac modules loaded this can be necessary to ensure that the initial query of the device information does not happen at the same time.
         type: integer
         default: 0
+      slac_init_timeout_ms:
+        description: Timeout for CM_SLAC_PARM.REQ. Any value below 20000 violates ISO15118-3.
+        type: integer
+        minimum: 10000
```
SebaLukas, line 90: In the description it says `Any value below 20000 violates ISO15118-3`. Theoretically then the minumum value should be `20000`? 😅

SebaLukas (commented):
Looking good 👍 
Look over my comment.

modules/EVSE/EvseSlac/manifest.yaml
FaHaGit: In my opinion, less than 20 seconds should be allowed, even if it is no longer standard-compliant. From my own experience, I don't think it's acceptable for legacy EV implementations to have to wait 20 seconds for charging to start. I think the description is fine as it is. I would be unhappy to increase it to 20 seconds.

SebaLukas: If 10000 ms is to be allowed, this must also be clearly stated in the description like so:
`All values between 10s and 20s can be set at your own risk, as they violate -3. Experience has shown that this should not cause any problems...`
An updated description would be appreciated :)

FaHaGit: I have improved the description as proposed

SebaLukas (approved):

PR #1898, opened 2026-02-27
Fix(slac): slac inline enum declaration

### LINKED ISSUE #1323: The SLAC interface uses the unsupported inline enum declaration
### Describe the bug

The `state` variable of the SLAC interface uses the unsupported inline enum declaration:
https://github.com/EVerest/everest-core/blob/72b342e3d55c191db9f1a2544acfec896f93c06c/interfaces/slac.yaml#L39-L45

Looks like this feature was implemented in https://github.com/EVerest/everest-utils/pull/12 but it's disabled by default (with no way to enable it).

### EVerest Domain

SLAC

### Affected EVerest Module

_No response_

### To Reproduce

_No response_

### Anything else?

_No response_

ThatsLucas:
Added types/slac.yaml to fix #1323 - The SLAC interface uses the unsupported inline enum declaration

Updated everything to work with the new types.

https://github.com/EVerest/everest-core/issues/1323

- [x ] I read the [contribution documentation](https://github.com/EVerest/EVerest/blob/main/CONTRIBUTING.md) and made sure that my changes meet its requirements


modules/Simulation/SlacSimulator/evse/slacImpl.cpp
```diff
@@ -12,7 +12,7 @@ void slacImpl::init() {
 }
 
 void slacImpl::ready() {
-    publish_state(state_to_string(state));
+    publish_state(types::slac::string_to_state(state_to_string(state)));
```
mhei: Converting forth and back again to string look complicated. Maybe we should get rid of the `utils/state.{c,hpp}`  in SlacSimulator completely now that we have a nice type?

ThatsLucas: hmm yea you'r right, i'll take a look at this

ThatsLucas: It is better now ?

modules/EV/EvManager/main/car_simulation.hpp
```diff
@@ -74,8 +74,12 @@ class CarSimulation {
         sim_data.pwm_duty_cycle = pwm_duty_cycle;
     }
 
-    void set_slac_state(std::string slac_state) {
-        sim_data.slac_state = std::move(slac_state);
+    void set_slac_state(const std::string& slac_state) {
+        sim_data.slac_state = slac_state;
+    }
```
SebaLukas: Why did you change std::move to a const ref?

modules/EV/EvManager/main/car_simulation.hpp
```diff
@@ -74,8 +74,12 @@ class CarSimulation {
...
+
+    void set_slac_state(types::slac::State slac_state) {
+        sim_data.slac_state = types::slac::state_to_string(slac_state);
```
SebaLukas: If we have now this nice `types::slac::State` type, then it should be used as well in the EvManager and not a `std::string`.

modules/EVSE/EvseSlac/main/slacImpl.cpp
```diff
@@ -65,7 +65,7 @@ void slacImpl::run() {
 
     callbacks.signal_dlink_ready = [this](bool value) { publish_dlink_ready(value); };
 
-    callbacks.signal_state = [this](const std::string& value) { publish_state(value); };
+    callbacks.signal_state = [this](const std::string& value) { publish_state(types::slac::string_to_state(value)); };
```
SebaLukas: My only problem with using `string_to_state()` that catching an exception is not handled here. As well in `EvSlac` module.

SebaLukas (changes requested):
Looking good 👍 
See my comments

modules/EV/EvManager/main/car_simulation.hpp
ThatsLucas: I saw that the function was not modifying slac_state so i added const ref.
And from what i understand its not logical to move a const value.

If you think it was better before i can change it back :p

modules/EV/EvManager/main/car_simulation.hpp
ThatsLucas: Yup you'r right gonna do that shortly

modules/API/slac_API/slac_API.cpp
```diff
@@ -15,6 +15,23 @@ namespace API_types_ext = ev_API_types::slac;
...
+    case API_types_ext::State::UNMATCHED:
+        return types::slac::State::UNMATCHED;
+    case API_types_ext::State::MATCHING:
+        return types::slac::State::MATCHING;
+    case API_types_ext::State::MATCHED:
+        return types::slac::State::MATCHED;
+    }
+
+    throw std::out_of_range("Unknown API slac::State value");
+}
```
cburandt: This should be in lib/everest/everest_api_types/src/everest_api_types/slac/wrapper.cpp, resp. ..../private_includes/everest_api_types/slac/wrapper.hpp.

Look for the to_internal_api/to_external_api functions of UploadLogsStatus in the respective system/wrapper.hpp/wrapper.cpp to see how the internal/external conversion routines are organized in the API types library.

You will need to:

in lib/everest/everest_api_types/
- add wrapper.hpp/cpp in the right place
- write to_int/external_api functions (use functions for UploadLogsStatus as a template)
- add cpp file to cmake target

in the slac_API module:
- include slac/wrapper.hpp
- call to_internal_api in the generate_api_var_state function

cburandt (changes requested):

modules/EVSE/EvseSlac/main/slacImpl.cpp
ThatsLucas: I didn't thought of that, now im catching it.

I just have a question here :
```
        try {
            publish_state(types::slac::string_to_state(value));
            EVLOG_error << fmt::format("Tried to publish unknown SLAC state '{}'. Error: {}", value, e.what());
            publish_dlink_ready(false);
        }
        
```
Should i keep publishing dlink false ? I guess it should never happen but just to be sure

modules/EV/EvManager/main/car_simulation.hpp
SebaLukas: In the previous function defintion it was not a const ref.
So I think move a `std::string` with `std::move` is valid. So please revert it back :+1:

modules/API/slac_API/slac_API.cpp
ThatsLucas: Okay i followed what you asked, i hope i did everything well, tell me if smth is missing.
Thanks for the help, it really helped me understand more how everest is working in general

modules/EV/EvManager/main/car_simulation.hpp
ThatsLucas: Should be good now

modules/EV/EvManager/main/car_simulation.hpp
ThatsLucas: I put the old definition back as it was. But about std::move im not sure how to do it.
Now that im catching a throw, i need slac_state in EVLOG_error and at this time, it might be empty with an std::move.

lib/everest/everest_api_types/src/everest_api_types/slac/wrapper.cpp
```diff
@@ -0,0 +1,39 @@
...
+    using TarT = SlacState_Internal;
+    switch (val) {
+    case SrcT::UNMATCHED:
+        return TarT::UNMATCHED;
+    case SrcT::MATCHING:
+        return TarT::MATCHING;
+    case SrcT::MATCHED:
+        return TarT::MATCHED;
+    }
+    throw std::out_of_range("Unexpected value for SlacState_Internal" + serialize(val));
```
cburandt:
```suggestion
    throw std::out_of_range("Unexpected value for everest::lib::API::V1_0::types::slac::SlacState_External");
```

lib/everest/everest_api_types/src/everest_api_types/slac/wrapper.cpp
```diff
@@ -0,0 +1,39 @@
...
+    using TarT = SlacState_External;
+    switch (val) {
+    case SrcT::UNMATCHED:
+        return TarT::UNMATCHED;
+    case SrcT::MATCHING:
+        return TarT::MATCHING;
+    case SrcT::MATCHED:
+        return TarT::MATCHED;
+    }
+    throw std::out_of_range("Unexpected value for SlacState_External");
```
cburandt:
```suggestion
    throw std::out_of_range("Unexpected value for everest::lib::API::V1_0::types::slac::SlacState_Internal");
```

cburandt (changes requested):
The code in wrapper.[hc]pp looks almost fine. The exception message texts where sort of mixed up, but that was my bad, since the example I pointed you at, had the same confusion. If you apply my suggestions below, this will be fixed.

lib/everest/everest_api_types/src/everest_api_types/slac/wrapper.cpp
ThatsLucas: No problem, done

ThatsLucas:
Do i have to do something about Codacy ?

barsnick:
> Do i have to do something about Codacy ?

These checks?
https://github.com/EVerest/everest-core/pull/1898/checks?check_run_id=65529027064

No, you cannot fix those, and they do not block the PR (unlike other failing checks).

ThatsLucas:
ok great

modules/EV/EvManager/main/car_simulatorImpl.cpp
```diff
@@ -344,7 +344,8 @@ void car_simulatorImpl::subscribe_to_variables_on_init() {
     // subscribe slac_state
     if (!mod->r_slac.empty()) {
         const auto& slac = mod->r_slac.at(0);
-        slac->subscribe_state([this](const auto& state) { car_simulation->set_slac_state(state); });
+        slac->subscribe_state(
+            [this](const auto& state) { car_simulation->set_slac_state(types::slac::state_to_string(state)); });
```
SebaLukas: Remove `types::slac::state_to_string()`

modules/EV/EvManager/main/car_simulation.hpp
```diff
@@ -75,7 +75,15 @@ class CarSimulation {
     }
 
     void set_slac_state(std::string slac_state) {
-        sim_data.slac_state = std::move(slac_state);
+        try {
+            sim_data.slac_state = types::slac::string_to_state(slac_state);
+        } catch (const std::exception& e) {
+            EVLOG_error << fmt::format("Tried to set unknown SLAC state '{}'. Error: {}", slac_state, e.what());
+        }
+    }
```
SebaLukas: Delete this function. We dont need this function (`std::string` as argument) anymore

modules/EVSE/EvseSlac/main/slacImpl.cpp
SebaLukas: Please remove `publish_dlink_ready(false);`

SebaLukas (approved):

cburandt (approved):

PR #1899, opened 2026-02-27
fix(OCPP16): remove brackets [] from MeterPublicKey format

Pietfried:
## Describe your changes
Change MeterPublicKey[n] to MeterPublicKeyn to match the format used by the OCA Eichrecht Whitepaper

While the format wasn't specifically clear from the requirements of the whitepaper, the example indicates that a format without [n] shall be used.

image

### DIFF
--- a/lib/everest/ocpp/lib/ocpp/v16/charge_point_configuration_base.cpp
+++ b/lib/everest/ocpp/lib/ocpp/v16/charge_point_configuration_base.cpp
@@ -258,11 +258,12 @@ ChargePointConfigurationBase::csvToMeasurandWithPhaseVector(const std::string& c
     return result;
 }
 
-std::optional<std::uint32_t> ChargePointConfigurationBase::extractConnectorId(const std::string& str) {
-    // example string "MeterPublicKey[1]"
+std::optional<std::uint32_t>
+ChargePointConfigurationBase::extractConnectorIdFromMeterPublicKey(const std::string& str) {
+    // example string "MeterPublicKey1"
 
     std::optional<std::uint32_t> result;
-    const std::regex id_regex(R"(^.+\[(\d+)\]$)");
+    const std::regex id_regex(R"(^MeterPublicKey(\d+)$)");
     std::smatch id_match;
 
     if (std::regex_match(str, id_match, id_regex)) {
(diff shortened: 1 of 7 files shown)


lib/everest/ocpp/lib/ocpp/v16/charge_point_configuration_base.cpp
```diff
@@ -259,10 +259,10 @@ ChargePointConfigurationBase::csvToMeasurandWithPhaseVector(const std::string& c
 }
 
 std::optional<std::uint32_t> ChargePointConfigurationBase::extractConnectorId(const std::string& str) {
-    // example string "MeterPublicKey[1]"
+    // example string "MeterPublicKey1"
 
     std::optional<std::uint32_t> result;
-    const std::regex id_regex(R"(^.+\[(\d+)\]$)");
+    const std::regex id_regex(R"(^.*[^\d](\d+)$)");;
```
james-ctc: extra semicolon on the line.
For the regex wouldn't the following be simpler: ^MeterPublicKey(\d+)$

lib/everest/ocpp/lib/ocpp/v16/charge_point_configuration_devicemodel.cpp
```diff
@@ -2613,8 +2613,8 @@ std::optional<KeyValue> ChargePointConfigurationDeviceModel::get(const CiString<
                 get_value = false;
             }
         }
-        // check keys starting MeterPublicKey[
-        if (key_str.find("MeterPublicKey[") == 0) {
+        // check keys starting MeterPublicKey
+        if (key_str.find("MeterPublicKey") == 0) {
```
james-ctc: rfind is used elsewhere - probably because it checks there is only one instance of the string
`key_str.find("MeterPublicKey") == 0` will probably match "MeterPublicKeyMeterPublicKey8"

lib/everest/ocpp/lib/ocpp/v16/charge_point_configuration_devicemodel.cpp
```diff
@@ -2941,7 +2941,7 @@ std::optional<ConfigurationStatus> ChargePointConfigurationDeviceModel::set(cons
             } else {
                 result = ConfigurationStatus::NotSupported;
             }
-        } else if (key_str.find("MeterPublicKey[") == 0) {
+        } else if (key_str.find("MeterPublicKey") == 0) {
```
james-ctc: rfind is probably better

lib/everest/ocpp/lib/ocpp/v16/charge_point_configuration_base.cpp
james-ctc: I've checked and this function is only used for extracting the connector from MeterPublicKey.
An alternate regexp: ^[^\d]+(\d+)$
My preference would be to rename the method so it is clear it is for MeterPublicKey only.

james-ctc (approved):

PR #1900, opened 2026-03-01
Fix cable check race condition causing permanent charger inoperability (#1392)

### LINKED ISSUE #1392: Race condition on EvseManager::fail_cable_check()
### Describe the bug

Hello,
I think the following happens while I interrupt the cable check phase by simulating an unplug event.
- `ErrorHandling::process_error()` will clear all errors and set `shared_context.error_prevent_charging_flag` to `false` as a consequence of the unplug event.
- Later `fail_cable_check()` will `call charger->set_hlc_error()` which sets `shared_context.error_prevent_charging_flag` to `true`.
- The flag stays true for the next charging session which won't proceed further and fail again during the cable check.

I know that this is a tricky corner case because it's impossible to remove the connector during the cable check as it is blocked by the vehicle. However, some vehicles have a manual switch that allows the connector to be unlocked and consequently removed during the charging phase.

### EVerest Domain

Charge Control

### Affected EVerest Module

Module is EvseManager. Files are EvseManager.cpp, Charger.cpp, ErrorHandling.cpp

### To Reproduce

_No response_

### Anything else?

_No response_

rishabhvaish:
## Summary
- Fixes a race condition between the detached `cable_check` thread and unplug handling in the state machine
- When a car is unplugged during DC cable check, the detached thread can call `fail_cable_check()` **after** `clear_errors_on_unplug()` has already run
- This leaves `MREC11CableCheckFault` permanently raised, making the charger inoperative until EVerest is restarted
- Fix: check if the session has already ended (state is `Idle` or `Finished`) before raising the cable check fault error

## Files changed
- `modules/EVSE/EvseManager/EvseManager.cpp` — added state check in `fail_cable_check()` before raising error

## Test plan
- [ ] Start a DC charging session and unplug during cable check phase
- [ ] Verify charger returns to Idle without permanent `MREC11CableCheckFault`
- [ ] Verify normal cable check failures (e.g., isolation fault) still raise the error correctly
- [ ] Verify next charging session starts successfully without reboot

Fixes #1392

### DIFF
--- a/modules/EVSE/EvseManager/EvseManager.cpp
+++ b/modules/EVSE/EvseManager/EvseManager.cpp
@@ -2535,13 +2535,18 @@ void EvseManager::fail_cable_check(const std::string& reason) {
         }
         r_hlc[0]->call_cable_check_finished(false);
     }
-
-    // Raising a cable check fault should not happen if a cancel_transaction (DeAuthorized) is triggered
-    // during cable check
+    // Raising a cable check fault should not happen if:
+    // - a cancel_transaction (DeAuthorized) is triggered during cable check
+    // - the car has already been unplugged (Idle/Finished), which prevents a race condition
+    //   where the detached cable check thread raises an error after clear_errors_on_unplug()
+    //   has already run, leaving the charger permanently inoperative (see GitHub issue #1392)
+    const auto current_state = charger->get_current_state();
     const auto last_stop_transaction_reason = charger->get_last_stop_transaction_reason();
-    if (not last_stop_transaction_reason.has_value() or
-        (last_stop_transaction_reason.has_value() and
-         last_stop_transaction_reason.value() != types::evse_manager::StopTransactionReason::DeAuthorized)) {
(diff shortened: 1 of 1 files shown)


SebaLukas:
Please resolve the conflicts before we can review your changes (updating your branch to latest main).

modules/EVSE/EvseManager/EvseManager.cpp
SebaLukas: Can you add your changes to this check. In your case if `current_state` is not `Idle` or `Finished` the error should be raised

rishabhvaish: Good point — folded the Idle/Finished check into the existing raise_cable_check_fault guard in 71707f2c instead of the separate early return.

SebaLukas (dismissed):

SebaLukas (changes requested):
Please remove this part of the commit messages:
`Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>`
Also please squash your commits into one. Then I think we can merge this one

SebaLukas (approved):

PR #1901, opened 2026-03-02
Refactor/framework message handler

djchhp:
## Describe your changes
This refactors the message handler in the EVerest framework to rely on the async utilities form lib/everest/util

It removes the explicit use of mutex and condition_var completely. thread_safe_queue and monitor are used instead.

### DIFF
--- a/lib/everest/framework/include/utils/message_handler.hpp
+++ b/lib/everest/framework/include/utils/message_handler.hpp
@@ -4,18 +4,18 @@
 #pragma once
 
 #include <atomic>
-#include <condition_variable>
+#include <chrono>
 #include <map>
-#include <memory>
-#include <mutex>
-#include <queue>
 #include <string>
 #include <thread>
 #include <unordered_map>
 #include <unordered_set>
 #include <vector>
 
+#include <everest/util/async/monitor.hpp>
 #include <everest/util/async/thread_pool_scaling.hpp>
+#include <everest/util/queue/thread_safe_queue.hpp>
+
 #include <utils/message_queue.hpp>
 #include <utils/types.hpp>
 
@@ -24,13 +24,19 @@ using CmdId = std::string;
 
 namespace Everest {
 
-constexpr int THREAD_POOL_SCALING_LATENCY_THRESHOLD_MS = 50;
-constexpr int THREAD_POOL_SCALING_LATENCY_THREAD_IDLE_TIMEOUT_S = 2;
-constexpr int THREAD_POOL_SCALING_MIN_THREAD_COUNT = 1;
+constexpr std::size_t THREAD_POOL_SCALING_LATENCY_THRESHOLD_MS = 50;
+constexpr std::chrono::seconds THREAD_POOL_SCALING_IDLE_TIMEOUT{2};
+constexpr std::size_t THREAD_POOL_SCALING_MIN_THREAD_COUNT = 1;
+constexpr std::size_t MAX_PENDING_MESSAGES_PER_TOPIC = 100;
 
-/// \brief Handles message dispatching and thread-safe queuing of different message types. This class uses two separate
-/// threads and message queues: one for operation messages (vars, cmds, errors, GetConfig, ModuleReady) and one for
-/// result messages (cmd results, GetConfig responses).
+/// \brief Handles message dispatching and thread-safe queuing of different message types.
+///
+/// Messages are routed to one of four channels based on their type:
+///   - operation_message_queue → operation_dispatcher_thread → operation_thread_pool
+///     (vars, cmds, errors, GetConfig, ModuleReady — parallel across topics, serial per topic)
+///   - result_message_queue    → result_worker_thread (cmd results, GetConfig responses — serial)
+///   - external_mqtt_message_queue → external_mqtt_worker_thread (external MQTT — serial)
+///   - GlobalReady             → ready_thread (one-shot, spawned per message)
 class MessageHandler {
 public:
     MessageHandler();
@@ -45,7 +51,31 @@ public:
     /// \brief Registers a \p handler for a specific \p topic
     void register_handler(const std::string& topic, std::shared_ptr<TypedHandler> handler);
 
+    using SharedTypedHandler = std::shared_ptr<TypedHandler>;
+    using SingleHandlerMap = std::map<MqttTopic, SharedTypedHandler>;
+    using MultiHandlerMap = std::map<MqttTopic, std::vector<SharedTypedHandler>>;
+
 private:
+    struct OperationTopics {
+        std::unordered_set<std::string> in_flight;
+        std::unordered_map<std::string, everest::lib::util::simple_queue<ParsedMessage>> pending_messages;
+    };
+
+    struct ResponseHandlers {
+        std::map<CmdId, std::shared_ptr<TypedHandler>> cmd; // cmd result handlers of module
+        std::shared_ptr<TypedHandler> config;               // get module config response handler of module
+    };
+
+    struct GenericHandlers {
+        MultiHandlerMap var;                // var handlers of module
+        SingleHandlerMap cmd;               // cmd handlers of module
+        MultiHandlerMap error;              // error handlers with wildcard support
+        SingleHandlerMap get_module_config; // get module config handler of manager
+        SharedTypedHandler global_ready;    // global ready handler of module
+        SingleHandlerMap module_ready;      // module ready handlers of manager
+        MultiHandlerMap external_var;       // external MQTT handlers of module
+    };
+
     void run_operation_dispatcher();
     void run_result_message_worker();
     void run_external_mqtt_worker();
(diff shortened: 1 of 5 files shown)


lib/everest/framework/lib/message_handler.cpp
```diff
@@ -1,11 +1,14 @@
 // SPDX-License-Identifier: Apache-2.0
 // Copyright Pionix GmbH and Contributors to EVerest
-
+#include "everest/util/misc/container.hpp"
```
Pietfried, line 4:
```suggestion
```

lib/everest/framework/include/utils/message_handler.hpp
```diff
@@ -5,15 +5,20 @@
 
 #include <atomic>
 #include <condition_variable>
-#include <functional>
 #include <map>
 #include <memory>
 #include <mutex>
 #include <queue>
```
Pietfried: Now stale includes can be removed: mutex, queue, condition_variable

lib/everest/framework/lib/message_handler.cpp
```diff
@@ -82,104 +88,141 @@ void MessageHandler::add(const ParsedMessage& message) {
...
     }
-    if (ready_thread.joinable()) {
-        ready_thread.join();
+}
+
+void MessageHandler::run_operation_dispatcher() {
+    while (auto message = operation_message_queue.wait_and_pop()) {
+        if (not message.has_value()) {
+            return;
+        }
```
Pietfried: This is dead code. We would not enter the while if message has no value
```suggestion
```

lib/everest/framework/lib/message_handler.cpp
```diff
@@ -82,104 +88,141 @@ void MessageHandler::add(const ParsedMessage& message) {
...
 
-        handle_result_message(message.topic, message.data);
+    schedule_operation_message(std::move(*next_message));
+}
+
+void MessageHandler::run_result_message_worker() {
+    while (auto message = result_message_queue.wait_and_pop()) {
+        if (not message.has_value()) {
+            return;
+        }
```
Pietfried: same here
```suggestion
```

lib/everest/framework/lib/message_handler.cpp
```diff
@@ -82,104 +88,141 @@ void MessageHandler::add(const ParsedMessage& message) {
...
 
 void MessageHandler::run_external_mqtt_worker() {
-    while (true) {
-        std::unique_lock<std::mutex> lock(external_mqtt_queue_mutex);
-        external_mqtt_cv.wait(lock, [this] { return !external_mqtt_message_queue.empty() || !running; });
-        if (!running) {
+    while (auto message = external_mqtt_message_queue.wait_and_pop()) {
+        if (not message.has_value()) {
             return;
         }
```
Pietfried: same here
```suggestion
```

lib/everest/framework/lib/message_handler.cpp
```diff
@@ -82,104 +88,141 @@ void MessageHandler::add(const ParsedMessage& message) {
...
-        }
-        result_cv.notify_all();
+        result_message_queue.push(message);
     } else if (msg_type == MqttMessageType::GlobalReady) {
         const auto topic_copy = message.topic;
         const auto data_copy = message.data.at("data");
 
-        ready_thread =
-            std::thread([this, topic_copy, data_copy] { (*global_ready_handler->handler)(topic_copy, data_copy); });
+        operation_thread_pool->run([this, topic_copy, data_copy] {
```
Pietfried: In the framework there is no contract that the ready function should return. For some/few EVerest modules the ready thread actually runs as long as the module and does not return, so a pool thread is permanently consumed for this. For that reason I'd rather keep the separate thread for this and use the pool only for those short-lived tasks.

Pietfried (changes requested):

lib/everest/framework/lib/message_handler.cpp
djchhp: done

lib/everest/framework/include/utils/message_handler.hpp
djchhp: done, also removed other obsolete includes

lib/everest/framework/lib/message_handler.cpp
djchhp: well spotted. done.

lib/everest/framework/lib/message_handler.cpp
djchhp: done

lib/everest/framework/lib/message_handler.cpp
djchhp: done

lib/everest/framework/lib/message_handler.cpp
djchhp: Fair point. I did not think about that. Reverted relevant commit.

Pietfried (dismissed):

mlitre (dismissed):

lib/everest/framework/include/utils/message_handler.hpp
```diff
@@ -24,13 +24,19 @@ using CmdId = std::string;
...
-constexpr int THREAD_POOL_SCALING_LATENCY_THREAD_IDLE_TIMEOUT_S = 2;
-constexpr int THREAD_POOL_SCALING_MIN_THREAD_COUNT = 1;
-
-/// \brief Handles message dispatching and thread-safe queuing of different message types. This class uses two separate
-/// threads and message queues: one for operation messages (vars, cmds, errors, GetConfig, ModuleReady) and one for
-/// result messages (cmd results, GetConfig responses).
+constexpr std::size_t THREAD_POOL_SCALING_LATENCY_THRESHOLD_MS = 50;
+constexpr std::chrono::seconds THREAD_POOL_SCALING_IDLE_TIMEOUT{2};
+constexpr std::size_t THREAD_POOL_SCALING_MIN_THREAD_COUNT = 1;
+constexpr std::size_t MAX_PENDING_MESSAGES_PER_TOPIC = 100;
```
mlitre, line 30: @Pietfried I would like your input on the size of each topic queue. I put 100 but I am not sure what is reasonable

lib/everest/framework/lib/message_handler.cpp
```diff
@@ -135,121 +197,101 @@ void MessageHandler::stop() {
...
-            handle_operation_message(message.topic, message.data);
-        } catch (const std::exception& e) {
-            EVLOG_error << "Exception while handling operation message on topic '" << message.topic
-                        << "': " << e.what();
-        } catch (...) {
-            EVLOG_error << "Unknown exception while handling operation message on topic '" << message.topic << "'";
-        }
-        on_operation_message_done(message.topic);
+        try_action_and_log(handle, "handling operation message", message.topic, message.data);
+        try_action_and_log(done, "on_opereation_message_done", message.topic);
```
Pietfried:
```suggestion
        try_action_and_log(done, "on_operation_message_done", message.topic);
```

lib/everest/util/include/everest/util/misc/bind.hpp
```diff
@@ -12,8 +12,55 @@ namespace everest::lib::util {
...
+ */
+template <typename R, typename C, typename... Args> auto bind_obj(R (C::*func)(Args...), C* obj) {
+    return [func, obj](Args&&... args) -> R { return (obj->*func)(std::forward<Args>(args)...); };
+}
+
+/**
+ * @brief Const-qualified version of bind_obj.
+ *
+ * Overload to support binding to member functions marked with 'const'.
+ * * @tparam R     The return type of the member function.
```
Pietfried: This occurs multiple times
```suggestion
 * @tparam R     The return type of the member function.
```

lib/everest/framework/lib/message_handler.cpp
```diff
@@ -61,9 +62,66 @@ bool check_topic_matches(const std::string& full_topic, const std::string& wildc
...
+    auto const* ptr = everest::lib::util::find_ptr(data, topic);
+    if (ptr != nullptr) {
+        handler_copy = ptr->second;
+    }
+    return handler_copy;
+}
+template <class FtorT, class... Args>
+void try_action_and_log(FtorT const& action, std::string const& error_source, std::string const& topic, Args... args) {
+    try {
+        action(topic, args...);
```
Pietfried: The previous code also passed by reference and since its a copy for every single message prcoessed its quite relevant
```suggestion
template <class FtorT, class... Args>
void try_action_and_log(FtorT const& action, std::string const& error_source, std::string const& topic, Args&&... args) {
    try {
        action(topic, std::forward<Args>(args)...);
```

lib/everest/framework/include/utils/message_handler.hpp
Pietfried: afaics it is used for warnings only and is not enforced. Currently a simple_queue is used. I'd be fine with a warning only for now and 100 feels high enough

Pietfried (changes requested):

lib/everest/framework/lib/message_handler.cpp
djchhp: done

lib/everest/util/include/everest/util/misc/bind.hpp
djchhp: I can only spot this single item and fixed it.

lib/everest/framework/lib/message_handler.cpp
djchhp: good catch. Fixed it.

lib/everest/framework/include/utils/message_handler.hpp
djchhp: I don't see a way to enforce it in a sensible way. bounded queue from utils would help, it blocks further pushing. But that just means things pile up elsewhere. Piling up of messages means, there is a resource problem. Only thing you can do is to terminate.

lib/everest/util/include/everest/util/misc/bind.hpp
Pietfried: ah I found it in other parts of the util lib (e.g. https://github.com/EVerest/everest-core/blob/main/lib/everest/util/include/everest/util/async/thread_pool_scaling.hpp#L113) but its nothing for this PR then.

Pietfried (approved):

PR #1915, opened 2026-03-04
feat(CarloGavazzi_EM580): add configuration for the public key format: binary vs der

florinmihut:

### DIFF
--- a/modules/HardwareDrivers/PowerMeters/CarloGavazzi_EM580/main/powermeterImpl.cpp
+++ b/modules/HardwareDrivers/PowerMeters/CarloGavazzi_EM580/main/powermeterImpl.cpp
@@ -220,12 +220,16 @@ void powermeterImpl::read_signature_config() {
     }
     EVLOG_info << "Signature type detected: " << signature_type_string;
 
-    this->m_public_key_hex = read_public_key_in_hex(this->m_public_key_length_in_bits);
-    EVLOG_info << "Public key (raw, hex): " << this->m_public_key_hex;
+    if (config.public_key_format == "binary") {
+        this->m_public_key_hex = read_public_key_in_hex(this->m_public_key_length_in_bits);
+        EVLOG_info << "Public key (raw, hex): " << this->m_public_key_hex;
+    } else if (config.public_key_format == "der") {
+        this->m_public_key_hex = read_public_key_der_in_hex(der_word_count);
+        EVLOG_info << "Public key (DER, hex): " << this->m_public_key_hex;
+    } else {
+        throw std::runtime_error("invalid public key format: " + config.public_key_format);
+    }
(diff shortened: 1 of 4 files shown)


pionix-juraed (dismissed):
I compiled and tested it, it looks good. The config option works.

modules/HardwareDrivers/PowerMeters/CarloGavazzi_EM580/main/powermeterImpl.cpp
```diff
@@ -220,12 +220,16 @@ void powermeterImpl::read_signature_config() {
...
-    EVLOG_info << "Public key (raw, hex): " << this->m_public_key_hex;
+    if (config.public_key_format == "binary") {
+        this->m_public_key_hex = read_public_key_in_hex(this->m_public_key_length_in_bits);
+        EVLOG_info << "Public key (raw, hex): " << this->m_public_key_hex;
+    } else if (config.public_key_format == "der") {
+        this->m_public_key_hex = read_public_key_der_in_hex(der_word_count);
+        EVLOG_info << "Public key (DER, hex): " << this->m_public_key_hex;
+    } else {
+        throw std::runtime_error("invalid public key format: " + config.public_key_format);
+    }
```
djchhp, line 232: this exception is caught here: https://github.com/EVerest/everest-core/blob/bb8cfdbb69bc1e54f9f11fb7120ead3079df98c9/modules/HardwareDrivers/PowerMeters/CarloGavazzi_EM580/main/powermeterImpl.cpp#L345
 The message printed is misleading. Is the public key strictly necessary to use the driver?

djchhp (changes requested):

modules/HardwareDrivers/PowerMeters/CarloGavazzi_EM580/main/powermeterImpl.cpp
florinmihut: True, the message might be partially misleading but both messages will be printed together so ... it is partially correct - a missconfiguration is stopping the driver to correctly configure the device.
Normally the exception should not happen - a misconfiguration should be detected and raised by the configuration parsing, when the system is starting. 
The pubic key binary format is the standard configuration by other power meters but the CGEM580 has a "special" format in addition available - DER. 
The only moment when the exception might be triggered is when somebody changes the manifest but will forget to change the implementation to actually react to that new option. 
To be honest I wanted a exception here because a warning might be easily ovelook but I did not want to make to big changes because this case is mostly theoretical.

florinmihut: I double checked with Kai, my assumption is wrong, the parser it will not catch it. Need to fix that, good catch.

djchhp (approved):

PR #1916, opened 2026-03-04
feat(EnergyNode): introduce capability limits

mhei:
## Describe your changes

EnergyNodes are currently used for modelling the AC side of grid connections, e.g. as grid_connection_point or for external APIs as reference point handle.

Sometimes the controlling module needs to know which is the maximum power at this point to e.g. calculate deratings of the whole station based on the rated maximum power.

At the moment, this is not easy possible so let's introduce a new capabilities interface variable which allows to tell about these caps.
    
While at, I re-ordered the manifest variables - feels now more intuitive to me.
I also ~~set the minimum phase count value from 0 to 1 and~~ added a default value which should fit most use-cases.

none 


mhei:
@corneliusclaussen , @Pietfried : I would appreciate some brief feedback on whether the direction is okay or complete nonsense. I have tried to explain the motivation, but if it is not clear enough, I am happy to provide further explanations.

If the approach is fine with you, then next step on my side would be to check how to adapt the API module which would be affected by this change. I'm not yet used to your API modules, is there a guide how to adapt? Do I need to manually create the code, or is there any helper tool (similar to ev-cli?) - newbie questions, will dig into this... 😄

Pietfried:
> @corneliusclaussen , @Pietfried : I would appreciate some brief feedback on whether the direction is okay or complete nonsense. I have tried to explain the motivation, but if it is not clear enough, I am happy to provide further explanations.
> 
> If the approach is fine with you, then next step on my side would be to check how to adapt the API module which would be affected by this change. I'm not yet used to your API modules, is there a guide how to adapt? Do I need to manually create the code, or is there any helper tool (similar to ev-cli?) - newbie questions, will dig into this... 😄

The proposed extensions to the `external_energy_limits`  make sense and look good to me!

There is no dedicated guide to adapt the API modules and there is also no code generator for this. You would therefore need to introduce the new type and operation in the [Async API documentation](https://github.com/EVerest/EVerest/blob/main/docs/source/reference/EVerest_API/external_energy_limits_consumer_API.yaml) and in the [everest_api_types lib](https://github.com/EVerest/EVerest/tree/main/lib/everest/everest_api_types) and add the new operation to the [external_energy_limits_consumer_API module](https://github.com/EVerest/EVerest/tree/main/modules/API/external_energy_limits_consumer_API).

[This PR](https://github.com/EVerest/EVerest/pull/2000) can serve as a good example/reference. It also adds a new variable to the internal and external APIs and includes all the required changes.

lib/everest/everest_api_types/src/everest_api_types/energy/codec.cpp
```diff
@@ -69,6 +69,11 @@ std::string serialize(EnforcedLimits const& val) noexcept {
     return result.dump(json_indent);
 }
 
+std::string serialize(CapabilityLimits const& val) noexcept {
+    json result = val;
+    return result.dump(json_indent);
```
Pietfried: With https://github.com/EVerest/EVerest/pull/2047 this has been optimized and can be  
```suggestion
    return nlohmann::json(val).dump(json_indent);
```

lib/everest/everest_api_types/src/everest_api_types/energy/codec.cpp
```diff
@@ -201,4 +211,10 @@ template <> EnforcedLimits deserialize(std::string const& val) {
     return obj;
 }
 
+template <> CapabilityLimits deserialize(std::string const& val) {
+    auto data = json::parse(val);
+    CapabilityLimits obj = data;
+    return obj;
```
Pietfried: same here
```suggestion
    return json::parse(val);
```

modules/EnergyManagement/EnergyNode/manifest.yaml
```diff
@@ -8,27 +8,29 @@ config:
     type: number
     minimum: 0.0
   phase_count:
-    description: phase count limit. Omit if not limited in this fuse.
+    description: phase count of this fuse
     type: integer
-    minimum: 0
+    minimum: 1
```
Pietfried: Albeit correct, this is technically a breaking change that should be avoided according to https://everest.github.io/nightly/project/releases/breaking-changes.html#id2

Pietfried (changes requested):
In general this looks good to me. See some comments in line.

modules/EnergyManagement/EnergyNode/manifest.yaml
mhei: Right. Then I would drop this line from this PR for now and we should address it later.

Pietfried (dismissed):

mhei:
Note: just force-pushed after rebase due to https://github.com/EVerest/EVerest/pull/1938

Pietfried:
> Note: just force-pushed after rebase due to #1938

Force-pushing alone shouldnt dismiss the review if the diff did not change. Have you had to resolve conflicts?

Pietfried:
> > Note: just force-pushed after rebase due to #1938
> 
> Force-pushing alone shouldnt dismiss the review if the diff did not change. Have you had to resolve conflicts?

Ah the module directory moved, got it!

Pietfried (approved):

Issue #1841, opened 2026-02-19
Adding ISO-2 state PaymentServiceSelection


SebaLukas:

Dalachowsky:
Work on this state will be continued by @razvanphp

PR #1919, opened 2026-03-05
Draft: libiso15118: d2: State ServiceDiscovery (#1840)

Dalachowsky:
## Describe your changes

Did a general draft of the ServiceDiscovery state.
Moved datatypes to msg_data_types.hpp.
Also noticed that there was a duplicate EnergyTransferMode
definition inside ServiceDiscovery and ChargeParameterDiscovery.

#1840 

### DIFF
--- a/lib/everest/iso15118/include/iso15118/d2/config.hpp
+++ b/lib/everest/iso15118/include/iso15118/d2/config.hpp
@@ -3,18 +3,35 @@
 #pragma once
 
 #include <string>
+#include <utility>
+#include <vector>
+
+#include <iso15118/message/d2/msg_data_types.hpp>
 
 namespace iso15118::d2 {
 
+namespace dt = msg::data_types;
+
 struct EvseSetupConfig {
     std::string evse_id;
+    dt::ChargeService charge_service;
+    std::vector<dt::PaymentOption> supported_payment_options;
+    dt::ServiceList offered_services;
 };
 
 // This should only have EVSE information
 struct SessionConfig {
-    explicit SessionConfig(EvseSetupConfig config) : evse_id(std::move(config.evse_id)) {};
+    // TODO(kd) move the constructor to .cpp file
+    explicit SessionConfig(EvseSetupConfig config) :
+        evse_id(std::move(config.evse_id)),
+        charge_service(std::move(config.charge_service)),
+        supported_payment_options(std::move(config.supported_payment_options)),
+        offered_services(config.offered_services){};
 
     std::string evse_id;
+    dt::ChargeService charge_service;
+    std::vector<dt::PaymentOption> supported_payment_options;
+    dt::ServiceList offered_services;
 };
 
 } // namespace iso15118::d2
--- a/lib/everest/iso15118/include/iso15118/d2/config.hpp
+++ b/lib/everest/iso15118/include/iso15118/d2/config.hpp
@@ -14,9 +14,9 @@ namespace dt = msg::data_types;
 
 struct EvseSetupConfig {
     std::string evse_id;
-    dt::ChargeService charge_service;
+    std::vector<dt::EnergyTransferMode> supported_energy_transfer_modes;
     std::vector<dt::PaymentOption> supported_payment_options;
-    dt::ServiceList offered_services;
+    std::vector<dt::ServiceID> offered_services;
 };
 
 // This should only have EVSE information
@@ -24,14 +24,14 @@ struct SessionConfig {
     // TODO(kd) move the constructor to .cpp file
     explicit SessionConfig(EvseSetupConfig config) :
         evse_id(std::move(config.evse_id)),
-        charge_service(std::move(config.charge_service)),
+        supported_energy_transfer_modes(std::move(config.supported_energy_transfer_modes)),
         supported_payment_options(std::move(config.supported_payment_options)),
-        offered_services(config.offered_services){};
+        offered_services(std::move(config.offered_services)){};
 
     std::string evse_id;
-    dt::ChargeService charge_service;
+    std::vector<dt::EnergyTransferMode> supported_energy_transfer_modes;
     std::vector<dt::PaymentOption> supported_payment_options;
-    dt::ServiceList offered_services;
+    std::vector<dt::ServiceID> offered_services;
 };
 
 } // namespace iso15118::d2
--- a/lib/everest/iso15118/include/iso15118/d2/state/payment_service_selection.hpp
+++ b/lib/everest/iso15118/include/iso15118/d2/state/payment_service_selection.hpp
@@ -7,8 +7,8 @@
 
 namespace iso15118::d2::state {
 
-struct PostServiceDiscovery : public StateBase {
-    PostServiceDiscovery(Context& ctx) : StateBase(ctx, StateID::PostServiceDiscovery) {
+struct PaymentServiceSelection : public StateBase {
+    PaymentServiceSelection(Context& ctx) : StateBase(ctx, StateID::PaymentServiceSelection) {
     }
 
     void enter() final{
--- a/lib/everest/iso15118/include/iso15118/d2/states.hpp
+++ b/lib/everest/iso15118/include/iso15118/d2/states.hpp
@@ -24,6 +24,7 @@ enum class StateID {
     Authorization,
     ServiceDetail,
     ServiceDiscovery,
+    PostServiceDiscovery,
     WeldingDetection,
     CableCheck,
     PowerDelivery,
--- a/lib/everest/iso15118/include/iso15118/d2/states.hpp
+++ b/lib/everest/iso15118/include/iso15118/d2/states.hpp
@@ -24,7 +24,7 @@ enum class StateID {
     Authorization,
     ServiceDetail,
     ServiceDiscovery,
-    PostServiceDiscovery,
+    PaymentServiceSelection,
     WeldingDetection,
     CableCheck,
     PowerDelivery,
--- a/lib/everest/iso15118/include/iso15118/message/d2/msg_data_types.hpp
+++ b/lib/everest/iso15118/include/iso15118/message/d2/msg_data_types.hpp
@@ -31,7 +31,9 @@ using EVSEID = std::string;                                     // Length: 7-37
 using MeterReading = uint64_t;                                  // Wh
 using SigMeterReading = std::string;                            // base 64 binary, MaxLength: 64
 using MeterStatus = int16_t;
-using TMeter = int64_t; // Unix timestamp format
+using TMeter = int64_t;           // Unix timestamp format
+using ServiceName = std::string;  // MaxLength: 32
+using ServiceScope = std::string; // MaxLength: 64
 
 enum class ResponseCode {
     OK,
@@ -72,6 +74,7 @@ enum class PaymentOption {
     Contract,
     ExternalPayment
 };
+using PaymentOptionList = std::vector<PaymentOption>; // [1 - 2]
 
 enum class EvseProcessing {
     Finished,
@@ -132,6 +135,38 @@ enum class IsolationLevel {
     Fault,
     No_IMD
 };
+
+enum class ServiceCategory {
+    EvCharging,
+    Internet,
+    ContractCertificate,
+    OtherCustom
+};
+
+enum class EnergyTransferMode {
+    AC_single_phase_core,
+    AC_three_phase_core,
+    DC_core,
+    DC_extended,
+    DC_combo_core,
+    DC_unique
+};
+using SupportedEnergyTransferMode = std::vector<EnergyTransferMode>; // MaxLength: 6
+constexpr auto SupportedEnergyTransferModeMaxLength = 6;
+
+struct Service {
+    ServiceID service_id;
+    std::optional<ServiceName> service_name{std::nullopt};
+    ServiceCategory service_category;
+    std::optional<ServiceScope> service_scope{std::nullopt};
+    bool FreeService;
+};
+struct ChargeService : Service {
+    SupportedEnergyTransferMode supported_energy_transfer_mode;
+};
+
+using ServiceList = std::vector<Service>; // [1 - 8]
+constexpr auto ServiceListMaxLength = 8;
 
 struct MeterInfo {
     MeterID meter_id;
--- a/lib/everest/iso15118/include/iso15118/message/d2/msg_data_types.hpp
+++ b/lib/everest/iso15118/include/iso15118/message/d2/msg_data_types.hpp
@@ -144,12 +144,12 @@ enum class ServiceCategory {
 };
 
 enum class EnergyTransferMode {
-    AC_single_phase_core,
-    AC_three_phase_core,
-    DC_core,
-    DC_extended,
-    DC_combo_core,
-    DC_unique
+    AcSinglePhaseCore,
+    AcThreePhaseCore,
+    DcCore,
+    DcExtended,
+    DcComboCore,
+    DcUnique
 };
 using SupportedEnergyTransferMode = std::vector<EnergyTransferMode>; // MaxLength: 6
 constexpr auto SupportedEnergyTransferModeMaxLength = 6;
@@ -161,12 +161,6 @@ struct Service {
     std::optional<ServiceScope> service_scope{std::nullopt};
     bool FreeService;
 };
-struct ChargeService : Service {
-    SupportedEnergyTransferMode supported_energy_transfer_mode;
-};
-
-using ServiceList = std::vector<Service>; // [1 - 8]
-constexpr auto ServiceListMaxLength = 8;
 
 struct MeterInfo {
     MeterID meter_id;
--- a/lib/everest/iso15118/src/iso15118/d2/state/service_discovery.cpp
+++ b/lib/everest/iso15118/src/iso15118/d2/state/service_discovery.cpp
@@ -7,6 +7,7 @@
 #include <iso15118/detail/d2/context_helper.hpp>
 #include <iso15118/detail/helper.hpp>
 #include <optional>
+#include <vector>
 
 namespace iso15118::d2::state {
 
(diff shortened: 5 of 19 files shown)


Dalachowsky:
Added unit tests for the ServiceDiscovery.
The next state is called PostServiceDiscovery because it either waits for ServiceDetail or ServicePaymentSelection. I guess it could just be named ServicePaymentSelection so that it adheres to #1841.

lib/everest/iso15118/src/iso15118/d2/state/service_discovery.cpp
```diff
@@ -16,6 +61,34 @@ Result ServiceDiscovery::feed(Event ev) {
...
+
+        m_ctx.respond(res);
+
+        if (res.response_code >= dt::ResponseCode::FAILED) {
+            // m_ctx.log("Error processing ServiceDiscoveryReq");
+            return {};
+        }
+
+        // TODO(kd): I wonder how to call the next state
+        // It waits either for ServiceDetailReq or ServicePaymentSelectionReq
```
SebaLukas: I dont like the name PostServiceDiscovery.
For me it is okay to go to ServicePaymentSelection state from here and handle in this state ServiceDetail and ServicePaymentSelection messages.

lib/everest/iso15118/include/iso15118/d2/state/payment_service_selection.hpp
SebaLukas, line 1: See my comment above

lib/everest/iso15118/include/iso15118/d2/states.hpp
```diff
@@ -24,6 +24,7 @@ enum class StateID {
     Authorization,
     ServiceDetail,
     ServiceDiscovery,
+    PostServiceDiscovery,
```
SebaLukas: Remove this line please

lib/everest/iso15118/include/iso15118/d2/config.hpp
```diff
@@ -3,18 +3,35 @@
...
-    explicit SessionConfig(EvseSetupConfig config) : evse_id(std::move(config.evse_id)) {};
+    // TODO(kd) move the constructor to .cpp file
+    explicit SessionConfig(EvseSetupConfig config) :
+        evse_id(std::move(config.evse_id)),
+        charge_service(std::move(config.charge_service)),
+        supported_payment_options(std::move(config.supported_payment_options)),
+        offered_services(config.offered_services){};
 
     std::string evse_id;
+    dt::ChargeService charge_service;
```
SebaLukas: In the `session_config` only the available EnergyTransferModes should be saved like `supported_payment_options`. Not the whole ChargeService struct.

lib/everest/iso15118/include/iso15118/d2/config.hpp
```diff
@@ -3,18 +3,35 @@
...
+    std::vector<dt::PaymentOption> supported_payment_options;
+    dt::ServiceList offered_services;
```
SebaLukas: Same as `charge_service`. Only the service ids should be saved. The parameters of these services should be seperate struct members. Look into D20 for reference.

lib/everest/iso15118/src/iso15118/d2/state/service_discovery.cpp
```diff
@@ -1,12 +1,57 @@
...
+        }
+        if (category.has_value() && service.service_category != category.value()) {
+            continue;
+        }
+        filtered_services.push_back(service);
+    }
+    return filtered_services;
+}
+
+d2::msg::ServiceDiscoveryResponse handle_request(const d2::msg::ServiceDiscoveryRequest& req, d2::Session& session,
```
SebaLukas, line 75: This handle_request function should create charger_service based on the energy_transfer_modes and offered_service based on the service list ids. ServiceName is known for CertificateInstallation and Internet.
Custom VAS should be accessed via the feedback(). Look into d20 service_detail state for the correct feedback function

lib/everest/iso15118/include/iso15118/d2/config.hpp
Dalachowsky: So i suppose that I should hardcode `service_id` of `ChargeService` to like 1? And what about `service_name` and `free_service`? Do we want to make `service_name` blank or add some generic value like "AC_DC_Charging"?

Dalachowsky: Ok I've found the Table 105 in ISO15118-2 which specifies the `service_id` and `service_name` so nevermind that. I've never encountered this table when scrolling the -2 so that's why I might have been a bit confused.

lib/everest/iso15118/src/iso15118/d2/state/service_discovery.cpp
Dalachowsky: Are we going to use the same `Feedback` class as d20 or are we going to make another one for d2? Because I see that most of the feedback functions use d20 data types

SebaLukas: I guess we could use the same Feedback class also for d2

lib/everest/iso15118/include/iso15118/d2/state/payment_service_selection.hpp
```diff
@@ -0,0 +1,27 @@
...
+    }
+
+    void enter() final{
+        // TODO
+    };
+
+    Result feed(Event) final {
+        // TODO
+        return Result{};
+    };
```
SebaLukas, line 21: Should be changed when `PaymentServiceSelection` is added

lib/everest/iso15118/include/iso15118/message/d2/msg_data_types.hpp
```diff
@@ -133,6 +136,38 @@ enum class IsolationLevel {
...
+};
+
+enum class EnergyTransferMode {
+    AC_single_phase_core,
+    AC_three_phase_core,
+    DC_core,
+    DC_extended,
+    DC_combo_core,
+    DC_unique
+};
```
SebaLukas, line 153: Please use PascalCase

lib/everest/iso15118/include/iso15118/message/d2/msg_data_types.hpp
```diff
@@ -133,6 +136,38 @@ enum class IsolationLevel {
...
+struct Service {
+    ServiceID service_id;
+    std::optional<ServiceName> service_name{std::nullopt};
+    ServiceCategory service_category;
+    std::optional<ServiceScope> service_scope{std::nullopt};
+    bool FreeService;
+};
+struct ChargeService : Service {
+    SupportedEnergyTransferMode supported_energy_transfer_mode;
+};
```
SebaLukas: I think you can move ChargeService back to service_discovery.hpp

lib/everest/iso15118/include/iso15118/message/d2/msg_data_types.hpp
```diff
@@ -133,6 +136,38 @@ enum class IsolationLevel {
...
+
+using ServiceList = std::vector<Service>; // [1 - 8]
+constexpr auto ServiceListMaxLength = 8;
```
SebaLukas: Same with `ServiceList` and `ServiceListMaxLength`

lib/everest/iso15118/src/iso15118/session/feedback.cpp
```diff
@@ -64,6 +65,13 @@ void Feedback::ev_termination(const std::string& ev_termination_code,
     call_if_available(callbacks.ev_termination, ev_termination_code, ev_termination_explanation);
 }
 
+std::optional<d2::msg::data_types::Service> Feedback::get_service_from_id(uint16_t service_id) const {
```
SebaLukas, line 68: Please add an unit test for this new function

lib/everest/iso15118/src/iso15118/session/feedback.cpp
```diff
@@ -64,6 +65,13 @@ void Feedback::ev_termination(const std::string& ev_termination_code,
...
+    if (not callbacks.get_service_from_id) {
+        return std::nullopt;
+    }
+    return std::invoke(callbacks.get_service_from_id, service_id);
```
SebaLukas, line 72: `call_if_available` is not working here?

lib/everest/iso15118/test/iso15118/d2/fsm/session_setup.cpp
```diff
@@ -17,9 +17,12 @@ SCENARIO("ISO15118-2 session setup state transitions") {
 
     // Move to helper function?
     const auto evse_id = std::string("everest se");
-    const d2::EvseSetupConfig evse_setup{evse_id};
+    d2::EvseSetupConfig evse_setup{};
+    evse_setup.evse_id = evse_id;
```
SebaLukas: Please change that lines to:
```
const std::vector<dt::EnergyTransferMode> supported_energy_transfer_modes = {dt::EnergyTransferMode::DcExtended};
const std::vector<dt::PaymentOption> supported_payment_options = {dt::PaymentOption::ExternalPayment};
const std::vector<dt::ServiceID> offered_services{};
const d2::EvseSetupConfig evse_setup{evse_id, supported_energy_transfer_modes, supported_payment_options,offered_services};
```

lib/everest/iso15118/test/iso15118/d2/fsm/service_discovery.cpp
```diff
@@ -0,0 +1,242 @@
...
+    const auto service_vas_internet_access = dt::Service{
+        2, "VAS_Internet_access", dt::ServiceCategory::Internet, std::nullopt, true,
+    };
+    // dt::ServiceList services = {service_vas_internet_access, service_foo};
+
+    d2::EvseSetupConfig evse_setup{};
+    evse_setup.evse_id = evse_id;
+    evse_setup.offered_services = {};
+    evse_setup.supported_energy_transfer_modes = {dt::EnergyTransferMode::DC_extended};
+    evse_setup.supported_payment_options = payment_options;
```
SebaLukas: Please change these lines to:
```cpp
const std::vector<dt::EnergyTransferMode> supported_energy_transfer_modes = {dt::EnergyTransferMode::DcExtended};
const std::vector<dt::PaymentOption> supported_payment_options = {dt::PaymentOption::ExternalPayment};
const std::vector<dt::ServiceID> offered_services{};
const d2::EvseSetupConfig evse_setup{evse_id, supported_energy_transfer_modes, supported_payment_options,offered_services};
```

lib/everest/iso15118/test/iso15118/d2/fsm/service_discovery.cpp
```diff
@@ -0,0 +1,242 @@
...
+
+using namespace iso15118;
+
+SCENARIO("ISO15118-2 service discovery state transitions") {
+
+    namespace dt = d2::msg::data_types;
+
+    // Move to helper function?
+    const auto evse_id = std::string("everest se");
+    const auto charge_service = dt::ChargeService{
```
SebaLukas: Change the name to `expected_charge_service`

lib/everest/iso15118/test/iso15118/d2/fsm/service_discovery.cpp
```diff
@@ -0,0 +1,242 @@
...
+            std::nullopt,
+            true,
+        },
+        {dt::EnergyTransferMode::DC_extended},
+    };
+    const auto payment_options = dt::PaymentOptionList{dt::PaymentOption::ExternalPayment};
+    const auto service_vas_internet_access = dt::Service{
+        2, "VAS_Internet_access", dt::ServiceCategory::Internet, std::nullopt, true,
+    };
+    // dt::ServiceList services = {service_vas_internet_access, service_foo};
```
SebaLukas: Remove these lines or add a test

lib/everest/iso15118/test/iso15118/d2/fsm/service_discovery.cpp
```diff
@@ -0,0 +1,242 @@
...
+        dt::Service{
+            1,
+            "AC_DC_Charging",
+            dt::ServiceCategory::EvCharging,
+            std::nullopt,
+            true,
+        },
+        {dt::EnergyTransferMode::DC_extended},
+    };
+    const auto payment_options = dt::PaymentOptionList{dt::PaymentOption::ExternalPayment};
```
SebaLukas: Rename this to `expected_payment_options` if you are using this variable in the REQUIRE()

lib/everest/iso15118/test/iso15118/d2/fsm/service_discovery.cpp
```diff
@@ -0,0 +1,242 @@
...
+        }
+    };
+
+    GIVEN("Good case - respond properly") {
+        auto state_helper = FsmStateHelper(d2::SessionConfig(evse_setup), feedback_callbacks);
+        auto ctx = state_helper.get_context();
+
+        fsm::v2::FSM<d2::StateBase> fsm{ctx.create_state<d2::state::ServiceDiscovery>()};
+
+        const auto header_req = d2::msg::Header{{0, 0, 0, 0, 0, 0, 0, 0}, std::nullopt};
```
SebaLukas: `SessionId` should not be zero

lib/everest/iso15118/test/iso15118/d2/fsm/service_discovery.cpp
```diff
@@ -0,0 +1,242 @@
...
+
+            const auto& session_setup_res = response_message.value();
+            REQUIRE(session_setup_res.response_code == dt::ResponseCode::OK);
+            REQUIRE(session_setup_res.charge_service.service_id == charge_service.service_id);
+            REQUIRE(session_setup_res.charge_service.service_name == charge_service.service_name);
+            REQUIRE(session_setup_res.charge_service.service_category == charge_service.service_category);
+            REQUIRE(session_setup_res.charge_service.service_scope == charge_service.service_scope);
+            REQUIRE(session_setup_res.charge_service.FreeService == charge_service.FreeService);
+            REQUIRE(session_setup_res.charge_service.supported_energy_transfer_mode ==
+                    charge_service.supported_energy_transfer_mode);
```
SebaLukas: Maybe it makes sense to overload `== operator` for `ChargeService`

lib/everest/iso15118/test/iso15118/d2/fsm/service_discovery.cpp
```diff
@@ -0,0 +1,242 @@
...
+        if (id == 10) {
+            return dt::Service{10, "Custom", dt::ServiceCategory::OtherCustom, "https://example.com", true};
+        } else {
+            return std::nullopt;
+        }
+    };
+
+    GIVEN("Good case - respond properly") {
+        auto state_helper = FsmStateHelper(d2::SessionConfig(evse_setup), feedback_callbacks);
+        auto ctx = state_helper.get_context();
```
SebaLukas: Please move this upper. `state_helper` and `ctx` should be created only once.

lib/everest/iso15118/test/iso15118/d2/fsm/service_discovery.cpp
```diff
@@ -0,0 +1,242 @@
...
+            REQUIRE(session_setup_res.charge_service.supported_energy_transfer_mode ==
+                    charge_service.supported_energy_transfer_mode);
+
+            REQUIRE(session_setup_res.payment_option_list == payment_options);
+            REQUIRE(session_setup_res.service_list.has_value() == false);
+        }
+    }
+
+    GIVEN("Good case - respond with offered services") {
+        evse_setup.offered_services = {2, 3, 10}; // Certificate, InternetAccess and Custom
```
SebaLukas: You can also use `ctx.session_config.offered_services = {2, 3, 10};`

SebaLukas (commented):
Some additional comments

lib/everest/iso15118/include/iso15118/d2/state/payment_service_selection.hpp
Dalachowsky: I was wondering how should we handle the following states. Other way would be to just comment out the transition from `ServiceDiscovery` to `PaymentServiceSelection` but this way we cannot introduce proper transition unit tests in the PR for `ServiceDiscovery`. So I think the best way would be to always make a skeleton of the next state(s).

lib/everest/iso15118/src/iso15118/session/feedback.cpp
Dalachowsky: `call_if_available` returns true/false and we have to return the callback result. Unless we handle it some other way.

lib/everest/iso15118/test/iso15118/d2/fsm/service_discovery.cpp
Dalachowsky: I'll fix it also in `session_setup.cpp`

Dalachowsky:
@SebaLukas  I've implemented all of the changes that you requested, except for overloading the `ChargeService` `==` operator as this would also require to overload this operator for `Service` so I've decided to leave it for now.

lib/everest/iso15118/src/iso15118/session/feedback.cpp
SebaLukas: You are right

lib/everest/iso15118/test/iso15118/d2/fsm/session_setup.cpp
```diff
@@ -29,7 +33,7 @@ SCENARIO("ISO15118-2 session setup state transitions") {
 
         fsm::v2::FSM<d2::StateBase> fsm{ctx.create_state<d2::state::SessionSetup>()};
 
-        const auto header_req = d2::msg::Header{{0, 0, 0, 0, 0, 0, 0, 0}, std::nullopt};
+        const auto header_req = d2::msg::Header{{0x02, 0xDB, 0x22, 0x07, 0x3B, 0x08, 0x4D, 0x2D}, std::nullopt};
```
SebaLukas: Thats not correct. This is a new session and the ev needs to send a `SessionID` with only zeros.
After that the Ev should send the `SessionId` from Evse. Communicated in the SessionSetupRes.
Only if the session is resumed the EV sends the old `SessionId`

lib/everest/iso15118/include/iso15118/d2/state/payment_service_selection.hpp
SebaLukas: A skeleton is totally fine and needed for the fsm tests. My comment was only a reminder that this code should be worked on in the next PRs

lib/everest/iso15118/src/iso15118/d2/state/service_discovery.cpp
```diff
@@ -16,6 +112,44 @@ Result ServiceDiscovery::feed(Event ev) {
...
+            }
+        }
+
+        const auto res = handle_request(*req, m_ctx.session, m_ctx.session_config.supported_energy_transfer_modes,
+                                        m_ctx.session_config.supported_payment_options, offered_services);
+
+        m_ctx.respond(res);
+
+        if (res.response_code >= dt::ResponseCode::FAILED) {
+            // m_ctx.log("Error processing ServiceDiscoveryReq");
```
SebaLukas, line 139: To answer your question from zulip: You totally can add `m_ctx.session_stopped = true;` if the ResponseCode is Failed.
Then the ,right now not existing iso.cpp that handles the fsm, knows to stop the session and close the tcp/tls connection

lib/everest/iso15118/src/iso15118/d2/state/service_discovery.cpp
```diff
@@ -1,12 +1,108 @@
...
+            res.service_list = offered_services;
+        }
+    }
+
+    // [V2G2-545]
+    return response_with_code(res, dt::ResponseCode::OK);
+}
+
 void ServiceDiscovery::enter() {
     // m_ctx.log.enter_state("ServiceDiscovery");
```
SebaLukas, line 107: I try to add log support also for d2 in the next weeks. Or you can also have a look into that?

SebaLukas (approved):
Looking good!
Except one small mistake in the session_setup fsm test

lib/everest/iso15118/test/iso15118/d2/fsm/session_setup.cpp
Dalachowsky: Ah, my bad. I just blindly "fixed" the Session IDs in all of the tests.

lib/everest/iso15118/src/iso15118/d2/state/service_discovery.cpp
Dalachowsky: I'd rather let you do the log support.

SebaLukas:
An unit test is failling: `220/394 Test #220: Scenario: Ser/Deserialize d2 current demand messages .............................................................................***Failed    0.00 sec`
Can you have a look?

Dalachowsky:
That's odd i've run the tests on my PC and 100% passed. Only difference is that I am only running tests from libiso15118.

SebaLukas:
Okay I retriggered the ci job. Lets see if it works. If not, then I will have a look

Dalachowsky:
I see that CI has passed. I think that we had a similar issue with the CI in the past.

PR #1930, opened 2026-03-09
Fix: Adding minor fixes so that EVerest can be built on Alpine

SebaLukas:
## Describe your changes
Adding fixes so that EVerest is building on Alpine.

EVerest is not building on Alpine. The changes fixes that.

### DIFF
--- a/modules/HardwareDrivers/PowerSupplies/Huawei_V100R023C10/fusion_charger_lib/goose-lib/libs/goose-ethernet/CMakeLists.txt
+++ b/modules/HardwareDrivers/PowerSupplies/Huawei_V100R023C10/fusion_charger_lib/goose-lib/libs/goose-ethernet/CMakeLists.txt
@@ -13,6 +13,12 @@ if (MACOSX)
   target_link_libraries(goose-ethernet PRIVATE pcap)
 endif()
 
+include(CheckIncludeFile)
+check_include_file(net/ethernet.h HAVE_NET_ETHERNET_H)
+if(HAVE_NET_ETHERNET_H AND USING_MUSL)
+    target_compile_definitions(goose-ethernet PRIVATE _NETINET_IF_ETHER_H)
+endif()
+
 if(FUSION_CHARGER_LIB_BUILD_TESTS)
   add_subdirectory(tests)
 endif()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -42,6 +42,7 @@ endif()
 
 # test whether we are building on a musl based system.
 # On these we need to increase the stack size as it is very low by default
+option(USING_MUSL "Using musl" OFF)
 find_program(LDD_EXECUTABLE ldd)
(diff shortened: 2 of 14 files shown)


modules/HardwareDrivers/PowerSupplies/Huawei_V100R023C10/fusion_charger_lib/goose-lib/libs/goose-ethernet/CMakeLists.txt
```diff
@@ -13,6 +13,12 @@ if (MACOSX)
   target_link_libraries(goose-ethernet PRIVATE pcap)
 endif()
 
+include(CheckIncludeFile)
+check_include_file(net/ethernet.h HAVE_NET_ETHERNET_H)
+if(HAVE_NET_ETHERNET_H AND EXISTS "/etc/alpine-release")
+    target_compile_definitions(goose-ethernet PRIVATE _NETINET_IF_ETHER_H)
+endif()
```
james-ctc, line 20: There may not be a better way, but ideally we should support MUSL regardless of whether it is Alpine Linux or not. I suggest a cmake option -DUSING_MUSL which would then work in all cases.

SebaLukas: Yes, that would be worth considering. So instead of checking whether alpine exists, check whether musl or glibc is active?

SebaLukas: Maybe a combination between a cmake define and an automatic detection for musl is the best?

james-ctc: this post has some other options:
https://github.com/dosemu2/dosemu2/issues/530

mlitre:
@SebaLukas there is a musl detection mechanism here: https://github.com/EVerest/everest-core/blob/main/CMakeLists.txt#L43
You could use that to update how it is done in other parts, except for that things look good to me

modules/HardwareDrivers/PowerSupplies/Huawei_V100R023C10/fusion_charger_lib/goose-lib/libs/goose-ethernet/CMakeLists.txt
SebaLukas: I added the cmake option and used the existing musl detection mechanism to set the option to `ON`

mlitre (approved):

james-ctc (approved):

Pietfried (approved):

PR #1931, opened 2026-03-09
feat: add configuration option for the IT field of LemDCBM400600

florinmihut:
## Describe your changes
feat(LemDCBM400600): add OCMF Identification Type (IT) configuration

Add a configuration option to set the OCMF Identification Type field
on the LEM DCBM 400/600 device at driver startup via PUT /v1/settings
with the ocmfId/IT payload.

The IT field is an integer (0-17) mapping to identification technologies
as defined in the LEM DCBM Communication Protocol section 4.2.10.3
(e.g. 3=ISO14443, 5=EMAID, 6=EVCCID). A value of -1 (default) skips
setting the field on the device.

Note: per device documentation, the IT field should only be set once
during the lifetime of the DCBM and not changed afterwards.

Changes:
- manifest.yaml: new integer config option `IT` (range -1..17, default -1)
- lem_dcbm_400600_controller.hpp: added IT to Conf struct, declared
  set_identification_type()
- lem_dcbm_400600_controller.cpp: implemented set_identification_type()
  sending PUT /v1/settings {"ocmfId":{"IT":<value>}}, called from init()
- powermeterImpl.cpp: pass IT config through to controller

https://pionix.atlassian.net/browse/ASPD-68


Pietfried (approved):

modules/HardwareDrivers/PowerMeters/LemDCBM400600/manifest.yaml
```diff
@@ -63,6 +63,23 @@ config:
     description: Used for a unique transaction tariff designation
     type: integer
     default: 0
+  IT:
```
Pietfried, line 66: An explanation why it helps to control this value (because it controls the value of ID) would be helpful.

modules/HardwareDrivers/PowerMeters/LemDCBM400600/main/lem_dcbm_400600_controller.cpp
```diff
@@ -12,6 +12,10 @@ void LemDCBM400600Controller::init() {
                     this->config.init_retry_wait_in_milliseconds);
     this->time_sync_helper->restart_unsafe_period();
     this->http_client->set_command_timeout(this->config.command_timeout_ms);
+
+    if (this->config.IT >= 0) {
+        this->set_identification_type(this->config.IT);
```
Pietfried: `set_identification_type` can throw. We don't want to crash here, either catch the exception and warn here or inside the function.

Pietfried (changes requested):
Given the device constraint about setting `IT`. Instead of writing on every start, could we read IT first and write only if the value is not the configured one?

modules/HardwareDrivers/PowerMeters/LemDCBM400600/manifest.yaml
florinmihut: Fair. I added a more clarifying comment.

modules/HardwareDrivers/PowerMeters/LemDCBM400600/main/lem_dcbm_400600_controller.cpp
florinmihut: This is by design. :)
The IT/ID fields are controlling OCMF and have a important role in identifing the end user which is charging the car. If setting IT in the Powermeter fails (whatever the reason), it should not allow charging the car since, potentially, this could cause not to be able to identify who charged the car. 
However, we should try multiple times (as we do for other operations) before throwing an exception. 
So, I added the call_with_retry on the operation.

Pietfried: We definitely do not want to crash here and allow normal operations even if this command fails. What we could potentially do is instead of crashing is to raise an error indicating that the powermeter configuration failed so that the EvseManager would not allow charging. 

If not, EVerest could end up in an inoperative state and would e.g. also not be reachable via OCPP anymore.
It could also very well be that no Eichrecht is relevant for this charging station, so we definitely want to be able to be operative.

florinmihut:
> Given the device constraint about setting `IT`. Instead of writing on every start, could we read IT first and write only if the value is not the configured one?

IT is written ONLY if configured to do so (<0). Otherwise it remains to the value configured before.
If I am checking its existing value, I am slowing down the init phase and the whole system will be slower.

I checked, the delay is about 10-20ms. 
I will implement the check first before setting.

modules/HardwareDrivers/PowerMeters/LemDCBM400600/manifest.yaml
```diff
@@ -63,6 +63,26 @@ config:
...
+    description: >-
+      OCMF Identification Type (set on device at startup via /settings/ocmfId/IT).
+      Values per LEM DCBM Communication Protocol 4.2.10.3:
+      0=NONE, 1=DENIED, 2=UNDEFINED, 3=ISO14443, 4=ISO15693,
+      5=EMAID, 6=EVCCID, 7=EVCOID, 8=ISO7812, 9=CARD_TXN_NR,
+      10=CENTRAL, 11=CENTRAL_1, 12=CENTRAL_2, 13=LOCAL, 14=LOCAL_1,
+      15=LOCAL_2, 16=PHONE_NUMBER, 17=KEY_CODE.
+      A value of -1 means the field is not set on the device.
+      By setting this value you can controll the content of the ID field of OCMF record.
```
Pietfried:
```suggestion
      By setting this value you can control the content of the ID field of OCMF record.  
```

modules/HardwareDrivers/PowerMeters/LemDCBM400600/manifest.yaml
```diff
@@ -63,6 +63,26 @@ config:
...
+      10=CENTRAL, 11=CENTRAL_1, 12=CENTRAL_2, 13=LOCAL, 14=LOCAL_1,
+      15=LOCAL_2, 16=PHONE_NUMBER, 17=KEY_CODE.
+      A value of -1 means the field is not set on the device.
+      By setting this value you can controll the content of the ID field of OCMF record.  
+      If this value is not set, ID is set to the transaction id.
+      Please check the device documentation for more details.
+
+      WARNING: The device documentation states: "For correct usage the IT fields shall 
+        be set once to the corresponding protocol used (or kept to NONE as per default value), 
+        and not be changed during the lifetime of the DCBM due to current software limitation."
```
Pietfried:
```suggestion
      WARNING: The device documentation states: "For correct usage the IT fields shall 
      be set once to the corresponding protocol used (or kept to NONE as per default value), 
      and not be changed during the lifetime of the DCBM due to current software limitation."
```

modules/HardwareDrivers/PowerMeters/LemDCBM400600/main/lem_dcbm_400600_controller.cpp
```diff
@@ -14,8 +14,18 @@ void LemDCBM400600Controller::init() {
     this->http_client->set_command_timeout(this->config.command_timeout_ms);
 
     if (this->config.IT >= 0) {
-        call_with_retry([this]() { this->set_identification_type(this->config.IT); },
-                        this->config.init_number_of_http_retries, this->config.init_retry_wait_in_milliseconds);
+        call_with_retry(
```
Pietfried, line 14: As discussed in https://github.com/EVerest/EVerest/pull/1931#discussion_r2930095899 I would recommend to try/catch here. The module can function without the IT field being verified or set, so aborting in init() should be avoided

florinmihut: We do not fail anymore, the call with retry, will retry few times the call (as it does with the other communication with the device) and we catch the exception in the retry loop where we either raise the communication error or we log again the error (this time as error). 
The init function of the controller.cpp is not called in the init function of the module, it is called in the ready loop.

However I think we should raise a VendorError from Powermeter in case DCBMUnexpectedResponseException is triggered by the device or we should try again to initialize the module (on the other hand if the module rejected the IT value previously will not change its mind to accept it unless of course it was an internal error).
The thing is that this is not really good documented in the device docs.

I will do some additional testing.

florinmihut: I decided to repeat the controller init (not the same as the module init) until one can set the IT correctly. Since the IT is controlling the OCMF, we can't allow transactions until it is correctly set. The documentation just says that the IT field should not be changed once it is configured but it doesn't says that the device will fail setting it. So, it we have a problem setting it, it should be just temporary (device not ready to set it for example).

In case we fail to set the IT, the rest of the system will run but transactions will not be permitted until correctly set.

Pietfried (dismissed):
lgtm, please remember to squash your commits into one.

Pietfried (approved):

hikinggrass:
@florinmihut could you squash your commits and merge please? :)

PR #1932, opened 2026-03-10
RsPaymentTerminal: Add backoff to the configure call

dorezyuk:

### DIFF
--- a/modules/HardwareDrivers/Payment/RsPaymentTerminal/src/main.rs
+++ b/modules/HardwareDrivers/Payment/RsPaymentTerminal/src/main.rs
@@ -46,7 +46,6 @@ use generated::{
     BankSessionTokenProviderClientSubscriber, Context, Module, ModulePublisher, OnReadySubscriber,
     PaymentTerminalServiceSubscriber, SessionCostClientSubscriber,
 };
-use std::cmp::min;
 use std::collections::HashMap;
 use std::hash::{Hash, Hasher};
 use std::sync::{mpsc::channel, mpsc::Sender, Arc, Mutex};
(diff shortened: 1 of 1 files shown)


modules/HardwareDrivers/Payment/RsPaymentTerminal/src/main.rs
```diff
@@ -271,17 +358,23 @@ impl PaymentTerminalModule {
 
         // Wait for the card.
         let mut read_card_loop = || -> CardInfo {
-            let mut timeout = std::time::Instant::now();
-            let mut backoff_seconds = 1;
-
+            let mut configure_backoff = backoff::Backoff::new(3600);
```
ydankner: Please document here why 3600, why 60. 
Also, nit - since the units are not clear from constructor calls, consider changing the name from `new` to, perhaps `from_seconds` or something of the sort?

modules/HardwareDrivers/Payment/RsPaymentTerminal/src/main.rs
```diff
@@ -271,17 +358,23 @@ impl PaymentTerminalModule {
...
+            let mut bank_token_backoff = backoff::Backoff::new(60);
             loop {
-                if let Err(inner) = self.feig.configure() {
-                    log::warn!("Failed to configure: {inner:?}");
-                    let inner: PTError = inner.into();
-                    publishers.payment_terminal.raise_error(inner.into());
-                    continue;
-                } else {
-                    publishers.payment_terminal.clear_all_errors();
+                if configure_backoff.is_ready() {
```
ydankner, line 365: Nit - I would find `is_elapsed` clearer, but feel free to leave as is.

modules/HardwareDrivers/Payment/RsPaymentTerminal/src/main.rs
```diff
@@ -271,17 +358,23 @@ impl PaymentTerminalModule {
...
+                    if let Err(inner) = self.feig.configure() {
+                        log::warn!("Failed to configure: {inner:?}");
+                        let inner: PTError = inner.into();
+                        publishers.payment_terminal.raise_error(inner.into());
+                        configure_backoff.record_failure();
+                        log::info!(
```
ydankner: Since we're logging after each record failure, maybe move the log into `record_failure`? up to you

modules/HardwareDrivers/Payment/RsPaymentTerminal/src/main.rs
```diff
@@ -58,6 +57,94 @@ use zvt_feig_terminal::feig::{CardInfo, Error};
...
+        }
+
+        pub fn record_failure(&mut self) {
+            self.backoff_secs = min(self.backoff_secs * 2, self.max_backoff_secs);
+            self.next_retry = Instant::now() + Duration::from_secs(self.backoff_secs);
+        }
+
+        pub fn record_success(&mut self) {
+            self.backoff_secs = 1;
+            self.next_retry = Instant::now();
```
ydankner: Nit - do we want to do `Instant::now() + Duration::from_secs(self.backoff_secs);` as well to avoid busy loop in case of success? Probably will not be an issue in our case, but could avoid future issues.

modules/HardwareDrivers/Payment/RsPaymentTerminal/src/main.rs
```diff
@@ -58,6 +57,94 @@ use zvt_feig_terminal::feig::{CardInfo, Error};
...
+        }
+
+        pub fn current_backoff_secs(&self) -> u64 {
```
ydankner: Nit - do we really need a getter here?

ydankner (approved):

modules/HardwareDrivers/Payment/RsPaymentTerminal/src/main.rs
dorezyuk: removed - good idea

hikinggrass (approved):

PR #1936, opened 2026-03-10
feat: Add temperature monitoring with warning/error thresholds to LEM DCBM 400/600

florinmihut:
Summary:
Implements configurable temperature monitoring for the LEM DCBM 400/600 powermeter module, monitoring temperatureH and temperatureL sensors against configurable warning and error thresholds with hysteresis and minimum exceedance duration.

Features:
- Temperature monitoring with independent warning and error thresholds
- Hysteresis-based clearing to prevent oscillation
- Minimum time requirement before raising events (configurable)
- Uses max(temperatureH, temperatureL) for evaluation
- Raises powermeter/VendorWarning (non-blocking) and powermeter/VendorError (stops charging) via error framework

Configuration Options (manifest.yaml):
- temperature_warning_level_C: Warning threshold in °C (default: 5000 = disabled)
- temperature_error_level_C: Error threshold in °C (default: 5000 = disabled)
- temperature_hysteresis_K: Hysteresis in Kelvin for clearing (default: 3)
- temperature_min_time_as_valid_ms: Minimum exceedance duration in ms (default: 5000)

Error Framework Integration:
- Added powermeter/VendorWarning to EvseManager ignore list (allows charging to continue)
- Updated EvseManager documentation to reflect new error types

https://pionix.atlassian.net/jira/servicedesk/projects/ASPD/queues/custom/470/ASPD-70


modules/HardwareDrivers/PowerMeters/LemDCBM400600/manifest.yaml
```diff
@@ -75,6 +75,35 @@ config:
     description: UD (OCMF/transaction fields)
     type: string
     default: ""
+  temperature_warning_level_C:
+    description: >-
+      Temperature warning threshold in degrees Celsius. If the maximum of temperatureH
```
mhei, line 80: Is this really in degree Celsius, or do you mean cent degree Celsius? The default value of 5000 seems a little bit too hot for "only degree". That said, I'd suggest to stick to "degree Celsius here", the conversion to whatever the meter expects should be done in software, not by human who configures this here to keep the error probability low.

florinmihut: I really mean degrees Celsius! The default value of 5000 is to keep the current behavior where the temperature is ignored (the user can react right now to it over EVerest API interface since the temperatures are ignored but published). If you would want the driver to react to the temperature, you would need to provide a valid temperature, obviously 5k is too high. So the 5k is on purpose because I would expect that if I put some default values, those will not fit for all cases and chargers.

Now sure that I understand what do you mean by conversion ... you mean if should be an integer instead of float? Or if a powermeter is having the temperature in other format like F or K, the conversion should be done in software? But this is what the change is doing. Sorry, can you please add more info I am not sure I can follow ...

florinmihut: Ohhh, now I got it! You thought that 5k is not real C degrees because the temp is ridiculously high. It is on purpose high to force sombody to put a valid temperature if they want / need the feature. Added "Default of 5000 effectively disables this".

mhei: Ah... sorry, I was assuming, that this is a number which is put/forwarded into some kind of register in the meter. But now I got it - the comment regarding conversion is then non-sense. Thanks for the explanation and the change - this should make it more clear. Thanks.

modules/HardwareDrivers/PowerMeters/LemDCBM400600/main/temperature_monitor.hpp
```diff
@@ -0,0 +1,99 @@
...
+
+    bool error_active_{false};
+    std::optional<std::chrono::steady_clock::time_point> error_exceeded_since_;
+
+    double last_max_temperature_{0.0};
+
+    void evaluate_level(bool& active, std::optional<std::chrono::steady_clock::time_point>& exceeded_since,
+                        double threshold, double temperature, std::chrono::steady_clock::time_point now,
+                        bool& raised_event, bool& cleared_event) {
+        last_max_temperature_ = temperature;
```
Pietfried: This could alternatively be set once in `update()` and then it also doesn't need to be passed to `evaluate_level()` which is excuted twice in a row

Pietfried (dismissed):
LGTM! See one nit in line.

Some addtional thoughts: 
Raising `MREC3HighTemperature` here would be more specific, but it would also cause some issues or additional handling in the consuming EvseManager module to differentiate between the warning and the error. In the future we may want to add more semantics to the `severity`.

I also noticed that we have the same errors defined multiple times now, e.g. VendorWarning, VendorError, CommunicationFault, MREC5OverVoltage... We may want to clean this up at some time, potentially introducing breaking changes though.

modules/HardwareDrivers/PowerMeters/LemDCBM400600/main/temperature_monitor.hpp
florinmihut: True, simplified!

Pietfried (dismissed):

Pietfried (approved):

PR #1938, opened 2026-03-10
Refactor(API): Move EVerestAPI modules into a common sub-folder

cburandt:
## Describe your changes
There are four API module categories implemented in EVerest:
- the "API" module
- the "EvAPI" module
- the "RpcApi" module
- around 20 modules that belong to the EVerestAPI

This PR rearranges the modules/API subfolders to collect all modules which belong to the EVerestAPI in one subfolder.
This should make it more obvious which modules belong together.

- moved all EVerestAPI modules into common sub-folder
- changed links in their handwritten docs to point an additional level up
- changed the names of the auto-generated toc-files for the module documentation (index -> autogenerated)
- as a result simplify the code generating these files (previously had to switch between different GLOB expressions)


mhei (dismissed):
In [modules/API/EVerestAPI/CMakeLists.txt](https://github.com/EVerest/everest-core/pull/1938/changes#diff-e67f696418e06eeea75b5195765ac4f779e6ef493c58c27d05282557e381b168) a trailing newline is missing. Other than that, really appreciated 👍

modules/API/CMakeLists.txt
```diff
@@ -1,23 +1,4 @@
...
-ev_add_module(isolation_monitor_API)
-ev_add_module(ocpp_consumer_API)
-ev_add_module(over_voltage_monitor_API)
-ev_add_module(power_supply_DC_API)
-ev_add_module(powermeter_API)
-ev_add_module(session_cost_consumer_API)
-ev_add_module(session_cost_API)
-ev_add_module(slac_API)
-ev_add_module(system_API)
+add_subdirectory(EVerestAPI)
```
cburandt, line 4: Looking at the documentation artifact: Having ev_add_module and add_subdirectory on the same level does not work when building the reference/modules index of the documentation...

cburandt:
double-checked the documentation artifact produced by the CI and found it to be good now.

mlitre (approved):

hikinggrass (approved):
👍 please squash your commits before merging

mlitre:
@Pietfried not sure if we should backport this though? It doesn't change anything but folder structure changes so that can be problematic development wise

PR #1939, opened 2026-03-10
feat(evse_manager_consumer_API): Add HlcSessionFailedReason

Pietfried:
## Describe your changes

Introduces HlcSessionFailedReasonEnum and HlcSessionFailedEvent to report ISO 15118-2 HLC session failures independently of the EvseManager state machine:
- EvseV2G maps the last V2G message at disconnect to a reason and publishes it via the ISO15118_charger interface
- IsoMux forwards it from the active protocol
- EvseManager attaches the session UUID and re-publishes
on the evse_manager interface.

Adds smoke tests covering ProtocolNegotiationFailed, AuthorizationFailed, ChargingParametersNotAccepted, and EnergyTransferSetupFailed (cable check fault). Extends plug_in_ac_iso/plug_in_dc_iso with an optional payment_type argument to support the new test scenarios.


modules/EVSE/Evse15118D20/charger/ISO15118_chargerImpl.cpp
```diff
@@ -19,6 +20,57 @@ std::mutex GEL; // Global EVerest Lock
 
 namespace dt = iso15118::message_20::datatypes;
 
+static types::evse_manager::ChargingPausedEVReasonEnum map_iso20_msg_to_ev_reason(iso15118::message_20::Type msg,
```
SebaLukas: Remove static and move this function to the nameless namespace in line 74

Pietfried:
@SebaLukas This is now ready for review. For now I dropped support in D20 for the HlcSessionFailed reason to keep the scope of this PR small. We should add a follow up for it: https://github.com/EVerest/everest-core/issues/1958

docs/source/reference/EVerest_API/evse_manager_consumer_API.yaml
cburandt, line 1: Just to be sure: HlcSessionFailedEvent is defined (+ all the conversion/serialization routines implemented), but it is never used, and the variable hlc_session_failed is intentionally not made available via the API?

Pietfried: My bad, thanks for catching this. In a previous approach the reason was incorporated as part of the SessionEvent. I'll add it to the API module

tests/core_tests/smoke_tests.py
```diff
@@ -52,19 +52,29 @@ class DcConfigAdjustmentStrategy(EverestConfigAdjustmentStrategy):
...
     """
 
-    def __init__(self, zero_power_ignore_pause: bool = False, hlc_charge_loop_without_energy_timeout_s: int = 300):
+    def __init__(self, zero_power_ignore_pause: bool = False, hlc_charge_loop_without_energy_timeout_s: int = 300, ev_d20_only = False, payment_enable_contract = True, force_payment_option = False, fail_cable_check=False):
         self.zero_power_ignore_pause = zero_power_ignore_pause
         self.hlc_charge_loop_without_energy_timeout_s = hlc_charge_loop_without_energy_timeout_s
+        self.ev_d20_only = ev_d20_only
+        self.payment_enable_contract = payment_enable_contract
+        self.force_payment_option = force_payment_option
+        self.fail_cable_check = fail_cable_check
```
cburandt:
```suggestion
        self.fail_cable_check = fail_cable_check
```

cburandt (commented):
LGTM, concerning the API part

(+ see the whitespace change below, detected by Codacy)

SebaLukas (dismissed):
Looking good!

modules/EVSE/EvseV2G/connection/connection.cpp
```diff
@@ -375,9 +413,20 @@ void connection_teardown(struct v2g_connection* conn) {
...
+    const V2gMsgTypeId last_msg = conn->last_v2g_msg_at_disconnect;
+    const bool evse_initiated_stop = conn->ctx->stop_hlc || conn->ctx->intl_emergency_shutdown;
+
     /* init charging session */
     v2g_ctx_init_charging_session(conn->ctx, true);
 
+    if (!evse_initiated_stop && (conn->d_link_action == dLinkAction::D_LINK_ACTION_TERMINATE ||
+                                 conn->d_link_action == dLinkAction::D_LINK_ACTION_ERROR)) {
+        if (const auto reason = map_v2g_msg_to_hlc_failed_reason(last_msg, conn->tls_handshake_failed)) {
+            EVLOG_info << "Publishing HLC session failed reason";
```
barsnick: Does this need a log message? If so, could it include the reason as well?

I was thinking the EvseManager would be the place to log the fail reason when it receives one.

barsnick (commented):
In EvseV2G's `din_validate_response_code()` / `iso_validate_response_code()`, the assigned response codes are evaluated. Some of the reasons for them are more detailed, but this abstraction looks okay.

modules/EVSE/EvseV2G/connection/connection.cpp
Pietfried: Good catch, I will remove it

SebaLukas (approved):

hikinggrass (approved):

PR #1940, opened 2026-03-11
feat(tests): Unify e2e test runner and enable parallel ISO 15118 execution

Pietfried:
## Describe your changes

Replace per-suite shell scripts with a single tests/run-tests.sh entry point covering all test suites (core, framework, async, ocpp). CI workflows are updated to invoke the new run-tests.sh directly.

ISO 15118 tests are grouped via xdist_group("ISO15118") to prevent conflicts when running without network isolation. When veth pairs are available (created by the new setup-network-isolation.sh), the NetworkIsolationPlugin strips the group markers and assigns each xdist worker a dedicated interface, enabling fully parallel execution. The NetworkIsolationStrategy auto-injects the assigned interface into EVerest configs for ISO 15118 modules.

pytest configuration is consolidated into tests/pytest.ini, which is always selected via --config-file so that both the unified runner and direct pytest invocations use the same rootdir and option definitions.


Pietfried:
https://github.com/EVerest/everest-ci/pull/150 will allow container to execute `ip link ... `  to create the veth pairs as part of the test orchestration.

.ci/build-kit/scripts/create_ocpp_tests_image.sh
```diff
@@ -1,5 +1,7 @@
 #!/bin/sh
 
+apt-get install -y --no-install-recommends iproute2
+
```
Pietfried:
```suggestion
```

.ci/build-kit/scripts/create_integration_image.sh
```diff
@@ -1,5 +1,7 @@
 #!/bin/sh
 
+apt-get install -y --no-install-recommends iproute2 # TODO: remove when we have it pre-installed in the base image
```
Pietfried:
```suggestion
```

tests/run-tests.sh
```diff
@@ -0,0 +1,230 @@
...
+        # Veth pairs already exist (e.g. leftover from a previous run or created by CI).
+        echo "Network isolation: detected pre-existing veth pairs."
+        ISOLATION_FLAG="--network-isolation"
+    elif [[ -x "$SETUP_SCRIPT" ]]; then
+        echo "Setting up network isolation (veth pairs)..."
+        # Try without sudo first (for CI).
+        # CTRL+C at the sudo prompt cancels it and falls back to sequential execution.
+        if "$SETUP_SCRIPT" setup "$WORKERS" 2>/dev/null; then
+            ISOLATION_FLAG="--network-isolation"
+        elif sudo "$SETUP_SCRIPT" setup "$WORKERS"; then
```
mlitre, line 93: We could probably set the -n or else it might hang here right?

applications/utils/everest-testing/src/everest/testing/core_utils/network_isolation/plugin.py
```diff
@@ -0,0 +1,196 @@
...
+        This allows ISO 15118 tests to be distributed freely across workers
+        instead of being forced into a single sequential group.
+
+        Markers can live on test functions, classes, or modules, so we strip
+        from the item and all its parent nodes.
+        """
+        if not self._active:
+            return
+
+        stripped_count = 0
```
mlitre: Do you need this, you don't do anything with it

tests/ocpp_tests/conftest.py
```diff
@@ -30,13 +30,19 @@
...
+    # (e.g. when invoked via run-tests.sh with --config-file pointing to tests/pytest.ini).
+    try:
+        parser.addoption(
+            "--everest-prefix",
+            action="store",
+            default="../../build/dist",
+            help="everest prefix path; default = '../../build/dist'",
+        )
+    except ValueError:
+        pass
```
mlitre: Maybe we should log something here or else we never know if there was an error

tests/core_tests/smoke_tests.py
```diff
@@ -550,6 +556,7 @@ async def test_iso15118_dc_session_error_before_session(
         "gcp": [Requirement("grid_connection_point", "external_limits")],
     }
 )
+@pytest.mark.xdist_group(name="ISO15118")
```
mlitre: Why is this needed here? Isn't this basic charging

tests/core_tests/smoke_tests.py
```diff
@@ -582,6 +589,7 @@ async def test_pwm_ac_session_no_energy_before_session(
...
```
mlitre: Same here

mlitre (commented):
Except for the small nits it looks good

tests/run-tests.sh
Pietfried: Its actually intentional for local developer use. The CI uses the non-sudo path that runs first. Its a bit fragile, which is why I added the comment above.

applications/utils/everest-testing/src/everest/testing/core_utils/network_isolation/plugin.py
Pietfried: Good catch, removed it

tests/ocpp_tests/conftest.py
Pietfried: added a log

tests/core_tests/smoke_tests.py
Pietfried: true, removed it

tests/core_tests/smoke_tests.py
Pietfried: true, removed it

.github/workflows/job_manifest-tests.yml
andistorm, line 1: Do we have a replacement for this test? Do we need one?

applications/utils/everest-testing/src/everest/testing/core_utils/network_isolation/strategy.py
```diff
@@ -0,0 +1,45 @@
...
+    @pytest.mark.everest_config_adaptions(NetworkIsolationStrategy("ev_test0"))
+"""
+
+from copy import deepcopy
+from typing import Dict
+
+from everest.testing.core_utils import EverestConfigAdjustmentStrategy
+
+# Module types that use the `device` config field for ISO 15118 communication
+ISO15118_MODULE_TYPES = frozenset({"EvseV2G", "Evse15118D20", "PyEvJosev"})
```
mlitre: @Pietfried I think your missing the network isolation for the IsoMux module, so if you just do the following it should fix your issue.
```suggestion
ISO15118_MODULE_TYPES = frozenset({"EvseV2G", "Evse15118D20", "IsoMux", "PyEvJosev"})
```

mlitre (commented):
The OCPP test most likely fails in the CI due to the strict timing so either we can put it in a serial run or relax the timing

applications/utils/everest-testing/src/everest/testing/core_utils/network_isolation/strategy.py
Pietfried: I thought so too, but its unfortunately not that easy. IsoMux connects to the -2 and -20 modules via `lo` on `::1` so the tests would conflict. To address this we need to make the proxy device IsoMux connects to not hardcoded but configurable. @SebaLukas fyi
@mlitre @andistorm I am putting this back to draft and will potentially add a PR for IsoMux which can then be incorporated in this PR.

Pietfried: See https://github.com/EVerest/EVerest/pull/2004

mlitre (dismissed):

mlitre:
@Pietfried some tests are still failing and there are conflicts now

Pietfried:
> @Pietfried some tests are still failing and there are conflicts now

Resolved the conflicts and avoid CP events to be published when in state F from the YetiDriver: https://github.com/EVerest/EVerest/pull/1940/changes/a04379ab6239600bd78f081a1cb00fc5b036a3c0

mlitre (approved):

andistorm (approved):
LGTM 🐧, but I currently have still some issues to run locally. It is not clear if those issues are related to this changes

Pietfried:
Manual backport PR to `stable/2026.02` https://github.com/EVerest/EVerest/pull/2068

PR #1941, opened 2026-03-11
EvseManager: fix missing type

dorezyuk:


types/evse_manager.yaml
```diff
@@ -97,6 +97,7 @@ types:
         type: array
         items:
           description: A single reason to be paused
+          type: object
```
cburandt:
```suggestion
          type: string
```

Since PauseChargingEVSEReasonEnum is of type string, this also needs to be of that type, doesn't it?

cburandt (changes requested):

types/evse_manager.yaml
dorezyuk: i understand that if its included that it must be a object - maybe I'm wrong

cburandt: True for composed object, but not here.
The referenced enum type is a string itself, not an object. So the item type must be string in order for the referenced enum to fit into it - it defines the allowed values these strings can take.
(or think about the expected result: This should definitively be an array of strings in the end. Not an array of objects that happen to contain one string each)

Also don't forget to update this file's hash in lib/everest/everest_api_types/tests/expected_types_file_hashes.csv
(cause of the CI failing atm)

cburandt (approved):

PR #1942, opened 2026-03-11
feat(EnergyNode): Enhancement to set per phase A limit when only W is set

mlitre:
We can now configure the energy node to convert the total W power to a current per phase limit. This is only useful for top level energy nodes.

### DIFF
--- a/modules/EnergyManagement/EnergyNode/manifest.yaml
+++ b/modules/EnergyManagement/EnergyNode/manifest.yaml
@@ -12,6 +12,23 @@ config:
     type: integer
     minimum: 0
     maximum: 3
+  enhance_external_schedule:
+    description: >-
+      When enabled, calculates per-phase current limits from total_power_W
+      and adds them to ac_max_current_A when processing external schedules.
+      ac_max_current_A is only enhanced in case it was not specified as part of
+      the external schedules. Uses nominal_voltage_V for calculations.
+    type: boolean
+    default: false
+  nominal_voltage_V:
+    description: >-
+      Nominal voltage in volts used for power to current calculations when
+      enhance_external_schedule is enabled.
+      This allows configuration for different regions (e.g., 120V, 230V, 400V).
+    type: number
+    minimum: 1.0
+    maximum: 1000.0
+    default: 230.0
 provides:
   energy_grid:
     description: This is the chain interface to build the energy supply tree
--- a/modules/EnergyManagement/EnergyNode/tests/CMakeLists.txt
+++ b/modules/EnergyManagement/EnergyNode/tests/CMakeLists.txt
@@ -1,0 +1,37 @@
+set(TEST_TARGET_NAME ${PROJECT_NAME}_energy_node_tests)
+# Compile the test file and the utility implementation
+add_executable(${TEST_TARGET_NAME})
+
+add_dependencies(${TEST_TARGET_NAME} ${MODULE_NAME})
+
+get_target_property(GENERATED_INCLUDE_DIR generate_cpp_files EVEREST_GENERATED_INCLUDE_DIR)
+
+set(INCLUDE_DIR
+    "${MODULE_DIR}/include"
+    "${MODULE_DIR}/tests"
+    ..
+    ${GENERATED_INCLUDE_DIR}
+    ${CMAKE_BINARY_DIR}/generated/modules/${MODULE_NAME}
+    ${CMAKE_CURRENT_SOURCE_DIR}/include
+)
+
+target_include_directories(${TEST_TARGET_NAME} PUBLIC
+    ${INCLUDE_DIR}
+    ${GENERATED_INCLUDE_DIR}
+)
+
+target_sources(${TEST_TARGET_NAME} PRIVATE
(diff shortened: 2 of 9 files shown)


modules/EnergyManagement/EnergyNode/manifest.yaml
```diff
@@ -12,6 +12,13 @@ config:
     type: integer
     minimum: 0
     maximum: 3
+  enhance_ext_schedule_with_current_phase_limits:
+    description: >-
+      When enabled, calculates per-phase current limits from total_power_W
+      and adds them to ac_max_current_A when processing external schedules.
+      Uses 230V as standard voltage for calculations.
```
Pietfried: I'd suggest to introduce another parameter for the nominal voltage that defaults to 230V. Otherwise that could lead to issues in different regions

modules/EnergyManagement/EnergyNode/tests/energy_node_tests.cpp
```diff
@@ -0,0 +1,292 @@
...
+
+    Conf config;
+
+    MockEnergyNode() {
+        config.fuse_limit_A = 32.0;
+        config.phase_count = 3;
+        config.enhance_ext_schedule_with_current_phase_limits = false;
+    }
+
+    void process_schedule_with_limits(std::vector<types::energy::ScheduleReqEntry>& schedule,
```
Pietfried: Instead of duplicating the function you could try to define it as a free function and pass the configuration to it, so you would actually test the production code.

modules/EnergyManagement/EnergyNode/tests/energy_node_tests.cpp
```diff
@@ -0,0 +1,292 @@
...
+    EXPECT_NEAR(schedule[0].limits_to_root.ac_max_current_A->value, 3000.0f / 230.0f, 0.01f);
+
+    // Entry 2: 6000W / (230V * 2) = 13.04A
+    EXPECT_TRUE(schedule[1].limits_to_root.ac_max_current_A.has_value());
+    EXPECT_NEAR(schedule[1].limits_to_root.ac_max_current_A->value, 6000.0f / (230.0f * 2.0f), 0.01f);
+
+    // Entry 3: 9000W / (230V * 3) = 13.04A
+    EXPECT_TRUE(schedule[2].limits_to_root.ac_max_current_A.has_value());
+    EXPECT_NEAR(schedule[2].limits_to_root.ac_max_current_A->value, 9000.0f / (230.0f * 3.0f), 0.01f);
+}
```
Pietfried:
```suggestion
}

```

modules/EnergyManagement/EnergyNode/tests/CMakeLists.txt
```diff
@@ -0,0 +1,35 @@
+set(TEST_TARGET_NAME ${PROJECT_NAME}_energy_node_tests)
+
+set(MODULE_DIR 
+    "${PROJECT_SOURCE_DIR}/modules/EnergyManagement/EnergyNode")
+
+# Only compile the test file, not the module sources
+add_executable(${TEST_TARGET_NAME} energy_node_tests.cpp)
+
+message("Current source dir: ${CMAKE_CURRENT_SOURCE_DIR}")
```
Pietfried:
```suggestion
```

modules/EnergyManagement/EnergyNode/manifest.yaml
Pietfried:
```suggestion
      When enabled, calculates per-phase current limits from total_power_W
      and adds them to ac_max_current_A when processing external schedules.
      ac_max_current_A is only enhanced in case it was not specified as part of 
      the external schedules. Uses 230V as standard voltage for calculations.
```

Pietfried (changes requested):

modules/EnergyManagement/EnergyNode/manifest.yaml
mlitre: Done

modules/EnergyManagement/EnergyNode/tests/CMakeLists.txt
mlitre: Done

modules/EnergyManagement/EnergyNode/tests/energy_node_tests.cpp
mlitre: Done

modules/EnergyManagement/EnergyNode/manifest.yaml
mlitre: Done

modules/EnergyManagement/EnergyNode/tests/energy_node_tests.cpp
mlitre: Done

Pietfried (dismissed):

Pietfried (approved):

Pietfried (approved):

SebaLukas (approved):
Nice one 👍

PR #1943, opened 2026-03-11
CI: add a workflow to check the types/interfaces hashes

### LINKED ISSUE #1923: CI: fail early on interface/type change hash check
### Describe the problem

When a PR changes interfaces and/or types, and misses updating the everest_API or at least its hashes, it will fail the CI checks quite late in the testcase everest-core_API_serialize_tests.

It would be great to have something that fails early, instead of running through the complete compile and test chain. Especially as this is relatively easy to script.

### EVerest Domain

Testing

### Affected EVerest Module

_No response_

### Describe your solution

_No response_

### Additional context

_No response_

barsnick:
## Describe your changes
Add and integrate a GitHub workflow which checks the interface/type checksum file for everest_api_types.
Add a bash script to execute the check and report warnings (files without registered checksum) and errors (mismatches) ,and fail the CI on error.

### Note
Uses the same runner as other lightweight jobs. We might want to switch to ubuntu-slim according to https://docs.github.com/en/actions/reference/runners/github-hosted-runners.

Fixes https://github.com/EVerest/everest-core/issues/1923

### DIFF
--- a/.github/workflows/scripts/check_api_source_file_hashes.sh
+++ b/.github/workflows/scripts/check_api_source_file_hashes.sh
@@ -1,0 +1,97 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+readonly ROOT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")/../../.." && pwd)"
+readonly TYPES_CSV="lib/everest/everest_api_types/tests/expected_types_file_hashes.csv"
+readonly INTERFACES_CSV="lib/everest/everest_api_types/tests/expected_interfaces_file_hashes.csv"
+
+declare -i failure_count=0
+
+trim() {
+    local value="$1"
+    value="${value#"${value%%[![:space:]]*}"}"
+    value="${value%"${value##*[![:space:]]}"}"
+    printf '%s' "${value}"
+}
+
+check_hashes() {
+    local label="$1"
+    local csv_path="$2"
+    local source_dir="$3"
+
+    declare -A expected_hashes=()
+    local line
+    local line_no=0
+
+    echo "Checking ${label} hashes using ${csv_path}"
+
+    while IFS= read -r line || [[ -n "${line}" ]]; do
+        line_no=$((line_no + 1))
+        line="$(trim "${line}")"
+
+        if [[ -z "${line}" || "${line}" == \#* ]]; then
+            continue
+        fi
+
+        if [[ "${line}" != *,* ]]; then
+            echo "ERROR: Invalid CSV line ${line_no} in ${csv_path}: ${line}"
+            failure_count=$((failure_count + 1))
+            continue
+        fi
+
+        local file_path="${line%%,*}"
+        local expected_hash="${line#*,}"
+        file_path="$(trim "${file_path}")"
+        expected_hash="$(trim "${expected_hash}")"
+
+        if [[ -n "${expected_hashes[${file_path}]+x}" ]]; then
+            echo "ERROR: Duplicate CSV entry for ${file_path} in ${csv_path}"
+            failure_count=$((failure_count + 1))
+            continue
+        fi
+
+        expected_hashes["${file_path}"]="${expected_hash}"
+        local absolute_file="${ROOT_DIR}/${file_path}"
+        if [[ ! -f "${absolute_file}" ]]; then
+            echo "ERROR: File listed in ${csv_path} is missing: ${file_path}"
+            echo "  expected: ${expected_hash}"
+            echo "  actual:   <missing file>"
+            failure_count=$((failure_count + 1))
+            continue
+        fi
+
+        local actual_hash
+        actual_hash="$(sha256sum "${absolute_file}" | awk '{print $1}')"
+        if [[ "${actual_hash}" != "${expected_hash}" ]]; then
+            echo "ERROR: SHA256 mismatch for ${file_path}"
+            echo "  expected: ${expected_hash}"
+            echo "  actual:   ${actual_hash}"
+            failure_count=$((failure_count + 1))
+        fi
+    done < "${ROOT_DIR}/${csv_path}"
+
+    while IFS= read -r absolute_file; do
+        local relative_file
+        relative_file="${absolute_file#${ROOT_DIR}/}"
+        if [[ -z "${expected_hashes[${relative_file}]+x}" ]]; then
+            local actual_hash
+            actual_hash="$(sha256sum "${absolute_file}" | awk '{print $1}')"
+            echo "WARNING: ${label} file is not tracked in ${csv_path}: ${relative_file}"
+            echo "  expected: <missing CSV entry>"
+            echo "  actual:   ${actual_hash}"
+        fi
+    done < <(find "${ROOT_DIR}/${source_dir}" -maxdepth 1 -type f -name '*.yaml' | sort)
+}
+
+check_hashes "type" "${TYPES_CSV}" "types"
+check_hashes "interface" "${INTERFACES_CSV}" "interfaces"
+
+if (( failure_count > 0 )); then
+    echo
+    echo "Source file hash check failed with ${failure_count} issue(s)."
+    exit 1
+fi
+
+echo
+echo "Source file hash check passed."
--- a/.github/workflows/scripts/check_api_source_file_hashes.sh
+++ b/.github/workflows/scripts/check_api_source_file_hashes.sh
@@ -71,17 +71,18 @@ check_hashes() {
         fi
     done < "${ROOT_DIR}/${csv_path}"
 
-    while IFS= read -r absolute_file; do
-        local relative_file
-        relative_file="${absolute_file#${ROOT_DIR}/}"
-        if [[ -z "${expected_hashes[${relative_file}]+x}" ]]; then
-            local actual_hash
-            actual_hash="$(sha256sum "${absolute_file}" | awk '{print $1}')"
-            echo "WARNING: ${label} file is not tracked in ${csv_path}: ${relative_file}"
-            echo "  expected: <missing CSV entry>"
-            echo "  actual:   ${actual_hash}"
-        fi
-    done < <(find "${ROOT_DIR}/${source_dir}" -maxdepth 1 -type f -name '*.yaml' | sort)
+# Check for untracked files - practical, but noisy
+#    while IFS= read -r absolute_file; do
+#        local relative_file
+#        relative_file="${absolute_file#${ROOT_DIR}/}"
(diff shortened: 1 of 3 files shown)


cburandt:
I like the idea in general.

For the other reviewers: Here one can see how it looks when it fails: https://github.com/EVerest/everest-core/actions/runs/22970405006/job/66685271604?pr=1946

However, I'm not sure if all the

"WARNING: type file is not tracked in lib/everest/everest_api_types/tests/expected_types_file_hashes.csv"

messages are helpful, but that may just be my taste. Took me a while to detect the actual error in the failing PR above.

barsnick:
Good point, that output is a bit wild.

There may be good reasons for omitting certain YAML files - mostly that they're not used by the stable API. So let me drop those warnings.

Oh, and I was hoping it might gate the other checks. (I thought the DCO check did that as well, but I may be wrong.) Would that be acceptable?

cburandt:
> Oh, and I was hoping it might gate the other checks. (I thought the DCO check did that as well, but I may be wrong.) Would that be acceptable?

As I'm not a codeowner, I cannot decide (and to be honest also lack the knowledge to evaluate the provided code) but it sounds quite reasonable to me.

.github/workflows/scripts/check_api_source_file_hashes.sh
```diff
@@ -0,0 +1,98 @@
...
+#        local relative_file
+#        relative_file="${absolute_file#${ROOT_DIR}/}"
+#        if [[ -z "${expected_hashes[${relative_file}]+x}" ]]; then
+#            local actual_hash
+#            actual_hash="$(sha256sum "${absolute_file}" | awk '{print $1}')"
+#            echo "WARNING: ${label} file is not tracked in ${csv_path}: ${relative_file}"
+#            echo "  expected: <missing CSV entry>"
+#            echo "  actual:   ${actual_hash}"
+#        fi
+#    done < <(find "${ROOT_DIR}/${source_dir}" -maxdepth 1 -type f -name '*.yaml' | sort)
```
andistorm, line 85: > However, I'm not sure if all the
> 
> "WARNING: type file is not tracked in lib/everest/everest_api_types/tests/expected_types_file_hashes.csv"
> 
> messages are helpful, but that may just be my taste. Took me a while to detect the actual error in the failing PR above.

I like the idea of detecting not tracked files. I would propose the following options:

1) Build a summary for the output of the script. Print at first all passed file hashes, then all warning, and at the end all errors. This way the errors are good to find and we could have even more debug output for passing entries.

2) Extend the csv files, by adding an additional column "tracked" or so and marking the explicitly not tracked files with "false" and a placeholder hash. This would even more enforce thinking about not tracked files and if they should be tracked

I like option 2 a bit more since it is more explicit and enforces thinking about each file

.github/workflows/scripts/check_api_source_file_hashes.sh
andistorm, line 1: In general I am not a big fan of bash scriptes with much functionality in it. I personally would prefer using python with a proper testing framework as pytest. Which then would already improve logging for example and is much more readable for non bash enthusiasts as me :)

But to not generate more work I would be fine with this bash script added some inline comments (see review suggestions)

.github/workflows/scripts/check_api_source_file_hashes.sh
```diff
@@ -0,0 +1,98 @@
...
+
+set -euo pipefail
+
+readonly ROOT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")/../../.." && pwd)"
+readonly TYPES_CSV="lib/everest/everest_api_types/tests/expected_types_file_hashes.csv"
+readonly INTERFACES_CSV="lib/everest/everest_api_types/tests/expected_interfaces_file_hashes.csv"
+
+declare -i failure_count=0
+
+trim() {
```
andistorm, line 11:
```suggestion
# A utility function to remove leading and trailing whitespace from a string.
trim() {
```

.github/workflows/scripts/check_api_source_file_hashes.sh
```diff
@@ -0,0 +1,98 @@
...
+    local value="$1"
+    value="${value#"${value%%[![:space:]]*}"}"
+    value="${value%"${value##*[![:space:]]}"}"
+    printf '%s' "${value}"
+}
+
+check_hashes() {
```
andistorm, line 18:
```suggestion
# The core logic function that compares actual file hashes against the CSV records.
check_hashes() {
```

.github/workflows/scripts/check_api_source_file_hashes.sh
```diff
@@ -0,0 +1,98 @@
...
+    local label="$1"
+    local csv_path="$2"
+    local source_dir="$3"
```
andistorm, line 21: This arg is not used at all at the moment, so it should be removed (also in the function calls)

.github/workflows/scripts/check_api_source_file_hashes.sh
```diff
@@ -0,0 +1,98 @@
...
+
+    declare -A expected_hashes=()
+    local line
+    local line_no=0
+
+    echo "Checking ${label} hashes using ${csv_path}"
+
+    while IFS= read -r line || [[ -n "${line}" ]]; do
```
andistorm, line 29:
```suggestion
    # Read the CSV file line by line. The '|| [[ -n "${line}" ]]' ensures the last line is read even if it lacks a newline.
    while IFS= read -r line || [[ -n "${line}" ]]; do
```

.github/workflows/scripts/check_api_source_file_hashes.sh
```diff
@@ -0,0 +1,98 @@
...
+        line_no=$((line_no + 1))
+        line="$(trim "${line}")"
+
```
andistorm, line 32:
```suggestion

# Skip empty lines or lines starting with '#' (comments)
```

.github/workflows/scripts/check_api_source_file_hashes.sh
```diff
@@ -0,0 +1,98 @@
...
+        if [[ -z "${line}" || "${line}" == \#* ]]; then
+            continue
+        fi
+
```
andistorm, line 36:
```suggestion

        # Ensure the line contains a comma (basic CSV validation).
```

.github/workflows/scripts/check_api_source_file_hashes.sh
```diff
@@ -0,0 +1,98 @@
...
+        if [[ "${line}" != *,* ]]; then
+            echo "ERROR: Invalid CSV line ${line_no} in ${csv_path}: ${line}"
+            failure_count=$((failure_count + 1))
+            continue
+        fi
+
```
andistorm, line 42:
```suggestion

        # Split the line into the file path and the hash using string manipulation.
```

.github/workflows/scripts/check_api_source_file_hashes.sh
```diff
@@ -0,0 +1,98 @@
...
+        local file_path="${line%%,*}"
+        local expected_hash="${line#*,}"
```
andistorm, line 44:
```suggestion
        local file_path="${line%%,*}" # Everything before the first comma.
        local expected_hash="${line#*,}" # Everything after the first comma.
```

.github/workflows/scripts/check_api_source_file_hashes.sh
```diff
@@ -0,0 +1,98 @@
...
+        file_path="$(trim "${file_path}")"
+        expected_hash="$(trim "${expected_hash}")"
+
```
andistorm, line 47:
```suggestion

        # Check for duplicate entries within the same CSV.
```

.github/workflows/scripts/check_api_source_file_hashes.sh
```diff
@@ -0,0 +1,98 @@
...
+        if [[ -n "${expected_hashes[${file_path}]+x}" ]]; then
+            echo "ERROR: Duplicate CSV entry for ${file_path} in ${csv_path}"
+            failure_count=$((failure_count + 1))
+            continue
+        fi
+
+        expected_hashes["${file_path}"]="${expected_hash}"
+        local absolute_file="${ROOT_DIR}/${file_path}"
+        if [[ ! -f "${absolute_file}" ]]; then
```
andistorm, line 56:
```suggestion
        # Verify the file actually exists on the filesystem.
        if [[ ! -f "${absolute_file}" ]]; then
```

.github/workflows/scripts/check_api_source_file_hashes.sh
andistorm, line 85: if not used it should be removed

.github/workflows/scripts/check_api_source_file_hashes.sh
```diff
@@ -0,0 +1,98 @@
...
+#            actual_hash="$(sha256sum "${absolute_file}" | awk '{print $1}')"
+#            echo "WARNING: ${label} file is not tracked in ${csv_path}: ${relative_file}"
+#            echo "  expected: <missing CSV entry>"
+#            echo "  actual:   ${actual_hash}"
+#        fi
+#    done < <(find "${ROOT_DIR}/${source_dir}" -maxdepth 1 -type f -name '*.yaml' | sort)
+}
+
+check_hashes "type" "${TYPES_CSV}" "types"
+check_hashes "interface" "${INTERFACES_CSV}" "interfaces"
```
andistorm, line 89: remove third arg

andistorm (changes requested):
As an addition to the "I prefer python comment":
I got the point of having this as early test, but I would suggest to unify this script with the already existing tests in `lib/everest/everest_api_types/tests/manual_tests/source_file_hash_check`

1) Use bash script for both
2) Use cmake script for both (an additional configure worklfow without building but with running this test could be added)
3) Switch to pytest and use it in both, in cmake for local detection and in the worklfow for ci check

cburandt:
In #2037 I changed the csv format to include all yaml files and flag them as (un)tracked. Test is now a pytest. The routines doing csv file creation and comparison (both python) can also be used without pytest.

.github/workflows/scripts/check_api_source_file_hashes.sh
cburandt: Concerning option 2, see my PR mentioned below (#2037).

barsnick:
> The routines doing csv file creation and comparison (both python) can also be used without pytest.

I'm very much willing - and tempted - to give up this Bourne shell solution in favor of @cburandt's Python solution (as a CI job).

barsnick:
I'm dropping this attempt in favor of @cburandt's https://github.com/EVerest/EVerest/pull/2037.

PR #1945, opened 2026-03-11
Feature/api retain var update messages

cburandt:
## Describe your changes

Some API modules publish variables to the external MQTT topics. If an API client starts and subscribes to such topics only after EVerest is up an running, it does not get current values for the associated variables, until updates are published.

**This PR changes the some EVerest framework classes to allow to set the retain flag when publishing to the external world.**
All modules which provide variables via the API are adapted and now offer a config option to allow for the variable values to be retained.

### Pros & Cons of this Solution

EVerest cannot clear the retained topics when being shut down (as there is not shutdown mechanism today, but also because it might have been forced to terminate). Therefore, it does so only on startup and an API client can still see all the retained messages until then. Clients need to consider this (e.g. by watching the API heartbeats).

### Possible Alternatives

Explicit republish command (publish latest value to the usual topic):
- can lead to collisions (what if a clients request to republish a variable but some ms before/after a real update comes in?)

Explicit get command:
- provide a command to get the value for some topics, but not by republishing to the original topic, but to some freely chosen replyTo (as in every other command) -> safer than a simple republish, but requires more overhead on both sides (client and the API modules)


hikinggrass:
I think we should split this PR into 2:
1. the module adapter and framework changes that allow modules to publish retained messages, this I think we can merge directly (and it's potentially useful for other things)
2. the API changes, here I think we should have another discussion and probably go with an alternative option

hikinggrass:
This PR has been replaced by the merged https://github.com/EVerest/EVerest/pull/2457 hasn't it?

cburandt:
Replaced by #2457

PR #1956, opened 2026-03-13
feat(evsesecurity): add opt-in bypass for unhandled critical extensions

mlitre:
Adds ignore_unhandled_critical_extensions to x509_verify_certificate_chain and the EvseSecurity module config. When enabled, X509_V_ERR_UNHANDLED_CRITICAL_EXTENSION is suppressed only if every critical extension on the certificate has a well-known RFC 5280 NID. Intended for interoperability with non-compliant CAs that mark some extensions as critical.

### DIFF
--- a/lib/everest/evse_security/lib/evse_security/crypto/openssl/openssl_crypto_supplier.cpp
+++ b/lib/everest/evse_security/lib/evse_security/crypto/openssl/openssl_crypto_supplier.cpp
@@ -501,6 +501,86 @@ X509Handle_ptr OpenSSLSupplier::x509_duplicate_unique(X509Handle* handle) {
     return std::make_unique<X509HandleOpenSSL>(X509_dup(get(handle)));
 }
 
+namespace {
+/// Verification callback that bypasses X509_V_ERR_UNHANDLED_CRITICAL_EXTENSION when every
+/// critical extension on the certificate has a well-known RFC 5280 NID.  Truly unknown/custom
+/// OID critical extensions still cause verification to fail.
+int critical_extension_bypass_callback(int ok, X509_STORE_CTX* ctx) {
+    if (!ok) {
+        const int error = X509_STORE_CTX_get_error(ctx);
+        X509* cert = X509_STORE_CTX_get_current_cert(ctx);
+
+        if (error == X509_V_ERR_UNHANDLED_CRITICAL_EXTENSION && cert != nullptr) {
+            // RFC 5280 extension NIDs that OpenSSL handles during certificate path
+            // validation and are therefore safe to have marked critical.
+            static const std::array<int, 12> handled_nids = {
+                NID_basic_constraints,
+                NID_key_usage,
+                NID_ext_key_usage,
+                NID_subject_alt_name,
+                NID_name_constraints,
+                NID_certificate_policies,
+                NID_policy_constraints,
+                NID_inhibit_any_policy,
+                NID_authority_key_identifier,
+                NID_subject_key_identifier,
+                NID_crl_distribution_points,
+                NID_info_access,
+            };
+
+            bool has_unknown_critical = false;
+            std::string nids_log;
+
+            const int num_exts = X509_get_ext_count(cert);
+            for (int i = 0; i < num_exts; i++) {
+                X509_EXTENSION* ext = X509_get_ext(cert, i);
+                if (!X509_EXTENSION_get_critical(ext)) {
+                    continue;
+                }
+                const int nid = OBJ_obj2nid(X509_EXTENSION_get_object(ext));
+                bool is_known = false;
+                for (const int known_nid : handled_nids) {
+                    if (nid == known_nid) {
+                        is_known = true;
+                        const char* nid_name = OBJ_nid2sn(nid);
+                        if (nid_name != nullptr) {
+                            if (!nids_log.empty()) {
+                                nids_log += ", ";
+                            }
+                            nids_log += nid_name;
+                            nids_log += "(";
+                            nids_log += std::to_string(nid);
+                            nids_log += ")";
+                        } else {
+                            if (!nids_log.empty()) {
+                                nids_log += ", ";
+                            }
+                            nids_log += "NID_" + std::to_string(nid);
+                        }
+                        break;
+                    }
+                }
+                if (!is_known) {
+                    has_unknown_critical = true;
+                    break;
+                }
+            }
+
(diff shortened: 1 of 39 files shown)


lib/everest/evse_security/lib/evse_security/crypto/openssl/openssl_crypto_supplier.cpp
```diff
@@ -561,6 +563,88 @@ CertificateValidationResult OpenSSLSupplier::x509_verify_certificate_chain(
         X509_STORE_CTX_set_flags(store_ctx_ptr.get(), X509_V_FLAG_NO_CHECK_TIME);
     }
 
+    // Bypass X509_V_ERR_UNHANDLED_CRITICAL_EXTENSION for known non-compliant certificates.
+    // Triggered only when every critical extension on the certificate has a well-known RFC 5280
+    // NID, so truly unknown/custom OID critical extensions still cause verification to fail.
+    if (ignore_unhandled_critical_extensions) {
```
james-ctc, line 666: It isn't clear why the following approach is not suitable:
```
X509_VERIFY_PARAM *param = X509_VERIFY_PARAM_new();
X509_VERIFY_PARAM_set_flags(param, X509_V_FLAG_IGNORE_CRITICAL);
SSL_CTX_set1_param(ctx, param);
X509_VERIFY_PARAM_free(param);
```

which looks like it simplifies to:
```
X509_STORE_set_flags(store_ctx_ptr.get(), X509_V_FLAG_IGNORE_CRITICAL);
```

lib/everest/evse_security/lib/evse_security/crypto/openssl/openssl_crypto_supplier.cpp
```diff
@@ -561,6 +563,88 @@ CertificateValidationResult OpenSSLSupplier::x509_verify_certificate_chain(
...
+                    }
+
+                    if (!has_unknown_critical) {
+                        char* subject = X509_NAME_oneline(X509_get_subject_name(cert), nullptr, 0);
+
+                        EVLOG_info << "Ignoring unhandled critical extension(s) with well-known NIDs [" << nids_log
+                                   << "] on certificate: " << (subject != nullptr ? subject : "unknown");
+                        OPENSSL_free(subject);
+                        X509_STORE_CTX_set_error(ctx, X509_V_OK);
+                        return 1;
```
james-ctc: this looks to be part of a for loop - should processing finish early?

lib/everest/evse_security/lib/evse_security/crypto/openssl/openssl_crypto_supplier.cpp
```diff
@@ -561,6 +563,88 @@ CertificateValidationResult OpenSSLSupplier::x509_verify_certificate_chain(
...
+                        }
+                    }
+
+                    if (!has_unknown_critical) {
+                        char* subject = X509_NAME_oneline(X509_get_subject_name(cert), nullptr, 0);
+
+                        EVLOG_info << "Ignoring unhandled critical extension(s) with well-known NIDs [" << nids_log
+                                   << "] on certificate: " << (subject != nullptr ? subject : "unknown");
+                        OPENSSL_free(subject);
+                        X509_STORE_CTX_set_error(ctx, X509_V_OK);
```
james-ctc: OpenSSL examples don't do this. Could this override another error?

lib/everest/evse_security/lib/evse_security/crypto/openssl/openssl_crypto_supplier.cpp
```diff
@@ -561,6 +563,88 @@ CertificateValidationResult OpenSSLSupplier::x509_verify_certificate_chain(
         X509_STORE_CTX_set_flags(store_ctx_ptr.get(), X509_V_FLAG_NO_CHECK_TIME);
     }
 
+    // Bypass X509_V_ERR_UNHANDLED_CRITICAL_EXTENSION for known non-compliant certificates.
+    // Triggered only when every critical extension on the certificate has a well-known RFC 5280
+    // NID, so truly unknown/custom OID critical extensions still cause verification to fail.
+    if (ignore_unhandled_critical_extensions) {
+        auto verify_callback = [](int ok, X509_STORE_CTX* ctx) -> int {
```
james-ctc: This doesn't need to be a lambda - it could be a standard function in an anonymous namespace

james-ctc (changes requested):

lib/everest/evse_security/lib/evse_security/crypto/openssl/openssl_crypto_supplier.cpp
mlitre: That is a possibility, but I intentionally did it this way to only ignore well know NIDs. In other words if there is an actually unknown critical extension we still fail.

lib/everest/evse_security/lib/evse_security/crypto/openssl/openssl_crypto_supplier.cpp
mlitre: This is just after the loop so the early return is fine here

lib/everest/evse_security/lib/evse_security/crypto/openssl/openssl_crypto_supplier.cpp
mlitre: No this only overrides the X509_V_ERR_UNHANDLED_CRITICAL_EXTENSION error and any other subsequent errors found are then processed regularly.

lib/everest/evse_security/lib/evse_security/crypto/openssl/openssl_crypto_supplier.cpp
mlitre: Changed

SebaLukas (dismissed):
On the iso (lib iso, tls and modules) side it looks good 👍

SebaLukas (approved):

hikinggrass:
What's the state of this PR @mlitre @james-ctc ?

mlitre:
> What's the state of this PR @mlitre @james-ctc ?

I'll move this to draft as I think it would be better to do this once the libiso / libtls refactor is done to avoid having to create the same config in multiple places. I also would still like to use a custom callback to at least log the "problematic" extensions. I still think the custom callback a viable solution to become a "default" since it does not actually cause issues imo.

PR #1961, opened 2026-03-16
fix(evse_security): correctly calculate total size of all cert store entries for garbage collection

mikuso:
## Describe your changes

This fixes a bug in `EvseSecurity::is_filesystem_full()` related to the calculation of the total size of all objects in the certificate store.

Previously, only the size of the final certificate in the set was being considered when comparing against the `max_fs_usage_bytes` limit, as the `total_size_bytes` accumulator was being overwritten at each iteration of the for loop.

With this fix, the accumulator accumulates as intended.


james-ctc (approved):
looks good to me. Well spotted

james-ctc (approved):

PR #1979, opened 2026-03-18
Use max power capabilities in PMaxSchedule

Pietfried:
## Describe your changes
This PR addresses the issue that inside an ISO15118-2 ChargeParameterDiscoveryRes DC session, EVerest reports the power that is available at that point in time. This can apparently cause EVs to use this as a maximum limit for the whole session. In order to improve real world compatibility, we set this value to the capabilities of the DC power supply. This is in line with the behavior in AC.

### DIFF
--- a/modules/EVSE/EvseV2G/din_server.cpp
+++ b/modules/EVSE/EvseV2G/din_server.cpp
@@ -590,11 +590,11 @@ static enum v2g_event handle_din_charge_parameter(struct v2g_connection* conn) {
 
     if ((unsigned int)1 == res->DC_EVSEChargeParameter.EVSEMaximumPowerLimit_isUsed) {
         res->SAScheduleList.SAScheduleTuple.array[0].PMaxSchedule.PMaxScheduleEntry.array[0].PMax =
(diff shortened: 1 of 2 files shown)


SebaLukas:
The PMax schedule in din_server.cpp should be changed as well

mlitre:
> The PMax schedule in din_server.cpp should be changed as well

I think we just have to be careful for the values in DIN since I know there was an issue with the size of the integer

SebaLukas:
> > The PMax schedule in din_server.cpp should be changed as well
> 
> I think we just have to be careful for the values in DIN since I know there was an issue with the size of the integer

There is already a check for that but thx for reminding :+1:

Pietfried:
> The PMax schedule in din_server.cpp should be changed as well

Added it :+1:

SebaLukas (approved):

ManuelJanocha:
@Pietfried we can test the fix on Monday at the earliest.

Pietfried:
> @Pietfried we can test the fix on Monday at the earliest.

Ok, we will include the fix in the upcoming rc4.

PR #1987, opened 2026-03-20
fix(message_queue): protect BootNotification from being dropped during queue overflow

mlitre:
When the websocket connection is delayed, transaction messages accumulate and trigger drop_messages_from_normal_message_queue(). Since BootNotification sits at the front of the normal queue (push_front'd at startup), it gets dropped first — deadlocking the queue because transaction messages with stall_until_accepted=true wait for registration that never completes.

Skip BootNotification in the drop loop using is_boot_notification_message(), and add a break guard in check_queue_sizes() to prevent infinite looping when only protected messages remain in the normal queue.

### DIFF
--- a/lib/everest/ocpp/tests/lib/ocpp/common/test_message_queue.cpp
+++ b/lib/everest/ocpp/tests/lib/ocpp/common/test_message_queue.cpp
@@ -21,7 +21,8 @@ enum class TestMessageType {
     NON_TRANSACTIONAL,
     NON_TRANSACTIONAL_RESPONSE,
     InternalError,
-    BootNotification
+    BootNotification,
+    BootNotificationResponse
 };
 
 static std::string to_string(TestMessageType m) {
@@ -42,6 +43,8 @@ static std::string to_string(TestMessageType m) {
         return "internal_error";
     case TestMessageType::BootNotification:
         return "boot_notification";
+    case TestMessageType::BootNotificationResponse:
+        return "boot_notificationResponse";
     }
     throw std::out_of_range("unknown TestMessageType");
 };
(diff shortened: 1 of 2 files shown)


lib/everest/ocpp/tests/lib/ocpp/common/test_message_queue.cpp
```diff
@@ -42,6 +43,8 @@ static std::string to_string(TestMessageType m) {
         return "internal_error";
     case TestMessageType::BootNotification:
         return "boot_notification";
+    case TestMessageType::BootNotificationResponse:
+        return "boot_notificationResponse";
```
Pietfried, line 47: Why was BootNotificationResponse added? I dont see it being used

mlitre: You are right, removed

lib/everest/ocpp/tests/lib/ocpp/common/test_message_queue.cpp
Pietfried, line 25: I think this can be removed too

Pietfried (dismissed):

lib/everest/ocpp/tests/lib/ocpp/common/test_message_queue.cpp
mlitre: They are actually required because it then tries to process the call result by appending Response to the original call and then converting to a string. Without the enum / conversion types the unit tests sigabort

Pietfried: Ah I just compiled and that looked fine. Sorry for the noise then

mlitre: No problem

Pietfried (approved):

PR #1998, opened 2026-03-24
Add comparison operations to lib util

djchhp:
## Describe your changes
Adds utils for comparison of floating point and optional floating point values to everest::lib::util

### DIFF
--- a/lib/everest/util/include/everest/util/math/comparison.hpp
+++ b/lib/everest/util/include/everest/util/math/comparison.hpp
@@ -1,0 +1,177 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2022 - 2026 Pionix GmbH and Contributors to EVerest
+
+#include <algorithm>
+#include <cmath>
+#include <limits>
+#include <optional>
+#include <type_traits>
+
+/**
+ * @file util.hpp
+ * @brief Mathematical and Optional utility functions for the EVerest framework.
+ */
+
+namespace everest::lib::util {
+
+// --- Floating Point Utilities ---
+
+/**
+ * @brief Calculates a decimal precision threshold (10^-n).
+ * * This function is used to generate an epsilon value based on a human-readable
+ * number of decimal digits. For example, a precision of 3 returns 0.001.
+ * * @tparam T The floating point type (float, double, long double).
+ * @param digits_of_precision The number of decimal places to consider.
+ * @return constexpr T The calculated threshold value.
+ */
+template <class T, std::enable_if_t<std::is_floating_point_v<T>>* = nullptr>
+constexpr T range_limit(int digits_of_precision) {
+    T result = static_cast<T>(1.0);
+    for (int i = 0; i < digits_of_precision; ++i) {
+        result /= static_cast<T>(10.0);
+    }
+    return result;
+}
+
+/**
+ * @brief Compares two floating point numbers for equality within a decimal precision.
+ * * Uses a fixed epsilon (absolute difference) approach. This is ideal for
+ * physical values (like Amps or Volts) where the scale of the numbers is
+ * relatively consistent and known.
+ * @detail std::abs is not guarantted to be constexpr in C++17, using the tenery operate instead.
+ * * @tparam Prec The number of decimal digits of precision to use for comparison.
+ * @tparam T1, T2 Floating point types of the input values.
+ * @param lhs Left hand side value.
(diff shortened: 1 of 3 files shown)


lib/everest/util/include/everest/util/math/comparison.hpp
```diff
@@ -0,0 +1,177 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2022 - 2026 Pionix GmbH and Contributors to EVerest
+
+#include <algorithm>
```
Pietfried, line 6:
```suggestion
// SPDX-License-Identifier: Apache-2.0
// Copyright 2022 - 2026 Pionix GmbH and Contributors to EVerest

#pragma once

#include <algorithm>
```

lib/everest/util/include/everest/util/math/comparison.hpp
```diff
@@ -0,0 +1,177 @@
...
+#include <cmath>
+#include <limits>
+#include <optional>
+#include <type_traits>
+
+/**
+ * @file util.hpp
```
Pietfried:
```suggestion
 * @file comparison.hpp
```

lib/everest/util/include/everest/util/math/comparison.hpp
```diff
@@ -0,0 +1,177 @@
...
+    }
+    return result;
+}
+
+/**
+ * @brief Compares two floating point numbers for equality within a decimal precision.
+ * * Uses a fixed epsilon (absolute difference) approach. This is ideal for
+ * physical values (like Amps or Volts) where the scale of the numbers is
+ * relatively consistent and known.
+ * @detail std::abs is not guarantted to be constexpr in C++17, using the tenery operate instead.
```
Pietfried:
```suggestion
 * @details std::abs is not guaranteed to be constexpr in C++17, using the tenery operate instead.
```

Pietfried (changes requested):

lib/everest/util/include/everest/util/math/comparison.hpp
```diff
@@ -0,0 +1,177 @@
...
+/**
+ * @brief Calculates a decimal precision threshold (10^-n).
+ * * This function is used to generate an epsilon value based on a human-readable
+ * number of decimal digits. For example, a precision of 3 returns 0.001.
+ * * @tparam T The floating point type (float, double, long double).
+ * @param digits_of_precision The number of decimal places to consider.
+ * @return constexpr T The calculated threshold value.
+ */
+template <class T, std::enable_if_t<std::is_floating_point_v<T>>* = nullptr>
+constexpr T range_limit(int digits_of_precision) {
```
barsnick, line 31: Maybe ensure `digits_of_precision` is non-negative by making it `unsigned int`? Same for all other functions with  `int Prec`.

barsnick (changes requested):
Very nice and welcomed!

I had just been fiddling with one of the many existing incarnations of `almost_eq()`, and wishing for this.

lib/everest/util/include/everest/util/math/comparison.hpp
barsnick: Additional nit: "ternary operator".

barsnick: Another remark: Isn't this an internal detail. not a detail of the APi. and shouldn't this therefore be an inline comment instead of doxygen?

djchhp: I guess that is correct and the more appropriate way to document it.

lib/everest/util/include/everest/util/math/comparison.hpp
djchhp: Thank you for the input. I considered that case and just gave it additional thought. Introducing unsigned parameters here can be quite annoying on the user side. You have to deal with implicit casts of int input, which easily results in millions of iterations here. Since the function is also constexpr this has the potential to crash the compiler. Sure, we could add sanity checks for the input, but the level would be arbitrary. 
I went for a different approach now. I kept the int as type, but made negative values actually legal, which results in almost_eq<-3>(247, 299) being almost equal. Tests have been added and documentation adjusted.

djchhp (commented):
Thank you for the input, all comments have been addressed.

Pietfried (approved):

barsnick (approved):

PR #2003, opened 2026-03-25
fix(v16): correct drop_transaction_data to drop every second element

### LINKED ISSUE #1990: 1.6 drop_transaction_data algorithm seems incorrect
### OCPP Version

OCPP1.6

### Describe the bug

The `drop_transaction_data` data algorithm in lib/ocpp/v16/utils.cpp should drop every second message (except the first and last) but in my testing seems to behave a bit stragely. When I tested it with 5 messages it kept message 1, 4 and 5 but I would expect it to keep 1, 3 and 5

### To Reproduce

_No response_

### Anything else?

_No response_

rishabhvaish:
## Describe your changes

Reopening https://github.com/EVerest/libocpp/pull/1178 in the monorepo as requested by @Pietfried in [this comment](https://github.com/EVerest/libocpp/pull/1178#issuecomment-4112708802).

The `drop_transaction_data` function in `v16/utils.cpp` was supposed to drop every second MeterValue while keeping the first and last (per OCPP 1.6 Appendix). The implementation iterated forward with `erase()`, but each removal shifted subsequent indices down, causing the loop to skip elements.

With 5 messages `[0, 1, 2, 3, 4]` the old code kept `[0, 3, 4]` (after two while-loop passes) instead of `[0, 2, 4]` in a single pass. The middle message was lost unnecessarily.

### Changes

- Switch to backward iteration from the highest odd index so that removals never affect the positions of elements still to be processed
- Updated the existing test expectation and added three new test cases covering 6-element, 7-element, and minimum-preservation scenarios

Fixes https://github.com/EVerest/EVerest/issues/1990

## Checklist

- [x] My changes follow the existing code style
- [x] New and existing tests pass locally with my changes

### DIFF
--- a/lib/everest/ocpp/lib/ocpp/v16/utils.cpp
+++ b/lib/everest/ocpp/lib/ocpp/v16/utils.cpp
@@ -13,7 +13,12 @@ size_t get_message_size(const ocpp::Call<StopTransactionRequest>& call) {
 void drop_transaction_data(size_t max_message_size, ocpp::Call<StopTransactionRequest>& call) {
     auto& transaction_data = call.msg.transactionData.value();
     while (get_message_size(call) > max_message_size && transaction_data.size() > 2) {
-        for (size_t i = 1; i < transaction_data.size() - 1; i = i + 2) {
+        // Drop every second message, keeping the first and last.
+        // Iterate backwards so that erase() does not shift the indices
+        // of elements we still need to remove.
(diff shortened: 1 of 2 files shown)


Pietfried:
Hi @rishabhvaish , the unit tests are failing. Can you have a look at it?

Pietfried:
Hi @rishabhvaish , lgtm! Can you please squash your commits into one?

Pietfried (approved):

PR #2004, opened 2026-03-25
feat(IsoMux): Make proxy device configurable

Pietfried:
## Describe your changes

* Add a new  config option `proxy_device` to IsoMux. When set, the proxy connects to the IPv6 link-local address of the given interface instead of the hardcoded ::1 loopback; Defaults to empty string to preserve existing behaviour.
* Required for parallel ISO 15118 test execution where multiple IsoMux instances would otherwise conflict on the shared loopback (required by https://github.com/EVerest/EVerest/pull/1940)

### DIFF
--- a/modules/EVSE/IsoMux/IsoMux.hpp
+++ b/modules/EVSE/IsoMux/IsoMux.hpp
@@ -34,6 +34,7 @@ struct Conf {
     int tls_timeout;
     int proxy_port_iso2;
     int proxy_port_iso20;
+    std::string proxy_device;
 };
 
 class IsoMux : public Everest::ModuleBase {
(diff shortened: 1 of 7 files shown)


mlitre:
This looks good to me we just need to fix the linting

mlitre (approved):

SebaLukas (approved):
Looks okay 👍
I'd prefer it if `proxy_if_name` took a `std::string` and, if necessary, a `std::optional` wrapped around the string. I find `const char*` a bit scary 😅

Pietfried:
Backport is required by https://github.com/EVerest/EVerest/pull/2068

PR #2016, opened 2026-03-27
Huawei_V100R023C10: Fix crash in case of malformed frame

florinmihut:
## Describe your changes
Fix crash in case of malformed frame


ManuelJanocha:
We have successfully tested the fix with the Huawei power unit.

Pietfried (approved):

PR #2023, opened 2026-03-31
Bazel: comply with the 10Gb github cache limit

dorezyuk:
## Describe your changes

The addition of the cross compilation runners overloaded the github cache size. Every build added ~3Gb of cache and the limit in github is 10Gb. 

This pr 
1. removes the fallback key from the cross compilation runners to the common cache
2. avoids in native builds to falling back to cross compilation cache
3. adds a ci config to more aggressively strip symbols from our artifacts (libocpp.a went from 350M to 60M)

This alone does still not suffice to actually comply with the 10Gb cache size limit and we still see cache evictions from valid caches.

The majority of the cache comes from the repo cache (~1.7Gb) - which is shared across all architectures. Splitting the cache into a shared repo cache and a dedicated disk-cache should help us to respect the 10Gb cache limit


hikinggrass (approved):

andistorm (approved):
🐧

PR #2032, opened 2026-04-01
Bazel: Update python dependencies

dorezyuk:
## Describe your changes

Our bazel dependencies got stale and diverged quite a bit from the dependencies defined in the various setup.cfg - this pr updates them to the same state

The only notable difference is that `applications/utils/everest-testing/setup.cfg` depeneds on ocpp 2.1.0 which requires python 3.11. Our current bazel setup uses python 3.10. To make it compatible I pinned the ocpp dependency to 2.0.0


hikinggrass:
> ## Describe your changes
> The only notable difference is that `applications/utils/everest-testing/setup.cfg` depeneds on ocpp 2.1.0 which requires python 3.11. Our current bazel setup uses python 3.10. To make it compatible I pinned the ocpp dependency to 2.0.0

Be aware that this will prevent you from running any of the OCPP integration tests for OCPP 2.1, since that has only been added to the ocpp dependency from 2.1.0 onwards

hikinggrass (approved):

PR #2038, opened 2026-04-02
Rust: Add `everestrs_main` proc macro

dorezyuk:
## Describe your changes

Relates to the https://github.com/EVerest/EVerest/pull/1948

In the wake of https://github.com/EVerest/EVerest/pull/1948 we discovered that the static lifetimes of the c++ ffi module and the subscriber actually result in a segfault, since the c++ fii module outlives the subscribers.

Logically the c++ ffi module and the Rust runtime object should be owned by the Rust module from the code gen. When this object goes out of scope, it should clean up the ffi module (shutdown mqtt comm) and then drop the runtime and the callbacks. 

We introduced the static objects to circumvent the problem that if users would actually drop their Rust Module, it would then kill the mqtt comms, resulting in a rust program which would run, but would not communicate anymore with EVerest 

```rs
fn main() {
    // `_` results in module being immediately dropped -> no comms to mqtt anymore
    let _ = Module::new(...);

   loop {
         std::thread::sleep(...);
   }
}
```

This pr addresses the issue in a different way - it adds the `everestrs_main` macro, which constructs the Rust module for the user and passes it as a ref to main. This way the module cannot be dropped from within the `main` function - preventing the problem above and still allowing us to shutdown everything gracefully

It furthermore adds a new convenience feature. In the Rust bindings the frist time the user code receives the publisher objects is in the `on_ready` callback. This should prevent user code from publishing to mqtt before all modules are ready. This however resulted in some complexity if users wanted to have publishers in their `main` (e.x. channel from `on_ready` to `main`). Here we make the `Module::start` blocking, and return the publishers once the `on_ready` has fired

```rs
#[everestrs::main]
fn main(module: &Module) -> Result<()> {
    // blocks until `on_ready` has been received
    let pubs = module.start(...);
    loop {
         ...
    }
}
```

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -198,6 +198,7 @@ everest_framework_crate_index.from_cargo(
         "//lib/everest/framework/everestrs:Cargo.toml",
         "//lib/everest/framework/everestrs/everestrs:Cargo.toml",
         "//lib/everest/framework/everestrs/everestrs-build:Cargo.toml",
+        "//lib/everest/framework/everestrs/everestrs-macro:Cargo.toml",
     ],
     supported_platform_triples = [
         "x86_64-unknown-linux-gnu",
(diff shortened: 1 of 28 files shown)


SirVer (approved):
That is a very nice design.

hikinggrass (approved):

PR #2040, opened 2026-04-02
Bazel: build io library and dependencies

hikinggrass:
## Describe your changes
In preparation for #1799 

Since the arm toolchains in the ci are quite old, don't build the "can" and "raw" parts of the io library for these targets

### DIFF
--- a/lib/everest/io/BUILD.bazel
+++ b/lib/everest/io/BUILD.bazel
@@ -1,0 +1,49 @@
+load("@rules_cc//cc:defs.bzl", "cc_library")
+
+cc_library(
+    name = "io",
+    srcs = glob(
+        [
+            "src/**/*.cpp",
+        ],
+        exclude = [
+            "src/can/*.cpp",
+            "src/raw/*.cpp",
+        ],
+    ) + select({
+        "@platforms//cpu:armv7": [],
+        "@platforms//cpu:aarch64": [],
+        "//conditions:default": glob([
+            "src/can/*.cpp",
+            "src/raw/*.cpp",
+        ]),
+    }),
+    hdrs = glob(
+        [
+            "include/everest/io/**/*.hpp",
+        ],
+        exclude = [
+            "include/everest/io/can/*.hpp",
+            "include/everest/io/raw/*.hpp",
+        ],
+    ) + select({
+        "@platforms//cpu:armv7": [],
+        "@platforms//cpu:aarch64": [],
+        "//conditions:default": glob([
+            "include/everest/io/can/*.hpp",
+            "include/everest/io/raw/*.hpp",
+        ]),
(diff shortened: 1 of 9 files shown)


lib/everest/io/src/can/socket_can_handler.cpp
```diff
@@ -33,9 +36,16 @@ int socket_can_handler::tx(uint32_t can_id, uint8_t len8_dlc, can_payload const&
 
     frame.can_id = can_id;
     auto const max_dlc_value = 15;
```
dorezyuk, line 38: that should be in the ifdef below I think

lib/everest/io/src/can/socket_can_handler.cpp
```diff
@@ -58,7 +68,11 @@ int socket_can_handler::rx(uint32_t& can_id, uint8_t& len8_dlc, can_payload& pay
             payload.clear();
             payload.assign(frame.data, frame.data + frame.can_dlc);
             can_id = frame.can_id;
+#ifdef CAN_MAX_RAW_DLC
             len8_dlc = frame.len8_dlc;
+#else
+            len8_dlc = 0;
```
dorezyuk: is this right?

third-party/bazel/extension.bzl
```diff
@@ -50,6 +50,14 @@ def _deps_impl(module_ctx):
         build_file = "@everest-core//third-party/bazel:BUILD.mqtt-c.bazel",
     )
 
+    maybe(
```
dorezyuk, line 54: maybe add a comment that we would need mosquitto 2.0 and not 2.1 which is currently in the regsitry

lib/everest/io/BUILD.bazel
```diff
@@ -0,0 +1,23 @@
+load("@rules_cc//cc:defs.bzl", "cc_library")
+
+cc_library(
```
dorezyuk, line 3: can you run buildifier on it - it will format it nicely

lib/everest/io/src/can/socket_can_handler.cpp
dorezyuk: if yes, then maybe a comment why we can set the len to 0 would help

dorezyuk (dismissed):

third-party/bazel/extension.bzl
hikinggrass: done

lib/everest/io/BUILD.bazel
hikinggrass: done

dorezyuk (approved):

james-ctc (approved):
looks good to me

PR #2047, opened 2026-04-07
Various performance improvements

hikinggrass:
## Describe your changes
Reduce copies in EVerest framework and in generated code

Use get_uuid from helpers in framework and ensure that the random generator isn't recreated (but still accessed in a thread safe manner) all the time

Add potentially faster date string parsing function

Cleanup EVerest API (de)serialization

Replace check_topic_matches with an optimized version

Add string_view usage instead of const std::string& in some places

Optimize some module code

Removing threads and mutexes in framework message handling


lib/everest/framework/lib/config.cpp
```diff
@@ -122,9 +132,9 @@ json get_serialized_module_config(const std::string& module_id, const ModuleConf
             serialized_mod_config["mappings"][fulfillment.module_id] = mapping;
         }
     }
-    const auto module_mapping = module_configurations.at(module_id).mapping;
+    const auto module_mapping = module_config.mapping;
     serialized_mod_config["mappings"][module_id] = module_mapping;
-    const auto telemetry_config = module_configurations.at(module_id).telemetry_config;
+    const auto telemetry_config = module_config.telemetry_config;
     if (telemetry_config.has_value()) {
```
hikinggrass, line 138: const ref

lib/everest/framework/lib/config.cpp
```diff
@@ -174,7 +184,7 @@ ParsedConfigMap parse_config_map(const json& config_map_schema,
     }
 
     std::set<std::string> unknown_config_entries;
-    const std::set<std::string> config_map_schema_keys = Config::keys(config_map_schema);
+    const std::set<std::string, std::less<>> config_map_schema_keys = Config::keys(config_map_schema);
```
hikinggrass: const ref

lib/everest/framework/lib/config.cpp
```diff
@@ -619,7 +648,7 @@ void ManagerConfig::load_and_validate_manifest(ModuleConfig& module_config) {
         }
     }
 
-    const std::set<std::string> provided_impls = Config::keys(this->manifests[module_name]["provides"]);
+    const std::set<std::string, std::less<>> provided_impls = Config::keys(this->manifests[module_name]["provides"]);
```
hikinggrass: use typedef

barsnick (commented):
Generally looks fine to me, and is very welcome. Some really good catches in there.

I'm not sure about the use of `std::string_view` though. I see how it can prohibit an implicit conversion of passing a `char *` into a function taking `const std::string&`. OTOH, the functions which now take a `std::string_view` argument later often use/convert this as a `std::string`, which implies a copy, IIUC. Is that not worse?

applications/utils/ev-dev-tools/src/ev_cli/templates/types.hpp.j2
```diff
@@ -125,27 +138,25 @@ struct {{ parsed_type.name }} {
...
-                        j["{{property.name}}"].push_back(val);
-                    }
         {% else %}
         {%- if property.enum %}
-        j["{{property.name}}"] = {{ enum_to_string(property.type) }}(k.{{ property.name }}.value());
+        j.emplace("{{property.name}}", {{ enum_to_string_view(property.type) }}(*k.{{ property.name }}));
             {%- else %}
                 {%- if property.type == 'DateTime' %}
-            j["{{property.name}}"] = k.{{property.name}}.value().to_rfc3339();
+            j.emplace("{{property.name}}", = *k.{{property.name}}.to_rfc3339());
```
Pietfried: stray `=` . Was this ever used?
```suggestion
            j.emplace("{{property.name}}", k.{{property.name}}->to_rfc3339());
```

modules/EVSE/EvseManager/EvseManager.cpp
```diff
@@ -266,7 +266,7 @@ void EvseManager::init() {
     r_bsp->subscribe_request_stop_transaction(
         [this](types::evse_manager::StopTransactionRequest r) { charger->cancel_transaction(r); });
 
-    r_bsp->subscribe_capabilities([this](types::evse_board_support::HardwareCapabilities c) {
+    r_bsp->subscribe_capabilities([this](types::evse_board_support::HardwareCapabilities const c) {
```
Pietfried:
```suggestion
    r_bsp->subscribe_capabilities([this](types::evse_board_support::HardwareCapabilities const& c) {
```

lib/everest/framework/lib/date.cpp
```diff
@@ -9,12 +12,90 @@ std::string to_rfc3339(const std::chrono::time_point<date::utc_clock>& t) {
...
+    if (input.size() < typical_len) {
+        return from_rfc3339_slow(t);
+    }
+    std::string_view y_str = input.substr(0, 4);
+    std::string_view m_str = input.substr(5, 7);
+    std::string_view d_str = input.substr(8, 10);
+    std::string_view h_str = input.substr(11, 13);
+    std::string_view M_str = input.substr(14, 16);
+    std::string_view s_str = input.substr(17, 19);
+    std::string_view ms_str = input.substr(20, 23);
```
Pietfried: The second arg of substr is `size_t len` not the end position. So this should rather be


```suggestion
    std::string_view y_str = input.substr(0, 4);
    std::string_view m_str = input.substr(5, 2);
    std::string_view d_str = input.substr(8, 2);
    std::string_view h_str = input.substr(11, 2);
    std::string_view M_str = input.substr(14, 2);
    std::string_view s_str = input.substr(17, 2);
    std::string_view ms_str = input.substr(20, 3);
```

lib/everest/framework/lib/config.cpp
```diff
@@ -484,12 +511,12 @@ const json& ConfigBase::get_settings() const {
     return this->settings;
 }
 
-const json ConfigBase::get_schemas() const {
+const json ConfigBase::get_schemas() const { // FIXME: why is this different?
```
Pietfried: different to what? is this FIXME needed?

lib/everest/framework/lib/config.cpp
```diff
@@ -484,12 +511,12 @@ const json& ConfigBase::get_settings() const {
...
     BOOST_LOG_FUNCTION();
     return this->schemas;
 }
 
-json ConfigBase::get_error_types() {
+json ConfigBase::get_error_types() { // FIXME: why is this different?
```
Pietfried: same here

lib/everest/framework/lib/config.cpp
```diff
@@ -410,56 +421,72 @@ ImplementationInfo extract_implementation_info(const std::unordered_map<std::str
...
 
     const auto info = extract_implementation_info(this->module_names, this->manifests, module_id, impl_id);
     return create_printable_identifier(info, module_id, impl_id);
 }
 
-std::string ConfigBase::get_module_name(const std::string& module_id) const {
-    return this->module_names.at(module_id);
+std::string ConfigBase::get_module_name(std::string_view module_id) const {
+    auto it = this->module_names.find(module_id);
+    return it->second; // FIXME: exception handling?
```
Pietfried:
```suggestion
    auto it = this->module_names.find(module_id);
    if (it == this->module_names.end()) {
        EVLOG_AND_THROW(EverestConfigError(fmt::format("Module id '{}' not found in config", module_id)));
    }
    return it->second;
```

Pietfried (changes requested):

hikinggrass:
> I'm not sure about the use of `std::string_view` though. I see how it can prohibit an implicit conversion of passing a `char *` into a function taking `const std::string&`. OTOH, the functions which now take a `std::string_view` argument later often use/convert this as a `std::string`, which implies a copy, IIUC. Is that not worse?

Yes that is debatable, I'll have another look at this :)

lib/everest/framework/lib/config.cpp
hikinggrass: basically the question was why a json was returned here instead of a const ref json, turns out that's just an oversight and unnecessary conversion since we can just return the Schemas type here.

lib/everest/framework/lib/config.cpp
hikinggrass: Similar fix to the other one

Pietfried (dismissed):

Pietfried (approved):

PR #2049, opened 2026-04-08
conformance: add handing of charge params in accordance with v2g2-477

marthauk:
## Describe your changes

To conform with the conformance test `TC_SECC_AC_VTB_ChargeParameterDiscovery_005` and the requirement `V2G2_477` add checking of `ChargingParameterDiscovery` parameters are within defined range.

Tested using conformance testing equipment running `TC_SECC_AC_VTB_ChargeParameterDiscovery_005`

---
This is **part 1 of 2 in a stack** made with GitButler:
- <kbd>&nbsp;2&nbsp;</kbd> #2214 
- <kbd>&nbsp;1&nbsp;</kbd> #2049 👈

### DIFF
--- a/lib/everest/cbv2g/include/cbv2g/iso_2/iso2_msgDefDatatypes.h
+++ b/lib/everest/cbv2g/include/cbv2g/iso_2/iso2_msgDefDatatypes.h
@@ -28,6 +28,7 @@ extern "C" {
 
 
 #include <stdint.h>
+#include <stdbool.h>
 
 #include "cbv2g/common/exi_basetypes.h"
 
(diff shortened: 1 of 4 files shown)


barsnick:
That requirement is quite strict. I see you defined the complete set of limits, implemented four of them. The test case only tests one of them, and only for negativity?

Many of these limits are considered outdated. ISO 15118-4 even says:
> [V2G2‐832] | Table 68 can only be tested partially Group_[V2G2‐832]

The mentioned test case TC_SECC_AC_VTB_ChargeParameterDiscovery_005 only tests EVMaxCurrent against '‐100'. That check may be okay. But we cannot charge an 800V battery system with 350 kW if we really limit the max current to 400 A, as given in Table 68.

If at all, this feature should be optional by configuration, and default disabled.

Mvh og takk for bidraget
Moritz

marthauk:
> That requirement is quite strict. I see you defined the complete set of limits, implemented four of them. The test case only tests one of them, and only for negativity?
> 
> Many of these limits are considered outdated. ISO 15118-4 even says:
> 
> > [V2G2‐832] | Table 68 can only be tested partially Group_[V2G2‐832]
> 
> The mentioned test case TC_SECC_AC_VTB_ChargeParameterDiscovery_005 only tests EVMaxCurrent against '‐100'. That check may be okay. But we cannot charge an 800V battery system with 350 kW if we really limit the max current to 400 A, as given in Table 68.
> 
> If at all, this feature should be optional by configuration, and default disabled.
> 
> Mvh og takk for bidraget Moritz

Merci & Guten Tag Moritz.

Your example of 800V battery system with 350kW is not directly relevant in this case. The test case is for AC, and the handling and checking of the ChargingParameter is specifically done within the if-case for AC. Please, if you might, enlighten me in which AC-system there is a risk of charging with >400A. The test case only test for negativity as you rightly point out, but that doesnt make it less applicable. It only means there should be something added to the test case, or make a separate test case.

I added   `iso2_PhysicalValueRange` as a way of increasing readability and also to be able to add any future checks somewhere else.

SebaLukas:
Well, you're adding a `PhysicalValueRange`, etc.. That already suggests that all `PhysicalValues` should be checked in the future, not just in this specific test case.

So I would suggest that this PR only adds the check for negative max current. A simple `if` statement is sufficient here. Then we wouldn't need `ValidatePhysicalValue`, etc., here.

In the past, we’ve had issues with several cars that didn’t adhere to the PhysicalValue min and max values. So I’d appreciate it if we could try to ensure that all cars can charge with EVerest.

marthauk:
@SebaLukas I understand your point, but I disagree that it implies that all PhysicalValues necessarily have to be checked in the future. I see for -20 they have not imposed such a strict value range for the PhysicalValue, and the reasoning for that is probably cases such as @barsnick refer to.

I understand for the DC-case that some of the values in the `iso2_PhysicalValueRange` is too narrow for real world scenarios. However, for the AC-case, I dont see this as the case. So it would make sense to use the `ValidatePhysicalValue` etc for the AC part of the -2, but might not for the DC-part.

And the underlaying thing here for EVerest as a whole is; dont we want to be having a principle of trying to conform to the given conformance tests? How else can we claim conformance to the ISO15118? In this case I dont see a real reason  to block this check. For AC, an EV that sends Parameter requests that violates the PhysicalValue thresholds should in my mind be seen as a failing/misbehaving system. If there are any real reason to adjust behaviour to let some EVs charge, there are already added so called "hacks_" config-variables, such as the `hack_skoda_enyaq`. 

When it comes to the suggestion to only add the check for negative max current; I think the four checks added in this PR better fullfills the requirement `v2g2-477` and `v2g2-832` than simply catering to the conformance test itself

SebaLukas:
Hi @marthauk 

An EVerest Project Manifesto will be released in the coming weeks, outlining EVerest’s mission and principles.
Two of these principles apply to this PR:
"Real-world interoperability beats pure standards compliance, but the choice is explicit" and "One stack to charge them all".

For me, this explicitly means for this PR that the `ValidatePhysicalValue` check, as described above, makes charging sessions with real cars impossible. Even if this is "just" about AC for you, the PR can be interpreted to mean that all `PhysicalValues` must be checked - regardless of whether it's AC or DC.

Nevertheless, the PR should add a check that, for example, negative values such as `ACMaxCurrent` result in a session abort. However, this should happen directly in the relevant code section in `iso_server.cpp` without an extra function like `ValidatePhysicalValue`.

SebaLukas:
What is the status of this PR? Do you want to work on this?

SebaLukas:
If you want to work on this again, then please reopen this PR and address our comments. For now I am closing this PR.

marthauk:
> If you want to work on this again, then please reopen this PR and address our comments. For now I am closing this PR.

Hi @SebaLukas. I was on paternal leave. I will probably reopen the PR shortly and address your comments.

marthauk:
I'm having a bit trouble finding the Project manifesto. Would appreciate if someone could link it to me :)

SebaLukas:
Welcome back :)
Unfortunately, the EVerest manifesto has not yet been published. I'll let you know as soon as it's published. However, I still assume that the two sentences above still apply.

PR #2061, opened 2026-04-09
feat(EnergyNode): Enhancement to set per phase A limit when only W is…

mlitre:
… set (#1942)

We can now configure the energy node to convert the total W power to a current per phase limit. This is only useful for top level energy nodes.

Furthermore, we've replaced the use of the mutex for the monitor.

(cherry picked from commit cb7426cb5080f3d435cf0218041e58d9a7c57741)


Pietfried:
@mlitre same here as https://github.com/EVerest/EVerest/pull/2062#issuecomment-4234468709

PR #2063, opened 2026-04-10
fix(libiso15118): A default parameter set is created in case of a failed response code

SebaLukas:
## Describe your changes
In the event of a failed response code, a default service parameter entry is now created so that libcbv2g can encode the EXI messages without errors. Additionally, all state tests have been moved to the fsm tests

In case of a sequence error, libcbv2g could not generate a service_detail_res message (-150 error).


lib/everest/iso15118/src/iso15118/d20/state/service_detail.cpp
mlitre, line 179: We could still end up with an empty list here since we clear the service list on line 92

lib/everest/iso15118/src/iso15118/message/service_detail.cpp
```diff
@@ -11,6 +11,10 @@
 namespace iso15118::message_20 {
 
 namespace datatypes {
+
+ParameterSet::ParameterSet(uint16_t _id) : id(_id), parameter({datatypes::Parameter{"", false}}) {
```
mlitre, line 15: Are we allowed to have an empty string? Is there min length on this?

SebaLukas: I looked into the schemas. There is only a `maxLength = 80` for this nameType/string defined. Not a `minLength`

lib/everest/iso15118/src/iso15118/d20/state/service_detail.cpp
SebaLukas: Thats right, thx for spotting. I will add a dummy `ParameterSet` after line 175 in the case that the ev sends a not recognizable service id

mlitre: I think you just need to check if it is still empty at the end and create the default in that case

mlitre (approved):

Issue #1678, opened 2025-12-19
libiso15118: Reduce the usage of std::vector


Dalachowsky:
### Describe the problem

Inside libiso15118 `std::vector` is extensively used to represent arrays of iso15118 messages. During conversion from libiso15118 to libcbv2g structs there is no bounds checking introduced. For me this is a bit of a red flag.
All of the arrays inside iso15118 have fixed maximum sizes, so there is no need of using `std::vector` to represent them. `std::vector` relies on dynamic allocation so it can be problematic if libiso15118 was to be used on an mcu, which I know that @razvanphp wants to do.

### EVerest Domain

ISO15118

### Affected EVerest Module

libiso15118

### Describe your solution

We could use `etl::vector` from the [etl](https://github.com/ETLCPP/etl) library. It introduces the SIZE parameter to the template and throws `etl::vector_full` if the size is exceeded. I think that this is the best solution if we consider using libiso15118 on an mcu. However this introduces another dependency for libiso15118.
Another solution would be to just introduce bounds checking in the `convert()` functions.

### Additional context

Example of this case would be the `ServiceList`
```cpp
struct Service {
    ServiceCategory service_id;
    bool free_service;
};
using ServiceList = std::vector<Service>; // max: 8
```
It is stated in the comment that the max size is 8, but nothing prevents you from exceeding this limit. Inside the `convert()` function to `iso20_` struct the conversion looks like this:
```cpp
    uint8_t index = 0;
    for (const auto& service : in.energy_transfer_service_list) {
        auto& out_service = out.EnergyTransferServiceList.Service.array[index++];
        cb_convert_enum(service.service_id, out_service.ServiceID);
        out_service.FreeService = service.free_service;
    }
    out.EnergyTransferServiceList.Service.arrayLen = in.energy_transfer_service_list.size();
```
The array inside `iso20_` struct has a fixed size of 8, so the code above can access it out of bounds.
```c
struct iso20_ServiceListType {
    // Service, ServiceType
    struct {
        struct iso20_ServiceType array[iso20_ServiceType_8_ARRAY_SIZE];
        uint16_t arrayLen;
    } Service;
};
```

PR #2064, opened 2026-04-10
Replacing std::vector with utils fixed_vector

SebaLukas:
## Describe your changes
Using fixed_vector instead of std::vector in the d20 messages.
Adding utils to libiso15118

#1678 

### DIFF
--- a/lib/everest/iso15118/cmake/exported-build.cmake
+++ b/lib/everest/iso15118/cmake/exported-build.cmake
@@ -1,3 +1,5 @@
+#TODO(SL): Adding util
+
 # download everest-core (source of libcbv2g)
 include(ExternalProject)
 ExternalProject_Add(
--- a/lib/everest/iso15118/src/iso15118/d20/state/authorization_setup.cpp
+++ b/lib/everest/iso15118/src/iso15118/d20/state/authorization_setup.cpp
@@ -21,7 +21,7 @@ message_20::AuthorizationSetupResponse handle_request(const message_20::Authoriz
 
     auto res = message_20::AuthorizationSetupResponse(); // default mandatory values [V2G20-736]
 
-    if (validate_and_setup_header(res.header, session, req.header.session_id) == false) {
+    if (not validate_and_setup_header(res.header, session, req.header.session_id)) {
         return response_with_code(res, dt::ResponseCode::FAILED_UnknownSession);
     }
 
@@ -31,7 +31,7 @@ message_20::AuthorizationSetupResponse handle_request(const message_20::Authoriz
         logf_warning("authorization_services was not set. Setting EIM as auth_mode");
         res.authorization_services = {dt::Authorization::EIM};
     } else {
-        res.authorization_services = authorization_services;
+        res.authorization_services = everest::lib::util::fixed_vector<dt::Authorization, 2>(authorization_services);
     }
 
     session.offered_services.auth_services = res.authorization_services;
--- a/lib/everest/iso15118/src/iso15118/d20/state/supported_app_protocol.cpp
+++ b/lib/everest/iso15118/src/iso15118/d20/state/supported_app_protocol.cpp
@@ -58,6 +58,7 @@ Result SupportedAppProtocol::feed(Event ev) {
 
         const auto res = handle_request(*req, m_ctx.session_config.custom_protocol);
         m_ctx.respond(res);
+
         m_ctx.ev_info.ev_supported_app_protocols = req->app_protocol;
 
         if (res.response_code == ResponseCode::Failed_NoNegotiation) {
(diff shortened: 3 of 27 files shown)


lib/everest/iso15118/cmake/exported-build.cmake
```diff
@@ -1,3 +1,5 @@
+#TODO(SL): Adding util
```
SebaLukas, line 1: @hikinggrass what do you think, should we tackle this right now or later?

lib/everest/iso15118/src/iso15118/message/schedule_exchange.cpp
```diff
@@ -198,7 +198,8 @@ template <> void convert(const struct iso20_AbsolutePriceScheduleType& in, datat
         CB2CPP_CONVERT_IF_USED(in_overstay_rules.OverstayPowerThreshold, out.overstay_rules->overstay_power_threshold);
 
         const auto& overstay_rules_in = in_overstay_rules.OverstayRule;
-        out.overstay_rules->overstay_rule.reserve(overstay_rules_in.arrayLen);
+        out.overstay_rules.emplace();
+        out.overstay_rules->overstay_rule.clear();
```
mlitre: Why are we overwriting / clearing what was already added to this?

lib/everest/iso15118/src/iso15118/d20/state/authorization_setup.cpp
```diff
@@ -31,11 +31,13 @@ message_20::AuthorizationSetupResponse handle_request(const message_20::Authoriz
...
     } else {
-        res.authorization_services = authorization_services;
+        res.authorization_services = everest::lib::util::fixed_vector<dt::Authorization, 2>(authorization_services);
     }
 
-    session.offered_services.auth_services = res.authorization_services;
-
+    for (const auto& service : res.authorization_services){
+        session.offered_services.auth_services.emplace_back(service);
+    }
```
mlitre: For sanity we should probably clear the offered services before emplacing, we could also change the type to fixed vector so we could just assign it or create a helper that can do the assignment from fixed_vector to std vector

lib/everest/iso15118/src/iso15118/d20/state/supported_app_protocol.cpp
```diff
@@ -58,7 +58,10 @@ Result SupportedAppProtocol::feed(Event ev) {
 
         const auto res = handle_request(*req, m_ctx.session_config.custom_protocol);
         m_ctx.respond(res);
-        m_ctx.ev_info.ev_supported_app_protocols = req->app_protocol;
+
+        for (const auto& app_protocol : req->app_protocol) {
+            m_ctx.ev_info.ev_supported_app_protocols.emplace_back(app_protocol);
+        }
```
mlitre: Same issue as before, we can either use a helper or change the type in ev_info

mlitre (changes requested):

lib/everest/iso15118/src/iso15118/message/schedule_exchange.cpp
SebaLukas: Because `overstay_rules` is a `std::optional`, without the emplace() we get an exception when accessing `overstay_rules` value

mlitre: Yes but just before we already did emplace with some values added lines 196-198

SebaLukas: Ahhh I missed line 196. Then I will remove emplace() here

modules/EVSE/Evse15118D20/charger/utils.cpp
```diff
@@ -245,12 +252,17 @@ void fill_v2x_charging_parameters(types::iso15118::V2XChargingParameters& out_pa
     out_params.ev_max_v2xenergy_request = convert_from_optional(ev_control_mode.max_v2x_energy);
 }
 
-std::vector<dt::ParameterSet>
+everest::lib::util::fixed_vector<dt::ParameterSet, 32>
 convert_parameter_set_list(const std::vector<types::iso15118_vas::ParameterSet>& parameter_set_list) {
-    std::vector<dt::ParameterSet> out;
-    out.reserve(parameter_set_list.size());
+    everest::lib::util::fixed_vector<dt::ParameterSet, 32> out;
+    out.clear();
```
mlitre: you don't need to initialize the fixed vector by calling clear, you can just initialize directly on the line above

SebaLukas: Because of #2063 I was overcautious, but I will remove the `clear()` :sweat_smile:

lib/everest/iso15118/cmake/exported-build.cmake
hikinggrass: I don't have a clear preference here, depends mostly on the downstream consumer(s) relying on it

SebaLukas: Then I would not tackle this right now.
Lets see if we need to add that quickly or not.

mlitre (approved):

hikinggrass (approved):

PR #2067, opened 2026-04-10
feat(ocpp): implement Q03 Central V2X control with charging schedule

mlitre:
Preserve operationMode through the composite schedule pipeline so that CentralSetpoint (and other V2X operation modes) propagate from charging profiles to the final schedule sent to the energy system.

Changes:
- Add operationMode field to IntermediatePeriod and period_entry_t
- Carry operationMode through all merge functions (tx/txDefault merge, lowest-limit merge) following the winning setpoint source
- Write operationMode to output in convert_intermediate_into_schedule
- Add V2X.05 validation: reject SetChargingProfile when setpoint lies outside the [dischargeLimit, limit] range
- Fix status document: correct Q03 title, add missing Q03.FR.03 and Q04 section, add V2X.06-V2X.10 generic rules
- Add unit tests for composite schedule operationMode preservation, V2X.05 validation, and update existing V21 test expectations

### DIFF
--- a/lib/everest/ocpp/include/ocpp/v2/profile.hpp
+++ b/lib/everest/ocpp/include/ocpp/v2/profile.hpp
@@ -54,14 +54,6 @@ struct period_entry_t {
     std::int32_t stack_level;
     ChargingRateUnitEnum charging_rate_unit;
     std::optional<float> min_charging_rate;
-
-    bool equals(const period_entry_t& other) const {
-        return (start == other.start) && (end == other.end) && (limit == other.limit) &&
-               (discharge_limit == other.discharge_limit) && (setpoint == other.setpoint) &&
-               (number_phases == other.number_phases) && (stack_level == other.stack_level) &&
-               (charging_rate_unit == other.charging_rate_unit) && (min_charging_rate == other.min_charging_rate) &&
-               (operationMode == other.operationMode);
-    }
 };
 
 /// \brief Calculate the number of seconds elapsed between \param to and \param from
@@ -69,6 +61,13 @@ std::int32_t elapsed_seconds(const ocpp::DateTime& to, const ocpp::DateTime& fro
 
 /// \brief Rounds down the \param dt to the nearest second
 ocpp::DateTime floor_seconds(const ocpp::DateTime& dt);
+
+/// \brief Map an optional OperationMode to its effective value.
+///
+/// An absent operationMode behaves as ChargingOnly. Use this helper in every
+/// comparator and dedup site so that nullopt and an explicit ChargingOnly are
+/// not treated as different.
+OperationModeEnum effective_mode(const std::optional<OperationModeEnum>& mode);
 
 /// \brief calculate the start times for the profile
 /// \param now the current date and time
--- a/lib/everest/ocpp/include/ocpp/v2/profile.hpp
+++ b/lib/everest/ocpp/include/ocpp/v2/profile.hpp
@@ -25,6 +25,7 @@ struct IntermediatePeriod {
     PeriodLimit power_setpoint;
     std::optional<std::int32_t> numberPhases;
     std::optional<std::int32_t> phaseToUse;
+    std::optional<OperationModeEnum> operationMode;
 };
 
 using IntermediateProfile = std::vector<IntermediatePeriod>;
@@ -49,6 +50,7 @@ struct period_entry_t {
     PeriodLimit setpoint;
     std::optional<std::int32_t> number_phases;
     std::optional<std::int32_t> phase_to_use;
+    std::optional<OperationModeEnum> operationMode;
     std::int32_t stack_level;
     ChargingRateUnitEnum charging_rate_unit;
     std::optional<float> min_charging_rate;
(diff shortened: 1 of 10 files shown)


lib/everest/ocpp/include/ocpp/v2/profile.hpp
TD-CompleoCS: This function does not appear to be used anywhere. Instead, comparing `IntermediatePeriod` is implemented twice in lib/everest/ocpp/lib/ocpp/v2/profile.cpp, however, without considering `operationMode`. I am not sure whether this is intentional.

lib/everest/ocpp/include/ocpp/v2/profile.hpp
```diff
@@ -57,7 +59,8 @@ struct period_entry_t {
         return (start == other.start) && (end == other.end) && (limit == other.limit) &&
                (discharge_limit == other.discharge_limit) && (setpoint == other.setpoint) &&
                (number_phases == other.number_phases) && (stack_level == other.stack_level) &&
-               (charging_rate_unit == other.charging_rate_unit) && (min_charging_rate == other.min_charging_rate);
+               (charging_rate_unit == other.charging_rate_unit) && (min_charging_rate == other.min_charging_rate) &&
+               (operationMode == other.operationMode);
```
TD-CompleoCS: When comparing the optional operationMode, wouldn't it make sense to consider `std::nullopt` the same as `OperationModeEnum::ChargingOnly` as per spec, Q02.FR.01 for example?

lib/everest/ocpp/lib/ocpp/v2/profile.cpp
TD-CompleoCS: Is it intentional to not compare `operationMode` here?

TD-CompleoCS: Same here, is it intentional to not compare `operationMode`? It may be fine for Q03, since it has to have a setpoint specified. But will it work in general, like Q10, for example? I think it would be a good idea to already consider this here.

Although not in scope of this PR, but what about `numberPhases`, why is it missing here?

lib/everest/ocpp/tests/lib/ocpp/v2/test_composite_schedule.cpp
```diff
@@ -1440,4 +1443,97 @@ TEST_F(CompositeScheduleTestFixtureV2, OfflineDuration_OfflineTooLong_InvalidAft
...
+
+    const DateTime start_time = ocpp::DateTime("2024-01-17T17:00:00");
+    const DateTime end_time = ocpp::DateTime("2024-01-17T18:00:00");
+
+    CompositeSchedule result = handler->calculate_composite_schedule(start_time, end_time, DEFAULT_EVSE_ID,
+                                                                     ChargingRateUnitEnum::W, false, false);
+
+    ASSERT_GE(result.chargingSchedulePeriod.size(), 2);
+    // ChargingOnly with no setpoint results in nullopt operationMode (nullopt defaults to ChargingOnly)
+    EXPECT_FALSE(result.chargingSchedulePeriod[0].operationMode.has_value());
```
TD-CompleoCS, line 1534: Even though it does not matter for this special case (`ChargingOnly`), I am worried by the fact that the `operationMode` is being lost here. What if you test this with `OperationModeEnum::LocalLoadBalancing` instead for `period1`? I am afraid it wouldn't make it through either, though it should. I am still wondering, whether the `merge_profiles_by_lowest_limit` function works as intended.

lib/everest/ocpp/lib/ocpp/v2/profile.cpp
```diff
@@ -654,11 +668,28 @@ IntermediateProfile merge_profiles_by_lowest_limit(const std::vector<Intermediat
...
                                    period.current_discharge_limit);
             get_set_setpoint_limit(period.power_setpoint, new_period.power_setpoint, period.power_limit,
                                    period.power_discharge_limit);
+
+            // operationMode follows the period that contributed the winning setpoint.
+            // If the new period has a setpoint and an operationMode, and the setpoint changed from
+            // the previous value, the new period's operationMode takes precedence.
+            const bool new_period_has_setpoint = !is_equal(new_period.current_setpoint.limit, NO_SETPOINT_SPECIFIED) ||
+                                                 !is_equal(new_period.power_setpoint.limit, NO_SETPOINT_SPECIFIED);
+            if (new_period_has_setpoint && new_period.operationMode.has_value()) {
```
TD-CompleoCS, line 686: I am not fully convinced by this logic here.

First, it only sets the `operationMode` in the presence of a setpoint, which may be fine for Q03, but not for other bidirectional operation modes, like `LocalLoadBalancing`, which does not come with a setpoint.

Second, the behavior here differs for `ChargingOnly`, depending on whether it is set explicitly in the `new_period` or not.

Furthermore, I am not sure yet, whether a V2X operation mode should be overwritten by `ChargingOnly`, `ExternalLimits` or `ExternalSetpoint` at all.

mlitre: I will review this as I think there might be more changes to be done, but in theory you should never have conflicting V2X operation modes as you could only get those in either TxProfile or TxDefault which get picked clearly on stack level. Where we do need to be careful is with AnyV2XMode vs ChargingOnly/ExternalLimits (this case if setpoint outside of limits then limits win) and AnyV2XMode vs ExternalSeptoint (config variable sets the winner here). In the end V2X operation modes are all setpoint based technically so all that really matters is whether or not the setpoint is outside of the default limits

mlitre: What I have implemented so far is that any V2X mode will take over the default and for now there will be a warning log if two V2X modes are used (this should only happen with setpoint priority between external and central, I am pushing back support for that for now).

TD-CompleoCS: Looks all good to me now. Thanks.

Pietfried (approved):
lgtm

PR #2073, opened 2026-04-13
Fix/ocpp/charging profiles non finite floats

Pietfried:
## Describe your changes

SetChargingProfile messages containing float values that overflow to +infinity (e.g. a max-double) are silently persisted as null by nlohmann::json. On the next startup, deserializing null into a numeric field throws an error, resulting in an EVerest crash.

Adds std::isfinite validation across all float fields in ChargingSchedule and ChargingSchedulePeriod for both v16 and v2x, rejecting affected profiles before they reach the database.

### DIFF
--- a/lib/everest/ocpp/lib/ocpp/v16/smart_charging.cpp
+++ b/lib/everest/ocpp/lib/ocpp/v16/smart_charging.cpp
@@ -3,6 +3,7 @@
 
 #include "ocpp/v16/charge_point_configuration_interface.hpp"
 #include <chrono>
+#include <cmath>
 #include <ocpp/common/constants.hpp>
 #include <ocpp/common/types.hpp>
 #include <ocpp/common/utils.hpp>
@@ -84,7 +85,16 @@ bool validate_schedule(const ChargingSchedule& schedule, const int charging_sche
         return false;
     }
 
+    if (!std::isfinite(schedule.minChargingRate.value_or(0.0))) {
+        EVLOG_warning << "INVALID SCHEDULE - Non-finite minChargingRate: " << schedule.minChargingRate.value_or(0.0);
+        return false;
+    }
+
     for (const auto& period : schedule.chargingSchedulePeriod) {
+        if (!std::isfinite(period.limit)) {
+            EVLOG_warning << "INVALID SCHEDULE - Non-finite limit: " << period.limit;
+            return false;
+        }
         if (period.numberPhases.has_value()) {
             const auto& number_phases = period.numberPhases.value();
             if (number_phases <= 0 or number_phases > DEFAULT_AND_MAX_NUMBER_PHASES) {
--- a/lib/everest/ocpp/lib/ocpp/v2/functional_blocks/smart_charging.cpp
+++ b/lib/everest/ocpp/lib/ocpp/v2/functional_blocks/smart_charging.cpp
@@ -3,6 +3,7 @@
 
 #include <ocpp/v2/functional_blocks/smart_charging.hpp>
 
+#include <cmath>
 #include <optional>
 
 #include <ocpp/common/constants.hpp>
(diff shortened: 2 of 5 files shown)


lib/everest/ocpp/lib/ocpp/v16/smart_charging.cpp
```diff
@@ -84,7 +85,16 @@ bool validate_schedule(const ChargingSchedule& schedule, const int charging_sche
         return false;
     }
 
+    if (schedule.minChargingRate.has_value() && !std::isfinite(schedule.minChargingRate.value())) {
```
james-ctc: `if` could be simplified:
```
if (!std::isfinite(schedule.minChargingRate.value_or(0.0)))
```

lib/everest/ocpp/lib/ocpp/v2/functional_blocks/smart_charging.cpp
```diff
@@ -67,6 +68,45 @@ void conform_validity_periods(ChargingProfile& profile) {
         profile.validTo = validTo;
     }
 }
+
+/// \brief Returns true if any float field in \p period is non-finite (infinity or NaN).
+/// Such values originate from float overflow (e.g. a max-double sent by a CSMS) and would be
+/// serialized as JSON null
+bool has_non_finite_float(const ChargingSchedulePeriod& period) {
+    auto non_finite = [](const std::optional<float>& v) { return v.has_value() && !std::isfinite(v.value()); };
```
james-ctc: perhaps a template would be more efficient. since this is used elsewhere a template would avoid code duplication or even use an inline function rather than a lambda:
```
inline bool non_finite(const std::optional<float>& v) {
    return v.has_value() && !std::isfinite(v.value()); 
};
```

lib/everest/ocpp/lib/ocpp/v2/functional_blocks/smart_charging.cpp
```diff
@@ -67,6 +68,45 @@ void conform_validity_periods(ChargingProfile& profile) {
...
+    if (non_finite(period.limit) || non_finite(period.limit_L2) || non_finite(period.limit_L3) ||
```
james-ctc: period.limit isn't an optional (a templated solution could address this)

lib/everest/ocpp/lib/ocpp/v2/functional_blocks/smart_charging.cpp
james-ctc:
```
return !std::isfinite(v.value_or(0.0));
```

james-ctc: template option (constexpr might not work in which case use inline):
```
template <typename T>bool non_finite(const T &value) = delete;

template <> constexpr bool non_finite<float>(const float &value) {
    return !std::isfinite(value);
}

template <> constexpr bool non_finite<std::optional<float>>(const std::optional<float> &value) {
    return !std::isfinite(value.value_or(0.0));
}

template <typename T, typename... Values> constexpr bool non_finite(const T& value, const Values... values) {
    return non_finite(value) || non_finite(values...);
}

// supports
if (non_finite(period.limit, period.limit_L2, period.limit_L3)) {
```

lib/everest/ocpp/lib/ocpp/v2/functional_blocks/smart_charging.cpp
```diff
@@ -67,6 +68,45 @@ void conform_validity_periods(ChargingProfile& profile) {
...
+        non_finite(period.dischargeLimit) || non_finite(period.dischargeLimit_L2) ||
+        non_finite(period.dischargeLimit_L3) || non_finite(period.setpoint) || non_finite(period.setpoint_L2) ||
+        non_finite(period.setpoint_L3) || non_finite(period.setpointReactive) ||
+        non_finite(period.setpointReactive_L2) || non_finite(period.setpointReactive_L3) ||
+        non_finite(period.v2xBaseline)) {
+        return true;
+    }
+    if (period.v2xFreqWattCurve.has_value()) {
+        for (const auto& point : period.v2xFreqWattCurve.value()) {
+            if (!std::isfinite(point.frequency) || !std::isfinite(point.power)) {
```
james-ctc: consider using `non_finite()` so that processing is consistent

lib/everest/ocpp/lib/ocpp/v2/functional_blocks/smart_charging.cpp
Pietfried: limit is actually optional for v2

james-ctc (approved):
LGTM

PR #2081, opened 2026-04-15
Feature/docs and docs deployment improvement

cburandt:
### Does
- Build and deploy docs for releases
- allow to trigger building+deploying and deleting the docs manually.
- Fix links to snapshot.yamls and add link to versions_index.html
- Add dynamic version selector to rendered html

### Does Not
- Deploy the docs for the stable branch.
- Viewing HTML, does not warn if an older than latest version is shown


.github/workflows/job_build-docs.yaml
```diff
@@ -122,7 +126,7 @@ jobs:
           --mount type=bind,source=$SSH_AUTH_SOCK,target=/ssh-agent \
           --env SSH_AUTH_SOCK=/ssh-agent \
           --env TRAILBOOK_everest_IS_RELEASE=OFF \
```
andistorm: Should also be set by workflow input, should be `ON` for builds triggered by tag/release

.github/workflows/job_build-docs.yaml
```diff
@@ -122,7 +126,7 @@ jobs:
...
-          --env TRAILBOOK_everest_INSTANCE_NAME=nightly \
+          --env TRAILBOOK_everest_INSTANCE_NAME="${{ inputs.instance_name }}" \
           --env TRAILBOOK_everest_OVERWRITE_EXISTING_INSTANCE=ON \
```
andistorm: Should also be set by workflow input, should be `OFF` for builds triggered by tag/release to ensure old releases are not overwritten

.github/workflows/job_setup-env.yml
```diff
@@ -61,6 +70,24 @@ jobs:
...
+          elif [[ "${GITHUB_REF}" == refs/heads/stable/* ]]; then
+            raw="${GITHUB_REF_NAME}"
+          elif [[ "${GITHUB_REF}" == refs/tags/* ]]; then
+            raw="release_${GITHUB_REF_NAME}"
+          else
+            raw="ref_${GITHUB_REF_NAME}"
+          fi
+          # Sanitize: lowercase, replace any character not in [a-z0-9_] with _
+          sanitized=$(echo "${raw}" | tr '[:upper:]' '[:lower:]' | tr -cs 'a-z0-9_' '_')
+          echo "instance_name=${sanitized}" >> $GITHUB_OUTPUT
```
andistorm: If this output is only used for doc build I would move it to `job_build-docs.yaml`. If used at multiple places I would find a better name `instance_name` is relativly generous without given doc context. Maybe `doc_instance_name`

andistorm: Maybe this logic is not needed if we restructure the triggering workflows a bit..

.github/workflows/on_workflow_dispatch.yaml
andistorm, line 1: I would split this up to `on_tag` or `on_release` and `on_workflow_dispatch`

`on_workflow_dispatch` would then not be restricted to doc build, when we want to add jobs later we can provide different configurations per inputs, for now doc only would be fine.

With this differentation we could determine instance names directly or keep it in setup-env which is also fine

cmake/trailbook-ext-everest/create-versions-json.cmake
andistorm, line 1: I did not checked the dependencies carefully yet, would like to do it in a following review

- [ ] Review cmake deps carefully

docs/source/_static/version-switcher.js
andistorm, line 1: @andistorm 
- [x] Review this in built documentation

cmake/trailbook-ext-everest/create-versions-json.cmake
andistorm, line 1: I would move this to trailbook, since multiversion feature is not everest specific and nice feature to have in standard trailbook framework. 

We could also discuss to switch to json format completly, then we should go up with the trailbook version if there is somewhere set one

andistorm (changes requested):

.github/workflows/job_setup-env.yml
cburandt: removed - the instance_names are part of the triggering workflows now

.github/workflows/on_workflow_dispatch.yaml
cburandt: done

cmake/trailbook-ext-everest/create-versions-json.cmake
cburandt: file removed - the installation happens withing add-trailbook.cmake now

cburandt: Did not switch to json for everything yet, but building the yaml and json at the same time now

.github/workflows/on_tag.yaml
```diff
@@ -0,0 +1,43 @@
...
+      SA_GITHUB_SSH_KEY: ${{ secrets.SA_GITHUB_SSH_KEY }}
+    with:
+      runner: ${{ inputs.runner || 'ubuntu-24.04' }}
+      is_fork: ${{ needs.setup-env.outputs.is_fork == 'true' }}
+      build_kit_artifact_name: ${{ needs.build-build-kit.outputs.build_kit_artifact_name }}
+      build_kit_image_tag: ${{ needs.build-build-kit.outputs.build_kit_image_tag }}
+      deploy_docs: true
+      instance_name: release_${{ github.ref_name }}
+      is_release: true
+      overwrite_existing_instance: false
```
andistorm:
```suggestion
      overwrite_existing_instance: false

```

.github/workflows/on_tag.yaml
```diff
@@ -0,0 +1,43 @@
...
+  build-docs:
+    name: Build and Deploy Documentation
+    needs:
+      - setup-env
+      - build-build-kit
+    uses: ./.github/workflows/job_build-docs.yaml
+    secrets:
+      SA_GITHUB_SSH_KEY: ${{ secrets.SA_GITHUB_SSH_KEY }}
+    with:
+      runner: ${{ inputs.runner || 'ubuntu-24.04' }}
```
andistorm:
```suggestion
      runner: ${{ env.runner }}
```
same for other jobs in this file, requires addition of: (or somethin similiar)
```
env:
    runner: ubuntu-24.04
```
As root yaml element

.github/workflows/on_workflow_dispatch.yaml
```diff
@@ -0,0 +1,47 @@
+name: Build and Deploy Documentation
+on:
+  workflow_dispatch:
+    inputs:
+      instance_name:
+        description: 'Override the documentation instance name (optional)'
+        required: false
+        type: string
+        default: ''
+
```
andistorm, line 29: I would suggest to add here the input `runner` again, so the user can optional use bigger runners

.github/workflows/on_workflow_dispatch.yaml
```diff
@@ -0,0 +1,47 @@
...
+      SA_GITHUB_SSH_KEY: ${{ secrets.SA_GITHUB_SSH_KEY }}
+    with:
+      runner: ${{ inputs.runner || 'ubuntu-24.04' }}
+      is_fork: ${{ needs.setup-env.outputs.is_fork == 'true' }}
+      build_kit_artifact_name: ${{ needs.build-build-kit.outputs.build_kit_artifact_name }}
+      build_kit_image_tag: ${{ needs.build-build-kit.outputs.build_kit_image_tag }}
+      deploy_docs: true
+      instance_name: ${{ inputs.instance_name || format('ref_{0}_{1}', github.ref_name, github.sha) }}
+      is_release: false
+      overwrite_existing_instance: true
```
andistorm:
```suggestion
      overwrite_existing_instance: false
```
Since sha is part of instance name there shouldn't be duplicates or the result would be exactly the same

cmake/trailbook/create_metadata_file.py
```diff
@@ -45,14 +56,21 @@ def main():
     )
     args = parser.parse_args()
 
+    if args.yaml_output_path is None and args.json_output_path is None:
+        parser.error("at least one of --yaml-output-path or --json-output-path is required")
+
```
andistorm: Is there any case in which only one of those both files should be created? If not I would restrict the script to not allow it

andistorm: Or document this specific usage

andistorm (changes requested):

cmake/trailbook/create_metadata_file.py
```diff
@@ -28,13 +29,23 @@ def main():
...
     )
     parser.add_argument(
-        '--output-path',
+        '--yaml-output-path',
         type=Path,
-        dest='output_path',
+        dest='yaml_output_path',
         action='store',
         required=True,
+        default=None,
```
andistorm:
```suggestion
```


default doesn't make sense combined with `required=true`

cmake/trailbook/create_metadata_file.py
```diff
@@ -45,14 +56,21 @@ def main():
     )
     args = parser.parse_args()
 
+    if args.yaml_output_path is None or args.json_output_path is None:
+        parser.error("at least one of --yaml-output-path or --json-output-path is required")
+
```
andistorm:
```suggestion

```
argparse should check this already and error text wouldn't match anymore

cmake/trailbook/create_metadata_file.py
```diff
@@ -73,8 +91,13 @@ def main():
         'versions': versions_list
     }
     # render yaml content
-    with args.output_path.open('w') as f:
-        yaml.dump(data, f, default_flow_style=False)
+    if args.yaml_output_path is not None:
+        with args.yaml_output_path.open('w') as f:
+            yaml.dump(data, f, default_flow_style=False)
```
andistorm:
```suggestion
    with args.yaml_output_path.open('w') as f:
        yaml.dump(data, f, default_flow_style=False)
```
Is already checked

cmake/trailbook/create_metadata_file.py
```diff
@@ -73,8 +91,13 @@ def main():
...
+
+    if args.json_output_path is not None:
+        with args.json_output_path.open('w') as f:
+            json.dump(data, f, indent=2)
```
andistorm:
```suggestion
    with args.json_output_path.open('w') as f:
        json.dump(data, f, indent=2)
```
Is already checked

cmake/trailbook/create_metadata_file.py
```diff
@@ -28,13 +29,23 @@ def main():
...
+        default=None,
         help='Path where the trailbook_metadata.yaml file will be created'
     )
+    parser.add_argument(
+        '--json-output-path',
+        type=Path,
+        dest='json_output_path',
+        action='store',
+        required=True,
+        default=None,
```
andistorm, line 132:
```suggestion
```

default doesn't make sense combined with `required=true`

andistorm (dismissed):

andistorm (approved):
LGTM 🐧

Issue #1851, opened 2026-02-20
Porting libiso15118 to MCU


SebaLukas:
### Describe the problem

The EVerest community asked whether libiso15118 could also be ported to microcontrollers. See zulip

### EVerest Domain

ISO15118

### Affected EVerest Module

lib/everest/iso15118
Evse15118D20

razvanphp:
As we are working on this port, we will share here our findings and refactoring needed:

1. **TLS 1.2/1.3 provider** = we need to use MbedTLS (lacks [OCSP support](https://github.com/Mbed-TLS/mbedtls/issues/880)) and/or WolfSSL (paid license, but supports TLS 1.3 and [OCSP](https://github.com/wolfSSL/wolfssl/commit/9a9eb2bf1d93ff9c8a5923cd789960e6d25d54b6)).

I see that OpenSSL was introduced in https://github.com/EVerest/libiso15118/commit/eb4bc83d073765856550f00e2b5c110a8bf72c93 and MbedTLS support was removed in https://github.com/EVerest/libiso15118/commit/9b77420b1a16ab6cf32c7526dce090bb19ba8eb0 and https://github.com/EVerest/libiso15118/commit/b22242daf381d3ccef1408283136dad81f72e73f 

We think we should refactor the io/connection_ classes to have plan + TLS, with different external providers with the same interface. Open discussion about OCSP, as it's different in TLS 1.3 and optional in 1.2 AFAIK.

2. Mutex / threading = Linux code often assumes pthread-style synchronization.

On Zephyr, we use:
k_mutex instead of pthread mutexes
k_sem for signaling
k_thread / workqueue-based execution instead of Linux threading

On Espressif IDF we might use other abstractions.

Should synchronization be abstracted behind a platform layer? Other ideas?

3. Event handling (eventfd) = Linux uses mechanisms like eventfd for wake-up / event notification.

On Zephyr, we use callbacks flags/semaphores and k_poll where needed. ESP-IDF has a [compatible implementation](https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference/storage/vfs.html) we could overload. 

4.  EMPTY state name in SDP enum = code works on Linux, but does not compile on MCU. ✅

On Zephyr, this expands to a macro, so we need to rename it for portability.

5. Multiple use of dynamic allocation & std::vector

As described in #1678, but we could switch to using `fixed_vector` implemented in https://github.com/EVerest/EVerest/pull/1795, merged in #2064 ✅

[...]

PR #2082, opened 2026-04-15
feat(libiso15118): use portable network byte-order APIs instead of endian.h (#1851)

razvanphp:
<endian.h> is not portable and is effectively Linux/glibc-oriented. This works for macOS and MCUs with TCP/IP stacks as well now.

The change is a portability swap, not a behavior change. Before, the code used Linux-style endian helpers. Those APIs are not reliably available on macOS. I replaced them with the standard socket byte-order APIs from <arpa/inet.h>:

- `htons` instead of `htobe16`
- `htonl` instead of `htobe32`
- `ntohs` instead of `be16toh`
- `ntohl` instead of `be32toh`

The protocol fields being encoded here are network-order fields, network byte order is big-endian & htons/htonl and ntohs/ntohl are specifically defined for host-to-network and network-to-host conversion.

Let's see if the tests pass now.

### DIFF
--- a/lib/everest/iso15118/README.md
+++ b/lib/everest/iso15118/README.md
@@ -37,14 +37,14 @@ ISO 15118 Support
 ISO15118 support is distributed accross multiple repositories and modules in EVerest. Please see the following references of other ISO15118 related development:
 
 - Some functionality of part 2 of ISO 15118 is integrated in the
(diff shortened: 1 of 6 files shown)


razvanphp:
Sorry, I force-pushed as I had to sign the commit.

SebaLukas (approved):

PR #2083, opened 2026-04-16
Fix(EvseV2G): Fixing buffer overflow in handle_cert_installation

SebaLukas:
## Describe your changes
When checking whether the base64 fits in the buffer, the 8 bytes from the `V2GTP_HEADER` are subtracted.
In addition, the payment option vector is now checked to see if it contains more than two elements. If so, only the first two elements are copied into the response message.

Previously, a buffer overflow could occur if, in the CertInstall state, a Base64 string was larger than `DEFAULT_BUFFER_SIZE-8`

### DIFF
--- a/modules/EVSE/EvseV2G/iso_server.cpp
+++ b/modules/EVSE/EvseV2G/iso_server.cpp
@@ -8,6 +8,7 @@
 #include <cbv2g/iso_2/iso2_msgDefEncoder.h>
 #include <everest/tls/openssl_util.hpp>
 
+#include <algorithm>
 #include <cstdint>
 #include <inttypes.h>
 #include <math.h>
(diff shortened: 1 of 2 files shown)


modules/EVSE/EvseV2G/iso_server.cpp
```diff
@@ -648,9 +648,14 @@ static enum v2g_event handle_iso_service_discovery(struct v2g_connection* conn)
              "PnC is not allowed without TLS-communication. Correcting value to '1' (ExternalPayment)");
     }
 
+    const auto payment_option_length =
```
james-ctc: could std::min() be used here?

SebaLukas: Yeah :+1:

james-ctc (approved):
looks good to me

mlitre (approved):
lgtm!

PR #2092, opened 2026-04-17
libiso15118: d2: State PaymentServiceSelection (#1841)

Dalachowsky:
## Describe your changes

Added state `PaymentServiceSelection` along with tests.

#1841 


Dalachowsky:
Same unit test failed as in #1919. If I had to guess then I'd say that there may be some uninitialized variable used, which sometimes leads to the tests failing, depending on what garbage is at the address. I'll take a look at the code but it would be nice to know which case is failing exactly.

Dalachowsky:
Ok that was quick. I think I found the culprit. Here would be the fix:
```
diff --git a/lib/everest/iso15118/src/iso15118/message/d2/dc_current_demand.cpp b/lib/everest/iso15118/src/iso15118/message/d2/dc_current_demand.cpp
index 4e59ec0af..5403a4438 100644
--- a/lib/everest/iso15118/src/iso15118/message/d2/dc_current_demand.cpp
+++ b/lib/everest/iso15118/src/iso15118/message/d2/dc_current_demand.cpp
@@ -12,6 +12,8 @@
 namespace iso15118::d2::msg {
 
 template <> void convert(const struct iso2_CurrentDemandReqType& in, DC_CurrentDemandRequest& out) {
+    init_iso2_CurrentDemandReqType(&in);
+
     convert(in.DC_EVStatus, out.ev_status);
     convert(in.EVTargetCurrent, out.ev_target_current);
     CB2CPP_CONVERT_IF_USED(in.EVMaximumCurrentLimit, out.ev_maximum_current_limit);
```

I'll introduce it in the next commit. For now you can review the current changes @SebaLukas.

lib/everest/iso15118/src/iso15118/d2/state/payment_service_selection.cpp
```diff
@@ -0,0 +1,153 @@
...
+                    std::find_if(parameters_list.value().begin(), parameters_list.value().end(),
+                                 [&](const auto& entry) {
+                                     return entry.parameter_set_id == selected_service.parameter_set_id.value();
+                                 }) == parameters_list.value().end()) {
+                    // m_ctx.log("Error processing PaymentServiceSelectionReq - selected unsupported parameter set id
+                    // for service");
+                    processing_ok = false;
+                    continue;
+                }
+            }
```
SebaLukas, line 99: I don't quite understand why `feedback` is being used here.
The `offered_services` should be cached in the `d2::Session` class, and then the `ServicePaymentSelection` state should check whether the selected service is also included in the `offered_services`.
`feedback` should only be used when a custom VAS is to be offered that is only available during runtime.

Dalachowsky: With the check at line 81 you're right I should have just used `offered_services`. However on line 88 i decided that this would be the way of checking if a proper parameter set was selected. As the parameter sets are ~~not standarized in ISO15118-2, the app using libiso15118 should be responsible for providing parameter sets for each service, no?~~ Ok it seems that there indeed are standarized parameters in tables 106, 107. I'll introduce them in the next commit. However for any other parameters we should use `feedback` interface so that the application can provide custom ones? I'm not really familiar with how theses are used, we never used parameters in our EVSE.

Unless we want to skip over them for now and open a ticket for introducing them in the future. So that we can focus on conducting a charging session in the simplest form.

SebaLukas: Skip them for now

lib/everest/iso15118/src/iso15118/d2/state/payment_service_selection.cpp
```diff
@@ -0,0 +1,153 @@
...
+            // [V2G2-539] Terminate session after ResponseCode::FAILED
+            m_ctx.session_stopped = true;
+            return {};
+        }
+
+        return m_ctx.create_state<Authorization>();
+    } else if (const auto req = variant->get_if<msg::ServiceDetailRequest>()) {
+        // [V2G2-547] process ServiceDetailRequest
+
+        bool service_supported = m_ctx.feedback.get_service_from_id(req->service_id).has_value();
```
SebaLukas: This should be only used for custom vas, defined outside of libiso.
For the other services (Internet and CertificateInstallation) we need to define the `paramters_list` here for D2 as well.

lib/everest/iso15118/src/iso15118/d2/state/payment_service_selection.cpp
```diff
@@ -0,0 +1,153 @@
...
+
+        for (const auto& selected_service : req->selected_service_list) {
+            if (not m_ctx.feedback.get_service_from_id(selected_service.service_id).has_value()) {
+                // m_ctx.log("Error processing PaymentServiceSelectionReq - selected unsupported service");
+                processing_ok = false;
+                continue;
+            }
+
+            if (selected_service.parameter_set_id.has_value()) {
+                const auto parameters_list = m_ctx.feedback.get_service_parameters_list(selected_service.service_id);
```
SebaLukas: Dont use `feedback.get_service_from_id` and `feedback.get_service_parameters_list`. This are only for custom vas, defined outside of libiso15118.
In this case the state should check if the service id was really offered in the ServiceDiscoveryRes and if yes then the service id and parameter list should be saved in the session (Except the `EvCharging` service)

lib/everest/iso15118/src/iso15118/d2/session.cpp
```diff
@@ -18,6 +18,20 @@ Session::Session() {
     }
 }
 
+void Session::select_service(dt::SelectedService service) {
+    for (const auto& selected_service : selected_services) {
+        if (service.service_id == selected_service.service_id) {
+            // TODO(kd) raise an exception? Overwrite the service?
```
SebaLukas: This make no sense. `selected_services` should be empty in this point

lib/everest/iso15118/src/iso15118/d2/session.cpp
```diff
@@ -18,6 +18,20 @@ Session::Session() {
     }
 }
 
+void Session::select_service(dt::SelectedService service) {
```
SebaLukas: If the EV selects a VAS, how is this handled?

lib/everest/iso15118/src/iso15118/d2/session.cpp
Dalachowsky: During first call yes, this check is supposed to handle cases when EV selects one ServiceID more than once

lib/everest/iso15118/include/iso15118/d2/session.hpp
```diff
@@ -31,13 +35,24 @@ class Session {
...
 
+    std::optional<dt::PaymentOption> selected_payment_option{std::nullopt};
+
 private:
     // NOTE (aw): could be const
     std::array<uint8_t, ID_LENGTH> id{};
 
     bool service_renegotiation_supported{false};
+
+    dt::SelectedServiceList selected_services{};
```
SebaLukas: I think we need to split the selected_services in two variables. One for the ChargeService and one for the optional vas.
For the charge service it is enough to use a bool to indicat that the Ev selected this service.
For the vas services we need only the selected parameters from the selected vas (e.g. internet protocol and port etc). Look here how I did it for d20: https://github.com/EVerest/EVerest/blob/main/lib/everest/iso15118/include/iso15118/d20/session.hpp#L83
With this approach it is hard to look into the vas parameters later in the fsm.

lib/everest/iso15118/src/iso15118/d2/session.cpp
SebaLukas: The EV only sends once PaymentServiceSelectionReq and selecting the energy service and optional vas in the selected_service_list.

SebaLukas: And in case of Pause/Resume the selected services should be saved before sending dlink_pause and selected again when the session resumed.

lib/everest/iso15118/include/iso15118/d2/session.hpp
```diff
@@ -31,13 +35,24 @@ class Session {
...
+    void select_service(dt::SelectedService service);
+    void select_service(dt::ServiceID service_id, std::optional<dt::ParameterSetID> parameter_set_id);
+
+    auto get_selected_services() const& {
+        return selected_services;
+    }
+
     ~Session();
 
+    std::optional<dt::PaymentOption> selected_payment_option{std::nullopt};
```
SebaLukas: I would prefer functions that could be called to select the payment_option and getting the option instead of a public variable

SebaLukas (changes requested):
Another review

lib/everest/iso15118/src/iso15118/d2/state/payment_service_selection.cpp
Dalachowsky: Regarding the InternetAccess service, do we want to only support the protocol/port combinations defined in -2 or do we want to also allow the user to specify additional combinations?

SebaLukas: In the beginning only the 4 options defined in -2

Dalachowsky:
I finally had time to rework the state. Tests will be added later, for now I wanted @SebaLukas to have a look on the implementation.
Regarding the Internet Access service I decided that the user shall provide the parameters list compliant with the Table 107. Mapping between the `struct InternetAccessParameter` and `struct ParameterSet` while automatically mapping the parameter set IDs was too much of a hassle for limited value added.

lib/everest/iso15118/src/iso15118/message/d2/msg_data_types.cpp
```diff
@@ -28,6 +28,25 @@ PhysicalValue from_float(const float in, const data_types::UnitSymbol unit) {
     return out;
 }
 
+ServiceCategory convert_service_id_to_service_category(const std::uint16_t service_id) {
+    switch (service_id) {
+    case 1:
+        return ServiceCategory::EvCharging;
+        break;
```
SebaLukas, line 35: If you return early then you dont need `break`

lib/everest/iso15118/src/iso15118/d2/state/payment_service_selection.cpp
```diff
@@ -0,0 +1,121 @@
...
+        if (payment_option_supported) {
+            m_ctx.session.set_selected_payment_option(req->selected_payment_option);
+            processing_ok = true;
+        } else {
+            // m_ctx.log("Payment option not supported");
+        }
+
+        for (const auto& selected_service : req->selected_service_list) {
+            m_ctx.session.select_service(selected_service);
+        }
```
SebaLukas, line 77: That could be moved to `handle_request`

lib/everest/iso15118/src/iso15118/d2/state/payment_service_selection.cpp
```diff
@@ -0,0 +1,121 @@
...
+
+        const auto res = handle_request(*req, m_ctx.session, processing_ok);
```
SebaLukas, line 79: A check if the ev selected charge_service is missing

lib/everest/iso15118/src/iso15118/d2/state/payment_service_selection.cpp
```diff
@@ -0,0 +1,121 @@
...
+            // m_ctx.log("Error processing PaymentServiceSelectionReq");
+            // [V2G2-539] Terminate session after ResponseCode::FAILED
+            m_ctx.session_stopped = true;
+            return {};
+        }
+
+        return m_ctx.create_state<Authorization>();
+    } else if (const auto req = variant->get_if<msg::ServiceDetailRequest>()) {
+        // [V2G2-547] process ServiceDetailRequest
+        std::optional<dt::ServiceParameterList> service_parameters(std::nullopt);
```
SebaLukas, line 93: `{}` instead of `(std::nullopt)` is enough

lib/everest/iso15118/src/iso15118/d2/state/payment_service_selection.cpp
```diff
@@ -0,0 +1,121 @@
...
+        service_parameters = m_ctx.session.get_service_parameter_list(req->service_id);
```
SebaLukas, line 94: You can combine both lines to
```cpp
const auto service_parameters = m_ctx.session.get_service_parameter_list(req->service_id);
```

lib/everest/iso15118/src/iso15118/d2/session.cpp
```diff
@@ -18,6 +21,84 @@ Session::Session() {
...
+        }
+    }
+    return ret;
+}
+
+void Session::select_service(dt::SelectedService service) {
+    select_service(service.service_id, service.parameter_set_id);
+}
+
+void Session::select_service(dt::ServiceID service_id, std::optional<dt::ParameterSetID> parameter_set_id) {
```
SebaLukas, line 71: If the ev sends accidently the same service_id twice, how is that handled here? -> The selected_vas_paramter is overwritten. So please add a check if service_id is already selected.

lib/everest/iso15118/include/iso15118/d2/session.hpp
```diff
@@ -4,9 +4,47 @@
...
+typedef uint16_t Port;
+} // namespace msg::data_types
+
+struct InternetAccessParameter {
+    dt::Protocol protocol;
+    dt::Port port;
+};
+
+struct OfferedServices {
+    std::vector<dt::CertificateService> certificate_service_parameter_list;
```
SebaLukas, line 38: `dt::CertificateService` -> `dt` is not needed

lib/everest/iso15118/include/iso15118/d2/session.hpp
```diff
@@ -4,9 +4,47 @@
...
+    // Parameter set IDs should comply with ISO 15118-2:2014 Table 107
+    std::map<dt::ParameterSetID, InternetAccessParameter> internet_access_parameter_list;
+};
+
+struct SelectedVasParameter {
+    std::optional<dt::CertificateService> certificate_service;
```
SebaLukas, line 44: `dt::CertificateService` -> `dt` is not needed

lib/everest/iso15118/include/iso15118/d2/session.hpp
Dalachowsky: It is, if you look at the top of the file `CertificateService` is namespaced inside `msg::data_types`. There is also a question pending whether should I put it in `msg_data_types.hpp`  or perhaps leave it in `session.hpp` without namespacing it as a data type

SebaLukas: Here `OfferedServices` is already in the data_types namespace and I guess it uses the enum class from line 18. Or it is another `CertificateService` type?

SebaLukas: Okay line 30 the data_type namespace ended. So its good.

lib/everest/iso15118/src/iso15118/d2/state/payment_service_selection.cpp
Dalachowsky: I still do not quite get how the logic should be separated between `feed()` and `handle_request()`. Most of the logic here is relying on `m_ctx` so i figured that it fits here.

lib/everest/iso15118/src/iso15118/d2/session.cpp
Dalachowsky: Do we want to respond with `ResponseCode::FAILED` or just omit the entry?

Dalachowsky:
I'm working on the tests for the `PaymentServiceSelection` and I realised that `offered_services` being kept in the `Session` object make it so that we cannot configure it with `SessionConfig`. Also moving it to `SessionConfig` would mean that some of the `Session::select_service()` logic would have to be moved to the fsm logic or we would have to pass something like `session_config.offered_services` into the `Session::select_service()`. How would you see it @SebaLukas ?

SebaLukas:
Can you try to update your branch?

SebaLukas:
Closing this one because @razvanphp will continue

PR #2093, opened 2026-04-17
yocto/scarthgap: add openssl patch for newer versions

Hgerner:
The patch originally provided in lib/staging/tls and lib/everest/tls was for openssl-3.0.8, and no longer applies to openssl version 3.2.x (yocto-5.0.16) and 3.5.x (post 5.0.17).

## Describe your changes
Convert the original SRC_URI from a URL pointing to a git blob to an actual patch file. Create a rebased patch file for 3.5.x.

Apply the patches conditionally, depending on the openssl version.

### DIFF
--- a/yocto/scarthgap/meta-everest/recipes-connectivity/openssl/openssl/openssl-3.5-feat-updates-to-support-status_request_v2.patch
+++ b/yocto/scarthgap/meta-everest/recipes-connectivity/openssl/openssl/openssl-3.5-feat-updates-to-support-status_request_v2.patch
@@ -1,0 +1,144 @@
+From e5f3cc0bf62f5ad58c419b6c46ac99800ae6b11c Mon Sep 17 00:00:00 2001
+From: Kai-Uwe Hermann <kai-uwe.hermann@pionix.de>
+Date: Fri, 7 Feb 2025 10:45:21 +0100
+Subject: [PATCH] feat: updates to support status_request_v2
+
+Upstream-Status: Pending
+
+Co-authored-by: James Chapman <james.chapman@pionix.de>
+Co-authored-by: Kai-Uwe Hermann <kai-uwe.hermann@pionix.de>
+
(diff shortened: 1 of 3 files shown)


barsnick:
Hej Håkan,
you need to sign off your commit (`git commit -s [--amend]` and force-push) for the DCO check to pass. (Use the same e-mail address as in your git Author tag.)

The description is a bit incorrect. A patch for 3.2 was already in scarthgap, and it applies until 3.5.4 - as your modifications to the recipe correctly reflect.

@james-ctc probably needs to verify whether the patch is okay for openssl-3.5.5+.

Hgerner:
Oh right. I'm sorry about that. 

I will fix the sign-off and commit message. It also seems like the 3.2 patch applies up to version 3.3.5. Then there's a gap where neither patch applies between 3.3.6-3.5.4. 

I'm unclear on the PV semantics when openembedded jumps versions. Either use <3.3.6 or <3.5.5.

barsnick:
Ah, I didn't check the 3.3 branch. I think openssl changed something in every minor branch (3.0, 3.1, 3.2, 3.3, ...) to break these patches.

That shouldn't be relevant for scarthgap though - they went directly from 3.2.6 to 3.5.5. Don't worry about the patch applying to anything else inbetween.

james-ctc (approved):

barsnick:
BTW, this provided patch will also apply cleanly to 3.5.6, but will not apply to 4.0.0. But we cannot expect 4.0.0 in kirkstone, scarthgap or wrynose, so this PR should be sufficient.

Hgerner:
I'm not familiar with GitHub's UI for testing and merging, but from what I could tell, the OCPP test had a ValueType error (which does not appear to be one I caused). Can I rerun the tests somehow, or do I just wait for you to do it (assuming I didn't cause it)?

barsnick:
> Can I rerun the tests somehow, or do I just wait for you to do it

Possibly, only admins/owners can retrigger the tests. I did so, and they passed.

PR #2094, opened 2026-04-17
Rust: Support shared linking

dorezyuk:
## Describe your changes

Bazel by default links everything statically. This change uses `cc_shared_library` in conjunction with `dynamic_mode=fully` to link to `libframework.so` and not to `libframework.a

To make sure that the Rust bindings don't link to boost log symbols we also move the logging registration into liblog. 

TBD: Maybe we should wire the log bindings through libframework to make the dynamic linking more robust

### DIFF
--- a/lib/everest/log/include/everest/logging.hpp
+++ b/lib/everest/log/include/everest/logging.hpp
@@ -33,11 +33,11 @@ enum severity_level {
 int init();
 
 /// \brief Initialize logger using the config pointed to by \p logconf
-/// \return minimum accepted severity level (-1=off, 0=verbose .. 4=error)
+/// \return minimum accepted severity level (-1=off, 0=verbose .. 5=critical)
 int init(const std::string& logconf);
 
 /// \brief Initialize logger using the config pointed to by \p logconf additionally setting the \p process_name
-/// \return minimum accepted severity level (-1=off, 0=verbose .. 4=error)
+/// \return minimum accepted severity level (-1=off, 0=verbose .. 5=critical)
 int init(const std::string& logconf, std::string process_name);
 
 /// \brief Logging function for foreign language interfaces.
--- a/lib/everest/log/include/everest/logging.hpp
+++ b/lib/everest/log/include/everest/logging.hpp
@@ -29,13 +29,19 @@ enum severity_level {
 };
 
 /// \brief Initialize a completely silenced logger
-void init();
+/// \return -1 for off.
+int init();
 
 /// \brief Initialize logger using the config pointed to by \p logconf
-void init(const std::string& logconf);
+/// \return minimum accepted severity level (-1=off, 0=verbose .. 4=error)
+int init(const std::string& logconf);
 
 /// \brief Initialize logger using the config pointed to by \p logconf additionally setting the \p process_name
-void init(const std::string& logconf, std::string process_name);
+/// \return minimum accepted severity level (-1=off, 0=verbose .. 4=error)
+int init(const std::string& logconf, std::string process_name);
+
+/// \brief Logging function for foreign language interfaces.
+void ffi_log(int level, int line, const std::string& file, const std::string& message);
 
 void update_process_name(std::string process_name);
 std::string trace();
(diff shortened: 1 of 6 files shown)


lib/everest/log/include/everest/logging.hpp
```diff
@@ -29,13 +29,19 @@ enum severity_level {
...
+/// \return minimum accepted severity level (-1=off, 0=verbose .. 4=error)
+int init(const std::string& logconf);
 
 /// \brief Initialize logger using the config pointed to by \p logconf additionally setting the \p process_name
-void init(const std::string& logconf, std::string process_name);
+/// \return minimum accepted severity level (-1=off, 0=verbose .. 4=error)
+int init(const std::string& logconf, std::string process_name);
+
+/// \brief Logging function for foreign language interfaces.
+void ffi_log(int level, int line, const std::string& file, const std::string& message);
```
dorezyuk, line 44: we should pipe that possibly through `framework/everest.hpp` or `framework/runtime.hpp` to make sure that if clients link against framewokr they don't depend on symbols from other libs

lib/everest/log/include/everest/logging.hpp
```diff
@@ -29,13 +29,19 @@ enum severity_level {
 };
 
 /// \brief Initialize a completely silenced logger
-void init();
+/// \return -1 for off.
+int init();
 
 /// \brief Initialize logger using the config pointed to by \p logconf
-void init(const std::string& logconf);
+/// \return minimum accepted severity level (-1=off, 0=verbose .. 4=error)
```
hikinggrass: doesn't this go up to 5=critical?

lib/everest/log/lib/logging.cpp
```diff
@@ -203,6 +204,16 @@ void init(const std::string& logconf, std::string process_name) {
 
     EVLOG_debug << "Logger " << (is_initialized ? "re" : "") << "initialized (using " << logconf << ")...";
     is_initialized = true;
+
+    // Probe the installed filter to find the minimum accepted severity level.
+    for (int ii = static_cast<int>(verbose); ii <= static_cast<int>(error); ++ii) {
```
hikinggrass: same here, this goes one further to `critical` doesn't it?

lib/everest/log/lib/logging.cpp
```diff
@@ -203,6 +204,16 @@ void init(const std::string& logconf, std::string process_name) {
...
+        auto rec = global_logger::get().open_record(boost::log::keywords::severity = static_cast<severity_level>(ii));
+        if (rec) {
+            return ii;
```
hikinggrass: maybe use a more descriptive name like `level`

lib/everest/log/lib/logging.cpp
```diff
@@ -203,6 +204,16 @@ void init(const std::string& logconf, std::string process_name) {
...
+        }
+    }
+
+    return -1;
```
hikinggrass, line 217: did you choose an int here on purpose for compatibility reasons? or could we return `Everest::Logging::severity_level` here? Would also save us the conversion above

hikinggrass (changes requested):

lib/everest/log/lib/logging.cpp
dorezyuk: yes I didn't want to leak this into the interface and also the level has no enum value for disabled

dorezyuk: ah actually it makes no sense what I wrote - so we can use severity_level - but then we need to somehow indicate the -1 case

lib/everest/log/lib/logging.cpp
dorezyuk: done

lib/everest/log/lib/logging.cpp
dorezyuk: done

lib/everest/log/include/everest/logging.hpp
dorezyuk: done

hikinggrass (approved):

PR #2098, opened 2026-04-20
fix(framework): add thread pool exception policy and use Rethrow in message handling

Pietfried:
## Describe your changes

Add ExceptionPolicy to thread_pool_scaling with Suppress and Rethrow modes. Initialize the MessageHandler operation thread pool with Rethrow to preserve pre-extension behavior. Keep exception handling decisions in everest.cpp handler wrappers, where forwarding policy is already defined. This avoids silently swallowed handler exceptions and lets modules terminate on unhandled failures.

### DIFF
--- a/lib/everest/framework/tests/test_message_handler.cpp
+++ b/lib/everest/framework/tests/test_message_handler.cpp
@@ -812,43 +812,6 @@ TEST_CASE("MessageHandler processes GetConfig messages", "[message_handler][conf
 // Test: Mixed Message Types
 // ============================================================================
 
-// ============================================================================
-// Test: Ordering preserved when a handler throws
-// ============================================================================
-
-TEST_CASE("MessageHandler preserves per-topic ordering after handler exception",
-          "[message_handler][ordering][exception]") {
-    MessageHandlerFixture handler;
-    ExecutionTracker tracker;
-
-    auto handler_func = std::make_shared<Handler>([&](const std::string& topic, const json& data) {
-        int seq = data.value("sequence", 0);
-        if (seq == 1) {
-            throw std::runtime_error("simulated handler failure");
-        }
-        tracker.record(topic, seq);
-    });
-    auto test_handler = std::make_shared<TypedHandler>(HandlerType::Call, handler_func);
-
-    handler->register_handler("test/topic", test_handler);
-
-    // Message 1 throws; messages 2 and 3 must still be processed in order.
-    handler->add(create_cmd_message("test/topic", 1));
-    handler->add(create_cmd_message("test/topic", 2));
-    handler->add(create_cmd_message("test/topic", 3));
-
-    tracker.wait_for_count(2);
-
-    auto events = tracker.get_events();
-    REQUIRE(events.size() == 2);
-    CHECK(events[0].sequence == 2);
-    CHECK(events[1].sequence == 3);
-}
-
-// ============================================================================
-// Test: Mixed Message Types
-// ============================================================================
-
 TEST_CASE("MessageHandler handles mixed message types concurrently", "[message_handler][mixed]") {
     MessageHandlerFixture handler;
     ExecutionTracker cmd_tracker;
--- a/lib/everest/util/include/everest/util/async/thread_pool_scaling.hpp
+++ b/lib/everest/util/include/everest/util/async/thread_pool_scaling.hpp
@@ -108,14 +108,35 @@ template <std::size_t ThresholdMs = 10> struct LatencyScaling {
 // |              |                          | when a specific depth limit is hit.    |
 // +--------------+--------------------------+----------------------------------------+
 
+// --- Exception Handling Policies ---
+
+/**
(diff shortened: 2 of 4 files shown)


lib/everest/util/include/everest/util/async/thread_pool_scaling.hpp
```diff
@@ -127,15 +140,17 @@ template <typename ScalingPolicy = LatencyScaling<10>> class thread_pool_scaling
      * @param[in] max Maximum allowed worker threads.
      * @param[in] timeout Idle duration before a surplus worker retires. Defaults to 60s.
      * @param[in] queue_limit Maximum tasks allowed in the queue. Defaults to 0 (unbounded).
+     * @param[in] exception_policy Exception handling policy for task execution. Defaults to Suppress.
      */
     template <class Rep, class Period>
     thread_pool_scaling(std::size_t min, std::size_t max,
                         std::chrono::duration<Rep, Period> timeout = std::chrono::seconds(60),
-                        std::size_t queue_limit = 0) :
+                        std::size_t queue_limit = 0, ExceptionPolicy exception_policy = ExceptionPolicy::Suppress) :
```
djchhp: As a policy, this should be a template parameter, no need for runtime checks

lib/everest/util/include/everest/util/async/thread_pool_scaling.hpp
```diff
@@ -265,8 +280,12 @@ template <typename ScalingPolicy = LatencyScaling<10>> class thread_pool_scaling
...
                         task_opt->func();
                     } catch (...) {
-                        // Suppress exception to prevent thread termination.
-                        // Fire-and-forget tasks are responsible for their own error handling.
+                        switch (m_exception_policy) {
+                        case ExceptionPolicy::Suppress:
+                            break;
+                        case ExceptionPolicy::Rethrow:
+                            throw;
+                        }
```
djchhp: Once the policy is a template parameter this can become `if constexpr` and handled at compile time

lib/everest/util/include/everest/util/async/thread_pool_scaling.hpp
```diff
@@ -108,6 +108,19 @@ template <std::size_t ThresholdMs = 10> struct LatencyScaling {
...
+ * @details Controls how the thread pool handles exceptions thrown by tasks:
+ *   - Suppress: Silently catch and discard exceptions (fire-and-forget semantics)
+ *   - Rethrow: Rethrow from worker context; uncaught thread exceptions terminate process
+ */
+enum class ExceptionPolicy {
+    Suppress, ///< Silently suppress exceptions, thread continues
+    Rethrow   ///< Rethrow exception from worker task context
+};
+
 /**
```
djchhp: I wonder if an enum as policy is sufficient, i.e. how generic can this become? Let's say you want special logging for those exception. That could be easily plugged in by user without touching thread pool, if this was a class based policy with a member function `handle_exception` or similar.

djchhp (changes requested):

lib/everest/framework/tests/test_message_handler.cpp
```diff
@@ -812,43 +812,6 @@ TEST_CASE("MessageHandler processes GetConfig messages", "[message_handler][conf
 // Test: Mixed Message Types
 // ============================================================================
 
-// ============================================================================
-// Test: Ordering preserved when a handler throws
```
mlitre, line 816: Shouldn't we just keep this test and maybe only use it on the supressed version?

mlitre (commented):
I would agree with all of @djchhp comments as well

lib/everest/framework/tests/test_message_handler.cpp
Pietfried: The policy is not configurable for the `MessageHandler` class and I'd like to avoid extending it only for this test case.

Pietfried:
@djchhp I adressed your suggestions and replaced the ExceptionPolicy enum with a template parameter

mlitre (dismissed):
This LGTM now

lib/everest/util/include/everest/util/async/thread_pool_scaling.hpp
```diff
@@ -108,14 +108,35 @@ template <std::size_t ThresholdMs = 10> struct LatencyScaling {
 // |              |                          | when a specific depth limit is hit.    |
 // +--------------+--------------------------+----------------------------------------+
 
+// --- Exception Handling Policies ---
+
+/**
+ * @brief Exception policy: silently swallow exceptions (fire-and-forget semantics).
+ */
+struct SuppressExceptions {
+    static void handle_exception(std::exception_ptr) noexcept {
```
djchhp:
```suggestion
    static void handle_exception([[maybe_unused]] std::exception_ptr) noexcept {
```
in order to suppress compiler warnings I suggest addind maybe_unused

djchhp (changes requested):
Just a minor formality left to make the compiler happy. Otherwise this is now very generic and user friendly.

djchhp (approved):

PR #2100, opened 2026-04-20
feat(ocpp v21): R04 DER Control

mlitre:
Implements OCPP 2.1 Edition 2 §R04 DER Control functional block: SetDERControl / GetDERControl / ClearDERControl handlers, NotifyDERStartStop dispatch (FR.20-22), priority- based supersede with deferred future-start activation, and SQLite persistence with migration, row cap, and schema CHECK constraints. Rejects non-finite floats, multi-field requests, and out-of-horizon startTimes. ACDERCtrlr / DCDERCtrlr device-model components added.


lib/everest/ocpp/tests/config/v2/DCDERCtrlr_1.json
Pietfried, line 1: Could this be retrieved dynamically just like `EVSE` and `Connector` components?

lib/everest/ocpp/include/ocpp/v2/database_handler.hpp
```diff
@@ -199,7 +199,88 @@ class DatabaseHandlerInterface {
...
+    /// \brief Retrieves a single DER control by control_id
+    virtual std::optional<std::string> get_der_control(const std::string& control_id) = 0;
+
+    /// \brief Total number of rows currently in DER_CONTROLS. Used by the DER control
+    /// handler to cap the table so a CSMS cannot grow the DB unbounded by issuing
+    /// SetDERControl with ever-unique controlIds.
+    virtual std::size_t count_der_controls() = 0;
+
+    /// \brief Retrieves all DER controls
+    virtual std::vector<std::string> get_all_der_controls() = 0;
```
Pietfried: nit: looks like this is never used

lib/everest/ocpp/lib/ocpp/v21/functional_blocks/der_control.cpp
```diff
@@ -0,0 +1,1049 @@
...
+    // Lambda holds a stopping handle for callback duration so destructor can wait for in-flight runs.
+    this->scheduled_control_timer.interval(
+        [this]() {
+            auto handle = this->stopping.handle();
+            if (*handle) {
+                return;
+            }
+            this->check_scheduled_controls();
+        },
+        std::chrono::seconds(30));
```
Pietfried: Should this be configurable or at least a constant definition?

lib/everest/ocpp/include/ocpp/v21/functional_blocks/der_control.hpp
```diff
@@ -0,0 +1,65 @@
...
+/// All DER controls are persisted in the database, no in-memory cache.
+class DERControl : public DERControlInterface {
+public:
+    explicit DERControl(const v2::FunctionalBlockContext& context);
+    ~DERControl() override;
+
+    void handle_message(const ocpp::EnhancedMessage<v2::MessageType>& message) override;
+
+    /// Check if a controlType is supported by any EVSE (DC or AC DERCtrlr)
+    bool is_control_type_supported(v2::DERControlEnum control_type) const;
```
Pietfried: afaics there is no need for this function to be public

lib/everest/ocpp/lib/ocpp/v21/functional_blocks/der_control.cpp
```diff
@@ -0,0 +1,1049 @@
...
+    std::vector<FixedPFGet> fixed_pf_injects;
+    std::vector<FixedVarGet> fixed_vars;
+    std::vector<FreqDroopGet> freq_droops;
+    std::vector<GradientGet> gradients;
+    std::vector<LimitMaxDischargeGet> limit_max_discharges;
+
+    for (const auto& control_json_str : control_jsons) {
+        try {
+            auto stored = json::parse(control_json_str);
+            const auto& request_json = stored.at("request");
```
Pietfried: There are quite so raw JSON operations which could be avoided by serializing to `SetDERControlRequest` directly.
```suggestion
            auto request = stored.at("request").get<SetDERControlRequest>();
```

lib/everest/ocpp/lib/ocpp/v21/functional_blocks/der_control.cpp
```diff
@@ -0,0 +1,1049 @@
...
+    case DERControlEnum::VoltVar:
+    case DERControlEnum::VoltWatt:
+    case DERControlEnum::WattPF:
+    case DERControlEnum::WattVar:
+        return req.curve.has_value() && validate_yunit(req.controlType, req.curve.value());
+    }
+    return false;
+}
+
+void DERControl::handle_set_der_control(ocpp::Call<SetDERControlRequest> call) {
```
Pietfried, line 413: The `database_handler` is used a lot in this implementation and can throw `QueryExecutionException` from many code paths. I would recommend to wrap `handle_set_der_control`, `handle_get_der_control` and `handle_clear_der_control` in a try-catch block to avoid terminating because of an unhandled exception is the DERControl impl

lib/everest/ocpp/include/ocpp/v21/functional_blocks/der_control.hpp
```diff
@@ -0,0 +1,65 @@
...
+#include <ocpp/v21/messages/SetDERControl.hpp>
+
+namespace ocpp::v21 {
+
+class DERControlInterface : public v2::MessageHandlerInterface {
+public:
+    ~DERControlInterface() override = default;
+};
+
+/// Functional block implementing OCPP 2.1 use case R04 "Configure DER control settings at Charging Station".
```
Pietfried: I think this is not only for R04 but for the full functional block R ?

lib/everest/ocpp/doc/v2/ocpp_2x_status.md
```diff
@@ -2756,39 +2756,57 @@ CSMS setpoint and Dynamic charging profiles from K01. There are no specific requ
...
+| R04.FR.37 | ✅     | |
+|           | **ClearDERControl** | |
+| R04.FR.40 |        | (removed in Edition 2) |
+| R04.FR.41 | ✅     | |
+| R04.FR.42 | ✅     | |
+| R04.FR.43 | ✅     | |
+| R04.FR.44 | ✅     | |
+| R04.FR.45 | ✅     | |
+| R04.FR.46 | ✅     | |
+|           | **yUnit validation (FR.50-56)** | |
```
Pietfried:
```suggestion
|           | **yUnit validation** | |
```

Pietfried (changes requested):

lib/everest/ocpp/include/ocpp/v21/functional_blocks/der_control.hpp
mlitre: Theoretically not wrong, but the actual control based on the grid codes will most likely be done outside of libocpp, so in the libocpp will only really handle the current grid codes being stored and retrieving the currently valid ones (similarly to charging profiles) and that will be passed on to the correct person may that be the EV (R02), a smart PSU that can handle the grid codes (R01 / R03) or a seperate module to handle applying the grid codes (R01/R03). I guess this functional block also implements R05, but that is mainly triggered from an outside source

lib/everest/ocpp/tests/config/v2/DCDERCtrlr_1.json
mlitre: This should be dynamically retrieved like EVSE and Connector, but this requires the new interface I will build for grid_support. Since that's not yet done I wanted the schemas to be defined already. Open to suggestions on this part. I was planning on just not importing the json anymore and building it dynamically from the result of the new interface that will be added later.

lib/everest/ocpp/lib/ocpp/v21/functional_blocks/der_control.cpp
mlitre: I've moved it to a constant definition, we could make it configurable, but I think that could be added later if needed.

lib/everest/ocpp/tests/config/v2/DCDERCtrlr_1.json
Pietfried: Ok. Would it be an option to just keep DCDERCtrlr_1.json to support the unit tests and drop the others? If we keep the other component configs they would be stored persistently in the device model db and this could cause issues when we switch to retrieving these components dynamically.

mlitre: Yes we can do that, eg keep in custom ad only in tests import them

mlitre: Should I keep the "generic" ones in the standardized parts or also remove them and have everything defined in code

Pietfried: If possible I'd keep the definitions in ctrlr component vars but drop the JSON configs. I guess they are currently needed for the e2e tests, maybe you can just mark those with pytest.mark.skip for now?

Pietfried (approved):
lgtm

PR #2102, opened 2026-04-21
Framework: Fix LatencyScaling in thread pools

dorezyuk:
## Describe your changes

The applications can "deadlock" if they don't have enough threads and don't receive new messages to retrigger the examination of the latency inside the thread pool.

The fix adds a new thread to periodically check on the latency of the queued messages

### DIFF
--- a/lib/everest/util/include/everest/util/async/thread_pool_scaling.hpp
+++ b/lib/everest/util/include/everest/util/async/thread_pool_scaling.hpp
@@ -12,7 +12,10 @@
 #include <functional>
 #include <future>
 #include <list>
+#include <optional>
 #include <thread>
+#include <type_traits>
+#include <variant>
 #include <vector>
 
 namespace everest::lib::util {
@@ -34,10 +37,15 @@ struct TrackedAction {
 
 // --- Scaling Policies ---
 
+// A policy advertises whether it needs a background supervisor (and at what
+// cadence) via a single constexpr: `supervisor_tick`. `std::nullopt` means no
+// supervisor; a value means "re-evaluate scaling every <tick> ms".
+
 /**
  * @brief Greedy scaling policy: grows whenever there is any backlog.
  */
 struct GreedyScaling {
+    static constexpr std::optional<std::chrono::milliseconds> supervisor_tick = std::nullopt;
     /**
      * @brief Decides to grow if there is any backlog.
      * @param current_workers Number of threads currently in the registry.
@@ -56,6 +64,7 @@ struct GreedyScaling {
  * @brief Conservative scaling policy: grows only when backlog is significant.
  */
 struct ConservativeScaling {
+    static constexpr std::optional<std::chrono::milliseconds> supervisor_tick = std::nullopt;
     static bool should_grow(std::size_t current_workers, std::size_t queue_size,
                             [[maybe_unused]] std::optional<std::chrono::steady_clock::time_point> oldest_task) {
         return queue_size > (current_workers * 2);
@@ -67,6 +76,7 @@ struct ConservativeScaling {
  * @tparam Limit The queue size threshold.
  */
 template <std::size_t Limit> struct FixedSizeScaling {
+    static constexpr std::optional<std::chrono::milliseconds> supervisor_tick = std::nullopt;
     static bool should_grow([[maybe_unused]] std::size_t current_workers, std::size_t queue_size,
                             [[maybe_unused]] std::optional<std::chrono::steady_clock::time_point> oldest_arrival) {
         return queue_size >= Limit;
@@ -75,12 +85,14 @@ template <std::size_t Limit> struct FixedSizeScaling {
 
 /**
  * @brief Latency-based scaling policy: grows if the oldest task has waited too long.
- * @tparam MaxWaitMs Maximum tolerable wait time in milliseconds.
+ * @tparam ThresholdMs Maximum tolerable wait time in milliseconds.
+ * @tparam TickMs Cadence at which the supervisor re-evaluates the policy.
  */
-template <std::size_t ThresholdMs = 10> struct LatencyScaling {
+template <std::size_t ThresholdMs = 10, std::size_t TickMs = 5> struct LatencyScaling {
+    static constexpr std::optional<std::chrono::milliseconds> supervisor_tick = std::chrono::milliseconds(TickMs);
     static bool should_grow([[maybe_unused]] std::size_t current_workers, std::size_t queue_size,
                             std::optional<std::chrono::steady_clock::time_point> oldest_arrival) {
-        if (queue_size <= 1 or not oldest_arrival.has_value()) {
+        if (queue_size < 1 or not oldest_arrival.has_value()) {
             return false;
         }
         const auto now = std::chrono::steady_clock::now();
@@ -127,6 +139,9 @@ public:
      * @param[in] max Maximum allowed worker threads.
      * @param[in] timeout Idle duration before a surplus worker retires. Defaults to 60s.
      * @param[in] queue_limit Maximum tasks allowed in the queue. Defaults to 0 (unbounded).
+     *
+     * The supervisor tick (if any) is carried by the ScalingPolicy itself; see
+     * @ref LatencyScaling for an example.
      */
     template <class Rep, class Period>
     thread_pool_scaling(std::size_t min, std::size_t max,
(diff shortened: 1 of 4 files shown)


lib/everest/util/tests/async/thread_pool_scaling_tests.cpp
```diff
@@ -30,13 +30,11 @@ TEST(ThreadPoolScalingTest, ScalesOnLatencyThreshold) {
     std::this_thread::sleep_for(5ms);
 
     std::atomic<bool> second_task_started{false};
+    // This should spawn a new thread already.
     pool.run([&]() { second_task_started = true; });
 
     ASSERT_FALSE(second_task_started.load());
 
-    std::this_thread::sleep_for(30ms);
-    pool.run([]() {}); // Trigger scaling check
```
dorezyuk, line 38: these were actually exposing the bug

lib/everest/util/include/everest/util/queue/thread_safe_bounded_queue.hpp
```diff
@@ -181,7 +189,7 @@ template <class T> class thread_safe_bounded_queue {
 
     simple_queue<T> m_queue;               ///< The underlying non-thread-safe container.
     const size_type m_max_size;            ///< Maximum capacity of the queue.
-    std::mutex m_mtx;                      ///< Mutex guarding access to the queue and state.
+    mutable std::mutex m_mtx;              ///< Mutex guarding access to the queue and state.
```
dorezyuk, line 192: m&m

lib/everest/util/include/everest/util/async/thread_pool_scaling.hpp
```diff
@@ -80,7 +80,7 @@ template <std::size_t Limit> struct FixedSizeScaling {
 template <std::size_t ThresholdMs = 10> struct LatencyScaling {
     static bool should_grow([[maybe_unused]] std::size_t current_workers, std::size_t queue_size,
```
djchhp:
```suggestion
template <std::size_t ThresholdMs = 10> struct LatencyScaling {
    static constexpr bool requires_supervisor = true;
    static bool should_grow([[maybe_unused]] std::size_t current_workers, std::size_t queue_size,
```

for time based policies, we need a supervisor. For purely size based policies, the supervisor is not necessary. Policies should expose if they need a supervisor or not.
The other policies need a `static constexpr bool requires_supervisor = false;` . Unfortunately github does not allow me to suggest the change there.

lib/everest/util/include/everest/util/async/thread_pool_scaling.hpp
```diff
@@ -127,51 +127,66 @@ template <typename ScalingPolicy = LatencyScaling<10>> class thread_pool_scaling
...
-        auto reg_h = m_reg.handle();
-        for (std::size_t i = 0; i < m_min_threads; ++i) {
-            spawn_worker_internal(reg_h);
+        {
+            auto reg_h = m_reg.handle();
+            for (std::size_t i = 0; i < m_min_threads; ++i) {
+                spawn_worker_internal(reg_h);
+            }
         }
+        m_supervisor = std::thread([this] { run_supervisor(); });
```
djchhp:
```suggestion
        if constexpr (ScalingPolicy::requires_supervisor) {
            m_supervisor = std::thread([this] { run_supervisor(); });
        }
```
Only start the supervisor if required.

lib/everest/util/include/everest/util/async/thread_pool_scaling.hpp
```diff
@@ -127,51 +127,66 @@ template <typename ScalingPolicy = LatencyScaling<10>> class thread_pool_scaling
...
         }
+        m_reg.notify_all();
         m_action_queue.stop();
 
-        // 2. Steal the active workers list. Explicitly clear the source so that any
+        // 2. Join the supervisor before tearing down the worker list: the supervisor
+        // can spawn new workers, and we must not race with the steal in step 3.
+        if (m_supervisor.joinable()) {
+            m_supervisor.join();
+        }
```
djchhp, line 184:
```suggestion
        // 2. Conditionally join the supervisor before tearing down the worker list: the supervisor
        // can spawn new workers, and we must not race with the steal in step 3.
        if constexpr (ScalingPolicy::requires_supervisor) {
            if (m_supervisor.joinable()) {
                m_supervisor.join();
            }
        }
```
Only if there is a supervisor, `join()` it

djchhp (changes requested):
This is a highly appreciate discovery and fix. However, before this can be merged a few points still need attendance. I created change suggestions, which all focus an making the supervisor optional (i.e. active based on the policy). They are like the minimal required change.

I think though, the changes should be more general.

1. ) The constructor should not expose the supervisor tick, if the policy does not require it
2.) The m_supervisor_tick and m_supervisor thread should be added/removed based on the policy

On 2.)

This could be achieved by moving those members to an external template
```
// Base template
template <bool RequiresSupervisor> struct SupervisorState;

// Specialization for TRUE: Contains the tick and the thread
template <> struct SupervisorState<true> {
    std::chrono::milliseconds tick;
    std::thread thread;
    explicit SupervisorState(std::chrono::milliseconds t) : tick(t) {
    }
};

// Specialization for FALSE: Completely empty
template <> struct SupervisorState<false> {
    // No variables. Takes up minimal space (1 byte due to C++ rules,
    // completely optimized away in many contexts).
    SupervisorState() = default;
};
```

On 1)

```
public:
    // with supervisor
    template <class Rep = int64_t, class Period = std::ratio<1>, class SP = ScalingPolicy,
              std::enable_if_t<SP::requires_supervisor, int> = 0>
    thread_pool_scaling(std::size_t min, std::size_t max,
                        std::chrono::duration<Rep, Period> timeout = std::chrono::seconds(60),
                        std::size_t queue_limit = 0,
                        std::chrono::milliseconds supervisor_tick = std::chrono::milliseconds(5)) :
        m_min_threads(min),
        m_max_threads(max),
        m_idle_timeout(std::chrono::duration_cast<std::chrono::milliseconds>(timeout)),
        m_supervisor_state(supervisor_tick), // Initialize the struct with the tick
        m_action_queue(queue_limit) {

        init_workers();
    }

    // without supervisor
    template <class Rep = int64_t, class Period = std::ratio<1>, class SP = ScalingPolicy,
              std::enable_if_t<!SP::requires_supervisor, int> = 0>
    thread_pool_scaling(std::size_t min, std::size_t max,
                        std::chrono::duration<Rep, Period> timeout = std::chrono::seconds(60),
                        std::size_t queue_limit = 0) :
        m_min_threads(min),
        m_max_threads(max),
        m_idle_timeout(std::chrono::duration_cast<std::chrono::milliseconds>(timeout)),
        m_supervisor_state(), 
        m_action_queue(queue_limit) {

        init_workers();
    }
private:
    
    void init_workers() {
        {
            auto reg_h = m_reg.handle();
            for (std::size_t i = 0; i < m_min_threads; ++i) {
                spawn_worker_internal(reg_h);
            }
        }
        if constexpr (ScalingPolicy::requires_supervisor) {
            m_supervisor_state.thread = std::thread([this] { run_supervisor(); });
        }
    }
    SupervisorState<ScalingPolicy::requires_supervisor> m_supervisor_state;
```
With those changes in place `init_workers()` needs
```
if constexpr (ScalingPolicy::requires_supervisor) {
            m_supervisor_state.thread = std::thread([this] { run_supervisor(); });
 }
```
the destrcutor
```
if constexpr (ScalingPolicy::requires_supervisor) {
            if (m_supervisor_state.thread.joinable()) {
                m_supervisor_state.thread.join();
            }
 }
```

and `run_supervisor()`

```
if (reg_h.wait_for([&]() { return reg_h->shutdown; }, m_supervisor_state.tick)) {
            return;
 }
```
I tried this locally and it still passes your changed tests.

dorezyuk:
@djchhp how about moving then the super visor into the policy?

djchhp:
> @djchhp how about moving then the super visor into the policy?

That could also be done. Latency scaling is currently the only time based scaling policy. Once you add another one, you likely want to share that. I guess this means the policies would essentially become templates themselves or some form of inheritance is involved.

lib/everest/util/include/everest/util/async/thread_pool_scaling.hpp
dorezyuk: done

lib/everest/util/include/everest/util/async/thread_pool_scaling.hpp
dorezyuk: done

lib/everest/util/include/everest/util/async/thread_pool_scaling.hpp
dorezyuk: done

djchhp (approved):
That is looking good. The way you solved conditional compilation with C++17 features is very clear and compact.

djchhp:
@Pietfried we need your review for the bazel build

Pietfried (approved):

PR #2103, opened 2026-04-22
Ocpp201: Convert the ResetType from Ocpp to EVerest

dorezyuk:
## Describe your changes

forwards the real ocpp reset type to the everest interface - has no real effect on the current system module -> the system module always triggers the same action for now

### DIFF
--- a/modules/EVSE/OCPP201/conversions.cpp
+++ b/modules/EVSE/OCPP201/conversions.cpp
@@ -1820,5 +1820,17 @@ std::vector<types::iso15118::EnergyTransferMode> to_everest_allowed_energy_trans
     return value;
 }
 
+types::system::ResetType to_everest_system_reset_type_enum(const ocpp::v2::ResetEnum& from) {
+    switch (from) {
+    case ocpp::v2::ResetEnum::OnIdle:
+        return types::system::ResetType::Soft;
+    case ocpp::v2::ResetEnum::Immediate:
+        [[fallthrough]];
+    case ocpp::v2::ResetEnum::ImmediateAndResume:
+        return types::system::ResetType::Hard;
+    }
+    return types::system::ResetType::NotSpecified;
+}
+
 } // namespace conversions
 } // namespace module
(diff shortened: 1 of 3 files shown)


mlitre (changes requested):
I think we should pause here and discuss how to extend the system reset types. I think we should take this time to actually update the functionality to have better support for OCPP 2.X reset functionality.

dorezyuk:
@mlitre  - can you elaborate what you have in mind?

modules/EVSE/OCPP201/conversions.cpp
```diff
@@ -1820,5 +1820,17 @@ std::vector<types::iso15118::EnergyTransferMode> to_everest_allowed_energy_trans
...
+types::system::ResetType to_everest_system_reset_type_enum(const ocpp::v2::ResetEnum& from) {
+    switch (from) {
+    case ocpp::v2::ResetEnum::OnIdle:
+        return types::system::ResetType::Soft;
+    case ocpp::v2::ResetEnum::Immediate:
+        [[fallthrough]];
+    case ocpp::v2::ResetEnum::ImmediateAndResume:
+        return types::system::ResetType::Hard;
+    }
+    return types::system::ResetType::NotSpecified;
```
Pietfried, line 1832: This conversion looks confusing. I think what @mlitre meant is that we should extend the internal EVerest `ResetType` to cover these additional requirements. The OCPP stack itself also contains internal processing though, e.g. if it sees `OnIdle` it will only schedule the actual reset command in EVerest. It's therefore also a question of responsibilities.

The System module implementations that we have in EVerest currently do not have any control over the charging process or state machine.

mlitre: @dorezyuk this is my point exactly, I think we should extend the everest reset type as well as review the interface to accept evseIds to properly support (if decided by the implementer of the system module) Evse specific resets. 
I agree with you Piet that afterwards we need to decide on a who  is responsible of what for implementing the decisions regarding reset. I lean towards keeping that in OCPP.

dorezyuk:
closed since the soft/hard is actually used by ocpp1.6

PR #2107, opened 2026-04-22
Rust: Break cyclic ownerships

dorezyuk:
## Describe your changes

Currently we have a cycling ownership of the runtime, where the ModuleInner would own the publishers and the publishers would own the runtime and the runtime would own the ModuleInner. This prevented us from dropping the runtime and the ModuleInner. 

This would prevent mockall test assertion evaluate their checkpoints. 

We break the cycle by using weak in the Publishers code

### DIFF
--- a/lib/everest/framework/everestrs/tests/modules/RsExample/rs_tests/mocked_test.rs
+++ b/lib/everest/framework/everestrs/tests/modules/RsExample/rs_tests/mocked_test.rs
@@ -60,7 +60,7 @@ mod some_module {
 
     #[everestrs::test(config = "config_probe.yaml", module = "example_1")]
     #[should_panic]
-    fn test_mocked_with_panic(module: &Module) {
+    fn test_mocked_with_panic_handler(module: &Module) {
         let mock_service = Arc::new(MockExampleServiceSubscriber::new());
 
         let (tx, rx) = std::sync::mpsc::channel();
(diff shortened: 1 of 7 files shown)


lib/everest/framework/everestrs/tests/modules/RsExample/rs_tests/mocked_test.rs
```diff
@@ -83,4 +83,22 @@ mod some_module {
...
+
+    #[everestrs::test(config = "config_probe.yaml", module = "example_1")]
+    #[should_panic]
+    fn test_mocked_with_panic_mocks(module: &Module) {
+        let mock_service = Arc::new(MockExampleServiceSubscriber::new());
+
+        let mut mock_client = MockExampleClientSubscriber::new();
+        mock_client
+            .expect_on_max_current()
+            .times(2..) // No one will call us twice.
```
dorezyuk, line 95: the test validating we fixed the bug

lib/everest/framework/include/utils/mqtt_abstraction_impl.hpp
```diff
@@ -97,6 +96,9 @@ class MQTTAbstractionImpl : public MQTTAbstraction {
     everest::lib::io::event::event_fd disconnect_event;
     everest::lib::io::event::fd_event_handler ev_handler;
 
+    // This must be destroyed first.
+    Thread mqtt_mainloop_thread;
```
dorezyuk, line 100: the thread must join first (in its destructor) before the ev_handler and similar gets destructed

hikinggrass (approved):

PR #2112, opened 2026-04-23
Jc/ev cli codegen changes

james-ctc:


applications/utils/ev-dev-tools/src/ev_cli/ev.py
```diff
@@ -184,7 +184,23 @@ def generate_tmpl_data_for_module(module, module_def):
     module_config = []
```
Pietfried, line 184: This should also be fine
```python
module_config = []
module_rwconfig = []

for conf_id, conf_info in module_def.get('config', {}).items():
    type_info = helpers.build_type_info(conf_id, conf_info['type'])
    if conf_info.get('mutability') == "ReadWrite":
        module_rwconfig.append(type_info)
    else:
        module_config.append(type_info)
```

lib/everest/framework/lib/mqtt_config_service.cpp
```diff
@@ -497,6 +497,61 @@ ConfigServiceClient::get_config_value(const everest::config::ConfigurationParame
...
+    SetResponse set_response;
+
+    try {
+        SetRequest set_request = data;
+        const auto& name = set_request.identifier.configuration_parameter_name;
+
+        if (const auto it = change_callbacks.find(name); it != change_callbacks.end()) {
+            // keep the value as a string. The module code applies the string to
+            // underlying type conversion (in the autogenerated code)
+            const auto result = it->second(set_request.value);
```
Pietfried, line 533: nit: this comment would have helped me
```suggestion
            // this exectues the actual on_ handler in the module
            const auto result = it->second(set_request.value);
```

lib/everest/framework/include/utils/mqtt_config_service.hpp
```diff
@@ -142,10 +163,16 @@ class ConfigServiceClient {
...
     GetConfigResult get_config_value(const everest::config::ConfigurationParameterIdentifier& identifier);
 
+    void register_config_change_handler(const std::string_view name, ConfigChangeHandler handler);
+
 private:
     std::shared_ptr<MQTTAbstraction> mqtt_abstraction;
     std::string origin;
     std::unordered_map<std::string, std::string> module_names;
+    std::map<std::string, ConfigChangeHandler> change_callbacks;
+    std::shared_ptr<TypedHandler> change_token;
```
Pietfried: Looks like we don't need `change_token` as a class member here

modules/EVSE/EvseManager/EvseManager.hpp
Pietfried, line 1: Are these changes to the EvseManager intentional with an empty RwConf ?

lib/everest/framework/lib/mqtt_config_service.cpp
```diff
@@ -497,6 +497,61 @@ ConfigServiceClient::get_config_value(const everest::config::ConfigurationParame
     return result;
 }
 
+void ConfigServiceClient::register_config_change_handler(const std::string_view name, ConfigChangeHandler handler) {
+    if (change_callbacks.empty()) {
+        // subscribe to the MQTT topic
+        const auto topic =
+            fmt::format("{}modules/{}/config/set_request", mqtt_abstraction->get_everest_prefix(), origin);
+        const auto handler = [this](const std::string& /*topic*/, const nlohmann::json& data) {
```
Pietfried: This variable name is shadowing the function parameter

lib/everest/framework/include/utils/types.hpp
```diff
@@ -90,6 +91,8 @@ enum class MqttMessageType {
     ClearError,            ///< Clear error message
     ConfigurationRequest,  ///< Configuration request (get or set)
     ConfigurationResponse, ///< Configuration response
+    SetConfigRequest,      ///< Set Configuration request
+    SetConfigResponse,     ///< Set Configuration response
```
Pietfried: Afaics these are not needed, `ConfigurationRequest` and `ConfigurationResponse` can be used instead.

lib/everest/framework/include/utils/types.hpp
```diff
@@ -39,6 +39,7 @@ using StringPairHandler = std::function<void(const std::string& topic, const std
 enum class HandlerType {
     Call,
     Result,
+    RuntimeConfigSet,
```
Pietfried: Not needed, `ConfigurationRequest` can be used.

lib/everest/framework/lib/mqtt_config_service.cpp
```diff
@@ -497,6 +497,61 @@ ConfigServiceClient::get_config_value(const everest::config::ConfigurationParame
...
+            mqtt_set_request(data);
+        };
+
+        change_token =
+            std::make_shared<TypedHandler>(HandlerType::RuntimeConfigSet, std::make_shared<Handler>(handler));
+
+        mqtt_abstraction->register_handler(topic, change_token, QOS::QOS2);
```
Pietfried, line 512: The message_handler.cpp is missing handling for `RuntimeConfigSet`, so this will result in `Unknown handler type for topic: ` . `HandlerType::ConfigurationRequest` shall be used instead.

Pietfried (changes requested):
The biggest issue I see is that the message handler of the target module is not set up to process the request because the mqtt_config_service.cpp uses `HandlerType::RuntimeConfigSet`.

modules/EVSE/EvseManager/EvseManager.hpp
james-ctc: it is what happens when you run `ev-cli` to update a module.
To keep the review simpler I haven't re-run `ev-cli` on all modules yet.
When you are happy with the review I'll update all modules so they are consistent

lib/everest/framework/lib/mqtt_config_service.cpp
james-ctc: topic is not a function parameter and not part of the lambda capture.
will update to avoid confusion though

Pietfried: Sorry, I was referring to `handler`

lib/everest/framework/include/utils/types.hpp
james-ctc: That was from the Framework Configuration in google drive. I'll remove it

Pietfried:
The lint fails, other than that lgtm!

james-ctc:
> The lint fails, other than that lgtm!

lint is failing on a line that doesn't exist - the CI seems to be broken

james-ctc:
The code has been manually merged to the feature/framework-configuration and the code will be revisited before integrating with main

PR #2113, opened 2026-04-23
fix(EnergyManager): Race Condition with energy_flow_request

Pietfried:
## Describe your changes

Eliminate data race and unnecessary copy in optimizer

Acquire energy_mutex before globals.init() so the entire run_optimizer is protected against concurrent on_energy_flow_request writes. Pass EnergyFlowRequest as const& throughout (run_optimizer, Market constructor and member) to avoid copying a potentially large object on every iteration.

### DIFF
--- a/modules/EnergyManagement/EnergyManager/EnergyManagerImpl.cpp
+++ b/modules/EnergyManagement/EnergyManager/EnergyManagerImpl.cpp
@@ -95,13 +95,13 @@ void EnergyManagerImpl::on_energy_flow_request(const types::energy::EnergyFlowRe
     }
 }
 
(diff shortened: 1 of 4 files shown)


SebaLukas (approved):

barsnick:
Could this have been the cause for spurious "power budget expired" phases under high load (but also sometimes when idle and not charging)? We have been chasing those, and perhaps this is a remedy.

Pietfried:
> Could this have been the cause for spurious "power budget expired" phases under high load (but also sometimes when idle and not charging)? We have been chasing those, and perhaps this is a remedy.

I acutally don't think that this is related. This fix was targeting a data race that I noticed a couple of times in e2e / integration tests at startup. The power budget expired log is likely just a consequence of too many MQTT messages queueing up in the EvseManager module.

barsnick:
Okay, thanks.

PR #2122, opened 2026-04-27
Add ChargeBridge systemd service

corneliusclaussen:


mhei (changes requested):
I'd prefer to bundle the chargebridge related files into its own package and also not to modify the existing `everest.service` file but to install a systemd drop-in in `/lib/systemd/system/everest.service.d/10-chargebridge.conf` with:
```
[Unit]
After=chargebridge.service
```
This way users can select or exclude this package as desired.

barsnick:
> I'd prefer to bundle the chargebridge related files into its own package

I had the same thought.

I could also imagine guarding this behind a `PACKAGECONFIG[chargebridge]`.

I'm not sure what to do about the `After=mosquitto.service chargebridge.service` though.

mhei:
> I'm not sure what to do about the `After=mosquitto.service chargebridge.service` though.

I think this could be nicely put into an extra file, the mentioned drop-in, which is packaged only in the chargebridge related package. Then it is installed in parallel to the existing stuff included in everest-core package and works seamlessly for systemd.

corneliusclaussen:
At the moment, the everest-core recipe builds and installs the chargebridge tool as well as the configs. So moving out only the chargebridge.service into a separate recipe does not help a lot. The cmake is not really set up for fully separate build. I think for now the best is the following: modify the everest-core.bb to split the output into separate rpm packages (and move chargebridge out of the everest.service, good idea with the drop-in!). Then the chargebridge.rpm would contain:
```
.
├── etc
│   └── chargebridge
│       ├── config-CB-EVAL-EV.yaml-example
│       ├── config-CB-EVAL-SIM.yaml-example
│       ├── config-CB-EVAL.yaml-example
│       ├── config-CB-SAT-AC.yaml-example
│       └── firmware
│           └── charge-bridge-fw_complete.cbfw
└── usr
    ├── bin
    │   └── pionix_chargebridge
    └── lib
        └── systemd
            ├── system
            │   ├── chargebridge.service
            │   └── everest.service.d
            │       └── 10-chargebridge.conf
            └── system-preset
                └── 98-chargebridge.preset
```

the everest.rpm would no longer contain the chargebridge files.

corneliusclaussen:
Implemented the changes as described above, pls review again.

mhei (approved):
That's exactly how I imagined it, thnx.
Not tested, but LGTM.

barsnick (approved):

PR #2124, opened 2026-04-27
[OCPP 2.0.1] missing EVerest ocpp interface implementation

james-ctc:
## Describe your changes

feat: EVerest OCPP interface implementations for OCPP201 module
- handle_stop
- handle_restart
- handle_change_availability

### DIFF
--- a/lib/everest/ocpp/lib/ocpp/v2/charge_point.cpp
+++ b/lib/everest/ocpp/lib/ocpp/v2/charge_point.cpp
@@ -371,6 +371,13 @@ bool ChargePoint::on_charging_state_changed(const std::uint32_t evse_id, const C
     return true;
 }
 
+ChangeAvailabilityResponse ChargePoint::on_change_availability(const ChangeAvailabilityRequest& request) {
+    bool transaction_active;
+    const auto response = availability->change_availability_req(transaction_active, request);
+    availability->action_change_availability_req(transaction_active, request, response);
+    return response;
+}
+
 std::optional<std::string> ChargePoint::get_evse_transaction_id(std::int32_t evse_id) {
     const auto& tx = this->evse_manager->get_evse(evse_id).get_transaction();
 
@@ -1231,6 +1238,14 @@ std::map<SetVariableData, SetVariableResult>
 ChargePoint::set_variables(const std::vector<SetVariableData>& set_variable_data_vector, const std::string& source) {
     // set variables and allow setting of ReadOnly variables
     return this->provisioning->set_variables(set_variable_data_vector, source);
+}
+
+void ChargePoint::register_variable_listener(
+    std::function<void(const std::unordered_map<std::int64_t, VariableMonitoringMeta>& monitors,
+                       const Component& component, const Variable& variable,
+                       const VariableCharacteristics& characteristics, const VariableAttribute& attribute,
+                       const std::string& value_previous, const std::string& value_current)>&& listener) {
+    device_model->register_variable_listener(std::move(listener));
 }
 
 GetCompositeScheduleResponse ChargePoint::get_composite_schedule(const GetCompositeScheduleRequest& request) {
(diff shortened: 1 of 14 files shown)


lib/everest/ocpp/lib/ocpp/v2/charge_point.cpp
```diff
@@ -371,6 +371,13 @@ bool ChargePoint::on_charging_state_changed(const std::uint32_t evse_id, const C
     return true;
 }
 
+ChangeAvailabilityResponse ChargePoint::on_change_availability(const ChangeAvailabilityRequest& request) {
+    bool transaction_active;
```
Pietfried, line 375: Same like above
```suggestion
    bool transaction_active{false};
```

lib/everest/ocpp/lib/ocpp/v2/functional_blocks/availability.cpp
```diff
@@ -94,79 +96,95 @@ void Availability::stop_heartbeat_timer() {
...
+                    }
                 }
             }
         }
     }
 }
 
+void Availability::handle_change_availability_req(Call<ChangeAvailabilityRequest> call) {
+    const auto request = call.msg;
+    bool transaction_active;
```
Pietfried, line 178: if `change_availability_req` returns early, this is uninitialized.
```suggestion
    bool transaction_active{false};
```

modules/EVSE/OCPP201/ocpp_generic/ocppImpl.hpp
```diff
@@ -46,6 +51,9 @@ class ocppImpl : public ocppImplBase {
 
     // ev@d2d1847a-7b88-41dd-ad07-92785f06f5c4:v1
     // insert your protected definitions here
+    MonitorList monitor_list;
```
Pietfried, line 54: `monitor_list` can be accessed from different threads:
- `handle_variable_changed` cmd handler
- from within an OCPP processing thread via that triggers `variable_changed`

Use e.g. `everest::lib::util::monitor` to protect it

modules/EVSE/OCPP201/ocpp_generic/ocppImpl.cpp
```diff
@@ -85,12 +116,73 @@ ocppImpl::handle_set_variables(std::vector<types::ocpp::SetVariableRequest>& req
...
+    using ChangeAvailabilityStatusEnum = ocpp::v2::ChangeAvailabilityStatusEnum;
+    ocpp::v2::ChangeAvailabilityResponse result;
+    result.status = ChangeAvailabilityStatusEnum::Rejected;
+
+    if (mod->charge_point == nullptr) {
+        EVLOG_warning << "ChargePoint not initialized, cannot handle change availability command";
+    } else {
+        bool process{true};
+        const auto ocpp_request = conversions::to_ocpp_change_availability_request(request);
+        if (request.evse.has_value()) {
```
Pietfried: While OCPP1.6 only defines two tiers (whole charging station + EVSE/Connector (in OCPP1.6 there is no difference between that)) OCPP2.x specifically defines three tiers (charging station, EVSE and Connector) so a request with an EVSE without a connector_id is valid and can be processed.

Pietfried (changes requested):

lib/everest/ocpp/lib/ocpp/v2/functional_blocks/availability.cpp
james-ctc: Correct, however on early return it is not used.
The only early return sets `response.status = ChangeAvailabilityStatusEnum::Rejected;`
`action_change_availability_req()` does nothing when 
`response.status == ChangeAvailabilityStatusEnum::Rejected`

lib/everest/ocpp/lib/ocpp/v2/charge_point.cpp
james-ctc: It isn't used on early return from `change_availability_req()`
Updated to give a default in `change_availability_req()`

Pietfried (approved):

PR #2128, opened 2026-04-27
feat: Add OCPP 1.6 config migration

Pietfried:
## Summary

Add configurable OCPP 1.6 storage backends and JSON-to-device-model migration

OCPP 1.6 always used a JSON file for configuration while OCPP 2.x used the device model SQLite database. This PR consolidates them by introducing configurable storage backends for the OCPP module and a one-time migration path.

## What changed

- Introduce a ConfigBackend string config key (`legacy` / `device_model` / `device_model_with_migration`). The existing default behavior is preserved: ConfigBackend defaults to `legacy`.
- Extract charge point configuration factory from OCPP.cpp into its own translation unit `charge_point_config_factory.cpp`.
- Move network connection keys (CentralSystemURI, ChargePointId, SecurityProfile, AuthorizationKey, HostName) from OCPP16LegacyCtrlr/InternalCtrlr into the standardized NetworkConfiguration slot component.
- Add check_integrity() to ChargePointConfigurationInterface, implemented in ChargePointConfigurationDeviceModel: verifies all required Core/Internal keys are present and NumberOfConnectors matches the connected EVSE count. Profile-conditional keys (SmartCharging, LocalAuthListManagement, PnC, CostAndPrice) are checked per SupportedFeatureProfiles; missing ones warn and strip the profile instead of failing.
- Auto-inject the NetworkConfiguration slot from the embedded default schema when absent, propagate value_source through initialize_database into the DB to prevent CSMS-set values from being overwritten on re-migration, fix ChargingScheduleAllowedChargingRateUnit notation conversion (Current/Power ↔ A/W).
- Add unit tests for value_source propagation, rate unit conversion, and integrity checks
- Wire device_model_with_migration into the OCPP 1.6 e2e test setup.
- Add how-to guide (ocpp-storage-migration.rst) and update the v16-to-v2 key mapping reference.

## Why

The OCPP and OCPP201 modules are planned to be merged into a single OCPP module. Migrating OCPP 1.6 configuration into the device model now means the same database can be reused directly when switching to OCPP201 or the future combined module without a further migration step.


lib/everest/ocpp/lib/ocpp/v16/charge_point_configuration.cpp
```diff
@@ -2444,6 +2444,24 @@ std::optional<std::string> ChargePointConfiguration::getDefaultPriceText(const s
     return std::nullopt;
 }
 
+std::optional<json> ChargePointConfiguration::getDefaultPriceText() {
```
james-ctc, line 2447: this could be simplifed:
```
std::optional<json> result;
try {
    auto default_price_texts = config["CostAndPrice"]["DefaultPriceText"];
    if (default_price_texts.is_object()) {
        result = default_price_texts;
    }
} catch(...) {
}
return result;
```

lib/everest/ocpp/lib/ocpp/v16/charge_point_configuration.cpp
```diff
@@ -2929,6 +2947,39 @@ std::optional<std::vector<KeyValue>> ChargePointConfiguration::getAllMeterPublic
     return key_values;
 }
 
+std::optional<std::string> ChargePointConfiguration::getMeterPublicKeysCsl() {
```
james-ctc, line 2944: try/catch would be simpler: see ChargePointConfiguration::getDefaultPriceText()

lib/everest/ocpp/lib/ocpp/v16/charge_point_configuration.cpp
```diff
@@ -2929,6 +2947,39 @@ std::optional<std::vector<KeyValue>> ChargePointConfiguration::getAllMeterPublic
...
+
+    std::vector<std::string> keys;
+    for (const auto& key : meter_public_keys) {
+        keys.push_back(key.get<std::string>());
+    }
+
+    return utils::to_csl(keys);
+}
+
+std::optional<json> ChargePointConfiguration::getMeterPublicKeys() {
```
james-ctc, line 2960: try/catch would be simpler: see ChargePointConfiguration::getDefaultPriceText()

lib/everest/ocpp/lib/ocpp/v16/charge_point_configuration_devicemodel.cpp
james-ctc, line 1: it would be nice to keep to the original style and not have multiple return statements.
But since the coding standards haven't been approved - I guess it is down to personal taste.

lib/everest/ocpp/lib/ocpp/v2/ocpp16_component_config_patcher.cpp
```diff
@@ -0,0 +1,276 @@
...
+
+bool is_same_variable(const DeviceModelVariable& v1, const DeviceModelVariable& v2) {
+    return ((v1.name == v2.name) && (v1.instance == v2.instance));
+}
+
+DeviceModelVariable* get_variable_to_patch(std::map<ComponentKey, std::vector<DeviceModelVariable>>& component_configs,
+                                           const Component& component, const Variable& variable) {
+
+    DeviceModelVariable variable_to_patch;
+    variable_to_patch.name = variable.name.get();
```
james-ctc, line 60: observation - `variable.name.get();` and `variable.instance->get();` look similar but one isn't optional. It would be nice if CiString didn't require get().

lib/everest/ocpp/lib/ocpp/v2/ocpp16_component_config_patcher.cpp
```diff
@@ -0,0 +1,276 @@
...
+
+namespace ocpp::v2 {
+
+namespace {
+
+bool is_same_variable(const DeviceModelVariable& v1, const DeviceModelVariable& v2) {
+    return ((v1.name == v2.name) && (v1.instance == v2.instance));
+}
+
+DeviceModelVariable* get_variable_to_patch(std::map<ComponentKey, std::vector<DeviceModelVariable>>& component_configs,
```
james-ctc: Rather than return a pointer, an iterator could be returned having the value component_configs.end() on error/not found.
C++ documentation on when iterators are invalidated gives more information than a pointer on its own.

lib/everest/ocpp/lib/ocpp/v2/ocpp16_component_config_patcher.cpp
```diff
@@ -0,0 +1,276 @@
...
+    if (readonly_override.has_value()) {
+        attribute_it->variable_attribute.mutability =
+            readonly_override.value() ? MutabilityEnum::ReadOnly : MutabilityEnum::ReadWrite;
+    }
+
+    attribute_it->value_source = "OCPP16Config";
+    attribute_it->variable_attribute.value = value;
+}
+
+void patch_variable_max_limit(std::map<ComponentKey, std::vector<DeviceModelVariable>>& component_configs,
```
james-ctc, line 150: if an iterator was used this could simplify to
```
void patch_variable_max_limit(std::map<ComponentKey, std::vector<DeviceModelVariable>>& component_configs,
                              const ocpp::v16::keys::DeviceModel_CV& device_model_cv, const int32_t max_limit) {

    if (auto it = get_variable_to_patch(component_configs, device_model_cv); it != component_configs.end()) {
        it->characteristics.maxLimit = static_cast<float>(max_limit);
    }
}
```

lib/everest/ocpp/lib/ocpp/v2/ocpp16_custom_config_mappings.cpp
```diff
@@ -0,0 +1,222 @@
...
+
+#include <nlohmann/json-schema.hpp>
+#include <ryml.hpp>
+#include <ryml_std.hpp>
+
+namespace ocpp::v2 {
+
+namespace {
+
+void yaml_error_handler(const char* msg, std::size_t len, ryml::Location loc, void*) {
```
james-ctc, line 17: it would be nice to avoid having another yaml loading approach.

lib/everest/ocpp/lib/ocpp/v2/ocpp16_custom_config_mappings.cpp
```diff
@@ -0,0 +1,222 @@
...
+    std::ifstream ifs(path);
+    const std::string content((std::istreambuf_iterator<char>(ifs)), (std::istreambuf_iterator<char>()));
+
+    const auto tree = ryml::parse_in_arena(ryml::to_csubstr(content));
+    return ryml_to_nlohmann_json(tree.rootref());
+}
+
+} // namespace
+
+nlohmann::json get_ocpp16_custom_mapping_schema() {
```
james-ctc, line 104: Perhaps load a file instead?

lib/everest/ocpp/lib/ocpp/v2/ocpp16_custom_config_mappings.cpp
```diff
@@ -0,0 +1,222 @@
...
+        }
+      }
+    }
+  }
+}
+)json");
+}
+
+Ocpp16CustomConfigMappings
+load_ocpp16_custom_config_mappings_from_yaml(const std::filesystem::path& mapping_file_path) {
```
james-ctc, line 175: I'm not sure how likely this is but json.at() can throw exceptions.

modules/EVSE/OCPP/manifest.yaml
```diff
@@ -19,6 +19,45 @@ config:
       about the database and its structure.
     type: string
     default: /tmp/ocpp_1_6_charge_point
+  DeviceModelDatabasePath:
```
james-ctc, line 22: this seems to be path + filename rather than just path.
`DeviceModelDatabase` would be a better name

modules/EVSE/OCPP/manifest.yaml
```diff
@@ -19,6 +19,45 @@ config:
...
+    type: string
+    default: device_model_migrations
+  DeviceModelConfigPath:
+    description: >-
+      Path to the OCPP device model component config directory used as the migration baseline before
+      applying the OCPP1.6 JSON configuration.
+      Relative paths are resolved against this module's share path.
+    type: string
+    default: component_config
+  DeviceModelConfigMappingsPath:
```
james-ctc: `DeviceModelConfigMappings` since it is referring to a file not directory

modules/EVSE/OCPP/OCPP.cpp
```diff
@@ -133,6 +141,184 @@ void create_empty_user_config(const fs::path& user_config_path) {
...
+        context.custom_mappings_path = custom_mappings_path.value();
+    }
+
+    return context;
+}
+
+bool should_run_ocpp16_to_device_model_migration(const fs::path& database_path) {
+    const auto device_model_database_exists = fs::exists(database_path);
+
+    if (!device_model_database_exists) {
```
james-ctc:
```
if (!fs::exists(database_path)) {
```

lib/everest/ocpp/lib/ocpp/v2/ocpp16_custom_config_mappings.cpp
Pietfried: I agree. I would suggest to follow up with a PR that adds the yaml tooling into lib/everest/ and then uses it at the corresponding parts.

lib/everest/ocpp/lib/ocpp/v2/ocpp16_custom_config_mappings.cpp
Pietfried: The previous schema validation ensures that the at() access is safe

james-ctc: A comment to explain that would be a good addition

modules/EVSE/OCPP/manifest.yaml
Pietfried: I agree that your suggestion is the better choice but I'd like to keep it because OCPP201 uses the same parameter and that makes it a bit easier to switch between OCPP1.6 and OCPP2.x when you just need to adjust the module name.

lib/everest/ocpp/lib/ocpp/v2/ocpp16_component_config_patcher.cpp
Pietfried: Replaced the raw pointer return with `std::optional<std::reference_wrapper<DeviceModelVariable>>`

lib/everest/ocpp/lib/ocpp/v2/ocpp16_component_config_patcher.cpp
Pietfried: This would only work with a single-level container lookup. Here we have two levels (map + vector) which is why I went with `std::optional<std::reference_wrapper<DeviceModelVariable>>`

james-ctc (approved):

lib/everest/ocpp/lib/ocpp/v2/ocpp16_custom_config_mappings.cpp
hikinggrass: since #2156 is now (on its way to be) merged, do you want to revisit this in this PR or in a follow-up?

hikinggrass (approved):

lib/everest/ocpp/lib/ocpp/v2/ocpp16_custom_config_mappings.cpp
Pietfried: I'll open up a follow up PR

lib/everest/ocpp/lib/ocpp/v16/charge_point_configuration_devicemodel.cpp
```diff
@@ -126,6 +120,23 @@ void raise_not_found(const std::string_view& section, const std::string_view& na
...
 
+bool has_parameter_value(DeviceModelInterface& storage, v16::keys::valid_keys key) {
+    const auto cv = v16::keys::convert_v2(key);
+    if (!cv) {
+        return false;
+    }
+    std::string value;
+    return storage.get_variable(cv->first, cv->second, v2::AttributeEnum::Actual, value) ==
+           v2::GetVariableStatusEnum::Accepted;
+}
```
james-ctc: I still would like the existing pattern to be used.
I've given up ob the coding standards ever being approved, but even so consistent approach in the file would be nice

james-ctc (changes requested):

james-ctc (dismissed):

lib/everest/ocpp/config/common/component_config/standardized/InternalCtrlr.json
```diff
@@ -873,13 +837,11 @@
   },
   "required": [
     "ChargeBoxSerialNumber",
```
mlitre, line 839: If we are removing fields, could we take a look at this serial number? Is it actually used because we have ChargeBox and ChargePointSerialNumber which is pretty confusing

lib/everest/ocpp/config/common/component_config/standardized/NetworkConfiguration_2.json
```diff
@@ -100,8 +100,7 @@
         }
       ],
       "description": "Per-slot identity override",
-      "type": "string",
-      "default": ""
```
mlitre, line 104: We should keep the default to blank as that indicates that we should fall back to SecurityCtrlr.Identity

lib/everest/ocpp/config/common/component_config/standardized/NetworkConfiguration_1.json
```diff
@@ -444,6 +444,21 @@
       ],
       "description": "Serial number of the CSMS root certificate",
       "type": "string"
+    },
+    "HostName": {
```
mlitre, line 448: I don't really see the use of making this configurable per connection, but maybe I am missing something. For me this should go into InternalCtrlr as it is custom

mlitre (commented):
Generally looks good a few comments that are worth a small discussion.
I will take a second look afterwards since the PR is quite large

lib/everest/ocpp/config/common/component_config/standardized/InternalCtrlr.json
james-ctc: `ChargeBoxSerialNumber` is in the OCPP 1.6 spec and should not be removed.
It is part of `BootNotification.req` §6.3

> Optional. This contains a value that identifies the serial number of
the Charge Box inside the Charge Point. Deprecated, will be
removed in future version

And is removed in 2.x

mlitre: Then let's remove ChargePointSerialNumber? I just don't understand why we have duplication. Let's just have one and each ocpp version pulls the right value. In the 1.6 case it is an "internal" variable anyways nothing strongly defined in spec to be called what it is. It's up to the implementer to have a way to pass the info in BootNotification (same as v2.x)

james-ctc: `ChargePointSerialNumber` is in the OCPP 1.6 spec and should not be removed.
It is part of `BootNotification.req` §6.3

> Optional. This contains a value that identifies the serial number of
the Charge Point.

mlitre: So both are part of the spec? I don't get it

mlitre: Nevermind OCPP 1.6 defined some strange values

lib/everest/ocpp/config/common/component_config/standardized/NetworkConfiguration_1.json
Pietfried: The advantage of defining it in `NetworkConfiguration` is that it can always be migrated because `HostName` is always defined. If the migration is executed on a target with existing `InternalCtrlr.json` without the definition of `HostName`, this parameter is not be migrated.

mlitre: Ok let's keep as is then

mlitre (dismissed):

Pietfried:
@hikinggrass or @dorezyuk can you as CODEOWNERs have a look at the bazel part?

mlitre (approved):

lib/everest/ocpp/tests/lib/ocpp/v16/v2config/memory_storage.cpp
```diff
@@ -282,7 +290,29 @@ std::optional<MemoryStorage::Storage::iterator> MemoryStorage::locate_v16(const
...
+std::optional<std::string> MemoryStorage::get_meter_public_keys_v16() const {
+    std::vector<std::string> meter_public_keys;
+    for (std::uint32_t i = 1;; i++) {
+        const auto key = "MeterPublicKey" + std::to_string(i);
+        auto it = locate_v16(key);
+        if (!it.has_value()) {
+            break;
+        }
+        meter_public_keys.push_back(it.value()->second);
+    }
```
hikinggrass, line 302: shouldn't this use a range-base for-loop, that for+break construct is a bit unintuitive

modules/EVSE/OCPP/OCPP.cpp
```diff
@@ -2,17 +2,18 @@
 // Copyright 2020 - 2022 Pionix GmbH and Contributors to EVerest
```
hikinggrass, line 2:
```suggestion
// Copyright 2020 - 2026 Pionix GmbH and Contributors to EVerest
```

hikinggrass (approved):

PR #2138, opened 2026-04-29
Add libcbv2g_json_wrapper application + AppHandshake converter

biubiuzhang:
## Summary
- Adds `applications/cbv2g_json_wrapper`, a native C library wrapping libcbv2g with a JSON-based EXI encode/decode API, intended to replace Josev's Java-based EXIficient codec on the EV simulation side.
- This is the first of 3 stacked PRs. PR 1 establishes the public API + dispatcher and the smallest protocol (App Handshake / SAP) so the surface can be reviewed before the larger ISO 15118-2 converter lands. ISO 15118-20 support is planned as a follow-up PR.
- Bundles cJSON 1.7.19 (MIT) under `third_party/`. Adds `libcbv2g` (`v0.3.2`) as a CMake dependency in `dependencies.yaml`, gated on `EVEREST_DEPENDENCY_ENABLED_LIBCBV2G`. The wrapper's `add_subdirectory` is gated on the same option so the default everest-core build is unaffected when libcbv2g is not enabled.
- Includes string-handling hardening (snprintf-bounded copies, strnlen-bounded reads) carried over from review feedback on the original PR EVerest/ext-switchev-iso15118#57 (CWE-120, CWE-126).

## Test plan
- [ ] `cmake -S . -B build -DEVEREST_BUILD_APPLICATIONS=ON -DEVEREST_DEPENDENCY_ENABLED_LIBCBV2G=ON -DBUILD_CBV2G_JSON_WRAPPER_TESTS=ON` configures cleanly with libcbv2g installed.
- [ ] `ctest -R cbv2g_test_apphand` runs round-trip + negative tests.
- [ ] Verify `libcbv2g_json_wrapper.so` installs to `lib/`.

Refs RFC EVerest/ext-switchev-iso15118#58. Replaces and supersedes EVerest/ext-switchev-iso15118#57. Companion PR for the Python codec swap: PR5A_PLACEHOLDER.


applications/cbv2g_json_wrapper/CMakeLists.txt
```diff
@@ -0,0 +1,82 @@
+cmake_minimum_required(VERSION 3.14)
+project(cbv2g_json_wrapper VERSION 1.0.0 LANGUAGES C)
+
+# Set C standard
+set(CMAKE_C_STANDARD 99)
+set(CMAKE_C_STANDARD_REQUIRED ON)
```
SebaLukas: Please use instead of `set` `target_compile_features` like so: `target_compile_features(my_lib PRIVATE c_std_99)`

applications/cbv2g_json_wrapper/CMakeLists.txt
```diff
@@ -0,0 +1,82 @@
...
+
+# Build as shared library
+set(CMAKE_POSITION_INDEPENDENT_CODE ON)
```
SebaLukas: Use `target_compile_options(my_lib PRIVATE -fPIC)` or `set_property(TARGET my_lib PROPERTY POSITION_INDEPENDENT_CODE ON)`

applications/cbv2g_json_wrapper/CMakeLists.txt
```diff
@@ -0,0 +1,82 @@
...
+target_include_directories(cjson PUBLIC
+    ${CMAKE_CURRENT_SOURCE_DIR}/third_party/cJSON
+)
+
+# Wrapper sources. Per-protocol converters are added in subsequent PRs.
+set(WRAPPER_SOURCES
+    src/cbv2g_json_wrapper.c
+    src/json_utils.c
+    src/apphand_converter.c
+)
```
SebaLukas: Please use `target_sources()`

applications/cbv2g_json_wrapper/CMakeLists.txt
```diff
@@ -0,0 +1,82 @@
...
+    target_link_libraries(cbv2g_test_apphand
+        cbv2g_json_wrapper
+        cjson
+    )
+    target_include_directories(cbv2g_test_apphand PRIVATE
+        ${CMAKE_CURRENT_SOURCE_DIR}/include
+        ${CMAKE_CURRENT_SOURCE_DIR}/third_party/cJSON
+    )
+    add_test(NAME cbv2g_test_apphand COMMAND cbv2g_test_apphand)
+endif()
```
SebaLukas, line 99: Maybe it makes more sense to use `catch2` or `gtest` here?

applications/cbv2g_json_wrapper/README.md
```diff
@@ -0,0 +1,68 @@
...
+Additional protocol support (DIN 70121, ISO 15118-2) is added in
+subsequent PRs in this stack. ISO 15118-20 support is planned as a
+follow-up PR.
+
+## Build
+
+The wrapper is built as part of EVerest's `applications/` tree when
+`-DEVEREST_BUILD_APPLICATIONS=ON` is set (the default). It links against
+the in-tree `cbv2g::*` targets exported by `lib/everest/cbv2g`, so no
+external dependency setup is required.
```
SebaLukas, line 53: If I want to build only `cbv2g_json_wrapper` standalone as shared library (run `cmake` in `cbv2g_json_wrapper` folder), I guess `libcbv2g` will be not found. In this case it make sense to get `libcbv2g` locally like so: https://github.com/EVerest/EVerest/blob/main/lib/everest/iso15118/cmake/local-build.cmake

applications/cbv2g_json_wrapper/tests/test_apphand.c
```diff
@@ -0,0 +1,185 @@
...
+    g_passed++; \
+} while (0)
+
+#define EXPECT(cond) do { \
+    if (!(cond)) { \
+        fprintf(stderr, "  FAIL: %s:%d: %s\n", __FILE__, __LINE__, #cond); \
+        g_failed++; \
+        return; \
+    } \
+} while (0)
```
SebaLukas: I would appreciate if you use `catch2` or `gtest` for the tests. Both test frameworks should only be included if the tests are built and running

applications/cbv2g_json_wrapper/src/apphand_converter.c
```diff
@@ -0,0 +1,326 @@
...
+    struct appHand_exiDocument doc;
+    init_appHand_exiDocument(&doc);
+
+    /* Determine message type and convert */
+    msg = cJSON_GetObjectItemCaseSensitive(root, "supportedAppProtocolReq");
+    if (msg != NULL) {
+        doc.supportedAppProtocolReq_isUsed = 1;
+        result = json_to_apphand_req(msg, &doc.supportedAppProtocolReq);
+        if (result != CBV2G_SUCCESS) {
+            goto cleanup;
```
SebaLukas: Please dont use `goto`. For me its fine to return early with the correct error and deleting `root`.

applications/cbv2g_json_wrapper/CMakeLists.txt
SebaLukas: I am curious: Why c99?

applications/cbv2g_json_wrapper/src/apphand_converter.c
```diff
@@ -0,0 +1,326 @@
...
+        }
+    } else {
+        msg = cJSON_GetObjectItemCaseSensitive(root, "supportedAppProtocolRes");
```
SebaLukas: For better readibility consider this alternative:
```
if (cJSON_HasObjectItem(root, "supportedAppProtocolReq") == 1){
    msg = cJSON_GetObjectItemCaseSensitive(root, "supportedAppProtocolReq");
    [...]
} else if (cJSON_HasObjectItem(root, "supportedAppProtocolRes") == 1) {
```
If `CaseSensitive` is important, then I think it okay to call `cJSON_GetObjectItemCaseSensitive` twice (first to check the `key` and second to get the `cJson*`)

applications/cbv2g_json_wrapper/src/apphand_converter.c
```diff
@@ -0,0 +1,326 @@
...
+        set_error("Failed to create JSON object");
+        return CBV2G_ERROR_JSON_GENERATE;
+    }
+
+    if (doc.supportedAppProtocolReq_isUsed) {
+        cJSON* req_json = apphand_req_to_json(&doc.supportedAppProtocolReq);
+        if (req_json == NULL) {
+            set_error("Failed to convert supportedAppProtocolReq to JSON");
+            result = CBV2G_ERROR_JSON_GENERATE;
+            goto cleanup;
```
SebaLukas: See above my comment about `goto`

applications/cbv2g_json_wrapper/src/apphand_converter.c
```diff
@@ -0,0 +1,326 @@
...
+        /* ProtocolNamespace - snprintf with %.*s is bounded and always
+         * null-terminates (CWE-120 / CWE-126). */
+        const char* ns = json_get_string(item, "ProtocolNamespace");
+        size_t ns_len = strnlen(ns, appHand_ProtocolNamespace_CHARACTER_SIZE);
+        int ns_written = snprintf(proto->ProtocolNamespace.characters,
+                                  appHand_ProtocolNamespace_CHARACTER_SIZE,
+                                  "%.*s", (int)ns_len, ns);
+        if (ns_written < 0) ns_written = 0;
+        if ((size_t)ns_written >= appHand_ProtocolNamespace_CHARACTER_SIZE)
+            ns_written = appHand_ProtocolNamespace_CHARACTER_SIZE - 1;
```
SebaLukas: Even if they're single-line `ifs`, please use `{}`

applications/cbv2g_json_wrapper/src/apphand_converter.c
```diff
@@ -0,0 +1,326 @@
...
+        res->ResponseCode = appHand_responseCodeType_OK_SuccessfulNegotiation;
+    } else if (strncmp(response_code, "OK_SuccessfulNegotiationWithMinorDeviation",
+                       sizeof("OK_SuccessfulNegotiationWithMinorDeviation")) == 0) {
+        res->ResponseCode = appHand_responseCodeType_OK_SuccessfulNegotiationWithMinorDeviation;
+    } else if (strncmp(response_code, "Failed_NoNegotiation",
+                       sizeof("Failed_NoNegotiation")) == 0) {
+        res->ResponseCode = appHand_responseCodeType_Failed_NoNegotiation;
+    } else {
+        /* Try as integer */
+        res->ResponseCode = json_get_int(json, "ResponseCode");
```
SebaLukas: This is not correct. In this case an error should be returned

applications/cbv2g_json_wrapper/src/apphand_converter.c
```diff
@@ -0,0 +1,326 @@
...
+            response_code_str = "OK_SuccessfulNegotiation";
+            break;
+        case appHand_responseCodeType_OK_SuccessfulNegotiationWithMinorDeviation:
+            response_code_str = "OK_SuccessfulNegotiationWithMinorDeviation";
+            break;
+        case appHand_responseCodeType_Failed_NoNegotiation:
+            response_code_str = "Failed_NoNegotiation";
+            break;
+        default:
+            response_code_str = "Unknown";
```
SebaLukas: In this case an error should be returned

SebaLukas (changes requested):
Overall I like the PR 👍 
I added some comments

applications/cbv2g_json_wrapper/CMakeLists.txt
biubiuzhang: Done

applications/cbv2g_json_wrapper/CMakeLists.txt
biubiuzhang: Done

applications/cbv2g_json_wrapper/CMakeLists.txt
biubiuzhang: Done

applications/cbv2g_json_wrapper/CMakeLists.txt
biubiuzhang: Switched to gtest. The find_package only fires when BUILD_CBV2G_JSON_WRAPPER_TESTS is on, so it stays out of regular builds.

applications/cbv2g_json_wrapper/README.md
biubiuzhang: Good call — added `cmake/local-build.cmake` based on the iso15118 one. It tries `-DCBV2G_SOURCE_DIR=...` first, then the in-tree path, then `find_package(cbv2g)`. Updated the README with a Standalone section. Small heads up: had to whitelist `applications/*/cmake/*-build.cmake` in `.gitignore`, the `*build*` glob was eating the file.

applications/cbv2g_json_wrapper/tests/test_apphand.c
biubiuzhang: Done, ported to `test_apphand.cpp` using gtest.

applications/cbv2g_json_wrapper/src/apphand_converter.c
biubiuzhang: Done

applications/cbv2g_json_wrapper/CMakeLists.txt
biubiuzhang: Mostly to stay in sync with libcbv2g — its targets export `c_std_99` PUBLIC, so anything linking `cbv2g::din` already inherits that floor. The wrapper itself doesn't actually need C11 either; only `_Thread_local` touches it and that's already gated with a `__thread` fallback for GCC/Clang. Happy to bump it if you'd rather move the whole stack though.

applications/cbv2g_json_wrapper/src/apphand_converter.c
biubiuzhang: Went with your 'call CaseSensitive twice' alternative, just stored the result in two named pointers up front so we don't actually call it twice. Same effect, drops the nested else and the `msg` reuse. I avoided `HasObjectItem` since it's case-insensitive in cJSON — would happily match `SupportedAppProtocolReq` but then the case-sensitive get returns NULL right after, leading to a segfault. Hope this works for you.

applications/cbv2g_json_wrapper/src/apphand_converter.c
biubiuzhang: Done

applications/cbv2g_json_wrapper/src/apphand_converter.c
biubiuzhang: Done — went through `json_utils.c` too while I was at it.

applications/cbv2g_json_wrapper/src/apphand_converter.c
biubiuzhang: Yeah, good catch. The fallthrough was silently coercing unknown strings to 0 which is `OK_SuccessfulNegotiation` — definitely a bug. Now rejects unknown strings with `CBV2G_ERROR_JSON_PARSE`; numeric `ResponseCode` form still works. Added a regression test.

applications/cbv2g_json_wrapper/src/apphand_converter.c
biubiuzhang: Done — returns NULL with `set_error` on unknown decoded values instead of emitting `"Unknown"`.

biubiuzhang:
Thanks for the thorough review @SebaLukas! Just pushed an update addressing all the inline comments — modernized the CMake bits, ported the tests over to gtest, dropped the gotos in favour of early returns, and fixed the silent fallback on unknown ResponseCode values (added a regression test for that one). Also added a `cmake/local-build.cmake` for the standalone build path you mentioned. Let me know if anything still looks off or if you'd prefer a different approach on any of them.

applications/cbv2g_json_wrapper/CMakeLists.txt
SebaLukas: Ah yeah `libcbv2g` uses also c_std_99. Then its fine. Maybe in the future we can move to another version

applications/cbv2g_json_wrapper/src/apphand_converter.c
SebaLukas: Fine for me :+1:

SebaLukas (approved):
Looking good 👍

PR #2139, opened 2026-04-29
Add DIN 70121 converter to libcbv2g_json_wrapper

biubiuzhang:
## Summary
- Adds the DIN 70121 (`urn:din:70121:2012:MsgDef`) JSON/EXI converter on top of EVerest/everest-core#2138.
- Extends the dispatcher with `PROTOCOL_DIN` and adds `NS_DIN_MSG_DEF` to the public header.
- Same string-handling hardening as PR 1 (CWE-120 / CWE-126).

**Stacked on EVerest/everest-core#2138 — please review that one first.** Until PR 1 merges, the diff in this PR will show PR 1's commits as well; the new code in this PR is the single commit on top.

## Test plan
- [ ] `ctest -R cbv2g_test_din` runs SessionSetupReq/Res, ServiceDiscoveryReq/Res, ChargeParameterDiscoveryReq, and the RFC-cited CurrentDemandReq round-trip (strict `cJSON_Compare`).


biubiuzhang:
`Run OCPP Tests` failure looks like flake in `plug_and_charge_tests.py` — different test fails each rerun and PR 1 cleared the same suite on its prior run. Could you rerun the failed job? I don't have write access here.

SebaLukas:
It would be nice if you can test `cbv2g_json_wrapper` against more tests.
Maybe you can use my `fuzzy-exi-generator` tool to generate multiple json strings and exi streams: https://codeberg.org/sebalukas/fuzzy-exi-generator
For the exi generator I am using Josev's java `EXICodec.jar`. So if `cbv2g_json_wrapper` can encode and decode every fuzzy message then Josev should not have a problem using `cbv2g_json_wrapper`.

applications/cbv2g_json_wrapper/src/din_converter.c
```diff
@@ -0,0 +1,1291 @@
...
+    cJSON* msg = NULL;
+
+    #define HANDLE_MSG_ENCODE(name, field, init_func, conv_func) \
+        msg = cJSON_GetObjectItemCaseSensitive(body, #name); \
+        if (msg != NULL) { \
+            init_func(&doc.V2G_Message.Body.field); \
+            doc.V2G_Message.Body.field##_isUsed = 1; \
+            result = conv_func(msg, &doc.V2G_Message.Body.field); \
+            goto encode; \
+        }
```
SebaLukas: Instead of macro, can you please define a helper function and dont use `goto`

applications/cbv2g_json_wrapper/src/din_converter.c
```diff
@@ -0,0 +1,1291 @@
...
+    #define HANDLE_MSG_DECODE(name, field, conv_func) \
+        if (doc.V2G_Message.Body.field##_isUsed) { \
+            cJSON* msg_json = conv_func(&doc.V2G_Message.Body.field); \
+            if (msg_json == NULL) { \
+                set_error("Failed to convert " #name " to JSON"); \
+                result = CBV2G_ERROR_JSON_GENERATE; \
+                goto cleanup; \
+            } \
+            cJSON_AddItemToObject(body_json, #name, msg_json); \
+        }
```
SebaLukas: See above

applications/cbv2g_json_wrapper/src/din_converter.c
```diff
@@ -0,0 +1,1291 @@
...
+
+    const char* unit = json_get_string(json, "Unit");
+    if (strcmp(unit, "h") == 0) pv->Unit = din_unitSymbolType_h;
+    else if (strcmp(unit, "m") == 0) pv->Unit = din_unitSymbolType_m;
+    else if (strcmp(unit, "s") == 0) pv->Unit = din_unitSymbolType_s;
+    else if (strcmp(unit, "A") == 0) pv->Unit = din_unitSymbolType_A;
+    else if (strcmp(unit, "V") == 0) pv->Unit = din_unitSymbolType_V;
+    else if (strcmp(unit, "W") == 0) pv->Unit = din_unitSymbolType_W;
+    else if (strcmp(unit, "Wh") == 0) pv->Unit = din_unitSymbolType_Wh;
+    else pv->Unit = din_unitSymbolType_W; /* Default */
```
SebaLukas: Missing {}

applications/cbv2g_json_wrapper/src/din_converter.c
```diff
@@ -0,0 +1,1291 @@
...
+static din_responseCodeType din_string_to_response_code(const char* str) {
+    if (strcmp(str, "OK") == 0) return din_responseCodeType_OK;
+    if (strcmp(str, "OK_NewSessionEstablished") == 0) return din_responseCodeType_OK_NewSessionEstablished;
+    if (strcmp(str, "OK_OldSessionJoined") == 0) return din_responseCodeType_OK_OldSessionJoined;
+    if (strcmp(str, "OK_CertificateExpiresSoon") == 0) return din_responseCodeType_OK_CertificateExpiresSoon;
+    if (strcmp(str, "FAILED") == 0) return din_responseCodeType_FAILED;
+    if (strcmp(str, "FAILED_SequenceError") == 0) return din_responseCodeType_FAILED_SequenceError;
+    if (strcmp(str, "FAILED_ServiceIDInvalid") == 0) return din_responseCodeType_FAILED_ServiceIDInvalid;
+    if (strcmp(str, "FAILED_UnknownSession") == 0) return din_responseCodeType_FAILED_UnknownSession;
+    /* ... add more as needed */
```
SebaLukas: There are some ResponseCode missing

applications/cbv2g_json_wrapper/src/din_converter.c
```diff
@@ -0,0 +1,1291 @@
...
+        case din_responseCodeType_FAILED_ChallengeInvalid: return "FAILED_ChallengeInvalid";
+        case din_responseCodeType_FAILED_ContractCanceled: return "FAILED_ContractCanceled";
+        case din_responseCodeType_FAILED_WrongChargeParameter: return "FAILED_WrongChargeParameter";
+        case din_responseCodeType_FAILED_PowerDeliveryNotApplied: return "FAILED_PowerDeliveryNotApplied";
+        case din_responseCodeType_FAILED_TariffSelectionInvalid: return "FAILED_TariffSelectionInvalid";
+        case din_responseCodeType_FAILED_ChargingProfileInvalid: return "FAILED_ChargingProfileInvalid";
+        case din_responseCodeType_FAILED_EVSEPresentVoltageToLow: return "FAILED_EVSEPresentVoltageToLow";
+        case din_responseCodeType_FAILED_MeteringSignatureNotValid: return "FAILED_MeteringSignatureNotValid";
+        case din_responseCodeType_FAILED_WrongEnergyTransferType: return "FAILED_WrongEnergyTransferType";
+        default: return "FAILED";
```
SebaLukas: Should be an error

applications/cbv2g_json_wrapper/src/din_converter.c
biubiuzhang: Done — replaced the macro with an explicit `json_to_din_body()` helper that dispatches via if-chain (one block per message type), and dropped the `goto`. Same approach on the decode side.

applications/cbv2g_json_wrapper/src/din_converter.c
biubiuzhang: Done — extracted a `din_body_to_json()` helper plus a tiny `append_decoded_msg()` for the per-message NULL check. No more `goto cleanup`.

applications/cbv2g_json_wrapper/src/din_converter.c
biubiuzhang: Done — went through the whole file with clang-tidy's `readability-braces-around-statements`, so the unit-symbol if/else chain and ~50 other single-line ifs are now braced.

applications/cbv2g_json_wrapper/src/din_converter.c
biubiuzhang: Yeah, good catch. Same silent-fallback pattern as the apphand one we discussed — the default was returning `"FAILED"` and quietly hiding a corrupt EXI stream. Changed it to `set_error` + return NULL, and added a small `add_response_code_string()` helper so the 11 `*Res_to_json` callers can propagate the error cleanly.

applications/cbv2g_json_wrapper/src/din_converter.c
biubiuzhang: Done — filled in all 15 missing entries so the from-string table now matches the to-string switch. Added a regression test (`ExtendedResponseCodesRoundTrip`) that walks through the previously-missing codes and verifies each one round-trips back to its own name instead of falling through to FAILED.

biubiuzhang:
Thanks for the pointer to `fuzzy-exi-generator` — that's a really nice idea and I want to do it properly rather than hand-wave. Setting up the Java/EXICodec.jar harness alongside this wrapper is a meaningful chunk of work though (basically a small CI job that runs the wrapper against generated fuzz cases and diffs against EXICodec.jar's output), so I'd like to track it as a follow-up rather than bolt it onto this PR. I'll open an issue against `ext-switchev-iso15118` linking back to your tool once this stack lands. Sound okay?

biubiuzhang:
Thanks for the review @SebaLukas! Just pushed an update addressing all five inline comments — replaced both `HANDLE_MSG_*` macros + `goto`s with explicit dispatch helpers, fixed the silent fallback in `din_response_code_to_string`, filled in the missing 15 `din_string_to_response_code` entries (with a regression test for them), braced all the single-line ifs, and ported `test_din.c` over to GoogleTest while I was at it. Replied separately on the `fuzzy-exi-generator` suggestion.

SebaLukas:
> Thanks for the pointer to `fuzzy-exi-generator` — that's a really nice idea and I want to do it properly rather than hand-wave. Setting up the Java/EXICodec.jar harness alongside this wrapper is a meaningful chunk of work though (basically a small CI job that runs the wrapper against generated fuzz cases and diffs against EXICodec.jar's output), so I'd like to track it as a follow-up rather than bolt it onto this PR. I'll open an issue against `ext-switchev-iso15118` linking back to your tool once this stack lands. Sound okay?

I think thats okay. With my fuzzy generator you can basically generate infinite json strings and exi streams saved in csv files. A CI job isn't necessary to begin with. A local test that loads the CSV test data and runs it against the generated code is enough for me.
Before we use this wrapper as a replacement for the Java Exi Codec, I'd still like to run as many tests as possible. The fuzzy generator can be a big help with that.

SebaLukas (commented):
Looking good!

biubiuzhang:
Hi @SebaLukas — circled back on this. Running the fuzz-CSV replay locally as you suggested; results below.

## What I added

Two commits pushed to this branch:

- **`test(cbv2g_json_wrapper): add CSV-driven fuzz replay against EXICodec.jar oracle`** (2b6133d75) — new `cbv2g_test_din_fuzz` GoogleTest binary that loads a `din.csv` produced by your generator and parametrically asserts:
  1. `decode(reference-EXI)` structurally equals the canonical JSON Body emitted by `EXICodec.jar`
  2. `decode(encode(reference-JSON))` structurally equals the reference JSON Body
  
  Encoded bytes are not compared to the reference because EXI permits multiple valid encodings of the same Infoset — byte-equality with the Siemens codec is too strict. The binary is opt-in via `CBV2G_FUZZ_CSV=/path/to/din.csv` and intentionally not registered with `add_test()`, so default `ctest` stays fast for contributors without the corpus.

- **`fix(cbv2g_json_wrapper): preserve Id and GenChallenge in ContractAuthenticationReq`** (2bf207ad4) — real bug your generator caught, see §2 below.

## Results — `--number-din-msgs 200` (4 400 rows / 8 801 tests, 313 ms)

| Suite | Result |
|---|---|
| `cbv2g_test_apphand` | 6 / 6 ✅ |
| `cbv2g_test_din` (hand-written) | 7 / 7 ✅ |
| `cbv2g_test_din_fuzz` | **3 485 / 8 801 pass** |

| Coverage scope | Hand-written only | + fuzz replay |
|---|---:|---:|
| `din_converter.c` lines | 40 % | **86 %** |
| Whole `cbv2g_json_wrapper/src` lines | 46 % | **83 %** |
| Whole `cbv2g_json_wrapper/src` functions | 52 % | **93 %** |

## 1 real bug fixed

`ContractAuthenticationReq` was 50/50 failing in **both** directions in round 1. Root cause was `din_converter.c`:

```c
static int json_to_din_contract_authentication_req(...) {
    (void)json; (void)msg;
    msg->Id_isUsed = 0;
    msg->GenChallenge_isUsed = 0;        // ← discarded input
    return CBV2G_SUCCESS;
}
static cJSON* din_contract_authentication_req_to_json(...) {
    (void)msg;
    return cJSON_CreateObject();         // ← discarded output
}
```

Both stubs replaced with the same bounded-copy `strnlen` + `snprintf` pattern already used for `ServiceScope` (CWE-120 / CWE-126). After the fix: `ContractAuthenticationReq` is 200 / 200 ✅ in both directions, no regression to the 13 hand-written tests, fix region executes 200× (input path) and 400× (output path) per `gcov`.

## Remaining ~5 300 failures — JSON dialect divergence, not bugs

The fuzz test surfaced one unresolved policy question I'd like your view on before we say the wrapper is a drop-in replacement.

The wrapper emits **integer enum values** for several fields; `EXICodec.jar` emits the corresponding **string names**. For example:

| Field | Wrapper output | EXICodec.jar output |
|---|---|---|
| `ServiceCategory` | `3` | `"OtherCustom"` |
| `EVErrorCode` | `0` | `"FAILED_ChargingVoltageOutOfRange"` |
| `EVSEStatusCode` | `"EVSE_NotReady"` (default-zero) | `"EVSE_Ready"` |
| `EVSENotification` | `"None"` (default-zero) | `"ReNegotiation"` |

The EXI bitstreams are equivalent — but downstream JSON consumers used to the Java codec's dialect won't see the same shape from the wrapper. There are two reasonable paths:

1. **Match the Java dialect** — wrapper translates enums to/from their string names. More work in `din_converter.c`, but it makes the wrapper truly drop-in.
2. **Document the divergence** — keep integer enums (they're closer to the cbv2g C struct layout), and update consumers of the wrapper.

Do you have a preference? I'm slightly leaning toward (1) given the framing of "replacement for the Java EXI Codec" in the PR description, but (2) is defensible if performance is the goal.

A couple of smaller things the corpus also flagged (low priority, not bugs):
- ≈ 5 % of generated `EVCCID` values are longer than DIN's 6-byte limit (`polyfactory` doesn't honour the XSD length restriction). The wrapper safely truncates via `strnlen`-bounded `hex_decode`. Harmless, but worth noting if you ever filter the corpus.
- JSON object key ordering differs between wrapper and reference; `cJSON_Compare` ignores it but it shows up in any diff that doesn't.

## Plan for the CI follow-up

Will open the issue against `ext-switchev-iso15118` referencing this work + your generator as soon as the PR lands, as discussed. The local opt-in replay is in place now so anyone with the CSV can re-run it locally.

PR #2140, opened 2026-04-29
Add ISO 15118-2 converter (incl. PnC messages) to libcbv2g_json_wrapper

biubiuzhang:
## Summary
- Adds the ISO 15118-2 (`urn:iso:15118:2:2013:MsgDef`) JSON/EXI converter on top of EVerest/everest-core#2139, including PnC: `PaymentDetailsReq/Res`, `SignedInfo` xmldsig fragment, the `Signature` block in `V2G_Message.Header`, and a fragment-encode fall-through for callers that pass bare body elements.
- Routes both `NS_ISO_V2_MSG_DEF` and `NS_XML_DSIG` to the iso2 dispatcher (xmldsig fragments ride on the iso2 fragment encoder).
- Same string-handling hardening (CWE-120 / CWE-126).

**Stacked on EVerest/everest-core#2138 → EVerest/everest-core#2139 — please review those first.** ISO 15118-20 support is planned as a follow-up PR.

## Test plan
- [ ] `ctest -R cbv2g_test_iso2` runs SessionSetupReq, ChargeParameterDiscovery, AuthorizationReq with GenChallenge, ChargingStatusReq/Res, CurrentDemandReq (strict `cJSON_Compare`), PaymentDetailsReq with cert chain, and SignedInfo fragment encode.
- [ ] PnC round-trip works against a real Josev EVCC (post-deploy smoke test).

### DIFF
--- a/applications/CMakeLists.txt
+++ b/applications/CMakeLists.txt
@@ -1,1 +1,2 @@
 add_subdirectory(pionix_chargebridge)
+add_subdirectory(cbv2g_json_wrapper)
--- a/applications/cbv2g_json_wrapper/CMakeLists.txt
+++ b/applications/cbv2g_json_wrapper/CMakeLists.txt
@@ -22,6 +22,7 @@ set(WRAPPER_SOURCES
     src/json_utils.c
     src/apphand_converter.c
     src/din_converter.c
+    src/iso2_converter.c
 )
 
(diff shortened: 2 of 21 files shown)


biubiuzhang:
`Run OCPP Tests` failure looks like flake in `plug_and_charge_tests.py` — different test fails each rerun and PR 1 cleared the same suite on its prior run. Could you rerun the failed job? I don't have write access here.

PR #2141, opened 2026-04-30
Add check of number of certificates

MatthiasContribute:
## Describe your changes

The InstallCertificate command is extended to check if the max number of certificates is reached.
If so, an error is returned.

### DIFF
--- a/lib/everest/evse_security/lib/evse_security/evse_security.cpp
+++ b/lib/everest/evse_security/lib/evse_security/evse_security.cpp
@@ -376,6 +376,12 @@ InstallCertificateResult EvseSecurity::install_ca_certificate(const std::string&
 
         X509CertificateBundle existing_certs(ca_bundle_path, EncodingFormat::PEM);
 
+        if (existing_certs.get_certificate_count() > max_fs_certificate_store_entries) {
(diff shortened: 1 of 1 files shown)


lib/everest/evse_security/lib/evse_security/evse_security.cpp
```diff
@@ -376,6 +376,12 @@ InstallCertificateResult EvseSecurity::install_ca_certificate(const std::string&
 
         X509CertificateBundle existing_certs(ca_bundle_path, EncodingFormat::PEM);
 
+        if (existing_certs.get_certificate_count() > max_fs_certificate_store_entries) {
```
Pietfried: Looks like `get_certificate_count` only gives you the number of certificates for the `existing_certs` bundle, which you probably want to retrieve the number of certificates of all domains.

It should also eventually be integrated with the OCPP variable CertificateEntries.maxLimit . That would be a follow up task though.

MatthiasContribute: Following your comment, the check is extended to check per call, over the sum over MO and V2G certificates.

hikinggrass (approved):

PR #2143, opened 2026-04-30
bugfix(Powermeter errors): Use powermeter/ definitions again instead of generic

corneliusclaussen:
The powermeter interface was recently changed to use the generic error definition instead of its separate powermeter.yaml. The powermeter drivers were not updated accordingly so the current state in main as well as stable branch is iconsistent which leads to errors being ignored.

This reverts back to separate powermeter error definition as the generic/CommunicationFault for powermeters causes trouble in the ocpp error mapping.


Pietfried:
@florinmihut I remember you observed similar issues when working on the LEM powermeter. Can you please review if these changes do not re-re-introduce issues?

Pietfried:
@corneliusclaussen the API serialize tests fail. The hashes used in the tests need to be updated according to https://github.com/EVerest/EVerest/tree/main/lib/everest/everest_api_types#automatic-detection-of-changes-of-everest-types-and-interfaces

Pietfried (approved):

PR #2145, opened 2026-04-30
Make use of CertificateStoreMaxLength to be used in EVSE-Security

MatthiasContribute:
## Describe your changes
Make use of CertificateStoreMaxLength from config.json/user_config.json


Pietfried (changes requested):
Hi @MatthiasContribute , thanks for the contribution. The unit tests can be fixed according to https://github.com/EVerest/EVerest/tree/main/lib/everest/everest_api_types#automatic-detection-of-changes-of-everest-types-and-interfaces . 

Your branch also now has conflicts with main.

MatthiasContribute:
Not clear, why the OCPP Run test failes?

MatthiasContribute:
Anything, i can do concerging the failure in Run OCPP Tests?

Pietfried:
> Anything, i can do concerging the failure in Run OCPP Tests?

You can analyze the ocpp-tests report here: https://github.com/EVerest/EVerest/actions/runs/25724104514/artifacts/6941756816

You introduced a new function in the internal yaml interfaces `set_max_fs_certificate_store_entries` and you have to make sure that the probe_module at least provides a default implementation for this. You can find other examples here: https://github.com/EVerest/EVerest/blob/main/tests/ocpp_tests/test_sets/ocpp16/ocpp_generic_interface_integration_tests.py#L125

interfaces/evse_security.yaml
```diff
@@ -226,6 +226,13 @@ cmds:
     result:
       description: True if verification succeeded, false if not
       type: boolean
+  set_max_fs_certificate_store_entries:
+    description: Command to set the maximum number of certificate store entries
+    arguments:
+      max_fs_certificate_store_entries:
+        description: Maximum number of certificate store entries. Must be greater than 0.
+        type: integer
+        minimum: 1
```
mlitre, line 235: This minimum number is probably worth a discussion, since having only 1 certificate is quite restrictive as a minimum.

MatthiasContribute:
Do not understand why the DCO check fails?

MatthiasContribute:
It is resolved, with a rebase.

MatthiasContribute:
The DCO check passe, but there is a timeout in the Bazel test i am not familiar with.

interfaces/evse_security.yaml
MatthiasContribute: It should be at least 1, in my opinion.

hikinggrass: I think that's probably fine, you can't set anything lower than 1 (so practically not 0) but are unbound above that

hikinggrass (dismissed):

hikinggrass (approved):

PR #2150, opened 2026-05-04
EvseManager: re-issue SessionStart if Ev is still plugged in at Sessi…

dorezyuk:
Currently the auth module would not consider an Ev as connected if the session stops and the customer is not unplugging. We reissue the SessionStart here to allow users to re-authenticate without unplugging

### DIFF
--- a/modules/EVSE/EvseManager/Charger.cpp
+++ b/modules/EVSE/EvseManager/Charger.cpp
@@ -964,6 +964,29 @@ void Charger::run_state_machine() {
                 break;
             }
 
+            // Cable still attached and the stop was user/CSMS-initiated: end
+            // the transaction, refresh the session, and route back to
+            // WaitingForAuthentication so a new RFID swipe can start a new
+            // transaction without requiring a replug. Fatal stops (errors,
+            // emergency, plug-out, etc.) fall through to Finished below and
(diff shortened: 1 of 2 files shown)


djchhp:
What is the state of this PR? If still valid, please rebase and ask for review.

dorezyuk:
@djchhp @Pietfried its still valid and I'm still waiting for review. I think rebasing makes probably not much sense at this point - I've done every so often a couple of rounds

Pietfried (changes requested):
This change introduces at least a violation of the StatusNotification lifecycle in OCPP1.6, by introducing a transient `Available` while plugged in. It also moves to `Preparing` without any user interaction.

The current definition of a session and a transaction in EVerest is:
1. A session starts with first user interaction (either EV plug in or provided authorization)
2. A transaction starts when EV is plugged in AND auth is present
3. A transaction stops when EV not plugged in OR Auth not present anymore
4. A session stops with EV plug out
5. A transaction can not be restarted once (3) happened -> A new transaction requires a replug

All implementations in Auth, EvseManager and the consumers like OCPP follow this contract (and are driven by respective session events),  which is why I am a little concered about event more side effects this change could introduce.

The behavior should at least be configurable (off by default) and properly integration tested with OCPP.

PR #2159, opened 2026-05-05
fix(mqtt): fix TOCTOU race causing messages to be silently dropped

Pietfried:
## Describe your changes

publish() checked `mqtt_is_connected` outside the mutex before queuing, so on_mqtt_connect() could drain and clear `messages_before_connected` between the check and the push, losing mqtt messages and potentially causing modules to time out on startup.

Fix by moving the check inside the mutex in publish(), draining the queue under the lock in on_mqtt_connect() before publishing outside it and making mqtt_is_connected atomic to eliminate remaining unsynchronised reads.

Should fix issue of E2E tests occasionally failing at startup.

### DIFF
--- a/lib/everest/framework/include/utils/mqtt_abstraction_impl.hpp
+++ b/lib/everest/framework/include/utils/mqtt_abstraction_impl.hpp
@@ -82,6 +82,7 @@ private:
     };
 
     bool mqtt_is_connected;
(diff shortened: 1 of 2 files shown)


hikinggrass (dismissed):
looks good, unfortunately has some conflicts after the recent merges on main

hikinggrass (approved):

PR #2162, opened 2026-05-05
fix: prevent crash when running in environments without O_DIRECT pipe support

nzbr:
## Describe your changes

According to [its man page](https://man7.org/linux/man-pages/man2/pipe.2.html), the pipe2 syscall returns `EINVAL` (via errno) if the `O_DIRECT` flag is used on a kernel that does not support it.
In addition to old versions of the Linux kernel, this is also the case with some sandbox implementations.
This PR introduces a fallback where it is attempted to create the pipe without `O_DIRECT` if the original call results in `EINVAL`

### DIFF
--- a/lib/everest/framework/src/system_unix.cpp
+++ b/lib/everest/framework/src/system_unix.cpp
@@ -206,8 +206,21 @@ std::string set_user_and_capabilities(const std::string& run_as_user, const std:
 SubProcess SubProcess::create(const std::string& run_as_user, const std::vector<std::string>& capabilities) {
     std::array<int, 2> pipefd{};
 
-    if (pipe2(pipefd.data(), O_CLOEXEC | O_DIRECT)) {
-        throw std::runtime_error(fmt::format("Syscall pipe2() failed ({}), exiting", strerror(errno)));
+    const auto flags = O_CLOEXEC;
+    // First, try to create a pipe with O_DIRECT
+    if (pipe2(pipefd.data(), flags | O_DIRECT)) {
+        auto errored = true;
+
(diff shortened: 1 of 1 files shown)


lib/everest/framework/src/system_unix.cpp
```diff
@@ -206,8 +206,20 @@ std::string set_user_and_capabilities(const std::string& run_as_user, const std:
 SubProcess SubProcess::create(const std::string& run_as_user, const std::vector<std::string>& capabilities) {
     std::array<int, 2> pipefd{};
 
-    if (pipe2(pipefd.data(), O_CLOEXEC | O_DIRECT)) {
-        throw std::runtime_error(fmt::format("Syscall pipe2() failed ({}), exiting", strerror(errno)));
+    auto flags = O_CLOEXEC | O_DIRECT;
```
james-ctc: could be const

lib/everest/framework/src/system_unix.cpp
```diff
@@ -206,8 +206,20 @@ std::string set_user_and_capabilities(const std::string& run_as_user, const std:
...
+    if (pipe2(pipefd.data(), flags)) {
```
james-ctc: an alternative
```
const flags = O_CLOEXEC;

// a bit clearer that O_DIRECT is being tried
if (pipe2(pipefd.data(), flags | O_DIRECT ) == -1) {

if (pipe2(pipefd.data(), flags ) == 0) {
```

james-ctc (dismissed):
looks fine, just some minor suggestions

nzbr:
done, thanks

james-ctc (approved):

PR #2164, opened 2026-05-05
Various CMake fixes for non-EDM (Yocto) builds

hikinggrass:
## Describe your changes

Add missing find_dependency for ryml and everest-tls

Fix typo in find_dependency for everest-evse_security

Use CMAKE_INSTALL_INCLUDEDIR instead of hardcoding include

Add version to everest_io lib so it doesn't produce an unversioned shared library

Remove ev_register_library_target in everest_util that accidentally pointed at everest_helpers


corneliusclaussen (approved):

Pietfried (approved):
Do we need to backport (parts of) this?

hikinggrass:
> Do we need to backport (parts of) this?

I don't think so, it seems these issues got introduced in #2047 and #2156 which are both not backported

PR #2165, opened 2026-05-06
test(tls): drop NUL terminator in base64_decode test inputs

### LINKED ISSUE #2131: OpenSSL base64 decode is broken on newer versions of OpenSSL
### Describe the bug

There are some unit tests for `lib/everest/tls/tests` that fail on newer versions of OpenSSL
test executable: tls_test

Debian 12: OpenSSL 3.0.19 27 Jan 2026 (Library: OpenSSL 3.0.19 27 Jan 2026)
works
```
[----------] Global test environment tear-down
[==========] 88 tests from 17 test suites ran. (3851 ms total)
[  PASSED  ] 88 tests.
```

Debian 13: OpenSSL 3.5.5 27 Jan 2026 (Library: OpenSSL 3.5.5 27 Jan 2026)
fails
```
[----------] Global test environment tear-down
[==========] 88 tests from 17 test suites ran. (3437 ms total)
[  PASSED  ] 86 tests.
[  FAILED  ] 2 tests, listed below:
[  FAILED  ] openssl.base64Decode
[  FAILED  ] openssl.base64DecodeNl
```

OpenSSL change log and git history have quite a few changes on base64 which make having a stable implementation difficult. e.g.

```
    Note finished state in cipher BIO EOF
    
    When the cipher BIO encounters a non-retriable EOF (or error), mark the
    state as "finished", else a subsequent BIO_flush() or attempted read may
    attempt to finalise the crypto state again, and fail, leading, for
    example, to users seeing erroneous apparent decryption failure.
    
    This is not a new problem, the fix should be backported to all supported
    releases.  It was made more apparent by recent changes to the base64
    BIO, which returns a non-retriable EOF when padding is seen at the end
    of the base64 data, even if the underlying next BIO is "retriable".
```

Simple test code using `EVP_DecodeBlock` or `BIO_read` return error conditions where the data should have been accepted.

I recommend using different base64 encode and decode functions and avoid the OpenSSL ones.

### EVerest Domain

Other, ISO15118

### Affected EVerest Module

affects TLS and evse_security libraries
EVSEV2G
15118
...

files affected:
- `lib/everest/tls/openssl_util.cpp`
- `lib/everest/evse_security/lib/evse_security/crypto/openssl/openssl_crypto_supplier.cpp`

### To Reproduce

Build EVerest with unit tests with OpenSSL 3.5.5 (other new versions may fail as well)
```
cd build/lib/everest/tls/tests
./tls_test
```

### Anything else?

_No response_

mlitre:
The fixture buffers are const char[] literals so sizeof() includes the trailing NUL. The cleanup loop in base64_decode only strips '\n', so NUL flowed into BIO_f_base64. OpenSSL <= 3.0 silently truncated at the NUL and returned the valid prefix, but OpenSSL 3.5 rejects the whole input on stray non-alphabet bytes and returns 0. Pass sizeof - 1 so the call mirrors what real callers do via std::string::size().

Fixes issue #2131 

### DIFF
--- a/lib/everest/tls/src/openssl_util.cpp
+++ b/lib/everest/tls/src/openssl_util.cpp
@@ -319,14 +319,25 @@ std::vector<std::uint8_t> base64_decode(const char* text, std::size_t len) {
     assert(text != nullptr);
     assert(len > 0);
 
-    // remove \n
+    // Strip whitespace + NUL; pass everything else to BIO_f_base64.
+    // Byte set matches EVP_Decode*'s B64_WS table so this stays drop-in
+    // equivalent to the evse_security path (consolidation target).
     auto input = std::make_unique<std::uint8_t[]>(len);
     std::size_t input_len{0};
 
     for (std::size_t i = 0; i < len; i++) {
         const auto item = text[i];
-        if (item != '\n') {
-            input.get()[input_len++] = item;
+        switch (item) {
+        case '\0':
+        case '\t':
+        case '\n':
+        case '\v':
+        case '\f':
+        case '\r':
+        case ' ':
+            continue;
+        default:
+            input.get()[input_len++] = static_cast<std::uint8_t>(item);
         }
     }
 
--- a/lib/everest/tls/src/openssl_util.cpp
+++ b/lib/everest/tls/src/openssl_util.cpp
@@ -316,8 +316,9 @@ bool sha_512(const void* data, std::size_t len, sha_512_digest_t& digest) {
 }
 
 std::vector<std::uint8_t> base64_decode(const char* text, std::size_t len) {
-    assert(text != nullptr);
-    assert(len > 0);
+    if (text == nullptr || len == 0) {
+        return {};
+    }
 
     // Strip whitespace + NUL; pass everything else to BIO_f_base64.
     // Byte set matches EVP_Decode*'s B64_WS table so this stays drop-in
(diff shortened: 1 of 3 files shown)


lib/everest/tls/src/openssl_util.cpp
```diff
@@ -316,17 +316,29 @@ bool sha_512(const void* data, std::size_t len, sha_512_digest_t& digest) {
 }
 
 std::vector<std::uint8_t> base64_decode(const char* text, std::size_t len) {
-    assert(text != nullptr);
-    assert(len > 0);
+    if (text == nullptr || len == 0) {
+        return {};
```
james-ctc: please avoid extra returns, the rest of the code doesn't do this

lib/everest/tls/src/openssl_util.cpp
```diff
@@ -316,17 +316,29 @@ bool sha_512(const void* data, std::size_t len, sha_512_digest_t& digest) {
...
+    // equivalent to the evse_security path (consolidation target).
     auto input = std::make_unique<std::uint8_t[]>(len);
     std::size_t input_len{0};
 
     for (std::size_t i = 0; i < len; i++) {
         const auto item = text[i];
-        if (item != '\n') {
-            input.get()[input_len++] = item;
+        switch (item) {
+        case '\0':
```
james-ctc: Perhaps \0 should terminate the loop. I can't think of a case where \0 would be in the middle of base64 encoded data. to a \0 is likely a length parameter error and continuing to process additional bytes is likely to fail anyway.

james-ctc (changes requested):
the description should link to the OpenSSL issue: https://github.com/EVerest/EVerest/issues/2131

other than that just some minor tidy ups.

james-ctc (approved):

PR #2170, opened 2026-05-06
[OCPP] EVSE Manager publishing updated energy limit

james-ctc:
## Describe your changes

The EVSE Manager publishes enforced_limits (which is used by OCPP). Unfortunately the ac_max_current_A published is before any adjustments have been made. This PR ensures that the updated limit is in ac_max_current_A in particular taking into account the PP current rating

### DIFF
--- a/modules/EVSE/EvseManager/Charger.cpp
+++ b/modules/EVSE/EvseManager/Charger.cpp
@@ -13,6 +13,7 @@
 #include <chrono>
 #include <generated/types/powermeter.hpp>
 #include <math.h>
+#include <optional>
 #include <string.h>
 #include <thread>
 #include <type_traits>
@@ -215,7 +216,7 @@ void Charger::run_state_machine() {
                     shared_context.hlc_charging_active = true;
                 }
                 shared_context.hlc_allow_close_contactor = false;
-                shared_context.max_current_cable = 0;
+                shared_context.max_current_cable.reset();
                 shared_context.hlc_charging_terminate_pause = HlcTerminatePause::Unknown;
                 shared_context.legacy_wakeup_done = false;
                 shared_context.hlc_d20_active = false;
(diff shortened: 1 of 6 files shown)


Pietfried (changes requested):
The changes look good to me, but it would be better to also include the PP value inside `request_energy_from_energy_manager`. This would already avoid that the EnergyManager distributes more current than PP allows. Your changes can stay, but would then only be input validation in case the EnergyManager made a mistake.

Something like this could be added to `request_energy_from_energy_manager`:

```cpp
// Cap by PP cable rating so the EnergyManager cannot allocate more than the cable allows
if (mod->config.charge_mode == "AC") {
        const float pp_rating = mod->bsp->read_pp_ampacity();
        const std::string source_pp = source_base + "/pp_ampacity";
        for (auto& e : energy_flow_request.schedule_import) {
            if (e.limits_to_root.ac_max_current_A.has_value() &&
                e.limits_to_root.ac_max_current_A.value().value > pp_rating) {
                e.limits_to_root.ac_max_current_A = {pp_rating, source_pp};
            }
        }
    }
```

Pietfried (dismissed):

james-ctc:
From the documentation there is a potential issue:
```
ac_pp_ampacity:
Current carrying capacity of the connected cable in ampere for AC 
charging with a socket. Publish whenever it changes.
This has no meaning for DC or AC charging with a fixed attached cable, 
it does not need to be implemented and the returned value is not used in
those cases.
```

Therefore the BSP module is not expected to publish PP values when there is a fixed cable.

Options:
- update EVSE Manager to ignore PP values when one has not been provided
- update BSP interface so that PP values must always be provided

@Pietfried what is your preference?

james-ctc:
@Pietfried I've updated in-line with the discussions yesterday and all tests are passing - ready for re-review

Pietfried (approved):

PR #2187, opened 2026-05-08
fix(ocpp1.6): reject invalid boolean ChangeConfiguration values

### LINKED ISSUE #2182: OCPP1.6 ChangeConfiguration silently accepts invalid boolean values (Accepted instead of Rejected)
### OCPP Version

OCPP1.6

### Describe the bug

In OCPP 1.6, `ChangeConfiguration.req` for a boolean `ConfigurationKey` silently coerces any non-`"true"` value to `false` and returns `status = Accepted`, instead of returning `status = Rejected` when the value does not parse as a valid boolean.

The CSMS therefore believes the requested value was accepted, while the charger has actually stored `false`. Operator and charger end up with divergent views of configuration state, with no error surfaced.

We would like to confirm with maintainers whether this is intentional behavior or a bug before submitting a PR. Our reading is that it is inconsistent with how the rest of the OCPP code in EVerest validates `ChangeConfiguration` input — see "Inconsistency with existing patterns" below.

> Note: this was originally found against `EVerest/libocpp` and re-verified against `EVerest/EVerest` `main` after the standalone repo was archived on 2026-04-27. The same `string_to_bool` and `ChangeConfiguration` handler is present at `lib/everest/ocpp/` in this repo, so the report applies as-is.

### To Reproduce

Tested against firmware 0.20.0 (libocpp version equivalent to migrated tree at `lib/everest/ocpp/`) on a Titan Rev1 charger with a MaEVe CSMS over MQTT.

1. Send `ChangeConfiguration` with a non-boolean value on a boolean key:

   ```
   {"type":2,"id":"cfg-badbool","action":"ChangeConfiguration",
    "request":{"key":"AuthorizeRemoteTxRequests","value":"maybe"}}
   ```

   Response:
   ```
   [3,"cfg-badbool",{"status":"Accepted"}]
   ```

2. Read it back with `GetConfiguration`:

   ```
   [3,"verify",{"configurationKey":[
     {"key":"AuthorizeRemoteTxRequests","readonly":false,"value":"false"}
   ]}]
   ```

The string `"maybe"` was silently coerced to `"false"` and stored, with `Accepted` returned.

### Root cause

`lib/everest/ocpp/lib/ocpp/common/types.cpp`:

```cpp
bool string_to_bool(const std::string& s) {
    std::string out = s;
    std::transform(out.begin(), out.end(), out.begin(), ::tolower);
    if (out == "true") {
        return true;
    }
    return false;   // any other input → false, no error signalled
}
```

This was last touched by libocpp PR #887 (case-insensitivity fix) but the silent-coercion behavior was kept.

`lib/everest/ocpp/lib/ocpp/v16/charge_point_configuration.cpp` then calls it directly for every boolean Core key without validation, e.g.:

```cpp
} else if (key == "AuthorizeRemoteTxRequests") {
    this->setAuthorizeRemoteTxRequests(ocpp::conversions::string_to_bool(value.get()));
}
```

The same pattern repeats for ~17 boolean keys (`LocalAuthorizeOffline`, `LocalPreAuthorize`, `AuthorizationCacheEnabled`, `StopTransactionOnInvalidId`, `UnlockConnectorOnEVSideDisconnect`, `AllowChargingProfileWithoutStartSchedule`, `DisableSecurityEventNotifications`, `ContractValidationOffline`, `ISO15118CertificateManagementEnabled`, `ISO15118PnCEnabled`, etc.). None of them validate the input before coercion, and all fall through to `Accepted`.

### Inconsistency with existing patterns

Other parts of the OCPP code already validate input and return `Rejected` for the same class of problem — only the OCPP 1.6 standard boolean keys are missing this:

| Path | Behavior on invalid input | Reference |
|---|---|---|
| OCPP 1.6 integer keys | `Rejected` via `is_positive_integer()` | `lib/everest/ocpp/lib/ocpp/v16/charge_point_configuration.cpp` (`BlinkRepeat`, etc.) |
| OCPP 1.6 custom keys | `Rejected` via JSON schema validation | libocpp PR #853 (merged) |
| OCPP 2.0.1 boolean | `Rejected` via `value == "true" or value == "false"` | `lib/everest/ocpp/lib/ocpp/v2/device_model.cpp` (`validate_value`) |
| **OCPP 1.6 standard boolean** | **`Accepted`, value silently set to `false`** | this issue |

### Expected behavior

The `ChangeConfiguration` handler should:

1. Validate that the value parses as a valid boolean (case-insensitive `"true"`/`"false"` only)
2. Return `ConfigurationStatus::Rejected` when validation fails
3. Leave the existing value unchanged on rejection

Correct response for the reproduction above:
```
[3,"cfg-badbool",{"status":"Rejected"}]
```

### Suggested fix direction

If maintainers agree this should be tightened, a minimal change would be:

1. Add `std::optional<bool> try_string_to_bool(const std::string&)` in `lib/everest/ocpp/lib/ocpp/common/types.cpp` that returns `std::nullopt` for anything not case-insensitive `"true"`/`"false"`. Keep `string_to_bool` for backward compatibility.
2. In `charge_point_configuration.cpp`, gate each boolean `ChangeConfiguration` branch with the optional, returning `ConfigurationStatus::Rejected` on `nullopt` — mirrors the existing `is_positive_integer` pattern used for `BlinkRepeat`.
3. Add a unit-test matrix in `lib/everest/ocpp/tests/lib/ocpp/v16/`: `"true"`, `"false"`, `"True"`, `"FALSE"` → Accepted; `"maybe"`, `""`, `"1"`, `"0"`, `"yes"` → Rejected.

Happy to submit a PR along these lines once the direction is confirmed.

### Anything else?

Related but distinct prior work in the (now-archived) `EVerest/libocpp` repo, for context:

- libocpp #877 / #887 — case-insensitivity fix; did not change the silent-coercion behavior
- libocpp #258 — precedent for hardening integer input (closed, not merged)
- libocpp #853 — schema validation for custom keys (merged), does not cover standard keys
- libocpp #485 — same class of validation-gap bug for string-list measurands (closed)

No safety impact (does not affect the charging safety envelope), but the divergence between CSMS and charger view of configuration state can mask real CSMS-side bugs and corrupt operator-visible config.

biubiuzhang:
## Summary

Fixes EVerest/EVerest#2182.

In OCPP 1.6, `ChangeConfiguration` on a boolean `ConfigurationKey` previously returned `status = Accepted` for any non-`"true"` value while silently storing `false`, because `string_to_bool()` coerces unknown input to `false` without signalling an error. The CSMS and the charger end up with divergent views of the configuration with no error surfaced.

This PR validates the input for the previously-unprotected boolean keys and returns `ConfigurationStatus::Rejected` for non-boolean values, leaving the existing value unchanged.

The implementation reuses the existing `ChargePointConfigurationBase::isBool()` helper (case-insensitive `"true"`/`"false"` only) and mirrors the validation pattern already used in the same file for:

- other boolean keys: `AllowOfflineTxForUnknownId`, `AuthorizationCacheEnabled`, `LocalAuthorizeOffline`, `LocalPreAuthorize`, `StopTransactionOnInvalidId`, `UnlockConnectorOnEVSideDisconnect`, `StopTransactionIfUnlockNotSupported`, `LocalAuthListEnabled`, `VerifyCsmsAllowWildcards`
- integer keys: `BlinkRepeat`, `ClockAlignedDataInterval`, etc. (via `is_positive_integer()`)

## Affected keys

- `AuthorizeRemoteTxRequests`
- `CentralContractValidationAllowed`
- `ContractValidationOffline`
- `DisableSecurityEventNotifications`
- `ISO15118CertificateManagementEnabled`
- `ISO15118PnCEnabled`
- `AllowChargingProfileWithoutStartSchedule`
- `CustomIdleFeeAfterStop`

## Behaviour change

Before:
```
ChangeConfiguration{ key: "AuthorizeRemoteTxRequests", value: "maybe" }
-> { status: "Accepted" }
GetConfiguration{ key: "AuthorizeRemoteTxRequests" }
-> { value: "false" }   // silently coerced
```

After:
```
ChangeConfiguration{ key: "AuthorizeRemoteTxRequests", value: "maybe" }
-> { status: "Rejected" }
GetConfiguration{ key: "AuthorizeRemoteTxRequests" }
-> { value: "true" }    // unchanged
```

Valid case-insensitive boolean literals (`"true"`, `"false"`, `"True"`, `"FALSE"`, etc.) continue to be accepted.

## Tests

Adds two test cases in `lib/everest/ocpp/tests/lib/ocpp/v16/test_configuration.cpp` keyed off `AuthorizeRemoteTxRequests`:

- `BooleanKeyAcceptsValidLiterals` — `"true"`, `"false"`, `"True"`, `"FALSE"`, `"tRuE"` -> `Accepted`
- `BooleanKeyRejectsInvalidLiterals` — `"maybe"`, `""`, `"1"`, `"0"`, `"yes"`, `"no"`, `"tru"`, `"falsey"` -> `Rejected`, and verifies the stored value is unchanged after each rejection.

## Notes

- No public API change. `string_to_bool()` is left untouched for backward compatibility; only the OCPP 1.6 `ChangeConfiguration` handler now validates input before calling it.
- The custom-key path at `charge_point_configuration.cpp:3015` is intentionally left untouched in this PR — that branch is covered by JSON schema validation (libocpp PR #853) and is out of scope for this issue.
- Issue #2182 was confirmed by @Pietfried as a reasonable fix direction.

### DIFF
--- a/lib/everest/ocpp/lib/ocpp/v16/charge_point_configuration.cpp
+++ b/lib/everest/ocpp/lib/ocpp/v16/charge_point_configuration.cpp
@@ -3574,7 +3574,11 @@ std::optional<ConfigurationStatus> ChargePointConfiguration::set(const CiString<
         EVLOG_warning << "Attempt to change AuthorizationKey to value with < 8 characters";
         return ConfigurationStatus::Rejected;
     } else if (key == "AuthorizeRemoteTxRequests") {
-        this->setAuthorizeRemoteTxRequests(ocpp::conversions::string_to_bool(value.get()));
+        if (isBool(value.get())) {
+            this->setAuthorizeRemoteTxRequests(ocpp::conversions::string_to_bool(value.get()));
+        } else {
+            return ConfigurationStatus::Rejected;
+        }
     } else if (key == "BlinkRepeat") {
         if (this->getBlinkRepeat() == std::nullopt) {
             return ConfigurationStatus::NotSupported;
@@ -3625,7 +3629,11 @@ std::optional<ConfigurationStatus> ChargePointConfiguration::set(const CiString<
         if (this->getCentralContractValidationAllowed() == std::nullopt) {
             return ConfigurationStatus::NotSupported;
         }
-        this->setCentralContractValidationAllowed(ocpp::conversions::string_to_bool(value.get()));
+        if (isBool(value.get())) {
+            this->setCentralContractValidationAllowed(ocpp::conversions::string_to_bool(value.get()));
+        } else {
+            return ConfigurationStatus::Rejected;
+        }
     } else if (key == "CertSigningWaitMinimum") {
         if (this->getCertSigningWaitMinimum() == std::nullopt) {
             return ConfigurationStatus::NotSupported;
@@ -3657,11 +3665,19 @@ std::optional<ConfigurationStatus> ChargePointConfiguration::set(const CiString<
             return ConfigurationStatus::Rejected;
         }
     } else if (key == "ContractValidationOffline") {
-        this->setContractValidationOffline(ocpp::conversions::string_to_bool(value.get()));
+        if (isBool(value.get())) {
+            this->setContractValidationOffline(ocpp::conversions::string_to_bool(value.get()));
+        } else {
+            return ConfigurationStatus::Rejected;
+        }
     } else if (key == "CpoName") {
         this->setCpoName(value.get());
     } else if (key == "DisableSecurityEventNotifications") {
-        this->setDisableSecurityEventNotifications(ocpp::conversions::string_to_bool(value.get()));
+        if (isBool(value.get())) {
+            this->setDisableSecurityEventNotifications(ocpp::conversions::string_to_bool(value.get()));
+        } else {
+            return ConfigurationStatus::Rejected;
+        }
     } else if (key == "HeartbeatInterval") {
         try {
             auto [valid, interval] = is_positive_integer(value.get());
@@ -3675,9 +3691,17 @@ std::optional<ConfigurationStatus> ChargePointConfiguration::set(const CiString<
             return ConfigurationStatus::Rejected;
         }
     } else if (key == "ISO15118CertificateManagementEnabled") {
-        this->setISO15118CertificateManagementEnabled(ocpp::conversions::string_to_bool(value.get()));
+        if (isBool(value.get())) {
+            this->setISO15118CertificateManagementEnabled(ocpp::conversions::string_to_bool(value.get()));
+        } else {
+            return ConfigurationStatus::Rejected;
+        }
     } else if (key == "ISO15118PnCEnabled") {
-        this->setISO15118PnCEnabled(ocpp::conversions::string_to_bool(value.get()));
+        if (isBool(value.get())) {
+            this->setISO15118PnCEnabled(ocpp::conversions::string_to_bool(value.get()));
+        } else {
+            return ConfigurationStatus::Rejected;
+        }
     } else if (key == "LightIntensity") {
         if (this->getLightIntensity() == std::nullopt) {
             return ConfigurationStatus::NotSupported;
@@ -3976,10 +4000,13 @@ std::optional<ConfigurationStatus> ChargePointConfiguration::set(const CiString<
             return ConfigurationStatus::NotSupported;
         }
     } else if (key == "AllowChargingProfileWithoutStartSchedule") {
-        if (this->getAllowChargingProfileWithoutStartSchedule().has_value()) {
+        if (!this->getAllowChargingProfileWithoutStartSchedule().has_value()) {
+            return ConfigurationStatus::NotSupported;
+        }
+        if (isBool(value.get())) {
             this->setAllowChargingProfileWithoutStartSchedule(ocpp::conversions::string_to_bool(value.get()));
         } else {
-            return ConfigurationStatus::NotSupported;
+            return ConfigurationStatus::Rejected;
         }
     } else if (key.get().find("DefaultPriceText") == 0) {
         const ConfigurationStatus result = this->setDefaultPriceText(key, value);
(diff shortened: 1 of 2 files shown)


lib/everest/ocpp/lib/ocpp/v16/charge_point_configuration.cpp
```diff
@@ -4007,7 +4034,11 @@ std::optional<ConfigurationStatus> ChargePointConfiguration::set(const CiString<
             return result;
         }
     } else if (key == "CustomIdleFeeAfterStop") {
-        this->setCustomIdleFeeAfterStop(ocpp::conversions::string_to_bool(value));
+        if (isBool(value.get())) {
+            this->setCustomIdleFeeAfterStop(ocpp::conversions::string_to_bool(value));
```
Pietfried:
```suggestion
            this->setCustomIdleFeeAfterStop(ocpp::conversions::string_to_bool(value.get()));
```

biubiuzhang: Done in 345dd8fa — switched to `value.get()` to match the form used by all the other boolean keys. Thanks!

Pietfried (dismissed):
lgtm! Please squash your commits into one.

biubiuzhang:
@Pietfried Thanks for the review! 

- ✅ **Squashed** into a single signed-off commit as requested.
- ✅ Also folded in a fix for the two Codacy/Cppcheck `missingIncludeSystem` warnings (dropped the redundant `<string>`/`<vector>` includes in the test — they're already pulled in transitively, matching the rest of the file).

CI is green except for **one unrelated flaky failure**: `ocpp_tests/test_sets/ocpp21/bidirectional.py::test_q02_no_service_renegotiation` (an OCPP 2.1 / ISO 15118-20 TLS test — `SSL_read_ex() unexpected eof` + a `NotImplementedError` in the CSMS harness). It's unrelated to this OCPP 1.6 `ChangeConfiguration` change. The relevant checks all pass: CMake/GCC unit tests, the new `test_configuration` cases, and DCO.

Ready for another approval + merge whenever you get a chance — a re-run of the "Run OCPP Tests" job should clear the flaky failure. 🙏

Pietfried (approved):

PR #2194, opened 2026-05-12
feat(CarloGavazzi_EM580): EM300 support, transaction replies, and ID docs

florinmihut:
## Describe your changes
Add runtime detection of EM300/ET300 via Carlo Gavazzi identification
register 300012: live metering with OCMF transactions disabled for known
non-Eichrecht codes; default path remains full OCMF for other devices.

Adapt reads for EM300 layout: 32-bit energy as integer/decimal registers,
and smaller instantaneous-value Modbus reads where large blocks return
Illegal data value.

Fix transaction API usage: TransactionStartResponse vs TransactionStopResponse
use different field ordering for the error string; trim redundant logging.

Build start/stop transaction responses with small helpers that set fields
explicitly (aligned with powermeter types), and propagate std::exception::what()
into TransactionStartResponse.error on failure.

Document register 300012: explicit EM300/ET300 code list and behaviour for
unknown codes (assumed OCMF-capable until added to the driver).


mhei (approved):
LGTM

Pietfried (approved):

PR #2199, opened 2026-05-13
fix(libframework): update cmake target properties for framework library versioning

ManuelJanocha:
## Describe your changes
Set both `VERSION` and `SOVERSION` on the `framework` CMake target so the build produces a properly versioned shared library with the full version encoded in the filename.

This improves ABI decoupling: consumers link against the SONAME (`libframework.so.0`), while the real file carries the full version (`libframework.so.0.25.0`), allowing parallel installation and easier rollback.

**Before** — no versioned file, only the bare SONAME:
```
-rwxr-xr-x  1 root root  2624864  /usr/lib/libframework.so.0
```

**After** — symlink chain with full version:
```
lrwxrwxrwx  1 root root       22  /usr/lib/libframework.so.0 -> libframework.so.0.25.0
-rwxr-xr-x  1 root root  2624864  /usr/lib/libframework.so.0.25.0
```

Consumers correctly reference the SONAME, not the full path:
```
$ readelf -d /usr/libexec/everest/modules/LemDCBM400600/LemDCBM400600
  (NEEDED) Shared library: [libframework.so.0]
```

### DIFF
--- a/lib/everest/framework/lib/CMakeLists.txt
+++ b/lib/everest/framework/lib/CMakeLists.txt
@@ -10,7 +10,10 @@ set(DEFAULT_LINKAGE SHARED)
 add_library(framework ${DEFAULT_LINKAGE})
 add_library(everest::framework ALIAS framework)
 
-set_target_properties(framework PROPERTIES SOVERSION ${PROJECT_VERSION})
+set_target_properties(framework PROPERTIES
+    VERSION ${PROJECT_VERSION}
+    SOVERSION ${PROJECT_VERSION_MAJOR}
+)
 
 target_sources(framework
     PRIVATE
--- a/lib/everest/framework/lib/CMakeLists.txt
+++ b/lib/everest/framework/lib/CMakeLists.txt
@@ -12,7 +12,7 @@ add_library(everest::framework ALIAS framework)
 
 set_target_properties(framework PROPERTIES
     VERSION ${PROJECT_VERSION}
-    SOVERSION ${PROJECT_VERSION_MAJOR}
+    SOVERSION ${PROJECT_VERSION_MAJOR}.${PROJECT_VERSION_MINOR}
 )
 
 target_sources(framework


barsnick:
Currently, libframework is not ABI-compatible across minor versions (e.g. 0.25.x to 0.24.x). So soname-ing the library with just its major "0" is not something I would do right now.

It would probably be okay to make the soname MAJOR.MINOR, and to assume that micro version changes imply ABI compatibility. (This is something which needs to be ensured in the development and release process.)

barsnick:
Actually, I don't really understand what this improves.

> **Before** — no versioned file, only the bare SONAME:
> 
> ```
> -rwxr-xr-x  1 root root  2624864  /usr/lib/libframework.so.0
> ```

I don't see this.

The way libframework currently builds, it already implements an SONAME. E.g. take the artifact from the GitHub job https://github.com/EVerest/EVerest/actions/runs/26487047098/job/77996768264 (compressed [dist artifact](https://github.com/EVerest/EVerest/actions/runs/26487047098/artifacts/7230718782)).

`objdump -p dist/lib/libframework.so.0.25.0` gives:
```
  SONAME               libframework.so.0.25.0
```
and it has the proper symbolic link:
```
$ ll dist/lib/libframework.so*
lrwxrwxrwx 1 dev dev      22 Mai 27 04:34 dist/lib/libframework.so -> libframework.so.0.25.0
-rw-r--r-- 1 dev dev 6347928 Mai 27 04:34 dist/lib/libframework.so.0.25.0
```
The difference to your approach is that you provide "compatability" across major+minor version, instead of major+minor+micro.

We don't have proper ABI change agnostic versioning in place yet anyway, so I think it's better to keep it the way it is.

ManuelJanocha:
The solution helps keep the applications independent from the library, even if the library is modified. This may be relevant if parts of Everest become subject to German Eichrecht.

The applications are always linked against the full library name, libframework.so.0.25.0.

Eichrecht requires the applications to remain stable. Any change must be reported to the notified body. A SONAME could help solve this.

ManuelJanocha:
We found an alternative solution for our specific use case. However, separation may still be beneficial for other customers.

It is up to you whether to close it or keep it open.

barsnick:
Could you elaborate on what you can achieve in perspective to the Eichrecht requirement? Is it that you could update the library without having the recompile the binary module(s)?

How does does this comply with German Eichrecht? Why is a change in the binary not okay, but in the library?

As mentioned, a MAJOR.MINOR binding vs. MAJOR.MINOR.MICRO as you have implemented it now would certainly be better. This would then work for patch releases. On the main development branch, you must be aware that incompatibilities may often be introduced and therefore the version will/should be bumped. And that there is no good ABI compatibility management yet.

PR #2202, opened 2026-05-13
fix(EvseManager): Enable Disable Handling

Pietfried:
## Describe your changes

add EVSEDisabled stop reason and repair disable during active session

Add EVSEDisabled to StopTransactionReason across types, everest_api_types, and OCPP 1.6/2.0.1 conversions (mapped to Other).

Fix the Charger disable path: replace the flag_authorized side-effect with an explicit flag_disable_requested flag checked in each active state (Charging, ChargingPausedEV/EVSE, PrepareCharging, WFA, StoppingCharging, Finished) so sessions are properly torn down before transitioning to Disabled. authorize() checks flag_disable_requested to reject late auth responses while disabled. enable_disable() drives run_state_machine() synchronously so the resulting state is visible to callers before returning like it was before. This is important for OCPP compatiblity


modules/EVSE/EvseManager/Charger.cpp
```diff
@@ -1667,13 +1694,24 @@ bool Charger::enable_disable(int connector_id, const types::evse_manager::Enable
...
-            shared_context.current_state = EvseState::Disabled;
-            signal_simple_event(types::evse_manager::SessionEventEnum::Disabled);
+            // Arm the state machine to tear down any active session and transition
+            // to Disabled. Each active charging state (Charging, ChargingPausedEV/EVSE,
+            // PrepareCharging) checks flag_disable_requested and routes to StoppingCharging
+            // or Finished. The Idle state transitions directly to Disabled.
+            // cp_state_F() and bsp->enable(false) are called from the Disabled state
+            // handler to guarantee correct ordering.
+            shared_context.flag_disable_requested = true;
+            shared_context.authorized_pnc = false;
```
florinmihut: I have a PR that deals with authorized_pnc #2203 
The problem with authorize_pnc is that it is not cleared in start_session / stop_session. That could cause problems (seen that in fixing the tests of the manager life-cycle PR).
If we clear the flag in the start/stop, I think the set here becoms redundant.

modules/EVSE/EvseManager/Charger.cpp
```diff
@@ -1299,6 +1321,7 @@ bool Charger::cancel_transaction(const types::evse_manager::StopTransactionReque
 void Charger::start_session(bool authfirst) {
     shared_context.session_active = true;
     shared_context.flag_authorized = false;
+    shared_context.flag_externally_cancelled = false;
```
florinmihut, line 1326: I think we should reset this flag not only here and in enable_disable, but as well in stop_session to be on the safe side.

modules/EVSE/EvseManager/Charger.cpp
florinmihut, line 636: We do not check the shared_context.flag_disable_requested in this state. 
Is this on purpose?
If we added the check on the other states, I think it should be added here too.

florinmihut (changes requested):

Pietfried:
@florinmihut : Conflicts resolved, ready for re-review

florinmihut (approved):

Pietfried:
@SebaLukas Can you do a review as a codeowner?

SebaLukas (approved):

mlitre (approved):

modules/EVSE/EvseManager/Charger.cpp
```diff
@@ -64,6 +64,8 @@ Charger::Charger(const std::unique_ptr<IECStateMachine>& bsp, const std::unique_
     shared_context.flag_transaction_active = false;
     shared_context.session_active = false;
 
+    internal_context.last_state_detect_state_change = EvseState::Idle;
```
TD-CompleoCS, line 67: This now causes the initial `Idle`-state to not be initialized, since `shared_context.current_state` is initialized with `EvseState::Idle` as well. See `initialize_state` in line 160 and its use in line 212.

Instead, `last_state_detect_state_change ` should be initialized with `EvseState::Disabled`, for example.

PR #2207, opened 2026-05-15
feat(io): allow binding udp/tcp clients to network device

mlitre:
Add optional `device` parameter (defaults to empty) to udp_client_socket, udp_server_socket and tcp_socket open/setup methods, plus the underlying open_udp_*_socket / open_tcp_socket[_with_timeout] helpers. When set, the socket is bound to the named interface via SO_BINDTODEVICE before connect/bind.

bind_socket_to_device degrades gracefully when SO_BINDTODEVICE is denied with EPERM/EACCES (no CAP_NET_RAW):
  - For client sockets, IP_UNICAST_IF (AF_INET) or IPV6_UNICAST_IF (AF_INET6) restricts the outgoing interface for unicast packets. No privilege required, works for both v4 and v6.
  - As a last resort for AF_INET, a source-IP bind() to the interface's IPv4 address. For UDP servers the fallback binds to (interface_ip:port) instead of the wildcard (IPv4 only).

For multicast destinations, additionally set IP_MULTICAST_IF (AF_INET, via ip_mreqn with ifindex) or IPV6_MULTICAST_IF (AF_INET6) and populate sin6_scope_id on the destination sockaddr. SO_BINDTODEVICE / IP[V6]_UNICAST_IF do not steer the multicast TX path; without these, connect() to a link-local group such as ff02::1 fails and the socket never reaches a usable state.

Threads transparently through generic_fd_event_client's variadic constructor, so udp_client(remote, port, timeout, "eth0") works without client-side changes.

### DIFF
--- a/lib/everest/io/src/socket/socket.cpp
+++ b/lib/everest/io/src/socket/socket.cpp
@@ -77,7 +77,7 @@ namespace socket {
 namespace {
 // Returns true when SO_BINDTODEVICE succeeded. Returns false on EPERM/EACCES
 // (caller decides on a fallback). Throws on any other failure.
-bool try_so_bindtodevice(int fd, std::string const& device) {
+bool apply_so_bindtodevice(int fd, std::string const& device) {
     if (setsockopt(fd, SOL_SOCKET, SO_BINDTODEVICE, device.c_str(), device.length()) == 0) {
         return true;
     }
@@ -102,7 +102,7 @@ sa_family_t get_socket_family(int fd) {
 // IPv6: IPV6_UNICAST_IF takes int (ifindex in host byte order).
 // Returns true on success; false if the socket family is neither AF_INET nor AF_INET6.
 // Throws if the interface lookup fails or if setsockopt fails with a non-permission error.
-bool try_unicast_if(int fd, std::string const& device) {
+bool apply_unicast_if(int fd, std::string const& device) {
     unsigned int ifindex = if_nametoindex(device.c_str());
     if (ifindex == 0) {
         throw std::runtime_error(build_errno_string("if_nametoindex(\"" + device + "\") failed"));
@@ -193,12 +193,12 @@ void bind_socket_to_device(int fd, std::string const& device) {
     if (device.empty()) {
         return;
     }
-    if (try_so_bindtodevice(fd, device)) {
+    if (apply_so_bindtodevice(fd, device)) {
         return;
     }
     // Try IP[_V6]_UNICAST_IF for outgoing unicast routing without privilege. Works for both
     // IPv4 and IPv6 client sockets.
-    if (try_unicast_if(fd, device)) {
+    if (apply_unicast_if(fd, device)) {
         return;
     }
     // Last resort for IPv4 sockets with an unusable family: bind a source IP belonging to the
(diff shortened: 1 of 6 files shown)


SebaLukas (dismissed):
I didn't find anything during my quick scan 👍 
Works on machine (udp client + ipv6 multicast supported device)

lib/everest/io/src/socket/socket.cpp
```diff
@@ -74,7 +74,148 @@ namespace socket {
...
+// (caller decides on a fallback). Throws on any other failure.
+bool try_so_bindtodevice(int fd, std::string const& device) {
+    if (setsockopt(fd, SOL_SOCKET, SO_BINDTODEVICE, device.c_str(), device.length()) == 0) {
+        return true;
+    }
+    if (errno == EPERM || errno == EACCES) {
+        return false;
+    }
+    auto msg = build_errno_string("Failed to bind socket to device " + device);
+    throw std::runtime_error(msg);
```
hikinggrass: (same for all other instances) you could just combine these lines into one. But if you prefer this for readability making this `const auto` would not hurt


```suggestion
    throw std::runtime_error(build_errno_string("Failed to bind socket to device " + device));
```

hikinggrass (dismissed):

lib/everest/io/src/socket/socket.cpp
mlitre: Fixed

SebaLukas (dismissed):

lib/everest/io/src/socket/socket.cpp
```diff
@@ -97,6 +229,28 @@ event::unique_fd open_udp_server_socket(std::uint16_t port) {
...
+            // SO_BINDTODEVICE not permitted: bind to (interface_ip:port) instead of wildcard,
+            // so the server only receives traffic addressed to that interface's IPv4.
+            try {
+                bind_socket_to_interface_address(socket_fd, device, port);
+            } catch (...) {
+                close(socket_fd);
+                throw;
+            }
+            return event::unique_fd{socket_fd};
+        }
```
djchhp:
```
for (auto* p = servinfo; p != NULL; p = p->ai_next) {
        auto socket_fd = event::unique_fd(::socket(p->ai_family, p->ai_socktype, p->ai_protocol));
        set_reuse_address(socket_fd);
        if (socket_fd == -1) {
            continue;
        }

        if(not device.empty()) {
               if (not try_so_bindtodevice(socket_fd, device)) {
                   bind_socket_to_interface_address(socket_fd, device, port);
                   return socket_fd;
               }
        }        

        if (bind(socket_fd, p->ai_addr, p->ai_addrlen) == -1) {
            continue;
        }

        return socket_fd;
    }
   throw std::runtime_error(std::string("Could not open a socket for localhost:") + std::to_string(port));
}
```

I was wondering if this much shorter code isn't equivalent

lib/everest/io/src/socket/socket.cpp
```diff
@@ -74,7 +74,139 @@ namespace socket {
 // socket.hpp implementations
 //
 
-event::unique_fd open_udp_server_socket(std::uint16_t port) {
+namespace {
+// Returns true when SO_BINDTODEVICE succeeded. Returns false on EPERM/EACCES
+// (caller decides on a fallback). Throws on any other failure.
+bool try_so_bindtodevice(int fd, std::string const& device) {
```
djchhp: I find the name a bit misleading. It suggest this cannot throw but maps any throws to boolean result.

lib/everest/io/src/socket/socket.cpp
```diff
@@ -74,7 +74,139 @@ namespace socket {
...
+    }
+    return ss.ss_family;
+}
+
+// Restrict the outgoing interface for unicast packets without requiring CAP_NET_RAW.
+// IPv4: IP_UNICAST_IF takes uint32_t (ifindex in network byte order via in_addr layout).
+// IPv6: IPV6_UNICAST_IF takes int (ifindex in host byte order).
+// Returns true on success; false if the socket family is neither AF_INET nor AF_INET6.
+// Throws if the interface lookup fails or if setsockopt fails with a non-permission error.
+bool try_unicast_if(int fd, std::string const& device) {
```
djchhp: same naming issue here

djchhp (approved):
Looks good to me.

PR #2222, opened 2026-05-20
Drop max charger current limitation

martinlukas84:
In MCS mode we are handling high current, power, and voltage values.
Up to now, there is a limitation of the max current for the charger of 1000A, defined in standard IEC61851-1. When the current exceeds the max current limitation, the current will not be updated anymore by the evsemanager. This leads to a rampdown of current to 0A and a warning message 'Power budget expire...', which is flooding the log. This PR should get rid of this behavior.


barsnick (approved):
I totally agree.

mhei (approved):
A little bit more background: in one of our MCS setups, we have the AC side running at 690 V with 3 phases. The MCS standard describes that plugs are certified up to 3.75 MW. Simple current calculation for the AC side is: `3750000 W / (3 x 690V) = 1811 A`. On lower voltage rails, the current would even be higher, so defining a static max limit here does not make sense which was the reason to propose the removal here.

Pietfried (approved):

PR #2225, opened 2026-05-20
refactor(tls/iso15118): ConnectionSSL adapter

mlitre:
## Describe your changes

Collapse `iso15118::io::ConnectionSSL` into a thin adapter over the shared
`tls::Server` stack, and grow `lib/everest/tls` with the capabilities the
ISO 15118 SECC path still needed. Before this change `ConnectionSSL` carried
its own ~500-line OpenSSL implementation (SSL_CTX setup, `client_hello`
callback, TLS 1.3 detection, keylog callback, private-key callback, peer-cert
SHA-512, BIO accept loop) that duplicated logic already living — and
unit-tested — in `tls::Server`. The duplication let the ISO 15118 TLS behavior
drift from the shared stack and meant multi-chain / certificate-authorities
selection would have to be reimplemented twice.

**ISO 15118 side**

- `lib/everest/iso15118/src/iso15118/io/connection_ssl.cpp` is rewritten as an
  adapter: it builds a `tls::Server::config_t` from `SSLConfig`, owns the
  listen socket and the `accept(2)` loop, hands each accepted fd to
  `tls::Server::wrap_accepted_fd`, and drives `accept` / `read` / `write` /
  `shutdown` through `tls::Connection`. TLS termination, cert-chain handling,
  OCSP stapling, key logging and the client-hello verify-mode upgrade all live
  in the `tls::` layer. The accepted socket is owned by the `tls::Connection`;
  the poll registration is keyed on `Connection::socket()`.
- Peer-certificate verification uses the V2G root as the primary trust anchor
  and the MO root as an additional anchor (via the new
  `verify_locations_additional_files`), matching the pre-refactor behavior of
  loading both.
- `lib/everest/iso15118/include/iso15118/config.hpp`: `SSLConfig` changes
  shape. The single `path_certificate_chain` / `path_certificate_key` /
  `private_key_password` fields are replaced by a `std::vector<ChainConfig>
  chains`, where each `ChainConfig` is one leaf+intermediate PEM, its key, an
  optional key password and OCSP DER files. This is the multi-chain shape
  needed for future TLS 1.3 multi-chain selection driven by the peer's
  `certificate_authorities` extension; **only a single chain is populated and
  used in this PR** — selection lands later.
- `modules/EVSE/Evse15118D20/charger/ISO15118_chargerImpl.cpp` is updated for
  the reshaped `SSLConfig` (it pushes a single `ChainConfig`) and now moves the
  config and callbacks into the controller.
- A peer-initiated close is surfaced explicitly in both phases:
  - During the data phase via `ReadResult::connection_closed` (EOF / TLS
    `close_notify`), routed into the canonical `Session::close()` — replacing
    the misleading "invalid packet state" throw on TLS and the silent stall on
    plain TCP.
  - During the TLS handshake, any fatal outcome (peer close, alert,
    verification failure) tears the connection down through the canonical
    `ConnectionSSL::close()`: the fd is unregistered, exactly one `CLOSED`
    event is emitted, and the `Session` treats `CLOSED` as terminal so the
    controller reaps it. Previously a mid-handshake close left the EOF'd fd
    registered in the level-triggered poll loop and the session alive, which
    busy-looped and blocked all future SDP sessions.
- The vehicle (peer) certificate SHA-512 is cached once on the `OPEN` event,
  since the live `tls::Connection` is gone after `close()`.
- `helper_ssl.{hpp,cpp}` and the private `is_tls_1_3` ClientHello parser are
  deleted; their roles move into the shared stack.

**libtls (`lib/everest/tls`) changes**

- `enforce_tls_1_3` flag on `Server::config_t`: pins the min protocol version
  to TLS 1.3 and clears `SSL_OP_ENABLE_MIDDLEBOX_COMPAT` (RFC 8446 D.4).
  Server init fails fast when the flag is set without an explicit TLS 1.3
  ciphersuites list, instead of producing a server that fails every handshake.
- `verify_locations_additional_files` on `Server::config_t`: extra CA
  trust-anchor PEM files loaded in addition to `verify_locations_file` (used
  by the SECC adapter for the MO root).
- Server owns the `client_hello` callback via `client_hello_cb_dispatch`,
  which fans out to per-feature handlers: `handle_tls_1_3_verify_upgrade` and
  `ServerStatusRequestV2::handle_client_hello`. The verify-mode upgrade is
  gated behind `Server::config_t::verify_client_on_tls13` (default `false`):
  when set, the server requires a peer certificate once the client negotiates
  TLS 1.3.
- `ServerConnection::peer_certificate_sha512()` accessor and
  `Server::wrap_accepted_fd(fd, ip, service)` factory for callers that own
  their own accept loop (ownership contract documented). The factory returns
  `nullptr` when the server has no initialized SSL_CTX (`init()` failed or was
  never called), instead of constructing a connection that would null-deref on
  a later `accept()`; on that failure path the fd is not closed and remains
  owned by the caller.
- `Connection::last_error()` with single-use OpenSSL error capture: the
  per-thread error queue is drained at detection time, stored per connection,
  and cannot bleed into a later operation or connection on the same thread.
  `result_t::closed` is documented as covering graceful close, TLS error and
  syscall error; a non-empty `last_error()` distinguishes an error from a
  graceful close.
- Internal cleanup of `tls.cpp`: the SSL operation wrappers return a result
  struct instead of threading an out-parameter, the four duplicated
  read/write/accept/connect drive loops collapse into one engine, error logs
  name the actual operation (previously everything logged as `SSL_read`), and
  `configure_ssl_ctx`'s positional booleans become an options struct.
- `openssl::is_tls_1_3()` in `openssl_util` — the shared, unit-tested
  replacement for the parser removed from `connection_ssl.cpp`.
- `Client::config_t` gains `min_proto_version`; `ConfigItem` gains a
  `const std::string&` convenience constructor.

**Note:** this changes the `iso15118::config::SSLConfig` struct API (a breaking
change for in-tree consumers of that struct). The only in-tree consumer,
`Evse15118D20`, is updated in the same PR.

**Behavior change (TLS 1.3 client auth):** the libiso15118 SECC adapter opts
into `verify_client_on_tls13`, so it requires the vehicle to present a client
certificate once TLS 1.3 is negotiated — a peer-cert-less TLS 1.3 handshake is
rejected. This is scoped to the adapter only: `EvseV2G` and `IsoMux` also build
on `tls::Server` but leave the flag at its default, so their TLS behavior is
unchanged.

None.


lib/everest/iso15118/src/iso15118/io/connection_ssl.cpp
```diff
@@ -2,321 +2,110 @@
...
-#include <fstream>
-#include <sstream>
-#include <unistd.h>
+#include <memory>
+#include <string>
+#include <string_view>
 #include <vector>
 
-#include <arpa/inet.h>
+#include <netdb.h>
```
SebaLukas, line 14: Needs to be changed back because of #2082

lib/everest/iso15118/src/iso15118/io/connection_ssl.cpp
```diff
@@ -336,37 +124,40 @@ ConnectionSSL::ConnectionSSL(PollManager& poll_manager_, const std::string& inte
...
-
-    const auto set_reuseport = setsockopt(ssl->fd, SOL_SOCKET, SO_REUSEPORT, &optval_tmp, sizeof(optval_tmp));
-    if (set_reuseport == -1) {
+    if (setsockopt(ssl->listen_fd, SOL_SOCKET, SO_REUSEPORT, &optval_tmp, sizeof(optval_tmp)) == -1) {
         log_and_throw("setsockopt(SO_REUSEPORT) failed");
     }
 
-    const auto bind_result = bind(ssl->fd, reinterpret_cast<const struct sockaddr*>(&address), sizeof(address));
-    if (bind_result == -1) {
+    if (bind(ssl->listen_fd, reinterpret_cast<const struct sockaddr*>(&address), sizeof(address)) == -1) {
```
SebaLukas: Change back to two lines instead of only one if

lib/everest/iso15118/src/iso15118/io/connection_ssl.cpp
```diff
@@ -336,37 +124,40 @@ ConnectionSSL::ConnectionSSL(PollManager& poll_manager_, const std::string& inte
...
         const auto error = "Failed to bind ipv6 socket to interface " + interface_name_;
         log_and_throw(error.c_str());
     }
 
-    const auto listen_result = listen(ssl->fd, DEFAULT_SOCKET_BACKLOG);
-    if (listen_result == -1) {
+    if (listen(ssl->listen_fd, DEFAULT_SOCKET_BACKLOG) == -1) {
```
SebaLukas: Same as above

lib/everest/tls/include/everest/tls/openssl_util.hpp
```diff
@@ -544,6 +544,14 @@ using log_handler_t = void (*)(log_level_t level, const std::string& err);
  */
 log_handler_t set_log_handler(log_handler_t handler);
 
+/**
+ * \brief return true iff the supported_versions extension payload lists TLS 1.3
```
SebaLukas: `iff` to `if`

lib/everest/tls/src/openssl_util.cpp
```diff
@@ -835,6 +835,48 @@ bool certificate_sha_1(openssl::sha_1_digest_t& digest, const X509* cert) {
...
+        return false;
+    }
+
+    if (length_supported_versions % 2 != 0) {
+        log_error("length_supported_versions is not divisible by 2");
+        return false;
+    }
+
+    bool result{false};
+
```
SebaLukas, line 848: I would add this comment from libiso:
```
// First Byte: length
// Byte 2+3 -> First Version (03, 04)
// Byte 4+5 -> Second Version (03, 03)
// ....
```

lib/everest/tls/src/tls.cpp
```diff
@@ -807,7 +908,24 @@ Connection::result_t ServerConnection::accept(int timeout_ms) {
...
         }
     }
-    return convert(result);
+    return capture_and_convert(convert(result));
+}
+
+std::optional<std::array<std::uint8_t, 64>> ServerConnection::peer_certificate_sha512() const {
+    assert(m_context != nullptr);
+    SSL* ssl = m_context->ctx.get();
+    X509* peer = SSL_get_peer_certificate(ssl);
```
SebaLukas: `SSL_get_peer_certificate` is deprecated in 3.0 and `SSL_get0_peer_certificate` used here instead

lib/everest/tls/src/tls.cpp
```diff
@@ -930,6 +1048,43 @@ bool Server::init_socket(const config_t& cfg) {
...
+    if (server_max != 0 && server_max < TLS1_3_VERSION) {
+        return SSL_CLIENT_HELLO_SUCCESS;
+    }
+
+    const unsigned char* data{nullptr};
+    std::size_t datalen{0};
+
+    if (SSL_client_hello_get0_ext(ssl, TLSEXT_TYPE_supported_versions, &data, &datalen) == 1) {
+        if (openssl::is_tls_1_3(data, datalen)) {
+            SSL_set_verify(ssl, SSL_VERIFY_PEER | SSL_VERIFY_FAIL_IF_NO_PEER_CERT, nullptr);
```
SebaLukas, line 1017: A log message is missing. I like to see if the tls connection switched to tls 1.3:
```
logf_info("Client supports TLS1.3: Change verify mode to SSL_VERIFY_PEER and "
                      "SSL_VERIFY_FAIL_IF_NO_PEER_CERT");
```

SebaLukas (changes requested):
First look done. See my comments :)
I'd like to start testing the tls essentials soon (look into unit tests, establish connections using OpenSSL as a client, and test within the SIL).

lib/everest/iso15118/src/iso15118/session/iso.cpp
```diff
@@ -58,9 +58,19 @@ void raise_invalid_packet_state(const io::SdpPacket& sdp_packet) {
     log_and_throw(error.c_str());
 }
 
-// NOTE (aw): this function return true, if it would block to read a complete packet
-//            if it returns false, the packet is complete
-bool read_single_sdp_packet(io::IConnection& connection, io::SdpPacket& sdp_packet) {
+namespace {
+enum class SdpReadResult {
```
SebaLukas: That should be called V2GTPReadResult

lib/everest/iso15118/src/iso15118/session/iso.cpp
```diff
@@ -58,9 +58,19 @@ void raise_invalid_packet_state(const io::SdpPacket& sdp_packet) {
...
+    would_block,       //!< more data is needed to complete the packet
+    connection_closed, //!< the peer closed the connection mid-read
+};
+} // namespace
+
+// NOTE (aw): this function reports a tri-state result:
+//            - would_block: it would block to read a complete packet
+//            - complete: the packet is complete
+//            - connection_closed: the peer closed the connection during the read
+SdpReadResult read_single_sdp_packet(io::IConnection& connection, io::SdpPacket& sdp_packet) {
```
SebaLukas: In the iso.cpp its `V2GTP` instead of `SDP`. At the start of the `libiso15118` development, Anton made an incorrect assumption and used `SDP` instead of `V2GTP` here

SebaLukas (commented):
Looking good! Only one minor changes requested

SebaLukas (dismissed):
Looking good!
Before we merge this one, we need more tests (SIL + HW + real cars if possible)

lib/everest/iso15118/src/iso15118/io/connection_ssl.cpp
```diff
@@ -2,321 +2,116 @@
...
-    const std::size_t max_pass_len = (size - 1); // we exclude the endline
-    const std::size_t max_copy_chars =
-        std::min(max_pass_len, password->length()); // truncate if pass is too large and buffer too small
-
-    std::memset(buf, 0, size);
-    std::memcpy(buf, password->c_str(), max_copy_chars);
-
-    return max_copy_chars;
+tls::Server::config_t make_tls_server_config(const config::SSLConfig& cfg, const std::string& interface_name,
+                                             int listen_fd, std::vector<tls::Server::certificate_config_t> chains) {
```
james-ctc: Passing objects on the stack can be problematic. A reference would be better

lib/everest/iso15118/src/iso15118/io/connection_ssl.cpp
```diff
@@ -380,179 +182,162 @@ Ipv6EndPoint ConnectionSSL::get_public_endpoint() const {
...
-    log_and_raise_openssl_error("Failed to SSL_read_ex(): " + std::to_string(ssl_error));
-
+    std::string msg = "Failed to read from TLS connection: ";
+    msg += result_name(result);
+    const auto ssl_err = ssl->connection->last_error();
+    if (not ssl_err.empty()) {
+        msg += " openssl=";
+        msg += ssl_err;
+    }
+    log_and_throw(msg.c_str());
```
james-ctc, line 238: something isn't right - log_and_throw implies that execution doesn't continue after this point.
So why the  `return {false, 0};`?

lib/everest/iso15118/src/iso15118/io/connection_ssl.cpp
```diff
@@ -380,179 +182,162 @@ Ipv6EndPoint ConnectionSSL::get_public_endpoint() const {
...
 
+    if (not handshake_complete) {
+        const auto result = ssl->connection->accept(/*timeout_ms=*/0);
+        switch (result) {
+        case tls::Connection::result_t::success:
+            break;
+        case tls::Connection::result_t::want_read:
+        case tls::Connection::result_t::want_write:
+        case tls::Connection::result_t::timeout:
+            return;
```
james-ctc, line 287: This is messy mixing return and break in the same switch

lib/everest/iso15118/src/iso15118/io/connection_ssl.cpp
```diff
@@ -380,179 +182,162 @@ Ipv6EndPoint ConnectionSSL::get_public_endpoint() const {
...
+        case tls::Connection::result_t::closed:
+            // convert() folds all fatal handshake outcomes (peer close, alert,
+            // protocol error) into closed, so this teardown handles every
+            // non-success, non-blocking handshake result.
+            logf_error("TLS handshake failed: connection closed");
+            this->close();
             return;
+        default:
+            log_and_throw("Failed to complete TLS handshake");
```
james-ctc, line 302: should have a break statement. 
There is no requirement for default to be the last item in a switch. 
Should something be added after then there may be unintended consequences

lib/everest/tls/src/openssl_util.cpp
```diff
@@ -835,6 +835,53 @@ bool certificate_sha_1(openssl::sha_1_digest_t& digest, const X509* cert) {
     return bResult;
 }
 
+bool is_tls_1_3(const std::uint8_t* in, std::size_t inlen) {
```
james-ctc, line 838: Please follow the coding guidelines especially since the rest of this file already does - so matching the existing style should be the default.
i.e. a single return statement

james-ctc (changes requested):
General:
Rather than having fixed key/certificates ... generate them as part of the build process.

lib/everest/iso15118/src/iso15118/io/connection_ssl.cpp
mlitre: Ok I made it an r value reference to avoid copying the vector

lib/everest/iso15118/src/iso15118/io/connection_ssl.cpp
mlitre: Good catch I remember adding that to avoid a warning but I just added [[noreturn]] instead

lib/everest/iso15118/src/iso15118/io/connection_ssl.cpp
mlitre: Fixed

lib/everest/iso15118/src/iso15118/io/connection_ssl.cpp
mlitre: Fixed

lib/everest/tls/src/openssl_util.cpp
mlitre: Done

SebaLukas:
Can you also update `cmake/ev-lib-dependencies.cmake` and add tls to the iso15118 deps

SebaLukas (dismissed):

james-ctc (dismissed):
all of my comments have been addressed. Many thanks

SebaLukas (dismissed):

lib/everest/tls/tests/tls_connection_tls13_test.cpp
```diff
@@ -0,0 +1,638 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2024 Pionix GmbH and Contributors to EVerest
```
hikinggrass:
```suggestion
// Copyright 2026 Pionix GmbH and Contributors to EVerest
```

lib/everest/iso15118/src/iso15118/io/connection_ssl.cpp
```diff
@@ -2,321 +2,116 @@
 // Copyright 2024 Pionix GmbH and Contributors to EVerest
```
hikinggrass:
```suggestion
// Copyright 2024-2026 Pionix GmbH and Contributors to EVerest
```

hikinggrass (dismissed):

hikinggrass (approved):

SebaLukas (approved):

PR #2226, opened 2026-05-20
feat(tls/iso15118): TLS 1.3 multi-chain + rotation

mlitre:
## Describe your changes

> **Stacked PR** — the stack is `refactor/tls-iso15118-adapter` (#2225) → `fix/iso15118-disconnect-teardown` (#2393) → this PR. #2225 lands the `ConnectionSSL` adapter + `Connection::last_error()` refactor this builds on; #2393 lands the iso15118 disconnect-teardown fixes. Review/merge them first. The GitHub base branch is still `refactor/tls-iso15118-adapter`, so the diff shown here includes #2393's commits until that PR merges; this PR's own changes are the TLS multi-chain + rotation work described below.

Enables **TLS 1.3 multi-chain certificate selection** and **live V2G certificate rotation** for the ISO 15118 SECC. A charger serving more than one V2G trust chain (e.g. `V2G_ROOT_CA` plus a grid-operator root) now presents the chain the connecting vehicle actually trusts, and picks up renewed/rotated leaf certificates without a process restart.

**Why:** the SECC previously always served `chains[0]` — the per-chain trust anchor never reached `tls::Server`, so its TLS 1.3 selection pool was permanently empty and multi-root deployments could fail the handshake against a vehicle that advertises a `certificate_authorities` list. Leaf rotation also required a module restart.

### libtls — TLS 1.3 chain selection

- `lib/everest/tls/extensions/trusted_ca_keys.{hpp,cpp}`: TLS 1.3 selection is driven by the peer's `certificate_authorities` extension via `select_by_dn_list()` (`SSL_get0_peer_CA_list`); the legacy custom `trusted_ca_keys` extension is scoped to TLS 1.2 and below.
- `handle_certificate_cb` branches on `SSL_version()` and both legs converge on a single `apply_selection_locked()` helper for the apply → fall-back-to-default → abort sequence. The lock requirement is structural: the helper takes a `std::lock_guard` witness, so it cannot be called without `m_mux` held.
- Failure semantics are explicit: when no configured chain matches the peer's advertised CAs the default chain is served and a warning names the decision; when the selected chain fails to apply, the server falls back to the default (logged with the failed chain's leaf subject) — unless the failed chain *is* the default, or no default chain is available at all, in which case the handshake aborts. The no-default abort matters because `use_certificate_and_key()` has already cleared the connection's certificates by that point — continuing would only reach an opaque OpenSSL error mid-handshake. Malformed peer DNs in the comparison are logged (capped at one line per selection, since the peer controls the list size).
- `lib/everest/tls/src/tls.cpp`: the SSL_CTX default certificate is now the **first chain that passed verification** in `init_certificates` (tracked as `m_default_chain_index`), not blindly `chains[0]` — so the no-match / no-extension fallback serves a chain clients can build a path for, and it agrees with the selection pool's `select_default()`. Skipping unusable leading chains is logged, as is the degenerate case where no chain verified at all.
- `init_certificates`: chains dropped at startup/rotation are now observable — a chain failing `verify_chain` logs its leaf subject, and a chain file yielding no certificates logs the file name.
- The libtls selection pool (in the refactor base this stacks on) admits a chain only when it carries a trust anchor — so populating the per-chain `trust_anchor_pem` from the iso15118 side is what activates TLS 1.3 selection.

### iso15118 library — per-chain trust anchor + thread-safe config

- `lib/everest/iso15118/include/iso15118/config.hpp`: `config::ChainConfig` gains `trust_anchor_pem` (the chain's self-signed root PEM). `ocsp_response_files` is now `std::vector<std::optional<std::string>>` — one entry per chain certificate in chain order, `std::nullopt` meaning "no OCSP staple at this position". This keeps the OCSP vector positionally aligned with the certificate chain (the evse_security producer emits placeholder entries; `tls::Server` requires equal sizes and pairs them by index), so a single certificate without OCSP data no longer silently disables stapling for the whole chain. Both `ChainConfig` and `SSLConfig` gain member-wise `operator==`.
- `lib/everest/iso15118/src/iso15118/io/ssl_config_builder.cpp` (declared in the install-excluded `include/iso15118/detail/io/ssl_config_builder.hpp`): `io::build_chain_configs` forwards `trust_anchor_pem` and each OCSP entry (`std::nullopt` → `tls::ConfigItem(nullptr)`) into `tls::Server::certificate_config_t`, so each chain enters the TLS 1.3 / `trusted_ca_keys` selection pool with its staples intact. Extracted from `connection_ssl.cpp` (which still calls it) into its own translation unit so it can be unit-tested.
- `lib/everest/iso15118/src/iso15118/tbd_controller.cpp`: `TbdController` holds the `SSLConfig` behind a monitor and exposes `set_ssl_config` + `ssl_config_snapshot` for thread-safe live updates. `connection_ssl_config()` is the single named seam every new secure connection reads its config from — rotation takes effect for the next incoming connection; established connections are unaffected. `TbdConfig::ssl` is documented as the initial value only.

### Evse15118D20 module — live rotation

- `modules/EVSE/Evse15118D20/charger/evse_security_chain_mapper.{hpp,cpp}`: `map_valid_chains` maps the multi-chain `GetCertificateFullInfoResult` into a fresh `SSLConfig`, forwarding each `CertificateInfo.certificate_root` into `trust_anchor_pem` (a chain without a root is still mapped — it can serve as the default — but logs that it is excluded from TLS 1.3 selection) and preserving OCSP entries verbatim in chain order. The rebuild-and-apply orchestration lives in a testable free function `resync_ssl_config(rebuild, apply)`: an empty rebuild preserves the last-good config, and **all** rebuild failures (`Everest::CmdTimeout` from the evse_security RPC, or any other exception) are caught and logged instead of escaping into the caller's thread — the SECC keeps serving its last-good certificates and retries on the next store update. `handle_certificate_store_update(event, rebuild, apply)` filters irrelevant events cheaply, then delegates to the same function.
- `modules/EVSE/Evse15118D20/charger/ISO15118_chargerImpl.{hpp,cpp}`: subscribes to `evse_security.certificate_store_update`; on a relevant event it rebuilds the config and pushes it into `TbdController` without a module restart. Right after registering the subscription, `ready()` runs a one-shot `resync_ssl_config` to close the startup window — a V2G certificate install completing between the initial config build and the subscription registration would otherwise not be served until the next store update (re-applying an identical config is harmless: apply is a single atomic write). The handler reads `this->controller` under a null guard (the controller is created once in `ready()` and never reset). The startup decision for an empty chain list is a pure helper: `ENFORCE_TLS` throws, other negotiation strategies warn and continue.

### Verification

All unit suites pass on this branch (rebuilt clean; suites re-run after the last commit):

- `tls_test` — 134 cases pass. Selection/apply coverage: `TrustedCaKeys.selectByDnList*` ×7 (first-match-wins, null/empty DN list, multi-anchor chains), `ServerTrustedCaKeysApply` ×4 (clean apply, fallback-to-default on key/leaf mismatch, abort when the only chain fails, no-match keeps default — driven through `handle_certificate_cb` on a raw `SSL*`). E2E: `Tls13MultiChainCertificateAuthorities` (peer CA list picks the alt chain), `Tls13NoMatchServesDefault` (no matching chain still completes the handshake on the default chain), `Tls13NoMatchSkipsUnverifiedDefault` (an unverifiable `chains[0]` is skipped and the first verified chain is served as the default), `MixedVersionServerSelectsAltChain` (one server, TLS 1.2 `trusted_ca_keys` leg + TLS 1.3 `certificate_authorities` leg, negotiated version asserted on each).
- iso15118 unit suite — `build_chain_configs` ×3 (`ssl_config_builder_test`, incl. OCSP staple order + key-password forwarding) and `TbdController SSL config monitor` ×4 (rotation seam, snapshot isolation, concurrent reader/writer) all pass.
- `evse_security_chain_mapper_test` — 22/22 pass (`MapValidChains` ×8 incl. OCSP order/nullopt preservation and rootless chains, `CertStoreUpdateFilter` ×3, `CertUpdateDecision` ×2, `HandleCertStoreUpdate` ×4 incl. RPC-timeout survival, `ResyncSslConfig` ×3 incl. the startup-window re-sync, `StartupEmptyChains` ×2).
- Over-the-wire multi-chain selection is exercised by the libtls TLS 1.3 E2E tests; full SIL validation additionally needs an EVCC that advertises the `certificate_authorities` extension.

_None tracked — TLS / ISO 15118 SECC feature work. Add a link here if there is an associated issue._

### DIFF
--- a/lib/everest/iso15118/include/iso15118/config.hpp
+++ b/lib/everest/iso15118/include/iso15118/config.hpp
@@ -31,6 +31,10 @@ struct ChainConfig {
     std::string path_certificate_key;                  //!< PEM file: private key for the leaf
     std::optional<std::string> private_key_password{}; //!< optional password for the private key
     std::vector<std::string> ocsp_response_files{};    //!< OCSP DER files in chain order
+    //!< inline PEM of the chain's self-signed root; drives TLS 1.3 chain selection
+    //!< (certificate_authorities) and TLS 1.2 trusted_ca_keys. Empty disables this chain
+    //!< from selection (it can still be the default chain when first in the list).
+    std::optional<std::string> trust_anchor_pem{};
 };
 
 /**
--- a/lib/everest/iso15118/include/iso15118/config.hpp
+++ b/lib/everest/iso15118/include/iso15118/config.hpp
@@ -30,11 +30,20 @@ struct ChainConfig {
     std::string path_certificate_chain;                //!< PEM file: leaf followed by intermediate CAs
     std::string path_certificate_key;                  //!< PEM file: private key for the leaf
     std::optional<std::string> private_key_password{}; //!< optional password for the private key
-    std::vector<std::string> ocsp_response_files{};    //!< OCSP DER files in chain order
-    //!< inline PEM of the chain's self-signed root; drives TLS 1.3 chain selection
-    //!< (certificate_authorities) and TLS 1.2 trusted_ca_keys. Empty disables this chain
-    //!< from selection (it can still be the default chain when first in the list).
+    /// OCSP DER files: one entry per chain certificate in chain order; std::nullopt
+    /// means no OCSP staple for that position
+    std::vector<std::optional<std::string>> ocsp_response_files{};
+    /// inline PEM of the chain's self-signed root; drives TLS 1.3 chain selection
+    /// (certificate_authorities) and TLS 1.2 trusted_ca_keys. Empty disables this chain
+    /// from selection (it can still be the default chain when first in the list).
     std::optional<std::string> trust_anchor_pem{};
+
+    bool operator==(const ChainConfig& other) const {
+        return path_certificate_chain == other.path_certificate_chain &&
+               path_certificate_key == other.path_certificate_key &&
+               private_key_password == other.private_key_password && ocsp_response_files == other.ocsp_response_files &&
+               trust_anchor_pem == other.trust_anchor_pem;
+    }
 };
 
 /**
@@ -57,6 +66,15 @@ struct SSLConfig {
     bool enable_tls_key_logging{false};           //!< write SSLKEYLOGFILE entries
     bool enforce_tls_1_3{false};                  //!< require TLS 1.3 minimum
     std::filesystem::path tls_key_logging_path{}; //!< destination directory for keylog
+
+    bool operator==(const SSLConfig& other) const {
+        return backend == other.backend && config_string == other.config_string && chains == other.chains &&
+               path_certificate_v2g_root == other.path_certificate_v2g_root &&
+               path_certificate_mo_root == other.path_certificate_mo_root &&
+               enable_ssl_logging == other.enable_ssl_logging &&
+               enable_tls_key_logging == other.enable_tls_key_logging && enforce_tls_1_3 == other.enforce_tls_1_3 &&
+               tls_key_logging_path == other.tls_key_logging_path;
+    }
 };
 
 } // namespace iso15118::config
--- a/lib/everest/iso15118/include/iso15118/detail/io/ssl_config_builder.hpp
+++ b/lib/everest/iso15118/include/iso15118/detail/io/ssl_config_builder.hpp
@@ -1,0 +1,19 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Pionix GmbH and Contributors to EVerest
+#pragma once
+
+#include <vector>
+
+#include <everest/tls/tls.hpp>
+
+#include <iso15118/config.hpp>
+
+namespace iso15118::io {
+
+/// \brief Map iso15118 SSLConfig chains to tls::Server certificate chain configs.
+/// Each ChainConfig.trust_anchor_pem (the chain's self-signed root PEM) is forwarded
+/// to certificate_config_t.trust_anchor_pem so tls::Server can admit the chain into its
+/// TLS 1.3 / trusted_ca_keys selection pool. Exposed (rather than file-local) for unit testing.
+std::vector<tls::Server::certificate_config_t> build_chain_configs(const config::SSLConfig& cfg);
+
+} // namespace iso15118::io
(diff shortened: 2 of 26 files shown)


corneliusclaussen:
Rebased onto #2675 (`fix/iso15118-visitor-internal-linkage`) and force-pushed as layer 4 of GitHub stack #2776; #2485 and the SAE stack now build on this PR directly (the squash-of-#2226 commit in #2683 is gone). Same 12 commits, no conflicts, new head 377ac1b66. Base branch is now `fix/iso15118-visitor-internal-linkage`, not main: please rebase onto that branch from now on. Local copy: `git fetch && git reset --hard origin/feat/tls-iso15118-multichain`.

a-w50 (dismissed):
From the perspective of functionality, this PR adds quite important features.
Concerning comments and tests:
- there are too much tests, that sometimes don't even pin the behaviour of the classes touched in this PR, but rather other things (for example the TbdController tests for the monitor basically only test the lib/util/monitor)
- comments are quite often explain reasoning, why something was changed.  They should rather document what is, and only if it is not obvious

Some other findings, which should be noted for future refactoring:
- `controller` cross-thread access is unenforced. Race-free today only because it is written once  before the subscription exists; that invariant lives in a comment. There is no destructor, the subscription is never unsubscribed, and the null guard at `ISO15118_chargerImpl.cpp:336` is unreachable
- lock "witness" parameter overclaims.** `std::lock_guard<std::mutex>` carries no mutex identity, so a  guard for *any* mutex satisfies `apply_selection_locked()`. The PR body's "structural … cannot be called without  `m_mux` held" is not true — I would drop the parameter and keep the `_locked` name.
- `init_ssl()` bounds guard misses the empty case. `tls.cpp:1125` falls back to index `0`, which is itself out of range when `cfg.chains` is empty. Pre-existing UB, but the new check makes it look handled — add an emptiness guard.
- could `m_default_chain_index` be retired? `SSL_CTX_use_cert_and_key` would let `init_ssl()` install the loaded default chain directly, removing the hand-maintained agreement between cfg.chains and m_chains
- the `evse_security_chain_mapper.cpp` got a lot of functionality that doesn't deal at all with chain mapping

Despite these issues I approve and hope these things will be cleaned and fixed in near future.  We nevertheless need to come up with a better design / separation of tls connections and it's integration into existing modules.

a-w50 (approved):

PR #2227, opened 2026-05-20
feat(io/tls): TLS client and server policies (superseded by #2558 to #2565)

mlitre:
## Superseded

Closed in favor of an eight PR series that replaces this branch. The work here was one branch touching
`lib/everest/io`, `lib/everest/tls` and `lib/everest/iso15118` at once; it has been split so each piece
sits on the smallest base it needs and can be reviewed by the people who own that library.

Rooted on `main`, reviewable in parallel:

- #2558 `fix(iso15118): extract the listen socket helper` — the `lib/everest/iso15118` changes, separated
  out as requested on this PR
- #2559 `feat(tls): wrap an already connecting descriptor` — the `lib/everest/tls` changes, and the root
  of the chain below
- #2560 `fix(io): carry the connect errno and name the retry delay` — independent connect errno fix

The chain, each on the one before it:

- #2561 `fix(io): make event handler registration truthful`
- #2562 `feat(io): tri-state connection state and bounded tx buffer`
- #2563 `feat(io): event loop driven handshake phase`
- #2564 `feat(io/tls): event loop driven TLS client and server` — the direct successor to this PR
- #2565 `refactor(chargebridge): drop tx gates and recover on error` — forks off #2562, not off #2563

#2564 is not a rebase of this branch. The TLS client was rebuilt as a one line alias over the generic
event client rather than a hand written class, so the version here that was added and then deleted within
the same branch never reaches a reviewer.

Each of the eight carries the full series in its own description, with its position relative to `main`.

### DIFF
--- a/lib/everest/io/BUILD.bazel
+++ b/lib/everest/io/BUILD.bazel
@@ -43,6 +43,7 @@ cc_library(
     }),
     visibility = ["//visibility:public"],
     deps = [
+        "//lib/everest/tls",
         "//lib/everest/util",
         "@mosquitto",
     ],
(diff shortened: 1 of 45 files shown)


SebaLukas:
The libiso15118 changes could also be handled in a seperated PR

PR #2229, opened 2026-05-20
feat(ocpp v21): K28 dynamic charging profiles

mlitre:
Implement OCPP 2.1 K28 adaptive Dynamic charging profile flow: periodic PullDynamicScheduleUpdate worker, CSMS-pushed UpdateDynamicSchedule handling, deadline tracking via the monitor-guarded DynamicScheduleManager, non-blocking teardown of in-flight workers.

Also harden the underlying sqlite layer that K28's concurrent writer exposed:
- sqlite::Statement::column_text throws on NULL instead of invoking std::string(nullptr) and aborting.
- Connection sets sqlite3_busy_timeout on open so SQLITE_BUSY is retried rather than surfaced raw to readers.
- v2 database_handler and smart_charging data-reading loops use `while (step() == SQLITE_ROW)` so non-row states surface as exceptions instead of dereferencing an invalid column.


mlitre:
Decomposed into:
- https://github.com/EVerest/EVerest/pull/2277
- https://github.com/EVerest/EVerest/pull/2278
- https://github.com/EVerest/EVerest/pull/2279

PR #2230, opened 2026-05-20
fix(ocpp): reply RpcFrameworkError on orphan call

mlitre:
## Describe your changes

When the OCPP message queue receives a `CallResult` or `CallError` whose unique id does not match the request currently in flight, it now replies to the CSMS with a `CallError(RpcFrameworkError)` instead of mishandling the frame:

- **No request in flight (unsolicited):** emit `CallError(RpcFrameworkError)` for the stray uid and return. Previously this was logged as "should not happen" and dropped without informing the CSMS.
- **A different request is in flight (uid mismatch):** emit `CallError(RpcFrameworkError)` for the stray uid and leave the in-flight request untouched. A mismatch is not treated as a timeout, so the in-flight request still completes when its own matching response arrives, and its own timeout timer stays responsible for retry/expiry. Previously the mismatched frame fell through into the response-handling path.

The behavior lives in the `CALLRESULT`/`CALLERROR` branch of `MessageQueue::receive()` in `lib/everest/ocpp/include/ocpp/common/message_queue.hpp`. Unit tests in `lib/everest/ocpp/tests/lib/ocpp/common/test_message_queue.cpp` cover the unsolicited case, and the uid-mismatch case for both `CallResult` and `CallError`: the framework error is emitted, the in-flight call is left pending, and it still completes on its real response.

Verified with the `MessageQueueTest` suite in `libocpp_unit_tests` (18/18 passing).

### DIFF
--- a/lib/everest/ocpp/include/ocpp/common/message_queue.hpp
+++ b/lib/everest/ocpp/include/ocpp/common/message_queue.hpp
@@ -849,24 +849,31 @@ public:
             }
         }
 
-        // TODO(kai): what happens if we receive a CallResult or CallError out of order?
         if (enhanced_message.messageTypeId == MessageTypeId::CALLRESULT ||
             enhanced_message.messageTypeId == MessageTypeId::CALLERROR) {
             {
                 const std::lock_guard<std::recursive_mutex> lk(this->next_message_mutex);
                 next_message_to_send.reset();
             }
-            // we need to remove Call messages from in_flight if we receive a CallResult OR a CallError
-
-            // TODO(kai): we need to do some error handling in the CallError case
             std::unique_lock<std::recursive_mutex> lk(this->message_mutex);
             if (this->in_flight == nullptr) {
-                EVLOG_error << "Received a CALLRESULT OR CALLERROR without a message in flight, this should not happen";
+                EVLOG_error << "Received a CALLRESULT OR CALLERROR without a message in flight; "
+                               "replying with RpcFrameworkError for uid "
+                            << enhanced_message.uniqueId;
+                this->push_call_error(CallError{enhanced_message.uniqueId, "RpcFrameworkError",
+                                                "Received unsolicited CallResult/CallError", json::object()});
                 return enhanced_message;
             }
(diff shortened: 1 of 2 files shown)


lib/everest/ocpp/include/ocpp/common/message_queue.hpp
```diff
@@ -849,24 +849,31 @@ template <typename M> class MessageQueue {
...
                 return enhanced_message;
             }
             if (this->in_flight->uniqueId() != enhanced_message.uniqueId) {
                 EVLOG_error << "Received a CALLRESULT OR CALLERROR with mismatching uid: "
                             << this->in_flight->uniqueId() << " != " << enhanced_message.uniqueId;
+                EnhancedMessage<M> stale;
+                stale.uniqueId = this->in_flight->uniqueId();
+                stale.offline = true;
+                this->in_flight->promise.set_value(stale);
+                this->reset_in_flight();
```
Pietfried: This looks like we just drop the message in flight without retrying. Especially for transaction related messages, retrying is prefered imo. 

I haven't tested, but maybe this approach would be better:

```cpp
if (this->in_flight->uniqueId() != enhanced_message.uniqueId) {
                EVLOG_error << "Received a CALLRESULT OR CALLERROR with mismatching uid: "
                            << this->in_flight->uniqueId() << " != " << enhanced_message.uniqueId;
                this->handle_timeout_or_callerror(std::nullopt);
                this->push_call_error(CallError{enhanced_message.uniqueId, "RpcFrameworkError",
                                                "messageId does not match in-flight call", json::object()});
                return enhanced_message;
}
```

mlitre: Probably a good idea to retry, I'll take a look. I would reuse the same max retries config key or this could go on forever

mlitre: Actually even better in my opinion I just respond with CallError and leave the timeout and retry mechanisms intact

mlitre:
Not allowed to send CallError after a CallResult

PR #2236, opened 2026-05-21
feat(yeti-simulator): add sending of fake SignedMeterValues

nzbr:
## Describe your changes

Makes the simulated powermeter, which is part of the YetiSimulator driver send out static, configurable meter value blobs. This will allow testing how they are handled throughout EVerest

## Checklist before requesting a review
- [x] I have performed a self-review of my code
- [ ] I have made corresponding changes to the documentation
- [x] I read the [contribution documentation](https://everest.github.io/nightly/project/contributing.html) and made sure that my changes meet its requirements

---
This is **part 1 of 5 in a stack** made with GitButler:
- <kbd>&nbsp;5&nbsp;</kbd> #2259 
- <kbd>&nbsp;4&nbsp;</kbd> #2239 
- <kbd>&nbsp;3&nbsp;</kbd> #2252 
- <kbd>&nbsp;2&nbsp;</kbd> #2214 
- <kbd>&nbsp;1&nbsp;</kbd> #2236 👈

PR #2214, opened 2026-05-18
Allow powermeter drivers to send signed meter values on transaction start

nzbr:
## Describe your changes

Adds the ability for powermeter drivers to send signed meter values on transaction start, instead of only after the transaction is stopped. The value is additionally exposed via the stable API and passed to a connected CSMS in the OCPP start transaction message.

## Checklist before requesting a review
- [x] I have performed a self-review of my code
- [ ] I have made corresponding changes to the documentation
- [x] I read the [contribution documentation](https://everest.github.io/nightly/project/contributing.html) and made sure that my changes meet its requirements

---
This is **part 2 of 5 in a stack** made with GitButler:
- <kbd>&nbsp;5&nbsp;</kbd> #2259 
- <kbd>&nbsp;4&nbsp;</kbd> #2239 
- <kbd>&nbsp;3&nbsp;</kbd> #2252 
- <kbd>&nbsp;2&nbsp;</kbd> #2214 👈 
- <kbd>&nbsp;1&nbsp;</kbd> #2236

PR #2239, opened 2026-05-22
test: add an OCPP test that checks if the signed meter values are actually sent

nzbr:
## Describe your changes

Adds a test that checks if the signed meter values from the simulated powermeter are correctly sent to the CSMS

---
This is **part 4 of 5 in a stack** made with GitButler:
- <kbd>&nbsp;5&nbsp;</kbd> #2259 
- <kbd>&nbsp;4&nbsp;</kbd> #2239 👈 
- <kbd>&nbsp;3&nbsp;</kbd> #2252 
- <kbd>&nbsp;2&nbsp;</kbd> #2214 
- <kbd>&nbsp;1&nbsp;</kbd> #2236

### DIFF
--- a/tests/ocpp_tests/test_sets/ocpp16/eichrecht.py
+++ b/tests/ocpp_tests/test_sets/ocpp16/eichrecht.py
@@ -9,9 +9,9 @@ from everest.testing.ocpp_utils.charge_point_utils import (
 )
 from everest.testing.ocpp_utils.central_system import ChargePoint16
 from everest.testing.ocpp_utils.fixtures import test_utility, charge_point_v16
-from everest_test_utils import get_everest_config_path_str
+from everest_test_utils import get_everest_config_path_str, OcppTestConfiguration
 
-from ocpp.v16 import call_result
+from ocpp.v16 import call, call_result
 
 
 @pytest.mark.asyncio
@@ -61,7 +61,7 @@ async def test_meter_public_key(
     assert any(
         entry['key'] == "MeterPublicKey2" and entry['value'] == "TESTPUBLICKEY2"
         for entry in response.configuration_key)
-    
+
     await charge_point_v16.get_configuration_req(key=["MeterPublicKey"])
 
     assert await wait_for_and_validate(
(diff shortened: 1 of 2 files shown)


tests/ocpp_tests/test_sets/ocpp16/eichrecht.py
```diff
@@ -125,3 +125,70 @@ async def test_meter_public_key(
...
+            test_utility,
+            charge_point_v16,
+            "StopTransaction",
+            {},
+        )
+    )
+
+    assert stop_transaction_msg is not None
+    assert stop_transaction_msg.transaction_data is not None
+    assert len(stop_transaction_msg.transaction_data) == 2
```
Pietfried: This is a bit fragile, depending on the time clock-aligned meter values could theoretically be part of `transaction_data` resulting in len > 2 . It would be fine if you just test for if meter values with context `Transaction.Begin` and `Transaction.End` are present **exactly** once and check the respective data of those.

lib/everest/ocpp/lib/ocpp/v16/charge_point_impl.cpp
```diff
@@ -4447,13 +4447,16 @@ void ChargePointImpl::on_transaction_started(const std::int32_t& connector, cons
...
-            get_signed_meter_value(signed_meter_value.value(), ReadingContext::Transaction_Begin, timestamp);
-        transaction->add_meter_value(meter_value);
+        meter_value = get_signed_meter_value(signed_meter_value.value(), ReadingContext::Transaction_Begin, timestamp);
+        transaction->add_meter_value(meter_value.value());
     }
 
     this->start_transaction(transaction);
+    if (meter_value) {
+        this->send_meter_value(connector, meter_value.value());
+    }
```
mlitre, line 4463: This shouldn't be needed as meter value timing / sends should be handled by the transaction already

nzbr: I had the wrong base-branch set. This should not have appeared in this PR. See my reply here: https://github.com/EVerest/EVerest/pull/2252#issuecomment-4594639072

nzbr:
If I interpreted that correctly, it looks like a part of Codeacy's analyzer crashed on their end

Pietfried (dismissed):

nzbr:
Needs a re-approve because I squashed the commits

Pietfried (approved):

PR #2243, opened 2026-05-22
Fix CMake packaging for non-EDM consumers

hikinggrass:
## Describe your changes

conversions, helpers, io, ocpp and yaml had some issues

Add `evc_setup_package_no_export` CMake function that replaces most hand-written config.cmake.in files

Turn generate_cpp_files into an INTERFACE library so it can be properly installed

Tested with Yocto scarthgap, everest-examples (EDM/Yocto) and an external application using EVerestAPI


lib/everest/conversions/ocpp/CMakeLists.txt
```diff
@@ -60,3 +77,25 @@ target_link_libraries(ocpp_conversions
...
+    ev_packaging(
+        NAME everest-ocpp-conversions
+        LIBRARY_NAME ocpp_conversions
+        NAMESPACE everest
+        TYPE STATIC
+    )
+endif()
+
+install(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/include/everest
+    DESTINATION include
```
andistorm:
```suggestion
    DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}
```
or should it be
```suggestion
    DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/everest
```
?

cmake/ev-packaging.cmake
```diff
@@ -0,0 +1,114 @@
...
+
+    set(LIBRARY_PACKAGE_NAME "${OPTNS_NAME}")
+    set(LIBRARY_PACKAGE_CMAKE_INSTALL_DIR ${CMAKE_INSTALL_LIBDIR}/cmake/${LIBRARY_PACKAGE_NAME})
+    set(LIBRARY_NAME "${OPTNS_LIBRARY_NAME}")
+    set(LIBRARY_TYPE "${OPTNS_TYPE}")
+
+    set(LIBRARY_PACKAGE_VERSION ${PROJECT_VERSION})
+    if (OPTNS_VERSION)
+        set(LIBRARY_PACKAGE_VERSION ${OPTNS_VERSION})
+    endif()
```
andistorm: Do we not want to enforce using the project version, since source is in the same repository?

cmake/ev-packaging.cmake
```diff
@@ -0,0 +1,114 @@
+function(ev_packaging)
```
andistorm: The name doesn't really describe what the function does. Why not staying with `setup_package` and naming it `ev_setup_package`.

Why not replacing `evc_setup_package` completly?

lib/everest/conversions/ocpp/CMakeLists.txt
```diff
@@ -29,6 +29,23 @@ target_link_libraries(ocpp_evse_security
...
+        LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
+        RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR}
+        INCLUDES DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/everest
+    )
+    ev_packaging(
+        NAME everest-ocpp-evse-security
+        LIBRARY_NAME ocpp_evse_security
+        NAMESPACE everest
+        TYPE STATIC
+    )
```
andistorm: Why is ev_packaging only applied for non EDM builds?

cmake/ev-packaging.cmake
hikinggrass: Renamed it to `evc_setup_package_no_export` to clarify the intent. If we integrate `everest-cmake` into the monorepo I would consider replacing it, until then it's a bit risky to change the behavior of this function when it's only needed in certain narrow usecases

cmake/ev-packaging.cmake
hikinggrass: I've dropped this completely, originally this is from the `evc_setup_package` definition in https://github.com/EVerest/everest-cmake/blob/e2a532602deaf0abdabbb340ee6fd5189573e169/packaging-helper.cmake#L79 where it apparently isn't used either

lib/everest/conversions/ocpp/CMakeLists.txt
hikinggrass: It addresses the same issues found in eg https://github.com/EVerest/EVerest/pull/1975 but without having to create individual `config.cmake.in` files with bespoke code

lib/everest/conversions/ocpp/CMakeLists.txt
hikinggrass: done

cmake/ev-packaging.cmake
hikinggrass: removed

andistorm (dismissed):
LGTM 🐧

djchhp (dismissed):

mlitre (dismissed):
LGTM!

djchhp (approved):

andistorm (approved):
🐧

mlitre (approved):

PR #2274, opened 2026-06-03
docs(project): add deprecation policy and active deprecation index

Pietfried:
## Describe your changes

* Introduces a deprecation policy for the EVerest public API
* Updates releases-and-versioning and breaking-changes with cross-references to the new documents and aligns RST heading style

### DIFF
--- a/docs/source/project/releases/deprecation-policy.rst
+++ b/docs/source/project/releases/deprecation-policy.rst
@@ -1,0 +1,207 @@
+.. _project-deprecation-policy:
+
+##############################
+Deprecation Policy for EVerest
+##############################
+
+This document describes how EVerest handles the deprecation of public API
+components. It complements the :ref:`project-release-and-versioning` and
+:ref:`project-breaking-changes` documents and applies to all components covered
+by the EVerest public API.
+
+********
+Overview
+********
+
+EVerest follows a pragmatic deprecation approach: public API components are not
+removed without warning, and integrators always have at least one stable release
+cycle to migrate before any breaking change occurs.
+
+*****
+Scope
+*****
+
+This policy applies to all components of the EVerest public API:
+
+- External AsyncAPIs
+- Energy Management JSON RPC API
+- Configuration, storage, and module contracts (EVerest module configuration and module availability,
+  OCPP configuration)
+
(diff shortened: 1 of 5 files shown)


docs/source/project/releases/deprecation-policy.rst
```diff
@@ -0,0 +1,142 @@
...
+   deprecated. A replacement (if applicable) is documented, and a planned
+   removal release is announced.
+3. **Removed**: The component is no longer available. Its removal is a breaking
+   change and follows the rules defined in :ref:`project-breaking-changes`.
+
+**************************
+Minimum Deprecation Period
+**************************
+
+A public API component **may not be removed without first being deprecated**.
```
james-ctc: I recommend using [RFC 2119](https://datatracker.ietf.org/doc/html/rfc2119) definitions.

`may` This word, or the adjective "OPTIONAL", mean that an item is truly optional.

In this instance `must` is a better choise.

dorezyuk:
i think if we really want to move in that direction we have also to add the "nightly" side - so newly merged features remain nightly until (add here some smart thoughts) and can be yanked without warning. Otherwise merging "experimental" features will become quite a burden

Pietfried:
> i think if we really want to move in that direction we have also to add the "nightly" side - so newly merged features remain nightly until (add here some smart thoughts) and can be yanked without warning. Otherwise merging "experimental" features will become quite a burden

We could introduce a tbd `experimental` marker for components of the public API to address this and decide case by case. Many features we develop (e.g. within ISO15118 or OCPP) are not adding / changing the public API though, so I am not expecting that we are running into many of such cases

Pietfried:
@dorezyuk @barsnick @corneliusclaussen I tried to incorporate Dimas feedback by introducing a concept of `experimental` public API components, allowing to keep those out of deprecation policy scope and stability guarantees. This allows the project to ship and iterate on new features quickly before committing to their long term stability

Pietfried:
@dorezyuk @barsnick @corneliusclaussen After the discussion in the TSC this is now ready for final review

dorezyuk (approved):

barsnick (approved):

corneliusclaussen (approved):

PR #2276, opened 2026-06-04
refactor(ConnectivityManager): Introduce ConnectivityManagerConfiguration replacing DeviceModelAbstract

Pietfried:
## Describe your changes

Decouple from DeviceModelAbstract via ConnectivityManagerConfiguration interface:

- Introduce the pure-virtual ConnectivityManagerConfiguration as the sole dependency of ConnectivityManager, replacing the direct reference to DeviceModelAbstract.
- DeviceModelAbstract and ChargePointConfigurationInterface both inherit from it, enabling a single ConnectivityManager to serve OCPP 1.6 and 2.x.
- Move get_websocket_connection_options and the identity/password resolution helpers into each implementation
- fix set_security_ctrl_* to use set_read_only_value for ReadOnly DM variables.

This PR prepares the usage of the ConnectivityManager also for OCPP1.6.

### DIFF
--- a/lib/everest/ocpp/include/ocpp/common/connectivity_manager_configuration.hpp
+++ b/lib/everest/ocpp/include/ocpp/common/connectivity_manager_configuration.hpp
@@ -1,0 +1,79 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2020 - 2024 Pionix GmbH and Contributors to EVerest
+
+#pragma once
+
+#include <ocpp/common/websocket/websocket_base.hpp>
+#include <ocpp/v2/ocpp_types.hpp>
+
+#include <cstdint>
+#include <optional>
+#include <string>
+#include <vector>
+
+namespace ocpp {
+
+/// \brief Configuration interface used by ConnectivityManager to persist and retrieve all network-configuration
+///        data it needs, independently of the OCPP version in use.
+class ConnectivityManagerConfiguration {
+public:
+    virtual ~ConnectivityManagerConfiguration() = default;
+
+    /// \brief Get the comma-separated network-slot priority string (e.g. "1,2,0").
+    /// \return The priority string, or an empty string when none is configured.
+    virtual std::string get_network_configuration_priority() = 0;
+
+    /// \brief Persist an updated priority string.
+    virtual void set_network_configuration_priority(const std::string& priority, const std::string& source) = 0;
+
+    /// \brief Read the NetworkConnectionProfile stored for \p slot.
+    /// \return The profile, or std::nullopt if the slot is absent or malformed.
+    virtual std::optional<ocpp::v2::NetworkConnectionProfile> read_network_connection_profile(std::int32_t slot) = 0;
+
+    /// \brief Persist a NetworkConnectionProfile for \p slot.
+    /// \return true on success.
+    virtual bool write_network_connection_profile(std::int32_t slot, const ocpp::v2::NetworkConnectionProfile& profile,
+                                                  const std::string& source) = 0;
+
+    /// \brief Erase all persisted data for \p slot.
+    virtual void clear_network_connection_profile(std::int32_t slot) = 0;
+
+    /// \brief Return true if security-profile 0 connections are permitted.
+    virtual bool get_allow_security_level_zero_connections() = 0;
+
+    /// \brief Return the currently active security profile level.
+    virtual std::int32_t get_security_profile() = 0;
+
+    /// \brief Return the timeout (in seconds) to wait for a network-configuration callback result,
+    ///        or std::nullopt to use the built-in default (60 s).
+    virtual std::optional<std::int32_t> get_network_config_timeout() = 0;
+
+    /// \brief Build and return the WebsocketConnectionOptions for the given \p slot.
+    ///
+    /// The implementation is responsible for reading the network connection profile for \p slot,
+    /// resolving the identity and basic-auth password (applying any per-slot overrides), parsing
+    /// and validating the CSMS URI, and populating all connection parameters (ciphers, ping/pong
+    /// intervals, TLS flags, retry settings, etc.).
+    ///
+    /// \return The populated options, or std::nullopt if the slot has no valid profile or the CSMS
+    ///         URL is malformed.
+    virtual std::optional<WebsocketConnectionOptions> get_websocket_connection_options(std::int32_t slot) = 0;
+
+    /// \brief Persist the security profile of the currently active connection.
+    virtual void set_active_security_profile(std::int32_t security_profile, const std::string& source) = 0;
+
+    /// \brief Persist the currently active network-profile slot.
+    virtual void set_active_network_profile_slot(std::int32_t slot, const std::string& source) = 0;
+
+    /// \brief Persist the negotiated OCPP protocol version for a specific slot.
+    virtual void set_per_slot_ocpp_version(std::int32_t slot, const std::string& version,
+                                           const std::string& source) = 0;
+
+    /// \brief Persist the security profile on the SecurityCtrlr component after a successful connect.
+    virtual void set_security_ctrl_security_profile(std::int32_t security_profile, const std::string& source) = 0;
+
+    /// \brief Persist the identity on the SecurityCtrlr component after a successful connect.
+    virtual void set_security_ctrl_identity(const std::string& identity, const std::string& source) = 0;
+};
+
+} // namespace ocpp
--- a/lib/everest/ocpp/include/ocpp/v16/charge_point_configuration_connectivity.hpp
+++ b/lib/everest/ocpp/include/ocpp/v16/charge_point_configuration_connectivity.hpp
@@ -1,0 +1,35 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2020 - 2026 Pionix GmbH and Contributors to EVerest
+
+#pragma once
+
+#include <ocpp/v16/charge_point_configuration_interface.hpp>
+
+namespace ocpp::v16 {
+
+/// \brief Implements the ConnectivityManagerConfiguration part of ChargePointConfigurationInterface for OCPP 1.6.
+///
+/// The behaviour is identical regardless of how the configuration is stored (JSON config file or device model):
+/// it depends only on the OCPP 1.6 getters/setters declared on ChargePointConfigurationInterface, which are
+/// resolved via virtual dispatch to the concrete implementation. This class remains abstract; it does not
+/// implement those getters/setters.
+class ChargePointConfigurationConnectivity : public ChargePointConfigurationInterface {
+public:
+    std::string get_network_configuration_priority() override;
+    void set_network_configuration_priority(const std::string& priority, const std::string& source) override;
+    std::optional<ocpp::v2::NetworkConnectionProfile> read_network_connection_profile(int32_t slot) override;
+    bool write_network_connection_profile(int32_t slot, const ocpp::v2::NetworkConnectionProfile& profile,
+                                          const std::string& source) override;
+    void clear_network_connection_profile(int32_t slot) override;
+    bool get_allow_security_level_zero_connections() override;
+    int32_t get_security_profile() override;
+    std::optional<int32_t> get_network_config_timeout() override;
+    std::optional<WebsocketConnectionOptions> get_websocket_connection_options(int32_t slot) override;
+    void set_active_security_profile(int32_t security_profile, const std::string& source) override;
+    void set_active_network_profile_slot(int32_t slot, const std::string& source) override;
+    void set_per_slot_ocpp_version(int32_t slot, const std::string& version, const std::string& source) override;
+    void set_security_ctrl_security_profile(int32_t security_profile, const std::string& source) override;
+    void set_security_ctrl_identity(const std::string& identity, const std::string& source) override;
+};
+
+} // namespace ocpp::v16
--- a/lib/everest/ocpp/lib/ocpp/common/connectivity_manager.cpp
+++ b/lib/everest/ocpp/lib/ocpp/common/connectivity_manager.cpp
@@ -6,9 +6,6 @@
 #include <fstream>
 
 #include <everest/logging.hpp>
-#include <ocpp/v2/ctrlr_component_variables.hpp>
-#include <ocpp/v2/device_model.hpp>
-#include <ocpp/v2/utils.hpp>
 
 namespace {
 const auto WEBSOCKET_INIT_DELAY = std::chrono::seconds(2);
@@ -21,20 +18,13 @@ constexpr std::int32_t default_network_config_timeout_seconds = 60;
 namespace ocpp {
 
 using NetworkConnectionProfile = ocpp::v2::NetworkConnectionProfile;
-using AttributeEnum = ocpp::v2::AttributeEnum;
-using DeviceModel = ocpp::v2::DeviceModel;
-using DeviceModelAbstract = ocpp::v2::DeviceModelAbstract;
 using OCPPInterfaceEnum = ocpp::v2::OCPPInterfaceEnum;
 using SetNetworkProfileRequest = ocpp::v2::SetNetworkProfileRequest;
-namespace ControllerComponentVariables = ocpp::v2::ControllerComponentVariables;
-namespace ControllerComponents = ocpp::v2::ControllerComponents;
-namespace NetworkConfigurationComponentVariables = ocpp::v2::NetworkConfigurationComponentVariables;
-namespace utils = ocpp::v2::utils;
 
-ConnectivityManager::ConnectivityManager(ocpp::v2::DeviceModelAbstract& device_model,
+ConnectivityManager::ConnectivityManager(ocpp::ConnectivityManagerConfiguration& configuration,
                                          std::shared_ptr<EvseSecurity> evse_security, const fs::path& share_path) :
-    device_model{device_model},
-    evse_security{evse_security},
+    configuration{configuration},
+    evse_security{std::move(evse_security)},
     share_path{share_path},
     websocket{nullptr},
     message_callback([](const std::string& message) {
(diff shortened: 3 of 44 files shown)


lib/everest/ocpp/include/ocpp/common/connectivity_manager_configuration.hpp
```diff
@@ -0,0 +1,79 @@
...
+    /// \brief Get the comma-separated network-slot priority string (e.g. "1,2,0").
+    /// \return The priority string, or an empty string when none is configured.
+    virtual std::string get_network_configuration_priority() = 0;
+
+    /// \brief Persist an updated priority string.
+    virtual void set_network_configuration_priority(const std::string& priority, const std::string& source) = 0;
+
+    /// \brief Read the NetworkConnectionProfile stored for \p slot.
+    /// \return The profile, or std::nullopt if the slot is absent or malformed.
+    virtual std::optional<ocpp::v2::NetworkConnectionProfile> read_network_connection_profile(int32_t slot) = 0;
```
james-ctc: observation:
should be `std::int32_t` and if slot numbers are always positive then should be `std::uint32_t`

lib/everest/ocpp/lib/ocpp/v16/charge_point_configuration_devicemodel.cpp
```diff
@@ -3298,4 +3299,118 @@ std::set<MessageType> ChargePointConfigurationDeviceModel::getSupportedMessageTy
     return supported_message_types_receiving;
 }
 
+// ConnectivityManagerConfiguration implementations
```
mlitre: We have exactly the same code block in lib/everest/ocpp/lib/ocpp/v16/charge_point_configuration.cpp, is that on purpose that we have this duplication?

lib/everest/ocpp/include/ocpp/common/connectivity_manager_configuration.hpp
```diff
@@ -0,0 +1,79 @@
...
+                                                  const std::string& source) = 0;
+
+    /// \brief Erase all persisted data for \p slot.
+    virtual void clear_network_connection_profile(int32_t slot) = 0;
+
+    /// \brief Return true if security-profile 0 connections are permitted.
+    virtual bool get_allow_security_level_zero_connections() = 0;
+
+    /// \brief Return the currently active security profile level.
+    virtual int32_t get_security_profile() = 0;
```
james-ctc: option: could be an enum especially as there are such a small number of valid values

lib/everest/ocpp/include/ocpp/common/connectivity_manager_configuration.hpp
```diff
@@ -0,0 +1,79 @@
...
+
+    /// \brief Return the timeout (in seconds) to wait for a network-configuration callback result,
+    ///        or std::nullopt to use the built-in default (60 s).
+    virtual std::optional<int32_t> get_network_config_timeout() = 0;
```
james-ctc: option: use std::chrono - it is then clearer what the value is

lib/everest/ocpp/lib/ocpp/v16/charge_point_configuration_devicemodel.cpp
```diff
@@ -3298,4 +3299,118 @@ std::set<MessageType> ChargePointConfigurationDeviceModel::getSupportedMessageTy
...
+    profile.ocppInterface = ocpp::v2::OCPPInterfaceEnum::Any; // not used for OCPP1.6
+    profile.ocppTransport = ocpp::v2::OCPPTransportEnum::JSON;
+    profile.messageTimeout = 10;
+    profile.identity = getChargePointId();
+    profile.basicAuthPassword = getAuthorizationKey();
+    return profile;
+}
+
+bool ChargePointConfigurationDeviceModel::write_network_connection_profile(
+    int32_t slot, const ocpp::v2::NetworkConnectionProfile& profile, const std::string& /*source*/) {
```
mlitre: nit: comment out profile as well since it is not used

lib/everest/ocpp/lib/ocpp/common/connectivity_manager.cpp
```diff
@@ -21,19 +18,12 @@ constexpr std::int32_t default_network_config_timeout_seconds = 60;
...
-namespace ControllerComponents = ocpp::v2::ControllerComponents;
-namespace NetworkConfigurationComponentVariables = ocpp::v2::NetworkConfigurationComponentVariables;
-namespace utils = ocpp::v2::utils;
 
-ConnectivityManager::ConnectivityManager(ocpp::v2::DeviceModelAbstract& device_model,
+ConnectivityManager::ConnectivityManager(ocpp::ConnectivityManagerConfiguration& configuration,
                                          std::shared_ptr<EvseSecurity> evse_security, const fs::path& share_path) :
-    device_model{device_model},
+    configuration{configuration},
     evse_security{evse_security},
```
james-ctc: evse_security{evse_security} should be evse_security{std::move(evse_security)}
or perhaps better:
ConnectivityManager::ConnectivityManager(ocpp::ConnectivityManagerConfiguration& configuration,
    const std::shared_ptr<EvseSecurity>& evse_security, const fs::path& share_path)

We should try and avoid copying objects onto the stack

lib/everest/ocpp/lib/ocpp/v16/charge_point_configuration.cpp
```diff
@@ -4040,4 +4041,117 @@ std::optional<ConfigurationStatus> ChargePointConfiguration::set(const CiString<
...
+    }
+    const auto security_profile = getSecurityProfile();
+    try {
+        auto uri = Uri::parse_and_validate(getCentralSystemURI(), charge_point_id, security_profile);
+
+        WebsocketConnectionOptions opts{{OcppProtocolVersion::v16},
+                                        uri,
+                                        security_profile,
+                                        getAuthorizationKey(),
+                                        std::chrono::seconds(10),
```
mlitre: Is there a reason why this is hardcoded here?

mlitre (commented):
Not in this PR, but related: `ChargePointImpl::get_ws_connection_options()` (charge_point_impl.cpp:404) already builds the same `WebsocketConnectionOptions` from `this->configuration`. The two new `get_websocket_connection_options(slot)` methods here make that three copies of the same thing, and the version_information.txt read below it overlaps with `ConnectivityManager::read_everest_version()`.

Could be worth having one builder on `ChargePointConfigurationInterface` and having this method call it, then just add the everest version. That gets us down to a single copy.

It's not a straight swap though: this returns by value and lets `Uri::parse_and_validate` throw, while the new methods return `optional` and catch it, so the caller handling would need adjusting.

Fine to leave as is and do it in a follow-up if that's cleaner.

lib/everest/ocpp/lib/ocpp/v2/device_model.cpp
```diff
@@ -1046,5 +1047,153 @@ std::int32_t DeviceModel::clear_custom_monitors() {
...
+        slot_pwd.has_value() && !slot_pwd->empty()) {
+        EVLOG_debug << "Using per-slot BasicAuthPassword for slot " << slot;
+        return *slot_pwd;
+    }
+    return get_optional_value<std::string>(ControllerComponentVariables::BasicAuthPassword);
+}
+
+std::optional<WebsocketConnectionOptions> DeviceModel::get_websocket_connection_options(int32_t slot) {
+    const auto charge_point_id = get_value<std::string>(ControllerComponentVariables::ChargePointId);
+    if (charge_point_id.find(':') != std::string::npos) {
```
james-ctc, line 1117: note `:` could be escaped `%3A`

james-ctc (dismissed):
some suggestions only

lib/everest/ocpp/lib/ocpp/v16/charge_point_configuration.cpp
Pietfried: The reason is just that it was also hardcoded before :hand_over_mouth:

lib/everest/ocpp/lib/ocpp/v16/charge_point_configuration_devicemodel.cpp
Pietfried: I just oversaw the potential of moving it into a common base class. Adressed in the last commit.

lib/everest/ocpp/include/ocpp/v16/charge_point_configuration_connectivity.hpp
```diff
@@ -0,0 +1,38 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2020 - 2026 Pionix GmbH and Contributors to EVerest
+
+#ifndef OCPP_V16_CHARGE_POINT_CONFIGURATION_CONNECTIVITY_HPP
+#define OCPP_V16_CHARGE_POINT_CONFIGURATION_CONNECTIVITY_HPP
```
mlitre: Replace with pragma once

lib/everest/ocpp/include/ocpp/v16/charge_point_configuration_connectivity.hpp
```diff
@@ -0,0 +1,38 @@
...
+/// The behaviour is identical regardless of how the configuration is stored (JSON config file or device model):
+/// it depends only on the OCPP 1.6 getters/setters declared on ChargePointConfigurationInterface, which are
+/// resolved via virtual dispatch to the concrete implementation. This class remains abstract; it does not
+/// implement those getters/setters.
+class ChargePointConfigurationConnectivity : public ChargePointConfigurationInterface {
+public:
+    std::string get_network_configuration_priority() override;
+    void set_network_configuration_priority(const std::string& priority, const std::string& source) override;
+    std::optional<ocpp::v2::NetworkConnectionProfile> read_network_connection_profile(int32_t slot) override;
+    bool write_network_connection_profile(int32_t slot, const ocpp::v2::NetworkConnectionProfile& profile,
```
mlitre, line 21: Let's be consistent and make these std::int32_t, but even so we could make them uint since slots and profiles are never negative

lib/everest/ocpp/lib/ocpp/v16/charge_point_configuration_connectivity.cpp
```diff
@@ -0,0 +1,129 @@
...
+    }
+    const auto security_profile = getSecurityProfile();
+    try {
+        auto uri = Uri::parse_and_validate(getCentralSystemURI(), charge_point_id, security_profile);
+
+        WebsocketConnectionOptions opts{{OcppProtocolVersion::v16},
+                                        uri,
+                                        security_profile,
+                                        getAuthorizationKey(),
+                                        std::chrono::seconds(10),
```
mlitre, line 79: Is this normal to be hardcoded to 10 seconds? I believe this sets the message timeout

lib/everest/ocpp/lib/ocpp/v2/device_model.cpp
```diff
@@ -1046,5 +1047,153 @@ std::int32_t DeviceModel::clear_custom_monitors() {
...
 }
 
+// ConnectivityManagerConfiguration implementations
+
+std::string DeviceModel::get_network_configuration_priority() {
+    return get_optional_value<std::string>(ControllerComponentVariables::NetworkConfigurationPriority).value_or("");
+}
+
+void DeviceModel::set_network_configuration_priority(const std::string& priority, const std::string& source) {
+    if (!ControllerComponentVariables::NetworkConfigurationPriority.variable.has_value()) {
```
mlitre, line 1058: Worth adding a warning here

lib/everest/ocpp/lib/ocpp/v2/device_model.cpp
```diff
@@ -1046,5 +1047,153 @@ std::int32_t DeviceModel::clear_custom_monitors() {
...
+        auto uri = Uri::parse_and_validate(profile.ocppCsmsUrl.get(), identity, profile.securityProfile);
+        const auto ocpp_versions = utils::get_ocpp_protocol_versions(
+            get_value<std::string>(ControllerComponentVariables::SupportedOcppVersions));
+
+        WebsocketConnectionOptions opts{
+            ocpp_versions,
+            uri,
+            profile.securityProfile,
+            basic_auth_password,
+            std::chrono::seconds(std::max(profile.messageTimeout, 1)),
```
mlitre, line 1141: Probably better to have a more sensible default here, we actually even have MessageTimeout variable with instance Default normally exactly for that purpose

lib/everest/ocpp/lib/ocpp/common/connectivity_manager.cpp
```diff
@@ -211,24 +199,16 @@ void ConnectivityManager::confirm_successful_connection() {
...
     const auto network_connection_profile = this->get_network_connection_profile(config_slot.value());
 
-    if (const auto& security_profile_cv = ControllerComponentVariables::SecurityProfile;
-        security_profile_cv.variable.has_value() and network_connection_profile.has_value()) {
-        this->device_model.set_read_only_value(security_profile_cv.component, security_profile_cv.variable.value(),
-                                               AttributeEnum::Actual,
-                                               std::to_string(network_connection_profile.value().securityProfile),
-                                               VARIABLE_ATTRIBUTE_VALUE_SOURCE_INTERNAL);
+    if (network_connection_profile.has_value()) {
+        this->configuration.set_active_security_profile(network_connection_profile.value().securityProfile,
```
mlitre, line 209: Here something changes due to the use of the slot vs general security profile, worth double checking what is the correct behavior

lib/everest/ocpp/lib/ocpp/v16/charge_point_configuration_connectivity.cpp
Pietfried: It was also like this before. I would adress this in a follow up.

lib/everest/ocpp/lib/ocpp/v2/device_model.cpp
Pietfried: Same like above, I didnt change the value here

lib/everest/ocpp/lib/ocpp/common/connectivity_manager.cpp
Pietfried: This just sets the `SecurityCtrlr::SecurityProfile` (telemetry) variable

Pietfried:
Changes adressed or commented.
@james-ctc @mlitre ready for re-review.

Pietfried:
@mlitre @james-ctc After approval of https://github.com/EVerest/EVerest/pull/2289 it has now been squash-merged into this branch. Requested changes have been adressed or commented on. Ready for re-review.

james-ctc (dismissed):
LGTM

james-ctc (dismissed):

james-ctc (dismissed):
LGTM

lib/everest/ocpp/lib/ocpp/v16/charge_point_configuration_connectivity.cpp
```diff
@@ -0,0 +1,129 @@
...
+
+std::optional<int32_t> ChargePointConfigurationConnectivity::get_network_config_timeout() {
+    return std::nullopt; // ConnectivityManager will apply its built-in 60 s default
+}
+
+std::optional<WebsocketConnectionOptions>
+ChargePointConfigurationConnectivity::get_websocket_connection_options(int32_t /*slot*/) {
+    const auto charge_point_id = getChargePointId();
+    if (charge_point_id.find(':') != std::string::npos) {
+        EVLOG_AND_THROW(std::runtime_error("ChargePointId must not contain ':'"));
```
mlitre: Should we throw or not here? If someone incorrectly configures a CS then it is just bricked since it will try on repeat to boot and fail. If we at least return std::nullopt we could fall back to another profile.

lib/everest/ocpp/lib/ocpp/v16/charge_point_configuration_connectivity.cpp
```diff
@@ -0,0 +1,129 @@
...
+                                        getAdditionalRootCertificateCheck().value_or(false),
+                                        getHostName(),
+                                        getVerifyCsmsCommonName(),
+                                        getUseTPM(),
+                                        getVerifyCsmsAllowWildcards(),
+                                        getIFace(),
+                                        getEnableTLSKeylog(),
+                                        getTLSKeylogFile()};
+        return opts;
+    } catch (const std::invalid_argument& e) {
```
mlitre, line 102: We should also catch DeviceModelErrors /Exceptions now that we can have a DM based config

lib/everest/ocpp/lib/ocpp/v16/charge_point_impl.cpp
```diff
@@ -1970,14 +1934,7 @@ ChargePointImpl::set_configuration_key_internal(CiString<50> key, CiString<500>
...
                 } else if (key == "WebSocketPingInterval") {
-                    auto websocket_ping_interval_option = configuration.getWebsocketPingInterval();
-
-                    if (websocket_ping_interval_option.has_value()) {
-                        auto websocket_ping_interval = websocket_ping_interval_option.value();
-                        auto websocket_pong_timeout = configuration.getWebsocketPongTimeout();
-
-                        websocket->set_websocket_ping_interval(websocket_ping_interval, websocket_pong_timeout);
-                    }
+                    this->connectivity_manager->set_websocket_connection_options_without_reconnect();
```
mlitre: This doesn't directly reset the ping interval, it only changes it on next reconnect

mlitre (changes requested):

lib/everest/ocpp/lib/ocpp/v16/charge_point_configuration_connectivity.cpp
Pietfried: Changed it so we fall back to another profile in this case

lib/everest/ocpp/lib/ocpp/v16/charge_point_configuration_connectivity.cpp
Pietfried: Added additional catch block for this

lib/everest/ocpp/lib/ocpp/v16/charge_point_impl.cpp
Pietfried: Fixed it. It was also not applied immediately for v2 as well, so I fixed this, too

mlitre (approved):
lgtm, maybe for a follow up PR I think maybe the configuration implementation for 1.6 shouldn't leak out exceptions. This would be cumbersome to add try catches everywhere we call a get or set variable

PR #2277, opened 2026-06-04
fix(db): harden sqlite reads against null/busy

mlitre:
## Describe your changes

Hardens the shared SQLite layer and the OCPP v2 profile reads against
concurrent access and malformed result handling. These are
general-purpose correctness fixes (they affect every consumer of
`lib/everest/sqlite`), surfaced while building concurrent charging-profile
workers but valuable on their own.

Three changes:

- **`sqlite::Connection`** sets `sqlite3_busy_timeout` (5s) on open, so a
  reader racing a concurrent writer on the same database retries briefly
  instead of getting a raw `SQLITE_BUSY` surfaced to the caller.

- **`sqlite::Statement::column_text`** throws `QueryExecutionException` when
  the column pointer is `NULL` instead of constructing `std::string(nullptr)`,
  which is undefined behavior and aborts the process. NULL arises both for a
  genuine NULL column value and when the statement isn't positioned on a row
  (e.g. `step()` returned `SQLITE_BUSY`/`ERROR`/`MISUSE`). Callers that expect
  nullable columns should continue to use `column_text_nullable`.

- **`v2::DatabaseHandler`** profile-reading loops use
  `while (step() == SQLITE_ROW)` instead of `while (step() != SQLITE_DONE)`,
  so a non-row state ends the loop (and surfaces as a typed exception via the
  change above) rather than being treated as "more rows" and dereferencing an
  invalid column.

This is the base of a 3-part stack splitting a larger change; it is
self-contained and reviewable in isolation.

### DIFF
--- a/lib/everest/ocpp/lib/ocpp/v2/database_handler.cpp
+++ b/lib/everest/ocpp/lib/ocpp/v2/database_handler.cpp
@@ -902,11 +902,15 @@ DatabaseHandler::get_charging_profiles_matching_criteria(const std::optional<std
             stmt->bind_int("@evse_id", evse_id.value());
         }
 
-        while (stmt->step() == SQLITE_ROW) {
+        int status = SQLITE_ERROR;
+        while ((status = stmt->step()) == SQLITE_ROW) {
             results.emplace_back(json::parse(stmt->column_text(1)), // profile
                                  stmt->column_int(0),               // EVSE ID
                                  CiString<20>(stmt->column_text(2)) // source
             );
+        }
+        if (status != SQLITE_DONE) {
+            throw QueryExecutionException(this->database->get_error_message());
         }
         return results;
     }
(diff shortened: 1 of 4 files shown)


lib/everest/ocpp/lib/ocpp/v2/database_handler.cpp
Pietfried, line 1: This could lead to truncated results, e.g. when exiting the while early because of `SQLITE_BUSY` or `SQLITE_ERROR`.

Could be mitigated with e.g.:

```cpp
int rc;
while ((rc = stmt->step()) == SQLITE_ROW) {
    auto profile = json::parse(stmt->column_text(0));
    profiles.push_back(profile);
}

if (rc != SQLITE_DONE) {
    throw QueryExecutionException(this->database->get_error_message());
}
```

lib/everest/sqlite/lib/everest/database/sqlite/connection.cpp
```diff
@@ -76,6 +76,12 @@ bool Connection::open_connection() {
         EVLOG_error << "Error opening database at " << this->database_file_path << ": " << sqlite3_errmsg(db);
         return false;
     }
+    // Retry briefly on SQLITE_BUSY instead of failing immediately. Without this any concurrent
+    // writer (e.g. background workers that update charging profiles) racing with a reader on the
+    // same connection surfaces as SQLITE_BUSY, and callers that loop while(step()!=DONE) on a
+    // non-row state would touch invalid column data.
```
Pietfried: libsqlite should not leak these ocpp library specifics

```suggestion
    // Retry briefly on SQLITE_BUSY instead of failing immediately. Without this any concurrent
    // writer racing with a reader on the
    // same connection surfaces as SQLITE_BUSY, and callers that loop while(step()!=DONE) on a
    // non-row state would touch invalid column data.
```

Pietfried (changes requested):
To address the reader/writer contention we should probably follow up with a PR that allows to sets journal_mode=WAL per database

Pietfried (approved):

PR #2278, opened 2026-06-04
refactor(ocpp): clear profiles returns deleted ids

mlitre:
## Describe your changes

Changes `DatabaseHandler::clear_charging_profiles_matching_criteria` to
return the ids of the rows it actually deleted instead of a `bool`. The
matching `SELECT` and the `DELETE` now run inside a single transaction and
share the exact same `WHERE` clause and bound parameters, so the returned
ids are guaranteed to be precisely the rows that were removed.

Motivation: callers that keep per-profile state (e.g. background tracking
of charging profiles) need to know *which* ids were cleared so they can
drop that state without re-deriving the delete filter on their side — a
mirrored filter could drift from the DB query and leak or over-drop state.
The previous `bool` only signalled "something / nothing was deleted."

Included in this change:

- `clear_charging_profiles_matching_criteria` returns
  `std::vector<std::int32_t>` (empty when nothing matched), with the
  SELECT+DELETE wrapped in one transaction.
- `DatabaseHandlerInterface` / `DatabaseHandler` signatures updated, and the
  GMock `database_handler_mock` updated to match.
- `DatabaseHandlerTest` clear-criteria assertions updated to check the
  returned ids.
- The one call site (`SmartCharging::clear_profiles`) updated to treat a
  non-empty result as "accepted".

The SQLite read-robustness change this previously stacked on has merged to
`main` (#2277), so this PR now applies directly on top of `main` and
contains only the refactor commit. The dynamic charging profiles consumer
that uses the returned ids stacks on top of this PR (#2279). Reviewable in
isolation.

### DIFF
--- a/lib/everest/ocpp/lib/ocpp/v2/database_handler.cpp
+++ b/lib/everest/ocpp/lib/ocpp/v2/database_handler.cpp
@@ -834,8 +834,12 @@ DatabaseHandler::clear_charging_profiles_matching_criteria(const std::optional<s
     if (!criteria.has_value()) {
         std::vector<std::int32_t> ids;
         auto select_all = this->database->new_statement("SELECT ID FROM CHARGING_PROFILES");
-        while (select_all->step() == SQLITE_ROW) {
+        int status;
+        while ((status = select_all->step()) == SQLITE_ROW) {
             ids.push_back(select_all->column_int(0));
+        }
+        if (status != SQLITE_DONE) {
+            throw QueryExecutionException(this->database->get_error_message());
         }
         if (!this->clear_charging_profiles()) {
             return {};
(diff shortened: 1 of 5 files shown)


lib/everest/ocpp/lib/ocpp/v2/database_handler.cpp
```diff
@@ -816,62 +816,85 @@ bool DatabaseHandler::clear_charging_profiles() {
...
+
+    // No criteria means clear all. Snapshot the ids first so the caller can drop their deadline
+    // tracking for exactly the rows that were removed.
     if (!criteria.has_value()) {
-        return this->clear_charging_profiles();
+        std::vector<std::int32_t> ids;
+        auto select_all = this->database->new_statement("SELECT ID FROM CHARGING_PROFILES");
+        while (select_all->step() == SQLITE_ROW) {
+            ids.push_back(select_all->column_int(0));
+        }
```
Pietfried: Also here it is probably be better to check for the status and throw before clearing if status is not `SQLITE_DONE`

lib/everest/ocpp/lib/ocpp/v2/functional_blocks/smart_charging.cpp
```diff
@@ -1162,8 +1162,9 @@ ClearChargingProfileResponse SmartCharging::clear_profiles(const ClearChargingPr
     ClearChargingProfileResponse response;
     response.status = ClearChargingProfileStatusEnum::Unknown;
 
-    if (this->context.database_handler.clear_charging_profiles_matching_criteria(request.chargingProfileId,
-                                                                                 request.chargingProfileCriteria)) {
+    const auto cleared_ids = this->context.database_handler.clear_charging_profiles_matching_criteria(
```
Pietfried: It was missing before, but a good opportunity to try-catch QueryExecutionException here.

Pietfried (changes requested):

Pietfried (approved):

PR #2279, opened 2026-06-04
feat(ocpp v21): K28 dynamic charging profiles

mlitre:
Adaptive Dynamic profiles: periodic pull worker, CSMS push handling, deadline tracking via DynamicScheduleManager, expiry filter, and a serialized+coalesced recompute in OCPP201.

## Describe your changes

Implements OCPP 2.1 **K28 Dynamic charging profiles** in libocpp, plus the
OCPP201 module change required to apply the resulting schedules safely.

### libocpp — Dynamic profile support

- New **`DynamicScheduleManager`** that owns the adaptive lifecycle of
  Dynamic charging profiles:
  - a periodic worker that issues `PullDynamicScheduleUpdate` when a
    profile's pull deadline (`dynUpdateTime + dynUpdateInterval`) elapses;
  - handling for CSMS-pushed `UpdateDynamicSchedule` requests;
  - deadline tracking guarded by a monitor, with a single reaper thread that
    polls in-flight pulls and tears them down without a blocking join;
  - per-profile single-flight so a profile with a pull in flight is not
    re-dispatched until it lands or times out.
- `SmartCharging` wiring: validation that a Dynamic profile carries exactly
  one schedule with one period, auto-population of `dynUpdateTime` on accept,
  and an expiry filter that drops a Dynamic profile from the composite
  calculation once its schedule duration has elapsed.
- `profile.cpp` gains single-source helpers for the Dynamic expiry and pull
  deadlines so the manager timer and the composite filter agree on the exact
  boundary instant.

### OCPP201 — serialized recompute

The charging-schedule recompute (`charging_schedules_timer_callback`) now
fires concurrently from three sources: the interval timer, the libocpp
message thread, and the K28 timer/reaper. It is guarded with a single-flight
`try_lock` plus a `recompute_pending` drain loop, so concurrent fires cannot
apply a stale composite over a fresh one and a burst collapses into one
recompute.

This module guard is the safety mechanism for the concurrency the
`DynamicScheduleManager` introduces, so the two land together in this PR by
design — the feature is never separated from the guard that makes it safe.

This stacks on #2278 (`clear` returning deleted ids — which the manager uses
to drop deadline tracking for exactly the cleared profiles). The SQLite
read-robustness change that was beneath both has merged to `main` (#2277), so
this PR now has a single base PR remaining.

### DIFF
--- a/lib/everest/ocpp/include/ocpp/v2/dynamic_schedule_manager.hpp
+++ b/lib/everest/ocpp/include/ocpp/v2/dynamic_schedule_manager.hpp
@@ -1,0 +1,230 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Pionix GmbH and Contributors to EVerest
+
+#pragma once
+
+#include <atomic>
+#include <cstdint>
+#include <functional>
+#include <future>
+#include <map>
+#include <memory>
+#include <optional>
+#include <thread>
+#include <vector>
+
+#include <date/date.h>
+#include <everest/timer.hpp>
+#include <everest/util/async/monitor.hpp>
+
+#include <ocpp/common/cistring.hpp>
+#include <ocpp/v2/message_handler.hpp>
+#include <ocpp/v2/ocpp_types.hpp>
+#include <ocpp/v2/types.hpp>
+
+namespace ocpp::v2 {
+
+struct FunctionalBlockContext;
+
+/// \brief A Dynamic profile's two K28 deadlines: when to pull the next schedule update and when the
+/// current schedule expires.
+struct ProfileDeadlines {
+    /// \brief K28.FR.10 pull deadline: absolute UTC time the next PullDynamicScheduleUpdateRequest
+    /// should be dispatched. Set from \c dynUpdateTime + \c dynUpdateInterval, or "now" on bootstrap
+    /// (missing \c dynUpdateTime); \c std::nullopt when the profile is not pull-tracked.
+    std::optional<date::utc_clock::time_point> pull;
+
+    /// \brief K28.FR.13 expiry deadline: \c dynUpdateTime + \c chargingSchedule[0].duration.
+    /// \c std::nullopt when the profile has no schedule duration or no \c dynUpdateTime.
+    std::optional<date::utc_clock::time_point> expire;
+};
+
+/// \brief Owns adaptive-pull and expiry deadline tracking for K28 Dynamic charging profiles plus
+/// the in-flight pull-response futures, multiplexed onto a single reaper thread. Composed inside
+/// \ref SmartCharging.
+///
+/// Public methods are safe to call from the smart-charging message-handler threads; the internal
+/// timer fires on its own io_context thread and all state is guarded by monitors.
+class DynamicScheduleManager {
+public:
+    /// \param context                         Functional-block context (device model, database,
+    ///                                        message dispatcher) shared with \ref SmartCharging.
+    /// \param set_charging_profiles_callback  Fired whenever the set of valid Dynamic profiles
+    ///                                        changes (K28.FR.06 push apply, K28.FR.13 expiry).
+    DynamicScheduleManager(const FunctionalBlockContext& context, std::function<void()> set_charging_profiles_callback);
+
+    /// Signals teardown so the reaper thread, possibly waiting on an in-flight CSMS response,
+    /// wakes within one poll slice instead of stalling for the full per-pull wait bound, then
+    /// joins the reaper (it captures \c this, so it must not outlive this object).
+    ~DynamicScheduleManager();
+
+    DynamicScheduleManager(const DynamicScheduleManager&) = delete;
+    DynamicScheduleManager& operator=(const DynamicScheduleManager&) = delete;
+    DynamicScheduleManager(DynamicScheduleManager&&) = delete;
+    DynamicScheduleManager& operator=(DynamicScheduleManager&&) = delete;
+
+    /// \brief K28.FR.10: rebuild pull- and expiry-deadline tracking by walking all persisted
+    /// profiles. Called once from \ref SmartCharging's constructor; also exposed so tests that
+    /// pre-populate the database can exercise the rebuild path.
+    void rebuild_from_db();
+
+    /// \brief Refresh adaptive-pull and expiry deadlines for \p profile, then rearm the timer.
+    /// K28.FR.14 reactivation is implicit: a refreshed \c dynUpdateTime moves both deadlines forward.
+    void update_tracking(const ChargingProfile& profile);
+
+    /// \brief Erase pull- and expiry-deadline tracking for \p charging_profile_id and rearm the
+    /// timer. Called from ClearChargingProfile and SetChargingProfile replace paths so a removed
+    /// or replaced profile cannot leave a stale deadline behind.
+    void erase_tracking(std::int32_t charging_profile_id);
+
+    /// \brief Handle a K28 UpdateDynamicScheduleRequest from the CSMS.
+    ///
+    /// Resolves the target profile by id and rejects with reasonCode \c "InvalidProfile" if missing
+    /// or non-Dynamic (K28.FR.11). Otherwise applies the update, refreshes pull-deadline tracking
+    /// (K28.FR.10), and fires the smart-charging callback only when fields actually changed. The
+    /// response is dispatched before the callback so the CSMS sees Accepted before downstream
+    /// effects.
+    void handle_update_dynamic_schedule_request(const ocpp::EnhancedMessage<MessageType>& message);
+
+private:
+    /// \brief Adaptive timer plus its shutdown flag, guarded as one unit so the two move and read
+    /// together. \c shutting_down lets a concurrent \ref reschedule_timer bail instead of re-arming
+    /// a timer the destructor is tearing down.
+    struct TimerState {
+        std::unique_ptr<Everest::SteadyTimer> timer;
+        bool shutting_down{false};
+    };
+
+    /// \brief One in-flight PullDynamicScheduleUpdate: the profile id, the response future the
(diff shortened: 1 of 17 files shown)


lib/everest/ocpp/lib/ocpp/v2/functional_blocks/smart_charging.cpp
```diff
@@ -1153,16 +1185,34 @@ SetChargingProfileResponse SmartCharging::add_profile(ChargingProfile& profile,
...
+        return response;
+    }
+
+    // K28.FR.10: refresh the adaptive-pull deadline for this profile id (handles Dynamic +
+    // non-Dynamic-replacing-Dynamic transitions). Run in its own try-block: the profile is already
+    // in the DB, so we must not lie to the CSMS with Rejected if pull-tracking refresh fails
+    // (e.g. std::bad_alloc on a map insert, std::system_error from timer rearm). Log the local
+    // mismatch and keep the Accepted response.
+    try {
+        this->dynamic_schedule_manager.update_tracking(profile);
```
Pietfried: Do we need to call this for all profiles or only for the ones with `ChargingProfileKindEnum::Dynamic`?

lib/everest/ocpp/tests/lib/ocpp/v2/functional_blocks/test_smart_charging.cpp
```diff
@@ -1476,7 +1476,8 @@ TEST_F(SmartChargingTest, K10_ClearChargingProfile_ClearsId) {
     auto profiles = database_handler->get_all_charging_profiles();
     EXPECT_THAT(profiles, testing::Contains(profile));
 
-    auto sut = smart_charging.clear_profiles(create_clear_charging_profile_request(DEFAULT_PROFILE_ID));
+    std::vector<std::int32_t> cleared_ids;
```
Pietfried, line 1479: nit: here and onwards you could verify the vlaues of `cleared_ids`

lib/everest/ocpp/include/ocpp/v2/dynamic_schedule_manager.hpp
```diff
@@ -0,0 +1,233 @@
...
+    /// effects.
+    void handle_update_dynamic_schedule_request(const ocpp::EnhancedMessage<MessageType>& message);
+
+private:
+    /// \brief Adaptive timer plus its shutdown flag, guarded as one unit by a monitor.
+    /// The timer drives K28.FR.10 PullDynamicScheduleUpdate dispatch and K28.FR.13 expiry
+    /// recompute. (Re)armed by \ref reschedule_timer at the earliest pull or expiry deadline
+    /// across all tracked profiles; fires \ref on_deadline on its own io_context thread.
+    /// \c shutting_down is set by the destructor so a concurrent \ref reschedule_timer bails
+    /// instead of re-arming a timer that is being torn down.
```
Pietfried: Most of this is not documenting the class itself but rather how it is used.

lib/everest/ocpp/lib/ocpp/v2/dynamic_schedule_manager.cpp
```diff
@@ -0,0 +1,583 @@
...
+        for (const auto& [id, entry] : *handle) {
+            if (entry.pull.has_value() && entry.pull.value() <= now) {
+                due_pull_ids.push_back(id);
+            }
+        }
+    }
+
+    for (const auto id : due_pull_ids) {
+        // Re-read profile to get fresh dynUpdateInterval (may have been mutated since deadline was set).
+        const auto match = this->lookup_dynamic_profile(id);
```
Pietfried, line 304: `lookup_dynamic_profile` catches `QueryExecutionException` and as a consequence returns `std::nullopt`. As a result, the profile is erased from `deadlines` map although it shouldn't.

lib/everest/ocpp/include/ocpp/v2/dynamic_schedule_manager.hpp
```diff
@@ -0,0 +1,233 @@
...
+#include <ocpp/v2/types.hpp>
+
+namespace ocpp::v2 {
+
+struct FunctionalBlockContext;
+
+/// \brief Per-profile adaptive-pull and expiry deadline tracking for K28 Dynamic profiles.
+/// Both fields are tracked under a single map entry so a refresh that touches both deadlines
+/// is observed atomically by the timer thread (no partial-state window).
+struct ProfileDeadlines {
```
Pietfried, line 31: Description rather applies to the member and not the class itself

lib/everest/ocpp/lib/ocpp/v2/dynamic_schedule_manager.cpp
```diff
@@ -0,0 +1,583 @@
...
+                    return true;
+                }
+                return false;
+            });
+            pulls.erase(split, pulls.end());
+        }
+
+        // Lock released: validate and apply each collected entry. Same five checks, in the same
+        // order, that the old per-pull worker performed; preserve the exact log wording so the K28
+        // error-path tests still match.
```
Pietfried: Comment is likely not up to date

modules/EVSE/OCPP201/OCPP201.cpp
```diff
@@ -1136,13 +1136,23 @@ void OCPP201::ready() {
...
+        return;
+    }
+    std::lock_guard<std::mutex> guard(recompute_mutex, std::adopt_lock);
+    while (recompute_pending.exchange(false)) {
+        const auto composite_schedule_unit = get_unit_or_default(config.RequestCompositeScheduleUnit);
+        const auto composite_schedules = charge_point->get_all_composite_schedules(
+            config.RequestCompositeScheduleDurationS, composite_schedule_unit);
+        publish_charging_schedules(composite_schedules);
+        set_external_limits(composite_schedules);
+    }
```
Pietfried, line 1172: There is still a raise when thread1 holds `recompute_mutex` until finishing the while and thread2's attempt to `try_lock` returns `false`. `recompute_pending` is then `true` and thread2's recompute is never run until the next trigger

lib/everest/ocpp/lib/ocpp/v2/functional_blocks/smart_charging.cpp
```diff
@@ -339,7 +350,10 @@ SmartCharging::SmartCharging(const FunctionalBlockContext& functional_block_cont
                              StopTransactionCallback stop_transaction_callback) :
     context(functional_block_context),
     set_charging_profiles_callback(set_charging_profiles_callback),
-    stop_transaction_callback(stop_transaction_callback) {
+    stop_transaction_callback(stop_transaction_callback),
+    dynamic_schedule_manager(functional_block_context, this->set_charging_profiles_callback) {
```
Pietfried: In order to save resources we should only intantiate `dynamic_schedule_manager` in case `SupportsDynamicProfiles` is `true`.

That would allow us to hide a large part of this behind that feature flag and save resources / threads.

The simplest solution would probably be to make the DynamicScheduleManager optional.

Pietfried (changes requested):

modules/EVSE/OCPP201/OCPP201.cpp
mlitre: I resolved the issue, let me know if the solution is ok for you

lib/everest/ocpp/lib/ocpp/v2/functional_blocks/smart_charging.cpp
mlitre: Yes it's needed if ever we replace the dynamic profile, e.g. replace with id being the same.

Pietfried (dismissed):

Pietfried:
@mlitre lgtm, still one conflict to resolve

mlitre:
> @mlitre lgtm, still one conflict to resolve

Ok will do

lib/everest/ocpp/lib/ocpp/v2/functional_blocks/smart_charging.cpp
```diff
@@ -1492,7 +1565,7 @@ bool SmartCharging::is_overlapping_validity_period(const ChargingProfile& candid
     overlap_stmt->bind_text(
         "@purpose", conversions::charging_profile_purpose_enum_to_string(candidate_profile.chargingProfilePurpose),
         everest::db::sqlite::SQLiteString::Transient);
-    while (overlap_stmt->step() != SQLITE_DONE) {
+    while (overlap_stmt->step() == SQLITE_ROW) {
```
Pietfried: We also need to check for the rc here after the while to not return an invalid result

modules/EVSE/OCPP201/OCPP201.cpp
```diff
@@ -1141,13 +1141,31 @@ void OCPP201::ready() {
...
+    while (recompute_pending.load()) {
+        if (!recompute_mutex.try_lock()) {
+            // Another thread holds the recompute; it will observe recompute_pending and run our update.
+            return;
+        }
+        {
+            std::lock_guard<std::mutex> guard(recompute_mutex, std::adopt_lock);
+            while (recompute_pending.exchange(false)) {
+                const auto composite_schedule_unit = get_unit_or_default(config.RequestCompositeScheduleUnit);
+                const auto composite_schedules = charge_point->get_all_composite_schedules(
```
Pietfried: This should never terminate so use try-catch

Pietfried (changes requested):

modules/EVSE/OCPP201/OCPP201.cpp
mlitre: Done

Pietfried (dismissed):
We still return invalid results in `smart_charging.cpp` (no way to distinguish e.g. no data and failure for caller) but since this was preexisting id suggest to merge and adress in a subsequent PR

Pietfried (approved):

PR #2282, opened 2026-06-05
feat(sqlite): enable WAL journal mode on file databases

Pietfried:
## Describe your changes

Set PRAGMA journal_mode=WAL when opening file-based connections so concurrent readers and a single writer stop serializing against each other. The mode is read back and the open fails if WAL cannot be established (WAL-or-fail); in-memory databases are detected via a single reused flag and skip WAL.

Add ConnectionWalTest covering WAL on a file database, in-memory open, and migration of a pre-existing rollback-journal database.

### DIFF
--- a/lib/everest/sqlite/tests/test_connection.cpp
+++ b/lib/everest/sqlite/tests/test_connection.cpp
@@ -1,0 +1,76 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2020 - 2025 Pionix GmbH and Contributors to EVerest
+
+#include <everest/database/sqlite/connection.hpp>
+#include <everest/database/sqlite/statement.hpp>
+#include <gtest/gtest.h>
+#include <sqlite3.h>
+
+#include <filesystem>
+#include <string>
+
+namespace fs = std::filesystem;
+
+namespace everest::db::sqlite {
+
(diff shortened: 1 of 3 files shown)


lib/everest/sqlite/tests/test_connection.cpp
```diff
@@ -0,0 +1,76 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2020 - 2025 Pionix GmbH and Contributors to EVerest
```
mlitre:
```suggestion
// Copyright 2020 - 2026 Pionix GmbH and Contributors to EVerest
```

mlitre (changes requested):
I think this option should be configurable or fallback to the current method, because this can break with some overlay mounts. This could have big consequences for setups with docker containers and such.

lib/everest/sqlite/tests/test_connection.cpp
Pietfried: Ok, ill implement a fallback.

Pietfried:
> I think this option should be configurable or fallback to the current method, because this can break with some overlay mounts. This could have big consequences for setups with docker containers and such.

Implemented as fallback without throwing, ready for rereview

mlitre (approved):

PR #2288, opened 2026-06-07
feature(BringUp): introduce BUEvBoardSupport

mhei:
## Describe your changes

This introduces a bringup module for the ev_board_support interface. It allows to control the CP state, enable/disable, 3ph/1ph switching and diode failure toggling.
The interface also has support for other features, but I decided to not add it here for now, because I don't use/need it at the moment.

none


modules/BringUp/BUEvBoardSupport/BUEvBoardSupport.cpp
```diff
@@ -0,0 +1,451 @@
...
+    }
+
+    return ev_info;
+}
+
+void BUEvBoardSupport::init() {
+}
+
+void BUEvBoardSupport::ready() {
+    auto screen = ScreenInteractive::Fullscreen();
```
cburandt, line 103: Being well aware, that the other BringUp modules do not set a good example: When exiting the ready() function there is some potential for seg-faulting (local variable screen captured by the lambdas which outlive the ready() function). Making it a member variable is nor a cure, as the r_* will still outlive it.
But this can only happen on exit and I consider it not worth blocking this PR. (but if a good solution is at hand, feel free to apply it)

cburandt (approved):
This is unclean on exit, but that's the same for other BringUp modules as well and can be fixed in a different PR.

modules/BringUp/BUEvBoardSupport/BUEvBoardSupport.cpp
mhei: Thanks for your review and the valuable feedback - you got me, I blindly copied from the other modules. A very simple solution in my eyes would be to make `screen` a global variable?

cburandt: Thinking twice, I opt for leaving it as is:
- 1st: It should be solved for all BU modules
- 2nd: the shutdown is broken in general (Loop() blocks, so there is a bigger change required to make this module shutdown in a clean way).

cburandt (approved):
see my comment below

mhei:
Ok, I agree - let's put it on the todo list but merge this one here as-is for now.

hikinggrass:
> Ok, I agree - let's put it on the todo list but merge this one here as-is for now.

@mhei I think this needs a rebase / force-push to retrigger the ci. Can you do that?

hikinggrass:
Thanks for the rebase, I think something went wrong with the DCO check run, could you edit/amend your commit and force-push again @mhei ? Sometimes the DCO check seems to pickup old base-hashes to compare against and that should fix it, sorry for the inconvenience :)

mhei:
No problem, seems it worked for this push now.

PR #2289, opened 2026-06-07
feat(ocpp16): Use ConnectivityManager instead of raw Websocket

Pietfried:
## Describe your changes

Drives the OCPP 1.6 `ChargePointImpl` websocket through a shared `ConnectivityManager` (mirroring the OCPP 2.x path) and, on top of that, makes the security-profile-switch revert window configurable.

**Connectivity: drive v16 websocket through `ConnectivityManager`**
- Replace `ChargePointImpl`'s raw `Websocket` ownership with a shared `ConnectivityManagerInterface`, moving connection handling completely into connectivity_manager (like for v2). The public constructors delegate to a new one that either accepts an externally provided manager or builds an internal one when none is given
- Extract the three websocket callbacks into public `on_websocket_connected` / `on_websocket_disconnected` / `on_websocket_connection_failed` methods and expose them on the `ChargePoint` class; `init_connectivity_manager()` wires them only when the connectivity_manager is built internally, and the message callback is registered unconditionally in `start()` before `connect()`.
- Because the manager has no signal to trigger the fallback, security-profile-switch failure detection moves to a v16 revert timer, serialized against the connected callback with a mutex.
- Relocate `ConnectivityManagerMock` from `v2/mocks` to shared `common/mocks` (namespace `ocpp`) so v16 tests can reuse it; add behavioral `ChargePointImpl` connectivity + security-switch unit tests (`tests/lib/ocpp/v16/test_charge_point_connectivity.cpp`).

**Feature: configurable security-profile-switch revert timeout** 
- Add OCPP 1.6 key `SwitchSecurityProfileConnectionTimeout` across the legacy JSON and device-model backends as a optional key. Registered in the v2 component-variable table, `known_keys`, and both backends 
- `switchSecurityProfile` arms the revert timer with the configured value, falling back to the existing 20 s `SECURITY_PROFILE_SWITCH_TIMEOUT` constant when unset.

I did not yet squash the commits to facilitate the review.

Base PR: https://github.com/EVerest/EVerest/pull/2276

### DIFF
--- a/lib/everest/ocpp/include/ocpp/v16/charge_point.hpp
+++ b/lib/everest/ocpp/include/ocpp/v16/charge_point.hpp
@@ -166,6 +166,22 @@ public:
     /// \brief Disconnects the the websocket connection to the CSMS if it is connected
     void disconnect_websocket();
 
+    /// \brief Notifies the charge point that the websocket is connected. This allows an external owner of an injected
+    /// ConnectivityManager to drive the charge point's websocket lifecycle (mirroring ocpp::v2::ChargePointInterface).
+    void on_websocket_connected(const int configuration_slot,
+                                const ocpp::v2::NetworkConnectionProfile& network_connection_profile,
+                                const ocpp::OcppProtocolVersion ocpp_version);
+
+    /// \brief Notifies the charge point that the websocket is disconnected. This allows an external owner of an injected
+    /// ConnectivityManager to drive the charge point's websocket lifecycle (mirroring ocpp::v2::ChargePointInterface).
+    void on_websocket_disconnected(const int configuration_slot,
+                                   const ocpp::v2::NetworkConnectionProfile& network_connection_profile);
+
+    /// \brief Notifies the charge point that the websocket connection failed. This allows an external owner of an
+    /// injected ConnectivityManager to drive the charge point's websocket lifecycle (mirroring
+    /// ocpp::v2::ChargePointInterface).
+    void on_websocket_connection_failed(ocpp::ConnectionFailedReason reason);
+
     /// \brief Calls the set_connection_timeout_callback that can be registered. This function is used to notify an
     /// Authorization mechanism about a changed ConnectionTimeout configuration key.
     void call_set_connection_timeout();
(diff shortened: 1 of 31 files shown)


lib/everest/ocpp/include/ocpp/v16/charge_point.hpp
```diff
@@ -77,6 +78,15 @@ class ChargePoint {
         const std::optional<SecurityConfiguration> security_configuration = std::nullopt,
         const std::function<void(const std::string& message, MessageDirection direction)>& message_callback = nullptr);
 
+    /// \brief Constructor that allows providing a \p connectivity_manager .
+    explicit ChargePoint(
+        ChargePointConfigurationInterface& cfg, const fs::path& share_path, const fs::path& database_path,
+        const fs::path& sql_init_path, const fs::path& message_log_path,
+        const std::shared_ptr<EvseSecurity> evse_security,
```
james-ctc, line 85: for consistency `const <type>&`
when not using a reference the whole object is copied to the stack

lib/everest/ocpp/lib/ocpp/v16/charge_point_configuration_devicemodel.cpp
```diff
@@ -674,6 +674,17 @@ ChargePointConfigurationDeviceModel::setInternalSupplyVoltage(const std::string&
     return set_value_check(isPositiveInteger, *storage, keys::valid_keys::SupplyVoltage, value);
 }
 
+ChargePointConfigurationDeviceModel::SetResult
+ChargePointConfigurationDeviceModel::setInternalSwitchSecurityProfileConnectionTimeout(const std::string& value) {
+    if (not getSwitchSecurityProfileConnectionTimeout().has_value()) {
+        return SetResult::UnknownVariable;
+    }
+    if (not isPositiveInteger(value) or std::stoi(value) < 1) {
```
james-ctc, line 682: it looks fine, however std::stoi can throw exceptions
an alternative
```
try {
    if (std::stoi(value) >=0) {
        return set_value_check(*storage, keys::valid_keys::SwitchSecurityProfileConnectionTimeout, value);
    }
} catch {...} {
}
return SetResult::Rejected;
```

james-ctc (approved):
looks good to me, some suggestions included

PR #2293, opened 2026-06-08
fix(Auth): persist EVSE-specific reservations and clear store on cancel

biubiuzhang:
## Describe your changes

Two related reservation-persistence fixes in `ReservationHandler`:

**1. EVSE-specific reservations were never persisted.** `store_reservations()` was only called from the global (no `evse_id`) branch of `make_reservation()`. The two EVSE/connector-specific accept paths returned `Accepted` without calling it, so a standalone EVSE-specific reservation (OCPP 1.6 `connectorId > 0` / OCPP 2.0.1 `ReserveNow` with `evseId` — the common case) was lost across a restart. Fix: call `store_reservations()` in both EVSE-specific accept paths.

**2. Cancelled/expired reservations resurrected on reboot.** `store_reservations()` skipped writing when the set was empty (`if (!reservations.empty())`), so cancelling/expiring the last reservation left the stale entry on disk to be reloaded on the next boot. Fix: always persist the (possibly empty) array. An empty array also loads cleanly (no "reservations is not a json array" warning).

Contradicts the persistence intent from #943 and OCPP reservation semantics (a reservation ends only on expiry, CancelReservation, or use — not on a reboot).

### Tests
- `store_load_evse_specific_reservation_without_global`
- `cancelled_reservation_not_resurrected_after_reload`

## How verified
New unit tests fail before / pass after. Reproduced and verified on hardware (rev1, SIL): EVSE-specific `reserve_now` now writes to the store and reloads after restart; before the fix it wrote 0 rows and was gone.

### DIFF
--- a/modules/EVSE/Auth/lib/ReservationHandler.cpp
+++ b/modules/EVSE/Auth/lib/ReservationHandler.cpp
@@ -136,6 +136,7 @@ ReservationHandler::make_reservation(const std::optional<uint32_t> evse_id,
                 EVLOG_info << "Created reservation for evse id " << evse_id.value() << ", connector type "
                            << types::evse_manager::connector_type_enum_to_string(
                                   reservation.connector_type.value_or(types::evse_manager::ConnectorTypeEnum::Unknown));
+                store_reservations();
                 return types::reservation::ReservationResult::Accepted;
             }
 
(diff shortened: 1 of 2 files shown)


Pietfried:
@biubiuzhang thanks, the changes look good. Could you rebase and force push to retrigger the CI?

biubiuzhang:
Thanks @Pietfried! Rebased onto the latest `main` and force-pushed — CI is re-running now. 🙏

Pietfried (approved):
LGTM! Please squash your commits so we can go ahead and merge the fix

biubiuzhang:
@Pietfried thanks for approving! Commits squashed.

PR #2298, opened 2026-06-10
Unify connector type enums: evse_manager ConnectorTypeEnum as sole source of truth (closes #2209)

### LINKED ISSUE #2209: Extend evse_board_support#/Connector_type to represent DC and J3400/NACS connectors
### Describe the problem

The `evse_board_support#/Connector_type` enum has only two values: `IEC62196Type2Cable` and `IEC62196Type2Socket`. Both describe AC Type 2 (Mennekes) connectors. There is no value that honestly describes a DC fast charger (CCS1, CCS2), a Tesla / J3400 / NACS connector, CHAdeMO, or MCS at the BSP layer.

In practice every DC BSP in the upstream tree hardcodes `IEC62196Type2Cable` as a placeholder:

- `modules/HardwareDrivers/EVSE/YetiDriver/board_support/evse_board_supportImpl.cpp:78`
- `modules/HardwareDrivers/EVSE/MicroMegaWattBSP/board_support/evse_board_supportImpl.cpp:52`
- `modules/HardwareDrivers/EVSE/TIDA010939/board_support/evse_board_supportImpl.cpp:87`
- `modules/HardwareDrivers/EVSE/PhyVersoBSP/connector_2/evse_board_supportImpl.cpp:27`

The placeholder is functionally inert today because the only consumers of `caps.connector_type` are three checks in `modules/EVSE/EvseManager/Charger.cpp` (lines 45, 292, 1748), all of which test for `IEC62196Type2Socket` specifically — i.e. "is this a detachable Type 2 AC socket where I need to read PP ampacity?". For a DC charger the check returns false and the PP-read branch is correctly skipped. The narrow enum is, in effect, a Type2Socket boolean dressed up as an enum.

This causes two ongoing pains:

1. **BSPs cannot honestly describe themselves.** Any consumer that wants to render a connector type from `hw_capabilities` (`modules/API/limit_decimal_places.cpp:246` does this) receives a misleading `IEC62196Type2Cable` string from a DC charger.
2. **Schema drift with the wider enum.** `evse_manager#/ConnectorTypeEnum` already covers `cCCS1`, `cCCS2`, `cTesla`, `cG105`, and (since PR #1696) `cMCS`. Implementers must use two parallel enums and a mapping layer to express a single concept.

### Affected component

- `types/evse_board_support.yaml` (`Connector_type` enum)
- `modules/EVSE/EvseManager/Charger.cpp` (consumers of `connector_type`)
- All BSPs under `modules/HardwareDrivers/EVSE/*`

### Describe your solution

Extend `types/evse_board_support.yaml#/Connector_type` with values mirroring the existing wider enum in `types/evse_manager.yaml#/ConnectorTypeEnum`:

```yaml
Connector_type:
  description: Type of charging connector available at this EVSE
  type: string
  enum:
    - IEC62196Type2Cable    # existing
    - IEC62196Type2Socket   # existing
    - cCCS1                 # new — CCS Combo 1
    - cCCS2                 # new — CCS Combo 2
    - cTesla                # new — legacy Tesla connector
    - cJ3400                # new — SAE J3400 / NACS (standardised 2023, distinct from legacy cTesla)
    - cG105                 # new — CHAdeMO
    - cMCS                  # new — Megawatt Charging System (IEC 63379)
```

The migration is non-breaking by addition. A landing PR should:

1. **Audit consumers.** The three checks at `Charger.cpp:45,292,1748` are written as `== IEC62196Type2Socket`. Those remain semantically correct (no new value equals Type2Socket). Confirm no other consumer does `!= IEC62196Type2Socket` or `== IEC62196Type2Cable` in a way that breaks under the new values — `grep -rn \"IEC62196Type2\" modules/ lib/` should be exhaustive.
2. **Update API wrappers.** `lib/everest/everest_api_types/src/everest_api_types/evse_board_support/{json_codec,wrapper}.cpp` need entries for the new values, mirroring the pattern PR #1696 used for `cMCS` in the `evse_manager` wrapper.
3. **Update the interface hash.** `lib/everest/everest_api_types/tests/expected_interfaces_file_hashes.csv` will need the new hash for `types/evse_board_support.yaml`.

### Alternatives considered

- **Leave the narrow enum AC-only and document it.** Treat `evse_board_support#/Connector_type` as a pure AC PP-socket discriminator (`Cable` vs `Socket`) and explicitly mark the field as not describing cable identity. DC BSPs continue publishing `IEC62196Type2Cable` as a deliberate placeholder. Current de-facto state but undocumented and confusing to integrators.
- **Deprecate the narrow enum and use `evse_manager#/ConnectorTypeEnum` at both layers.** Cleaner long-term but breaks the BSP interface contract and requires migrating every BSP.

The proposed extension is the least invasive option and matches the precedent set by PR #1696 for `cMCS` in the wider enum.

### Additional context

- **Precedent:** PR #1696 (`feat(MCS): add MCS related types and helpers`, merged 2026-02-02) extended `evse_manager#/ConnectorTypeEnum` with `cMCS` but explicitly did not touch the narrow `evse_board_support#/Connector_type`. The MCS BSP faces the same placeholder problem documented here.
- **Recent activity in this area:** PR #1490 (`Remove ac_read_pp_ampacity from evse_board_support interface`, merged 2025-11-10) shows the BSP interface is being actively cleaned up.
- **Standards reference:** SAE J3400 was published in December 2023 as the standardised NACS specification, distinct from the legacy proprietary Tesla connector. The existing `cTesla` in `evse_manager#/ConnectorTypeEnum` predates J3400 standardisation; a separate `cJ3400` value distinguishes the standardised connector. Integrators wanting them treated identically can map both.
- Happy to open the PR for this if there's interest in the proposal.

biubiuzhang:
## Describe your changes

Implements the unification @mhei proposed in #2209: make `evse_manager#/ConnectorTypeEnum` the **sole source of truth** for connector types, extend it to the full OCPP 2.1 value set, and remove the narrow two-value `evse_board_support#/Connector_type` enum so BSPs can announce their real connector type and have it propagate through all components.

### Type definitions
- **`types/evse_manager.yaml`**: `ConnectorTypeEnum` extended with the values OCPP 2.1 has on top of the current set: `cChaoJi`, `cGBT_DC`, `cLECCS`, `cNACS`, `cNACS_CCS1`, `cUltraChaoJi` (29 values total, now a 1:1 superset match of `ocpp::v2::ConnectorEnumStringType`). `cNACS` covers SAE J3400; the pre-existing `cTesla` stays for the legacy proprietary connector — integrators wanting them treated identically can map both.
- **`types/evse_board_support.yaml`** *(breaking)*: `HardwareCapabilities.connector_type` now `$ref`s `/evse_manager#/ConnectorTypeEnum` (same pattern `types/reservation.yaml` already uses).

### Consumers
- **EvseManager/Charger**: the three `== IEC62196Type2Socket` checks (PP ampacity read) become `== sType2` — semantically identical, no new value equals `sType2`. Placeholder defaults (`Charger` member init, EvseManager safe-default capabilities) use `Undetermined` instead of pretending to be a Type 2 cable. Additionally, when no `connector_type` is configured on EvseManager, the BSP-announced type is now propagated to the `evse` interface (`get_evse()`), so the honest type actually reaches consumers; an explicitly configured type still wins.
- **All in-tree BSPs** announce honest values: YetiDriver/AdAcEvse22KwzKit/YetiSimulator → `cType2`; TIDA010939 + PhyVerso keep their socket/cable distinction as `sType2`/`cType2`; MicroMegaWattBSP (DC charging tester) → `cCCS2`; BringUp dummy caps → `Undetermined`.
- **OCPP201**: `ReserveNow.connectorType` was previously fed into the generated `string_to_connector_type_enum`, which throws on the hyphenated OCPP wire spellings (`sCEE-7-7`, `s309-1P-16A`, …) — a latent crash for any CSMS reserving with such a type. Replaced with an explicit `to_everest_connector_type()` conversion covering all 29 wire values, falling back to `Unknown` with a warning.
- **libocpp v2 `types.hpp`**: removed stray trailing spaces in the `s309-1P-16A `/`s309-1P-32A ` string constants (would never match spec-compliant CSMS values).
- **everest_api_types**: external `evse_board_support::Connector_type` becomes an alias of `evse_manager::ConnectorTypeEnum`; its hand-rolled 2-value codec/wrapper code is removed and the evse_manager codecs (extended with the new values) take over. `expected_types_file_hashes.csv` updated for both changed type files.
- **RpcApi**: module-local `ConnectorTypeEnum` extended with the new values (this also adds the previously missing `cMCS`); unknown values continue to degrade gracefully to `Unknown`.
- **chargebridge** configs and **EVerest_API reference docs** updated to the new wire values.

### Breaking changes
- `types::evse_board_support::Connector_type` no longer exists. Out-of-tree BSPs must publish `types::evse_manager::ConnectorTypeEnum` in `HardwareCapabilities.connector_type` (`IEC62196Type2Cable` → `cType2`, `IEC62196Type2Socket` → `sType2`).
- External API (`everest_api_types` V1_0) consumers of `evse_board_support` `hw_capabilities` see the new wire strings (`cType2`/`sType2`/... instead of `IEC62196Type2Cable`/`Socket`).
- To ease migration (in the spirit of the deprecation policy proposed in #2274), the external API still **accepts** `IEC62196Type2Cable`/`IEC62196Type2Socket` on input and maps them to `cType2`/`sType2`, so clients publishing hardware capabilities keep working unmodified. This legacy input acceptance can be dropped after a deprecation period.

### Tests
- The auto-generated everest_api_types serialization tests are derived from the API headers, so they now round-trip **all 29 values** through the json codecs and cover the cross-namespace `HardwareCapabilities.connector_type` field (verified by running the test autogenerator locally).
- `ChargerTest` updated to the unified enum (`sType2`).
- Verified locally: all touched YAML parses, `ev-cli types generate-headers` resolves the new `$ref` correctly (generated `HardwareCapabilities` carries `types::evse_manager::ConnectorTypeEnum`), and the modified everest_api_types codecs/wrappers compile cleanly against the generated headers.

### Issue ticket number and link

Closes #2209

### Checklist before requesting a review


james-ctc:
I recommend reading and perhaps contributing to the deprecation policy that covers breaking changes:
https://github.com/EVerest/EVerest/pull/2274/changes

biubiuzhang:
Thanks for the pointer @james-ctc — good timing. I checked the PR against the proposed policy: the BSP-side changes fall under the internal interfaces/types exclusion, but the external API wire values were indeed in scope. I've pushed a commit so the external API still accepts the legacy `IEC62196Type2Cable`/`Socket` values on input (mapped to `cType2`/`sType2`), with the old spellings treated as deprecated. Happy to add an entry to the Active Deprecation Index once #2274 lands.

mhei:
Now that we have a proper type in the capabilities and it is not optional, we should drop the configuration option from EvseManager ([manifest.yaml](https://github.com/biubiuzhang/everest-core/blob/616fc60c29751f6cf9d4c269c6e4cfb7a49bbf48/modules/EVSE/EvseManager/manifest.yaml#L10-L12)), right?

biubiuzhang:
Thanks @mhei, @james-ctc for the review and the pointers.

You're right that with a proper, non-optional connector type in the capabilities the `EvseManager` config option should go. Thinking it through, though, the full unification here (removing `evse_board_support#/Connector_type` as a breaking change, plus the deprecation-index plumbing gated on #2274) is a bigger, cross-cutting refactor than I can properly own and shepherd through review right now.

Rather than let it sit half-finished, I'm going to close this. The underlying idea from #2209 is still worth doing — happy for someone on the Car WG to pick it up, and I'm glad to help review. If our downstream need for the additional connector-type values comes back, I'll open a smaller, additive PR that just extends the enum without the breaking removal.

Appreciate the time on this.

PR #2300, opened 2026-06-10
feat(libiso15118): Adding graceful shutdown

SebaLukas:
## Describe your changes
Adding the ability to gracefully exit the infinite `loop()` from the `TbdController`. During an active session, either the PowerDelivery state or the active charge loop is terminated. If no session is active, the loop() is exited immediately.

I have actively decided against terminating the session in every state based on a failed response code. Theoretically, terminating the session before the PowerDelivery state is not an error, and therefore, as the standard specifies, the session should be terminated in the PowerDelivery state.

Before the changes there was no way to stop the endless `loop()` from the `TbdController` and shutdown gracefully the whole `libiso15118`.

### DIFF
--- a/lib/everest/iso15118/include/iso15118/d20/context.hpp
+++ b/lib/everest/iso15118/include/iso15118/d20/context.hpp
@@ -140,6 +140,14 @@ public:
         current_timeout = timeout;
     }
 
+    void request_shutdown() {
+        requested_shutdown = true;
+    }
+
+    [[nodiscard]] bool shutdown_requested() const {
+        return requested_shutdown;
+    }
+
     const session::Feedback feedback;
 
     session::SessionLogger& log;
@@ -170,6 +178,8 @@ private:
     Timeouts& timeouts;
 
     std::optional<TimeoutType> current_timeout{std::nullopt};
+
+    bool requested_shutdown{false};
 };
 
 } // namespace iso15118::d20
--- a/lib/everest/iso15118/src/iso15118/io/sdp_server.cpp
+++ b/lib/everest/iso15118/src/iso15118/io/sdp_server.cpp
@@ -99,7 +99,19 @@ SdpServer::~SdpServer() {
     // FIXME (aw): rather use some RAII class for this!
     logf_info("Shutting down SDP server!");
     if (fd != -1) {
-        close(fd);
+        close();
+    }
+}
+
+void SdpServer::close() {
+    logf_info("Closing Sdp Server");
+
+    const auto close_sdp = ::close(fd);
+    if (close_sdp == -1) {
+        logf_error("Sdp server close() failed");
+    } else {
+        logf_info("Sdp server closed gracefully");
+        fd = -1;
     }
 }
 
--- a/lib/everest/iso15118/src/iso15118/session/iso.cpp
+++ b/lib/everest/iso15118/src/iso15118/session/iso.cpp
@@ -325,4 +325,16 @@ void Session::close() {
     ctx.session_stopped = true;
 }
 
+void Session::request_shutdown() {
+    if (not state.connected) {
+        logf_info("Shutdown requested before an EV connected");
+        ctx.session_stopped = true;
+        connection->close();
+        ctx.feedback.signal(session::feedback::Signal::DLINK_TERMINATE);
+    } else {
+        push_control_event(d20::StopCharging{true}); // Stopping active charge loop
+        ctx.request_shutdown();
+    }
+}
+
 } // namespace iso15118
--- a/lib/everest/iso15118/src/iso15118/tbd_controller.cpp
+++ b/lib/everest/iso15118/src/iso15118/tbd_controller.cpp
@@ -34,8 +34,18 @@ TbdController::TbdController(TbdConfig config_, session::feedback::Callbacks cal
     }
 }
 
+TbdController::~TbdController() {
+    if (config.enable_sdp_server) {
+        poll_manager.unregister_fd(sdp_server->get_fd());
+        sdp_server->close();
+    }
+}
+
 void TbdController::loop() {
     static constexpr auto POLL_MANAGER_TIMEOUT_MS = 50;
+
+    shutdown_active.store(false);
+    bool shutdown_signaled{false};
 
     if (not config.enable_sdp_server) {
         auto connection = std::make_unique<io::ConnectionPlain>(poll_manager, interface_name);
@@ -45,7 +55,7 @@ void TbdController::loop() {
 
     auto next_event = get_current_time_point();
 
-    while (true) {
+    while (session or not shutdown_active.load()) {
         const auto poll_timeout_ms = get_timeout_ms_until(next_event, POLL_MANAGER_TIMEOUT_MS);
 
         try {
@@ -66,6 +76,11 @@ void TbdController::loop() {
             callbacks.signal(session::feedback::Signal::DLINK_ERROR);
         }
 
+        if (session and shutdown_active.load() and not shutdown_signaled) {
+            session->request_shutdown(); // Stopping the session
+            shutdown_signaled = true;
+        }
+
         if (session) {
             try {
                 const auto next_session_event = session->poll();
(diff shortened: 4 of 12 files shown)


lib/everest/iso15118/include/iso15118/d20/context.hpp
```diff
@@ -170,6 +178,8 @@ class Context {
     Timeouts& timeouts;
 
     std::optional<TimeoutType> current_timeout{std::nullopt};
+
+    bool requested_shutdown{false};
```
mlitre, line 182: Why aren't we using atomic_bool like in tbd_controller? Can't the context also be used / read on different threads? E.g someone calls shutdown from a different thread and then we read on the poll thread?

lib/everest/iso15118/src/iso15118/tbd_controller.cpp
```diff
@@ -34,9 +34,19 @@ TbdController::TbdController(TbdConfig config_, session::feedback::Callbacks cal
...
+    if (config.enable_sdp_server) {
+        poll_manager.unregister_fd(sdp_server->get_fd());
+        sdp_server->close();
+    }
+}
+
 void TbdController::loop() {
     static constexpr auto POLL_MANAGER_TIMEOUT_MS = 50;
 
+    shutdown_active = false;
```
mlitre: This applies to all usage of atomic bool, we should use `store` and `load` for accessing and replacing value.

mlitre: Also do we actually need to reset here? Wouldn't just be a better design if we want to restart the loop to actually rebuild the tbd controller?

lib/everest/iso15118/src/iso15118/tbd_controller.cpp
mlitre, line 183: Probably just a weird edge case, but should we also guard on this function for shutdown to avoid starting a session and getting stuck in a loop again?

lib/everest/iso15118/src/iso15118/io/sdp_server.cpp
```diff
@@ -99,7 +99,18 @@ SdpServer::~SdpServer() {
...
+}
+
+void SdpServer::close() {
+    logf_info("Closing Sdp Server");
+
+    const auto close_sdp = ::close(fd);
+    if (close_sdp == -1) {
+        logf_error("Sdp server close() failed");
+    } else {
+        logf_info("Sdp server closed gracefully");
```
mlitre, line 113: We should set fd to -1 or else we double close. 
Or we can just use RAII pattern and let the default destructor handle it properly

lib/everest/iso15118/src/iso15118/tbd_controller.cpp
```diff
@@ -79,14 +93,27 @@ void TbdController::loop() {
...
     }
+    logf_info("Existing TbdController loop gracefully");
+}
+
+void TbdController::shutdown() {
+    logf_info("Trigger graceful shutdown");
+    if (session) {
+        logf_info("Session is active. Stop the session right now.");
+        session->push_control_event(d20::StopCharging{true}); // Stopping active charge loop
+        session->request_shutdown(); // Stopping the session before charge loop in power delivery
```
mlitre: We should set shutdown_active before because we can process the control event and recreate a session at line 96 above

mlitre (changes requested):
General idea to simplify the shutdown process: instead of adding context.shutdown_requested we just reuse the progress::stop or stop flags and then keep shutdown_active. This could reduce the loop to `while (session or not shutdown_active)`. All that's needed is probably handling StopCharging / stop in PowerDelivery and pushing the stop charging control event in the loop (eg create an initiatiate_shutdown method that calls push control event stop or close if we are connected or not. You would also need a little extra handling in the loop to catch whether a shutdown process was initiated to avoid trying to close or stop multiple times. This would also simplify the shutdown method where it would just  set the shutdown flag.

SebaLukas:
I will have another thought about your suggestion. Maybe we can improve my PR

lib/everest/iso15118/src/iso15118/tbd_controller.cpp
SebaLukas: > Also do we actually need to reset here? Wouldn't just be a better design if we want to restart the loop to actually rebuild the tbd controller?

This depends on the user. I guess as a user you want to use the same controller after existing the loop and start again.

lib/everest/iso15118/src/iso15118/tbd_controller.cpp
SebaLukas: Good point

lib/everest/iso15118/src/iso15118/io/sdp_server.cpp
SebaLukas: I will set fd to -1.

lib/everest/iso15118/include/iso15118/d20/context.hpp
SebaLukas: Thats a good point. I guess every variable needs a guard when someone is calling a TbdController function from a different thread. But that should be solved in another PR

SebaLukas:
The problem without using a ctx variable is, that we can not store the control event later for the fsm. Control events are designed like fire and forget. So if we call the StopCharging Event in case of a graceful shutdown we need a way to store this event. It was easier for that case to add `requested_shutdown` instead of storing and handling the control event in iso/loop()

lib/everest/iso15118/include/iso15118/d20/context.hpp
SebaLukas: With the new changes requested_shutdown can be bool and not an atomic_bool

lib/everest/iso15118/src/iso15118/tbd_controller.cpp
SebaLukas: I added a check

lib/everest/iso15118/src/iso15118/tbd_controller.cpp
SebaLukas: The order does not matter

lib/everest/iso15118/src/iso15118/io/sdp_server.cpp
```diff
@@ -112,6 +112,7 @@ void SdpServer::close() {
     } else {
         logf_info("Sdp server closed gracefully");
     }
+    fd = -1;
```
mlitre: We should only mark fd as -1 if the close was successful

lib/everest/iso15118/src/iso15118/io/sdp_server.cpp
```diff
@@ -99,8 +99,20 @@ SdpServer::~SdpServer() {
...
+void SdpServer::close() {
+    logf_info("Closing Sdp Server");
+
+    const auto close_sdp = ::close(fd);
+    if (close_sdp == -1) {
+        logf_error("Sdp server close() failed");
+    } else {
+        logf_info("Sdp server closed gracefully");
     }
+    fd = -1;
```
mlitre: To improve this overall let's just use the sdp_server's close method and that is the only one that calls `::close`. And in this close method we guard on fd != -1 to call `::close` and we set -1 if the close was successful.
In my opinion there are too many places where we try to close and it can lead to issues

mlitre (changes requested):

lib/everest/iso15118/src/iso15118/io/sdp_server.cpp
SebaLukas: Done

lib/everest/iso15118/src/iso15118/io/sdp_server.cpp
SebaLukas: Good point :+1: 
Only `close()` calls `::close()`. But I didnt add the guard to `close()` because only the descructor of `SdpServer` and `TbdController` are calling `close()` and `SdpServer` had already the guard

lib/everest/iso15118/src/iso15118/session/iso.cpp
```diff
@@ -325,4 +325,16 @@ void Session::close() {
...
 
+void Session::request_shutdown() {
+    if (not state.connected) {
+        logf_info("Shutdown requested before an EV connected");
+        ctx.session_stopped = true;
+        connection->close();
+        ctx.feedback.signal(session::feedback::Signal::DLINK_TERMINATE);
+    } else {
+        push_control_event(d20::StopCharging{true}); // Stopping active charge loop
+        ctx.request_shutdown();
```
mlitre, line 336: Isn't it better to flag the shutdown first and then push the control event? In this order could we miss a moment where we have to process this flag?

mlitre: Discussed does not matter

mlitre (approved):

PR #2303, opened 2026-06-11
OCPP201: get component from the error

dorezyuk:
## Describe your changes

Add the mapping of the component to the error handling in OCPP.
With this change the error contains the binary name and the module id, which are mapped from the error

### DIFF
--- a/lib/everest/ocpp_module_common/src/device_model/everest_device_model_storage.cpp
+++ b/lib/everest/ocpp_module_common/src/device_model/everest_device_model_storage.cpp
@@ -900,6 +900,45 @@ EverestDeviceModelStorage::get_variable_attributes(const ocpp::v2::Component& co
     return this->device_model_storage->get_variable_attributes(component_id, variable_id, attribute_enum);
 }
 
+std::optional<ocpp::v2::Component>
+EverestDeviceModelStorage::get_component_for_module(const std::string& module_id,
+                                                    const std::string& implementation_id) const {
+    const auto module_config_it =
+        std::find_if(this->module_configs.begin(), this->module_configs.end(),
+                     [&module_id](const auto& entry) { return entry.first.module_id == module_id; });
+    if (module_config_it == this->module_configs.end()) {
+        return std::nullopt;
+    }
+
+    Component component;
+    component.name = module_config_it->first.module_type;
+    component.instance = module_id;
+
+    const auto mapping_it = this->mappings.find(module_id);
+    if (mapping_it == this->mappings.end()) {
+        return component;
+    }
+
+    auto mapping = mapping_it->second.module;
+    const auto implementation_mapping_it = mapping_it->second.implementations.find(implementation_id);
+    if (implementation_mapping_it != mapping_it->second.implementations.end() and
+        implementation_mapping_it->second.has_value()) {
+        mapping = implementation_mapping_it->second;
+    }
+
+    // in OCPP2.x the id and connectorId of the EVSEType must be > 0
+    if (mapping.has_value() and mapping->evse > 0) {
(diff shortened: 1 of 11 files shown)


hikinggrass:
@dorezyuk @Pietfried should we move this towards OCPPmulti, since it has conflicts now

dorezyuk:
I guess the files are now under lib/everest/ocpp_module_common and were moved here https://github.com/EVerest/EVerest/pull/2396 @Pietfried @hikinggrass let me know when you have time to review the code so I can rebase

modules/EVSE/OCPP201/OCPP201.cpp
```diff
@@ -184,6 +184,23 @@ int32_t get_connector_id_from_error(const Everest::error::Error& error) {
...
 
+ocpp::v2::Component OCPP201::get_component_from_error(const Everest::error::Error& error) const {
+    // the device model storage is created in ready(); error handlers registered in init() can fire earlier
+    if (this->everest_device_model_storage != nullptr) {
+        const auto component = this->everest_device_model_storage->get_component_for_module(
+            error.origin.module_id, error.origin.implementation_id);
+        if (component.has_value()) {
+            return component.value();
+        }
+        EVLOG_warning << "Failed to map the error to a valid component; Using fallback for error mapping";
```
Pietfried, line 195: Since this can occur on every error without the access configuration for the ocpp module, this should rather be EVLOG_debug

lib/everest/ocpp_module_common/src/device_model/everest_device_model_storage.cpp
```diff
@@ -900,6 +900,45 @@ EverestDeviceModelStorage::get_variable_attributes(const ocpp::v2::Component& co
     return this->device_model_storage->get_variable_attributes(component_id, variable_id, attribute_enum);
 }
 
+std::optional<ocpp::v2::Component>
+EverestDeviceModelStorage::get_component_for_module(const std::string& module_id,
```
Pietfried, line 904: The registration builds the ComponentKey from the module mapping only, see https://github.com/EVerest/EVerest/blob/00c241e96c0d75266a23fdd62daf000e20e482a1/lib/everest/ocpp_module_common/src/device_model/everest_device_model_storage.cpp#L612 "impl mappings are not taken into account at the moment".

This function prefers the impl mapping when present. When they differ, the emitted Component matches no component in the reported device model. Either register per-impl component keys too, or use the module mapping here and drop the implementation_id parameter

lib/everest/ocpp_module_common/src/device_model/everest_device_model_storage.cpp
```diff
@@ -900,6 +900,45 @@ EverestDeviceModelStorage::get_variable_attributes(const ocpp::v2::Component& co
...
+    const auto module_config_it =
+        std::find_if(this->module_configs.begin(), this->module_configs.end(),
+                     [&module_id](const auto& entry) { return entry.first.module_id == module_id; });
+    if (module_config_it == this->module_configs.end()) {
+        return std::nullopt;
+    }
+
+    Component component;
+    component.name = module_config_it->first.module_type;
+    component.instance = module_id;
```
Pietfried, line 915: Component::name and ::instance are CiString<50> so this can throw a `StringConversionException`

Pietfried (changes requested):
- This introduces different behavior between OCPP201 and OCPPmulti for how the component is derived for an error.  We prefer to add these kind of changes to be added to the new OCPPmulti only
- The component can only be derived via the config_service_client if the OCPP module gets the respective access config block in the YAML file . This should at least be documented.

PR #2304, opened 2026-06-11
feat(util): implement emplace and optimize move semantics for queues

djchhp:
## Describe your changes

Add perfect-forwarding emplace functions across the queue hierarchy and
optimize pop internals to consistently utilize standard move semantics.

- Implement template `emplace(Args&&...)` using perfect forwarding in
  `simple_queue`, `thread_safe_queue`, and `thread_safe_bounded_queue`.
- Update bounded and unbounded thread-safe `emplace` implementations to
  block appropriately and return the queue size post-insertion for API 
  consistency with existing `push` overloads.
- Fix a subtle bug where `std::forward` was erroneously applied to concrete 
  rvalue references (`value_type&&`) in move `push` functions; replaced 
  with `std::move`.
- Transition the storage wrapper pipeline to native `std::optional` 
  conversions, bypassing outdated fallback `std::unique_ptr` dynamic allocations.
- Remove redundant `std::move` calls on value-returned temporaries in `pop()` 
  to cleanly leverage standard library RVO/NRVO rules.
- Add comprehensive Google Test coverage validating emplace backpressure, 
  thread synchronization, parameter forwarding, and strict move-only guarantees.

### DIFF
--- a/lib/everest/util/include/everest/util/queue/thread_safe_bounded_queue.hpp
+++ b/lib/everest/util/include/everest/util/queue/thread_safe_bounded_queue.hpp
@@ -9,14 +9,16 @@
 #include <chrono>
 #include <condition_variable>
 #include <mutex>
+#include <optional>
 
 namespace everest::lib::util {
 
 /**
  * A thread safe bounded queue implemented on top of \ref queue::simple_queue. <br>
  * The common resource \ref simple_queue is guarded by a mutex in every member function.
- * A caller blocking on \p push will be unblocked when space becomes available via \p pop.
- * A caller blocking on \p pop or \p try_pop will be unblocked when new data is made available via \p push.
+ * A caller blocking on \p push or \p emplace will be unblocked when space becomes available via \p pop.
+ * A caller blocking on \p pop or \p try_pop will be unblocked when new data is made available via \p push or \p
+ * emplace.
  * @tparam T Datatype held by the queue
  */
 template <class T> class thread_safe_bounded_queue {
@@ -47,21 +49,8 @@ public:
      * @param[in] value data
      * @return The size of the queue after push. Returns 0 if the queue is stopped.
      */
-    size_type push(const value_type& value) {
-        std::unique_lock lock(m_mtx);
-        if (m_max_size > 0) {
-            m_cv_producer.wait(lock, [this]() { return m_queue.size() < m_max_size || m_stop; });
-        }
-
-        if (m_stop) {
-            return 0;
-        }
-
-        m_queue.push(value);
-        auto result = m_queue.size();
-        lock.unlock();
-        m_cv_consumer.notify_one();
-        return result;
+    size_type push(value_type const& value) {
+        return emplace(value);
     }
 
     /**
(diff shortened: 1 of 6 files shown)


lib/everest/util/include/everest/util/queue/thread_safe_bounded_queue.hpp
```diff
@@ -80,7 +82,30 @@ template <class T> class thread_safe_bounded_queue {
             return 0;
         }
 
-        m_queue.push(std::forward<value_type>(value));
+        m_queue.push(std::move(value)); // Fixed: Changed std::forward to std::move
```
mlitre: nit we can remove the comment

lib/everest/util/include/everest/util/queue/thread_safe_queue.hpp
```diff
@@ -80,7 +98,9 @@ template <class T> class thread_safe_queue {
      * @return An element from the queue.
      */
     value_type pop() {
-        return pop_impl(-1).value();
+        // Moving the temporary result of pop_impl into an rvalue reference
+        // ensures that value() returns an rvalue as well, effectively avoiding a copy.
+        return std::move(pop_impl(-1)).value();
```
mlitre: I don't think this move is necessary, you didn't do it for the thread_safe_bounded_queue and the PopStrictlyMovesAndNeverCopies test still passes

lib/everest/util/include/everest/util/queue/thread_safe_bounded_queue.hpp
mlitre, line 87: This another copy of the same blocking mechanism as for the two push methods, I suggest using this new improved emplace as the new basis for push.
`size_type push(const value_type& value) { return emplace(value); }` and `size_type push(value_type&& value)      { return emplace(std::move(value)); }`
This properly forwards the move or copy semantics and reduces boilerplate that can diverge. This same fix could be applied to thread_safe_queue and simple_queue.

lib/everest/util/include/everest/util/queue/thread_safe_queue.hpp
```diff
@@ -44,15 +49,28 @@ template <class T> class thread_safe_queue {
...
+     * @param[in] args Arguments forwarded to construct the data element.
+     * @return The size of the queue after emplace
+     */
+    template <class... Args> size_type emplace(Args&&... args) {
+        std::unique_lock lock(m_mtx);
+        m_queue.emplace(std::forward<Args>(args)...);
+        auto result = m_queue.size();
+        lock.unlock();
+        m_cv.notify_one();
+        return result;
```
mlitre, line 67: When looking into this I noticed we properly handle m_stop in the pop methods and in thread_safe_bounded_queue pop/push/emplace methods. Is there a reason why we don't handle m_stop in the push/emplace methods here? Shouldn't we if we are stopping, block new pushes to avoid excess growth?

lib/everest/util/include/everest/util/queue/thread_safe_queue.hpp
mlitre:
```suggestion
 * or \p try_pop will be unblocked when new data made available via \p push or emplace
```

mlitre (changes requested):

lib/everest/util/include/everest/util/queue/thread_safe_queue.hpp
djchhp: Thinking about this again you are right. Since we never assign an name to the result of pop_impl it is an rvalue already.

lib/everest/util/include/everest/util/queue/thread_safe_bounded_queue.hpp
djchhp: done

lib/everest/util/include/everest/util/queue/thread_safe_bounded_queue.hpp
djchhp: Much cleaner. Implemented as suggested.

lib/everest/util/include/everest/util/queue/thread_safe_queue.hpp
djchhp: done

lib/everest/util/include/everest/util/queue/thread_safe_queue.hpp
djchhp: That one is interesting. I think that is correct. As a first shot I added behavior similar to the bounded queue, like return 0 when stopped and reject the value. Not sure why it has not been, there. Also cannot think of any reason.

mlitre (approved):
Looks good to me!

PR #2308, opened 2026-06-12
OCPP201: Allow remote start with empty evse-id

dorezyuk:
## Describe your changes

Adds support for the `RequestStartTransaction` without an evse-id. In this case we leave the connector list empty and leave the decision which connector to pick to the auth manager

### DIFF
--- a/lib/everest/ocpp/lib/ocpp/v2/functional_blocks/remote_transaction_control.cpp
+++ b/lib/everest/ocpp/lib/ocpp/v2/functional_blocks/remote_transaction_control.cpp
@@ -33,7 +33,7 @@ namespace {
 /// \param evse Evse to check.
 /// \return True if at least one connector is not faulted or unavailable.
 ///
-bool is_evse_connector_available(EvseInterface& evse) {
+bool is_evse_connector_available(const EvseInterface& evse) {
     if (evse.has_active_transaction()) {
         // If an EV is connected and has no authorization yet then the status is 'Occupied' and the
         // RemoteStartRequest should still be accepted. So this is the 'occupied' check instead.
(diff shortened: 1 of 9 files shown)


Pietfried:
@dorezyuk Can you look into the OCPP e2e test that is failing? It would also be nice to have an additional e2e test for the EVSE 0 scenario because it would be nice to cover this use case through the OCPP, EvseManager + Auth modules

lib/everest/ocpp/lib/ocpp/v2/functional_blocks/remote_transaction_control.cpp
```diff
@@ -193,11 +193,29 @@ void RemoteTransactionControl::handle_remote_start_transaction_request(Call<Requ
...
+                                         ocpp::ReservationCheckStatus::ReservedForOtherToken;
+
+                if (is_evse_connector_available(evse) and !is_reserved) {
+                    response.status = RequestStartStopStatusEnum::Accepted;
+                    this->transaction.set_remote_start_id_for_evse(0, msg.idToken, msg.remoteStartId);
+                    break;
+                }
+            }
+
+            if (response.status == RequestStartStopStatusEnum::Rejected) {
```
Pietfried, line 215: It would be nice to include a `statusInfo` in the response like its done in L190

Pietfried (commented):
Note that even if we respond `Accepted` here, there can still be cases and race conditions in which the Auth module can not assign to an EVSE (because of potential state missmatch between Auth and OCPP state machine). In this case we respond with `Accepted` via OCPP but a transaction can never start. I still think this is acceptable.

lib/everest/ocpp/lib/ocpp/v2/functional_blocks/remote_transaction_control.cpp
dorezyuk: in line 190 we pass that from the add_profile_response which we don't have here. From the spec it seems to not obvious what the statusinfo should be in this case - can you be more specific

Pietfried: It's optional, but can be very helpful for the CSMS to get some additional information. It could just be something like you provide in the log statement, e.g.

`ReasonCode`: NoEvseAvailable
`additionalInfo`: RequestStartTransactionRequest requested without providing an EVSE, but no EVSE is available.

Pietfried (approved):

hikinggrass:
@dorezyuk can you squash your commits before merging?

PR #2311, opened 2026-06-12
feat(EvseV2G): Adding telemetry

djchhp:
## Describe your changes

  Adds EVSE/V2G telemetry support for EvseManager, EvseSecurity, and EvseV2G.

  - Adds telemetry API types and JSON codecs in `everest_api_types`.
  - Publishes typed telemetry payloads through the existing EVerest telemetry wrapper path.
  - Extends framework telemetry values to support `nlohmann::json` entries.
  - Adds EvseManager control telemetry and EvseSecurity certificate status telemetry.
  - Adds V2G transport, EV electrical, payment service, and charger status telemetry.
  - Guards V2G telemetry state updates with a publisher-owned mutex.
  - Resets session-scoped V2G telemetry between sessions.
  - Reports DIN authorization requests consistently with ISO authorization telemetry.

  Verification:
  - `cmake --build build/TESTING --target EvseV2G din_server_test v2g_ctx_test sdp_test -j24`
  - `ctest --test-dir build/TESTING -R "din_server_test|v2g_ctx_test|sdp_test" --output-on-failure`

  N/A

### DIFF
--- a/lib/everest/everest_api_types/include/everest_api_types/telemetry/API.hpp
+++ b/lib/everest/everest_api_types/include/everest_api_types/telemetry/API.hpp
@@ -1,0 +1,224 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Pionix GmbH and Contributors to EVerest
+
+#pragma once
+
+#include <cstdint>
+#include <optional>
+#include <string>
+
+namespace everest::lib::API::V1_0::types::telemetry {
+
+enum class ChargeProgress {
+    Start,
+    Stop,
+    Renegotiate,
+    Pause,
+    Terminate,
+};
+
+enum class V2gDin70121CommunicationState {
+    WaitForSessionSetup = 0,
+    WaitForServiceDiscovery = 1,
+    WaitForPaymentServiceSelection = 2,
+    WaitForAuthorization = 3,
+    WaitForChargeParameterDiscovery = 4,
+    WaitForCableCheck = 5,
+    WaitForPreCharge = 6,
+    WaitForPreChargePowerDelivery = 7,
+    WaitForCurrentDemand = 8,
+    WaitForCurrentDemandPowerDelivery = 9,
+    WaitForWeldingDetectionSessionStop = 10,
+    WaitForSessionStop = 11,
+    WaitForTerminatedSession = 12,
+};
+
+enum class V2gIso15118AcCommunicationState {
+    WaitForSessionSetup = 0,
+    WaitForServiceDiscovery = 1,
+    WaitForServiceDetailPaymentServiceSelection = 2,
+    WaitForPaymentDetailsCertificateInstallCertificateUpdate = 3,
+    WaitForPaymentDetails = 4,
+    WaitForAuthorization = 5,
+    WaitForChargeParameterDiscovery = 6,
+    WaitForPowerDelivery = 7,
+    WaitForChargingStatus = 8,
+    WaitForChargingStatusPowerDelivery = 9,
+    WaitForMeteringReceipt = 10,
+    WaitForSessionStop = 11,
+    WaitForTerminatedSession = 12,
+};
+
+enum class V2gIso15118DcCommunicationState {
+    WaitForSessionSetup = 0,
+    WaitForServiceDiscovery = 1,
+    WaitForServiceDetailPaymentServiceSelection = 2,
+    WaitForPaymentDetailsCertificateInstallCertificateUpdate = 3,
+    WaitForPaymentDetails = 4,
+    WaitForAuthorization = 5,
+    WaitForChargeParameterDiscovery = 6,
+    WaitForCableCheck = 7,
+    WaitForPreCharge = 8,
+    WaitForPreChargePowerDelivery = 9,
+    WaitForCurrentDemandPowerDelivery = 10,
+    WaitForCurrentDemand = 11,
+    WaitForMeteringReceipt = 12,
+    WaitForWeldingDetectionSessionStop = 13,
+    WaitForTerminatedSession = 14,
+};
+
+struct V2gCommunicationState {
+    std::optional<V2gDin70121CommunicationState> din70121;
+    std::optional<V2gIso15118AcCommunicationState> iso15118_ac;
+    std::optional<V2gIso15118DcCommunicationState> iso15118_dc;
+};
+
+inline bool operator==(V2gCommunicationState const& lhs, V2gCommunicationState const& rhs) {
+    return lhs.din70121 == rhs.din70121 && lhs.iso15118_ac == rhs.iso15118_ac && lhs.iso15118_dc == rhs.iso15118_dc;
+}
+
+inline bool operator!=(V2gCommunicationState const& lhs, V2gCommunicationState const& rhs) {
+    return !(lhs == rhs);
+}
+
+enum class V2gMessageState {
+    SupportedAppProtocol = 0,
+    SessionSetup = 1,
+    ServiceDiscovery = 2,
+    ServiceDetail = 3,
+    PaymentServiceSelection = 4,
+    PaymentDetails = 5,
+    Authorization = 6,
+    ChargeParameterDiscovery = 7,
+    MeteringReceipt = 8,
+    CertificateUpdate = 9,
+    CertificateInstallation = 10,
+    ChargingStatus = 11,
+    CableCheck = 12,
+    PreCharge = 13,
+    PowerDelivery = 14,
+    CurrentDemand = 15,
+    WeldingDetection = 16,
+    SessionStop = 17,
+    Unknown = 18,
+};
+
+enum class V2gServerStatus {
+    Inactive = 0,
+    Active = 1,
+};
+
+enum class V2gEvErrorCode {
+    NO_ERROR = 0,
+    FAILED_RESSTemperatureInhibit = 1,
+    FAILED_EVShiftPosition = 2,
+    FAILED_ChargerConnectorLockFault = 3,
+    FAILED_EVRESSMalfunction = 4,
+    FAILED_ChargingCurrentdifferential = 5,
+    FAILED_ChargingVoltageOutOfRange = 6,
+    Reserved_A = 7,
+    Reserved_B = 8,
+    Reserved_C = 9,
+    FAILED_ChargingSystemIncompatibility = 10,
+    NoData = 11,
+};
+
+struct CertChainState {
+    bool configured{false};
+    bool synced{false};
+    int32_t num_files{0};
+    int32_t num_useful_files{0};
+};
+
+struct CertTelemetry {
+    bool config_complete{false};
+    bool sync_complete{false};
+    CertChainState secc_chain;
+    CertChainState mo_root;
+};
+
+struct EvseControlStatus {
+    bool authorisation_finished{false};
+    bool emergency_stop{false};
+    bool error_stop{false};
+    bool normal_stop{false};
+    bool contract_payment_enabled{false};
+    bool free_charging_enabled{false};
+};
+
+struct V2gTransport {
+    V2gCommunicationState comm_state;
+    V2gMessageState message_state{V2gMessageState::SupportedAppProtocol};
+    V2gServerStatus udp_server_status{V2gServerStatus::Inactive};
+    V2gServerStatus tcp_listener_status{V2gServerStatus::Inactive};
+    V2gServerStatus tcp_server_status{V2gServerStatus::Inactive};
+    bool tcp_connection_established{false};
+    bool tcp_discovery_enable{false};
+    bool tcp_security_enable{false};
+    bool tcp_security_required{false};
+    int32_t tcp_port_number{0};
+    bool session_setup_requested{false};
+    bool authorization_requested{false};
+    bool charge_parameter_discovery_requested{false};
+    bool cable_check_requested{false};
+    bool pre_charge_requested{false};
+    bool current_demand_requested{false};
+    bool welding_detection_requested{false};
+};
+
+struct V2gEvElectrical {
+    ChargeProgress charge_progress{ChargeProgress::Start};
+    int32_t battery_soc_percent{0};
+    V2gEvErrorCode error_code{V2gEvErrorCode::NO_ERROR};
+    float maximum_current_A{0.};
+    float maximum_power_W{0.};
+    float maximum_voltage_V{0.};
+    float maximum_rated_current_A{0.};
+    float maximum_rated_power_W{0.};
+    float maximum_rated_voltage_V{0.};
+    float target_current_A{0.};
+    float target_voltage_V{0.};
+    int32_t remaining_time_bulk_min{0};
+    int32_t remaining_time_full_min{0};
+    float energy_request_Wh{0.};
+    float energy_capacity_Wh{0.};
+};
+
+struct V2gPaymentService {
+    bool certificate_install_requested{false};
+    bool certificate_update_requested{false};
+    bool charging_service_requested{false};
+    bool external_payment_requested{false};
+    bool contract_payment_requested{false};
+    bool contract_payment_approved{false};
+    bool contract_payment_error{false};
+    bool internet_ftp20_requested{false};
+    bool internet_ftp21_requested{false};
+    bool internet_http80_requested{false};
+    bool internet_https443_requested{false};
+};
+
+struct V2gChargerStatus {
+    std::string evcc_id;           // "XX:XX:XX:XX:XX:XX"
+    std::string selected_protocol; // "DIN70121" / "ISO15118-2-2010" / "ISO15118-2-2013"
+    bool cable_check_status{false};
+    bool param_discovery_finished{false};
+    std::string isolation_status;
+    float dynamic_max_current_A{0.};
+    float dynamic_max_power_W{0.};
+    float dynamic_max_voltage_V{0.};
+};
+
+struct V2gEvseElectrical {
+    float present_current_A{0.};
+    float present_voltage_V{0.};
+    float maximum_rated_current_A{0.};
+    float maximum_rated_power_W{0.};
+    float maximum_rated_voltage_V{0.};
+    float minimum_rated_current_A{0.};
+    float minimum_rated_voltage_V{0.};
+    float current_ripple_A{0.};
+    float current_tolerance_A{0.};
+};
+
+} // namespace everest::lib::API::V1_0::types::telemetry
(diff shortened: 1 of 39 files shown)


SebaLukas:
Bazel ci jobs are failing

lib/everest/everest_api_types/include/everest_api_types/telemetry/API.hpp
```diff
@@ -0,0 +1,117 @@
...
+    bool contract_payment_enabled{false};
+    bool free_charging_enabled{false};
+};
+
+struct V2gTransport {
+    int32_t comm_state{0};
+    int32_t message_state{0};
+    int32_t udp_server_status{0};
+    int32_t tcp_listener_status{0};
+    int32_t tcp_server_status{0};
```
SebaLukas: It would make sense to define enum classes for the different int32_t values.

lib/everest/everest_api_types/include/everest_api_types/telemetry/API.hpp
```diff
@@ -0,0 +1,117 @@
...
+    bool cable_check_requested{false};
+    bool pre_charge_requested{false};
+    bool current_demand_requested{false};
+    bool welding_detection_requested{false};
+};
+
+struct V2gEvElectrical {
+    ChargeProgress charge_progress{ChargeProgress::Start};
+    int32_t battery_soc_percent{0};
+    int32_t error_code{0};
```
SebaLukas: Here as well

modules/EVSE/EvseManager/evse/evse_managerImpl.cpp
```diff
@@ -7,6 +7,7 @@
 #include <date/tz.h>
 #include <utils/date.hpp>
 
+#include <everest_api_types/telemetry/json_codec.hpp>
```
SebaLukas, line 10: This include is not really needed because `evse_managerImpl.hpp` has also the same include

modules/EVSE/EvseManager/evse/evse_managerImpl.cpp
```diff
@@ -22,6 +23,18 @@ bool str_to_bool(const std::string& data) {
     return false;
 }
 
+void evse_managerImpl::publish_control_telemetry() {
```
SebaLukas: You could pass a local `control_status` variable as a parameter to the `publish` function here.
In the current implementation, it might not be thread-safe, as `control_status` could be written to and/or read from at the same time.
As I understand the changes, in most cases only one member of `control_status` is true and the rest are false. Except `contract_payment_enabled` and `free_charging_enabled`

modules/EVSE/EvseV2G/charger/ISO15118_chargerImpl.cpp
```diff
@@ -4,6 +4,7 @@
 #include "ISO15118_chargerImpl.hpp"
 #include "log.hpp"
 #include "sdp.hpp"
+#include "telemetry_publisher.hpp"
```
SebaLukas, line 7: Are these includes really necessary in the cpp files?

modules/EVSE/EvseV2G/din_server.cpp
```diff
@@ -231,6 +254,33 @@ static void publish_din_charge_parameter_discovery_req(
...
+        if (ctx->telemetry_publisher) {
+            ctx->telemetry_publisher->update_ev_electrical([&](auto& telemetry) {
+                telemetry.maximum_current_A = evMaximumCurrentLimit;
+                telemetry.maximum_power_W = evMaximumPowerLimit;
+                telemetry.maximum_voltage_V = evMaximumVoltageLimit;
+                telemetry.maximum_rated_current_A = evMaximumCurrentLimit;
+                telemetry.maximum_rated_power_W = evMaximumPowerLimit;
+                telemetry.maximum_rated_voltage_V = evMaximumVoltageLimit;
+                if (v2g_charge_parameter_discovery_req->DC_EVChargeParameter.EVEnergyCapacity_isUsed ==
+                    (unsigned int)1) {
```
SebaLukas: I think you can remove all c style cast `(unsigned int)` or ``(int)`

modules/EVSE/EvseV2G/EvseV2G.cpp
```diff
@@ -52,6 +54,7 @@ void EvseV2G::init() {
     (void)openssl::set_log_handler(log_handler);
     tls::Server::configure_signal_handler(SIGUSR1);
     v2g_ctx->tls_server = &tls_server;
+    v2g_ctx->telemetry_publisher = telemetry_publisher.get();
```
SebaLukas, line 58: Maybe it make sense to use `std::move` instead of a pointer here because only `v2g_ctx` is using `telemetry_publisher`

modules/EVSE/EvseV2G/EvseV2G.cpp
```diff
@@ -108,6 +111,8 @@ void EvseV2G::ready() {
     invoke_ready(*p_charger);
     invoke_ready(*p_extensions);
 
+    telemetry_publisher->publish_all();
```
SebaLukas, line 115: Needs to be called in the `v2g_ctx` if `std::move` is used

modules/EVSE/EvseV2G/iso_server.cpp
```diff
@@ -396,6 +445,33 @@ static void publish_iso_charge_parameter_discovery_req(
...
+        if (ctx->telemetry_publisher) {
+            ctx->telemetry_publisher->update_ev_electrical([&](auto& telemetry) {
+                telemetry.maximum_current_A = evMaximumCurrentLimit;
+                telemetry.maximum_power_W = evMaximumPowerLimit;
+                telemetry.maximum_voltage_V = evMaximumVoltageLimit;
+                telemetry.maximum_rated_current_A = evMaximumCurrentLimit;
+                telemetry.maximum_rated_power_W = evMaximumPowerLimit;
+                telemetry.maximum_rated_voltage_V = evMaximumVoltageLimit;
+                if (v2g_charge_parameter_discovery_req->DC_EVChargeParameter.EVEnergyCapacity_isUsed ==
+                    (unsigned int)1) {
```
SebaLukas: Same here, remove all c style cast

SebaLukas (changes requested):

modules/EVSE/EvseManager/evse/evse_managerImpl.cpp
djchhp: I can't find that include in evse_managerImpl.hpp, that includes onlye the API.hpp definitions, but not the json codec.

modules/EVSE/EvseV2G/charger/ISO15118_chargerImpl.cpp
djchhp: Yes, they are necessary. Even in EvseV2G.hpp for the unique_ptr

modules/EVSE/EvseV2G/din_server.cpp
djchhp: agreed

lib/everest/everest_api_types/include/everest_api_types/telemetry/API.hpp
djchhp: agreed

lib/everest/everest_api_types/include/everest_api_types/telemetry/API.hpp
djchhp: agreed

modules/EVSE/EvseV2G/EvseV2G.cpp
djchhp: v2g_context is currently allocated with calloc/free, so moving a std::unique_ptr into it would require changing context construction/destruction. I think it is better to keep it non owned. The context also already references other module owned objects.

modules/EVSE/EvseV2G/EvseV2G.cpp
djchhp: look above

modules/EVSE/EvseManager/evse/evse_managerImpl.cpp
djchhp: Agreed

modules/EVSE/EvseV2G/EvseV2G.cpp
SebaLukas: Yeah you are right. I had the same issue. Then lets keep it that way :+1:

lib/everest/everest_api_types/include/everest_api_types/telemetry/API.hpp
```diff
@@ -0,0 +1,177 @@
...
+    StateId6 = 6,
+    StateId7 = 7,
+    StateId8 = 8,
+    StateId9 = 9,
+    StateId10 = 10,
+    StateId11 = 11,
+    StateId12 = 12,
+    StateId13 = 13,
+    StateId14 = 14,
+};
```
SebaLukas: StateID0-14 are which states?

modules/EVSE/EvseV2G/din_server.cpp
```diff
@@ -765,6 +885,11 @@ static enum v2g_event handle_din_cable_check(struct v2g_connection* conn) {
     struct din_CableCheckResType* res = &conn->exi_out.dinEXIDocument->V2G_Message.Body.CableCheckRes;
     enum v2g_event nextEvent = V2G_EVENT_NO_EVENT;
 
+    if (conn->ctx->telemetry_publisher) {
+        conn->ctx->telemetry_publisher->update_transport(
+            [&](auto& transport) { transport.cable_check_requested = true; });
```
SebaLukas: That is called every time for every cable check req message. I guess that should be called only once?

modules/EVSE/EvseV2G/din_server.cpp
```diff
@@ -828,6 +953,11 @@ static enum v2g_event handle_din_pre_charge(struct v2g_connection* conn) {
     struct din_PreChargeResType* res = &conn->exi_out.dinEXIDocument->V2G_Message.Body.PreChargeRes;
     enum v2g_event nextEvent = V2G_EVENT_NO_EVENT;
 
+    if (conn->ctx->telemetry_publisher) {
+        conn->ctx->telemetry_publisher->update_transport(
+            [&](auto& transport) { transport.pre_charge_requested = true; });
```
SebaLukas: Same as cable check

modules/EVSE/EvseV2G/din_server.cpp
```diff
@@ -866,6 +996,11 @@ static enum v2g_event handle_din_current_demand(struct v2g_connection* conn) {
     struct din_CurrentDemandResType* res = &conn->exi_out.dinEXIDocument->V2G_Message.Body.CurrentDemandRes;
     enum v2g_event nextEvent = V2G_EVENT_NO_EVENT;
 
+    if (conn->ctx->telemetry_publisher) {
+        conn->ctx->telemetry_publisher->update_transport(
+            [&](auto& transport) { transport.current_demand_requested = true; });
```
SebaLukas: Same as cable check and pre charge

modules/EVSE/EvseV2G/din_server.cpp
```diff
@@ -920,6 +1055,11 @@ static enum v2g_event handle_din_welding_detection(struct v2g_connection* conn)
     struct din_WeldingDetectionResType* res = &conn->exi_out.dinEXIDocument->V2G_Message.Body.WeldingDetectionRes;
     enum v2g_event nextEvent = V2G_EVENT_NO_EVENT;
 
+    if (conn->ctx->telemetry_publisher) {
+        conn->ctx->telemetry_publisher->update_transport(
+            [&](auto& transport) { transport.welding_detection_requested = true; });
```
SebaLukas: See above

modules/EVSE/EvseV2G/iso_server.cpp
```diff
@@ -268,6 +310,14 @@ static void publish_iso_authorization_req(struct iso2_AuthorizationReqType const
     // V2G values that can be published: Id, Id_isUsed, GenChallenge, GenChallenge_isUsed
 }
 
+static void publish_iso_authorization_req(struct v2g_context* ctx, struct iso2_AuthorizationReqType const* const req) {
+    publish_iso_authorization_req(req);
+
+    if (ctx->telemetry_publisher) {
+        ctx->telemetry_publisher->update_transport([&](auto& transport) { transport.authorization_requested = true; });
```
SebaLukas: Same as CableCheck, etc.

modules/EVSE/EvseV2G/din_server.cpp
```diff
@@ -289,11 +359,12 @@ static void publish_din_current_demand_req(struct v2g_context* ctx,
...
 
-    publish_dc_ev_target_voltage_current(ctx,
-                                         calc_physical_value(v2g_current_demand_req->EVTargetVoltage.Value,
-                                                             v2g_current_demand_req->EVTargetVoltage.Multiplier),
-                                         calc_physical_value(v2g_current_demand_req->EVTargetCurrent.Value,
-                                                             v2g_current_demand_req->EVTargetCurrent.Multiplier));
+    const auto target_voltage = calc_physical_value(v2g_current_demand_req->EVTargetVoltage.Value,
+                                                    v2g_current_demand_req->EVTargetVoltage.Multiplier);
+    const auto target_current = calc_physical_value(v2g_current_demand_req->EVTargetCurrent.Value,
+                                                    v2g_current_demand_req->EVTargetCurrent.Multiplier);
```
SebaLukas, line 382: That could lead spamming the mqtt bus. This also applies to other messages and paramaters that are called repeatedly (CableCheck, PreCharge, CurrentDemand, WeldingDetection

modules/EVSE/EvseV2G/din_server.cpp
```diff
@@ -478,6 +582,11 @@ enum v2g_event states::handle_din_contract_authentication(struct v2g_connection*
         &conn->exi_out.dinEXIDocument->V2G_Message.Body.ContractAuthenticationRes;
     enum v2g_event nextEvent = V2G_EVENT_NO_EVENT;
 
+    if (conn->ctx->telemetry_publisher) {
+        conn->ctx->telemetry_publisher->update_transport(
+            [](auto& transport) { transport.authorization_requested = true; });
```
SebaLukas: Same as the others

modules/EVSE/EvseV2G/iso_server.cpp
```diff
@@ -1904,6 +2061,11 @@ static enum v2g_event handle_iso_cable_check(struct v2g_connection* conn) {
     struct iso2_CableCheckResType* res = &conn->exi_out.iso2EXIDocument->V2G_Message.Body.CableCheckRes;
     enum v2g_event next_event = V2G_EVENT_NO_EVENT;
 
+    if (conn->ctx->telemetry_publisher) {
+        conn->ctx->telemetry_publisher->update_transport(
+            [&](auto& transport) { transport.cable_check_requested = true; });
```
SebaLukas: Same

modules/EVSE/EvseV2G/iso_server.cpp
```diff
@@ -1953,6 +2115,11 @@ static enum v2g_event handle_iso_pre_charge(struct v2g_connection* conn) {
     struct iso2_PreChargeResType* res = &conn->exi_out.iso2EXIDocument->V2G_Message.Body.PreChargeRes;
     enum v2g_event next_event = V2G_EVENT_NO_EVENT;
 
+    if (conn->ctx->telemetry_publisher) {
+        conn->ctx->telemetry_publisher->update_transport(
+            [&](auto& transport) { transport.pre_charge_requested = true; });
```
SebaLukas: Same

modules/EVSE/EvseV2G/iso_server.cpp
```diff
@@ -1991,6 +2158,11 @@ static enum v2g_event handle_iso_current_demand(struct v2g_connection* conn) {
     struct iso2_CurrentDemandResType* res = &conn->exi_out.iso2EXIDocument->V2G_Message.Body.CurrentDemandRes;
     enum v2g_event next_event = V2G_EVENT_NO_EVENT;
 
+    if (conn->ctx->telemetry_publisher) {
+        conn->ctx->telemetry_publisher->update_transport(
+            [&](auto& transport) { transport.current_demand_requested = true; });
```
SebaLukas: Same

modules/EVSE/EvseV2G/iso_server.cpp
```diff
@@ -2137,6 +2309,11 @@ static enum v2g_event handle_iso_welding_detection(struct v2g_connection* conn)
     struct iso2_WeldingDetectionResType* res = &conn->exi_out.iso2EXIDocument->V2G_Message.Body.WeldingDetectionRes;
     enum v2g_event next_event = V2G_EVENT_NO_EVENT;
 
+    if (conn->ctx->telemetry_publisher) {
+        conn->ctx->telemetry_publisher->update_transport(
+            [&](auto& transport) { transport.welding_detection_requested = true; });
```
SebaLukas: Same

SebaLukas (changes requested):

FaHaGit:
I've also looked over the PR and have some questions about it. Did I understand correctly that the new telemetry interface is intentionally used outside the regular EVerest interfaces, so a consumer has to subscribe directly to MQTT topics based on the configured telemetry.id and the specific topic name? If so, is there a central place where the available topics and their JSON payloads are documented, or does a consumer currently have to derive them from the source code and the corresponding publish sites? At the moment, my impression is that this telemetry interface is primarily intended for debugging and testing purposes, for example via an MQTT observer tools (like MQTT explorer), rather than for other EVerest modules to consume directly. Is that understanding correct?

djchhp:
@SebaLukas Telemetry is only published when a telemetry id is set for the module. User who are not interested are not affected. My approach would be to keep it as is. If issues arise from too much traffic, we should fix it on the framework level (i.e. implement on change publish only, or a minimal wait interval).

djchhp:
@SebaLukas All comments have been addressed. Publication of telemetry can be configured as on change. 
@FaHaGit I added asyncapi docs for the telemetry

djchhp:
@SebaLukas Could you have another look at this PR please?

modules/EVSE/EvseV2G/manifest.yaml
```diff
@@ -75,6 +75,11 @@ config:
       Enable the built-in SDP server
     type: boolean
     default: true
+  publish_telemetry_only_on_change:
+    description: >-
+      Only publish telemetry blocks when values changed.
+    type: boolean
+    default: false
```
SebaLukas: Maybe change that to true. The previous behaivour did not include any telemetry data and now its only updated if values changed

SebaLukas (dismissed):
One small change we should discuss. Except that 👍

SebaLukas (approved):

hikinggrass (approved):

PR #2315, opened 2026-06-16
InfyPower: fix communication failure recovery in FIXED_ADDRESS mode 

IvarWithoutBones:
## Describe your changes

The InfyPower driver tracks whether each converter module it manages responds to CAN messages, and if it doesn't for more than `device_connection_timeout_s` the module is considered inactive and won't be used for commands like `setExportVoltageCurrent`. 

In `OperatingMode::GROUP_DISCOVERY` the driver periodically polls the module count and updates which modules are active accordingly, making it possible to recover from temporary communication failure. However, in `OperatingMode::FIXED_ADDRESS` the driver would never attempt to communicate with inactive modules again, locking up the system until EVerest is restarted. This PR fixes that by polling the status of all configured modules, in case they came back online. 

It also includes fix that makes it possible for multiple InfyPower devices to exist on the same CAN bus, while the driver only manages a subset of them. It'd previously process incoming responses from any device, even when in `FIXED_ADDRESS` mode where you explicitly specify which converters it should work with. This messes up the driver state, since it'll use response data from unrelated converters. It now only processes messages from configured addresses in `FIXED_ADDRESS` mode.

### DIFF
--- a/modules/HardwareDrivers/PowerSupplies/InfyPower/can_driver_acdc/InfyCanDevice.cpp
+++ b/modules/HardwareDrivers/PowerSupplies/InfyPower/can_driver_acdc/InfyCanDevice.cpp
@@ -60,8 +60,9 @@ void InfyCanDevice::set_config_values(const std::string& addrs, int group_addr, 
     EVLOG_info << "Infy: Operating with controller address: 0x" << std::hex << controller_address;
     if (!addrs.empty()) {
         operating_mode = OperatingMode::FIXED_ADDRESS;
-        active_module_addresses = parse_module_addresses(addrs);
-        expected_module_count = active_module_addresses.size();
+        module_addresses = parse_module_addresses(addrs);
+        expected_module_count = module_addresses.size();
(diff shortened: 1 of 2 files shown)


florinmihut (changes requested):
Looks good to me, nice catch!

However there might be a possible a case where a devices after disappearing and reappearing might be ignored by the driver. It hasn't directly something with your changes from the PR, just improve the functionality so that the hotplug works better. 
I have prepared a patch fixing it here: https://github.com/EVerest/EVerest/pull/2427
I kept your commits and messages. You can cherry-pick that commit here and I will approve it.

IvarWithoutBones:
Ah right, good catch!

I've cherry-picked the commit like you suggested, and rebased my own on top of main (hence the force push). I'm not sure that properly preserved the authorship information though, since GitHub shows a tag saying I committed it. Feel free to close this PR in favor of your own if you prefer.

florinmihut (approved):
LGTM

PR #1935, opened 2026-03-10
Replace the SLAC stack with an event-driven implementation

djchhp:
## Describe your changes
This replaces the ISO 15118-3 SLAC data link negotiation with a modernized implementation. The EVSE and EV side are re-implemented as **EvseSlac** and **EvSlac** on top of a new **lib/everest/slac**:
- event-driven single-threaded I/O based on everest::lib::io
- the state machines re-written in boost::msm

The implementation is fully unit-testable without hardware. Beyond the port, it closes conformance gaps of the previous implementation (CM_VALIDATE BCB-toggle validation, CM_AMP_MAP, matched-state link supervision) and hardens the behavior seen with real-world EVs and modems.

The legacy stack (the old lib/everest/slac, lib/everest/fsm, modules/EVSE/EvseSlac, modules/EV/EvSlac) is removed and the new code takes over its names. Existing configurations keep working without change: the manifests are a strict superset of the old ones (same keys, types and defaults), same interfaces, same implementation id. All six in-tree configs on `EvseSlac`/`EvSlac` now run the new modules.

### Architecture
  - Let the SLAC modules start without the PLC interface being present
  - Reduce thread count: I/O and the SLAC state machine run in a single thread
  - Reduce code by leveraging everest::lib::io and everest::lib::util
  - Re-implement the SLAC state machine in terms of boost::msm
  - New lib/everest/slac library, fully unit-testable without hardware — the FSMs run headless in tests (~5.7k lines of unit/regression tests)
  - EV-side counterpart: EvSlac module on the same library, usable for simulators and self-tests
  - Automatic modem vendor detection at startup (Qualcomm via OP_ATTR, Lumissil via NSCM_GET_VERSION), device info logged
  - SLAC session telemetry via new everest_api_types telemetry types; INFO-level session narration in the log

### ISO 15118-3 conformance (validated against the ISO 15118-5 SLAC conformance suite)
  - CM_VALIDATE: full two-step BCB-toggle validation with timed observation window — new count_bc command on the slac interface, EvseManager pushes its B→C transition count per session
  - CM_AMP_MAP (A.9.6 transmit-power limitation): always responds to incoming requests; optional SECC-initiated direction with an operator-provided amplitude-map YAML (amp_map_file)
  - CM_SLAC_MATCH.REQ bounded by TT_match_sequence after validation; late requests get no CNF
  - CM_ATTEN_CHAR.IND sent immediately after the last sound (no needless wait for the sounding timeout)
[...]

PR #2324, opened 2026-06-17
EvseSlac: Handle CM_SET_KEY failures, add retries and improve internal key state

KyotoFox:
## Describe your changes

Hello!

We have been doing some tests with EVerest and ISO 15118-2 with EvseV2G and EvseSlac, and sometimes we see CM_SET_KEY fail with our QCA7005.

Right now, EvseSlac doesn't validate if the CM_SET_KEY request is successful or not.
This PR adds validation and retries (up to a configurable amount) if it does fail.

Not only is this more robust to "reset" the PLC network between sessions, 
it also improves failures in conformance tests like TC_SECC_CMN_VTB_PLCLinkStatus_002 
(this conformance test first performs a valid SLAC match, then verifies that the EVSE leaves the PLC network after the EV disconnects/goes to state A).

In our case, this conformance test would fail because the CM_SET_KEY-request was unsuccessful. 
I'm not sure what the exact root cause behind the failure is, but retrying seems to fix it.
The CM_SET_KEY failure isn't consistent either. The first CM_SET_KEY.REQ is often successful, but in the above conformance test it seems to fail consistently.
My theory is that the CM_SET_KEY in this test occurs very quickly after SLAC matching is done, so there might be some timing that the QCA doesn't like when CM_SET_KEY.REQ is sent with this timing.

Also of note is that [Qualcomm](https://github.com/qca/open-plc-utils/blob/46c3506453c15b873fd6ed3e76c9872cea5e143a/slac/evse_cm_set_key.c#L170-L173) INVERTS the result field in CM_SET_KEY.CNF.
While the HomePlugAV GreenPhy specifies 0x00 as success, Qualcomm treats 0x00 as failure and 0x01 as success 🙃
I did consider checking `modem_vendor` for this inversion, but as I understand it, this is the "de-facto" standard for CM_SET_KEY.RES in PLC modems? ([See also](https://github.com/qca/open-plc-utils/issues/161#issuecomment-1130054300))

During our testing, we have also seen occasional cases where SLAC matching succeeds, but the EV & EVSE fails to join the network. 
I think one source of this failure could be caused by EvseSlac and the PLC modem having a mismatching NMK.
I made an improvement where the new NMK isn't "committed" internally until CM_SET_KEY is successful.
(The internal NMK is used in CM_SLAC_MATCH.CNF for the next session)

This _does_ lead to the case where if CM_SET_KEY _always_ fails, then EvseSlac will always use the same NMK.
This could also be improved upon, but I'm not sure if there is any reason to do so? 
Something is clearly wrong if CM_SET_KEY consistently fails, and the NMK isn't really critical for security anyways.

There could also be the reverse case, where CM_SET_KEY _fails_, but the PLC modem _does_ update the key. Then the next SLAC session will have the wrong NMK, but at least there will have been warnings about it in the logs.
A further improvement here could also be to actually read the NMK back out, but I know this might not always be possible (e.g. if you configure the QCA to update the NMK without writing to flash)

This is my first PR in EVerest, so I'm happy to get some feedback!

### DIFF
--- a/lib/everest/slac/fsm/evse/include/everest/slac/fsm/evse/context.hpp
+++ b/lib/everest/slac/fsm/evse/include/everest/slac/fsm/evse/context.hpp
@@ -162,8 +162,12 @@ struct EvseSlacConfig {
     // flag for using 5% PWM in AC mode
     bool ac_mode_five_percent{true};
 
-    // timeout for CM_SET_KEY.REQ
+    // timeout for CM_SET_KEY.REQ; also used as the interval between retries
     int set_key_timeout_ms = 500;
+
+    // maximum number of CM_SET_KEY.REQ attempts before giving up (a new attempt is made every
+    // set_key_timeout_ms if the previous one timed out or was rejected by the modem)
+    int set_key_max_attempts = 10;
 
     // timeout for CM_SLAC_PARM.REQ
     int slac_init_timeout_ms = slac::defs::TT_EVSE_SLAC_INIT_MS;
--- a/lib/everest/slac/fsm/evse/include/everest/slac/fsm/evse/context.hpp
+++ b/lib/everest/slac/fsm/evse/include/everest/slac/fsm/evse/context.hpp
@@ -156,7 +156,6 @@ struct EvseSlacConfig {
     uint8_t plc_peer_mac[ETH_ALEN] = {0x00, 0xB0, 0x52, 0x00, 0x00, 0x01};
 
     // FIXME (aw): we probably want to use std::array here
-    void generate_nmk();
     uint8_t session_nmk[slac::defs::NMK_LEN]{};
 
     // flag for using 5% PWM in AC mode
--- a/lib/everest/slac/fsm/evse/include/everest/slac/fsm/evse/states/others.hpp
+++ b/lib/everest/slac/fsm/evse/include/everest/slac/fsm/evse/states/others.hpp
@@ -25,6 +25,10 @@ struct ResetState : public FSMSimpleState {
     int set_key_attempts{0};
     // time the last CM_SET_KEY.REQ was sent, used to space out retries
     std::chrono::steady_clock::time_point last_attempt_time{};
+
+    // Candidate NMK currently being programmed into the modem. 
+    // Promoted to slac_config.session_nmk when CM_SET_KEY.CNF succeeds, or else we continue with the old key.
+    uint8_t pending_nmk[slac::defs::NMK_LEN]{};
 };
 
 struct ResetChipState : public FSMSimpleState {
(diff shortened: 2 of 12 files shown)


lategoodbye:
@KyotoFox Could you please provide more details about the error itself:
Which QCA7005 firmware is affected?
How often does this occur?
Can you provide more details about the QCA7005 PIB configuration?
Does this issue occur with or without regular reset of the QCA7005 by EVerest?

djchhp:
Ported (experimentally) to https://github.com/EVerest/EVerest/pull/1935 and tested with QCA 3.x and Lumissil. Unfortunately I could not observe a single case of set key failures.

KyotoFox:
Thanks for testing with Lumissil as well!

In my configuration I have been running with a QCA7005 with the latest firmware, 3.5.0:
`MAC-QCA7005-3.5.0.25-00-20251127-CS`.

The PIB has "nothing special" in it, just the default `QCA7005-EVSE-SpiSlave-Automotive-3.5.0.pib` with:
- Normal EVSE-side SLAC config
- Coax Medium
- Always CCo
- Options set by evse.sh in `open-plc-utils`
- "Resolve NID collision without reboot" (new option) turned on
- Event Notifications for link loss, etc... turned on

I don't have super-good data at the moment for how often it happens, but it seems to happen fairly reliably for short sessions in my setup. If i recall correctly it doesn't always result in the key not being set (e.g. the chip returns an error, but the new key _is_ set). I'll try to get back with some better data on this.

We are running _without_ reset (both `do_chip_reset` and `reset_instead_of_fail` set to false).
I haven't tested much with reset yet (although I did validate that the patch works with reset, as that changes the state machine)
From some earlier testing we found it more stable to not do the reset.

lategoodbye:
@KyotoFox Thank you. In case you have access to Avitar, could you please check the state of the following PIB settings:

"Skip Nvm PibUpdate For SetNmk"
"Local NMK Change Applied Immediately (no reboot)"
"BootFromHost BackGround Update"

In case you don't have Avitar maybe you can ask your supplier.

djchhp:
@KyotoFox I just got assigned to review this PR. Thank you very much for your contribution.

I had a first look at the code again and currently have 2 major concerns.
1.) The 0x01 result code is the only accepted value, this was previously unchecked. I expect all chips and will all firmware versions handle it this way, but we cannot test this. In order to not to break existing installations, the original behavior must be the default for all existing configurations. This includes accepting both result values and not using retries at all.
2.) After giving up, the FSM now continues into matching while advertising a key that was never sent to the modem in this cycle, on first boot not at all. If the modem did apply the key but the confirmation was lost or misclassified, this guarantees a link failure where the old code stayed consistent. On give-up, either fall back to the key that was actually transmitted, or fail loudly instead of silently entering Idle.

Please also make sure, that the tests you added are picked up by the CI.

djchhp:
This PR is superseded by #1935 which re-implements EvseSlac and libslac. The complete logic implemented here has been integrated and attributed. Thank you for you contribution.

PR #2325, opened 2026-06-17
feat(energy): add temperature-based AC current derating

florinmihut:
Add AcTemperatureDerating module that applies AC current limits to an EnergyNode from temperature_sensor readings using configurable derating curves (linear interpolation, minimum across sensors). Includes a new temperature_sensor interface with providers in CarloGavazzi_EM580 (body temperature) and YetiSimulator (load-derived), a SIL config, docs and unit tests for the curve logic.

Limits are recomputed on each reading and on a periodic timer so stale sensors fall back conservatively; decreases apply immediately while increases are debounced.


interfaces/temperature_sensor.yaml
```diff
@@ -0,0 +1,12 @@
+description: >-
+  Generic temperature sensor reporting interface. Providers publish one or more
+  temperature readings that can be consumed independently of powermeter data.
```
mhei: I think we should drop the reference to power meter data here - a temperature sensor is completely independent and it's only the first public relation here.

interfaces/temperature_sensor.yaml
```diff
@@ -0,0 +1,12 @@
...
+vars:
+  temperatures:
+    description: >-
+      Temperature readings in degree Celsius. The meaning of each entry is product
+      specific and may be identified via the identification or location fields.
+    type: array
+    items:
+      type: object
+      $ref: /temperature#/Temperature
```
mhei, line 12: At chargebyte, we already have a very similar custom interface defined: https://github.com/chargebyte/everest-chargebyte/blob/main/interfaces/cb_temperatures.yaml
The missing part here is our error codes, see https://github.com/chargebyte/everest-chargebyte/blob/main/errors/cb_temperatures.yaml . I'm wondering whether we could add these errors here too? Then it would be possible to drop our own interface and switch over to this one here.

modules/EnergyManagement/CMakeLists.txt
```diff
@@ -1,2 +1,3 @@
 ev_add_module(EnergyManager)
 ev_add_module(EnergyNode)
+ev_add_module(AcTemperatureDerating)
```
mhei: we should sort alphabetically, will make merges later easier

modules/EnergyManagement/AcTemperatureDerating/manifest.yaml
```diff
@@ -0,0 +1,61 @@
...
+      clamps below the first / above the last point.
+
+      module_id is the temperature provider instance id from active_modules /
+      connections.temperature.module_id (not the module type name and not
+      implementation_id). identification is Temperature.identification from the
+      published reading. Each Temperature.identification in a batch needs a matching curve key.
+      Readings without identification log a warning and use fallback_max_current_A.
+
+      Every connected temperature provider must have at least one curve key with prefix
+      module_id. Missing curves at runtime log a warning and apply fallback_max_current_A.
```
mhei: I find that difficult: if you have a provider that has very different sources, then you might only want to evaluate exactly one, but not have a warning for all the others.

modules/EnergyManagement/AcTemperatureDerating/manifest.yaml
```diff
@@ -0,0 +1,61 @@
...
+    default: 1000
+requires:
+  temperature:
+    interface: temperature_sensor
+    min_connections: 1
+    max_connections: 16
+  energy_node:
+    interface: external_energy_limits
+    min_connections: 1
+    max_connections: 1
```
mhei, line 89: Isn't `min_connection == 1 && max_connection == 1` the default? So we can omit this here?

interfaces/temperature_sensor.yaml
florinmihut: Good point, thanks!

interfaces/temperature_sensor.yaml
florinmihut: I think we need to discuss this a bit. This is going a bit longer post probably.

You want that the temperature provider to raise those errors? 
**SelftestFailed** - that is equivalent to a communication failure from logical point of view because it basically telling me that it cant't provide a temperature (in this particular case because of a self test) but it is equivalent to the communication failure. 
**ChargingAbort** - you want to stop the session? Or this is an information that the session was aborted? Why would a temperature provider raise an error like this? Or should the derating component raise this error?

Who should react on those errors?

For the first error, this component it is already detecting it - imagine that the temperature provider is stalled and it is not sending anything. Sure, an error raised saying that it will not provide values is a nicer way to specify it, but at the end the stalling mechanism is still needed and the error is not bringing anything new. If the stalling is detected, you can specify a fallback value - it can even be 0A.

For the second error - this is a bit complicated. We designed this to be generic. You have temperature sensors that are generic and they should not take decisions. That is why this error could be raised by this derating component. However, canceling the session might not be easily possible if you have multiple connectors since this is connected to an EnergyNode  and not to the EVSEManager. What we could do, we can reduce the current to 0A which is going to pause the charging. If the temperature goes under, we can again continue charging.  So aborting the session might not be necessary or practical. 

What remains: what should happen if the temperature reaches a level that is dangerous high? In that case we don't talk about regulating we talk about safety critical. If the temperature raise is safety critical - should be implemented in a different entity since EVerest is not safety certified.

modules/EnergyManagement/AcTemperatureDerating/manifest.yaml
florinmihut: You are right. 
If you have a driver that can provide 2 temperature values, it would be stupid to implement twice the temperature provided interface. That leads us to this solution with the identification and location.
Now, what I want to avoid is people using the derating component wrongly configured and then complaining that something is not working. That means I am validating a bit strict how it is configured. 
The only way right now to ignore the second source in a particular provider is to provide a curve to a high value for the current. 
```
"my_meter.Ambient": [
    {"temp_C": -20, "max_current_A": 32},
    {"temp_C": 80, "max_current_A": 32}
  ]
```
This would cause derating to "ignore" it.
Not nice. I could add an ignore identifications list and that might simplify it.
I am open to suggestions.

modules/EnergyManagement/AcTemperatureDerating/manifest.yaml
florinmihut: I am not sure what the default is. Theoretically, yes, I could omit this, but it is no harm either to leave it.

interfaces/temperature_sensor.yaml
mhei: Sorry for delayed response and thanks for your thoughts...

It’s right; the idea behind our current interface is for it to raise these two specific errors.

Regarding `SelftestFailed`: I don’t agree that this error is synonymous with a communication error. Yes, the effect or reaction is the same, but isn't the fundamental purpose of (standardized) error codes to enable better diagnostics? I understand that the meter's temperature register doesn't have this `SelftestFailed` state: either communication works—meaning you can read a value that should generally be valid—or it doesn't, in which case it’s a communication error. However, our interface communicates with a safety controller via UART, so it makes sense to distinguish between being unable to communicate at all and an internal test failure.

As for the second error, the original idea was to indicate that a specific sensor caused the charging process to abort. But I would agree that this doesn't really belong in this interface, so I’m fine with ignoring it here.

modules/EnergyManagement/AcTemperatureDerating/manifest.yaml
mhei: > I want to avoid ... people ... complaining that something is not working

They will exist one way or another, but I understand the goal.

Personally, it would annoy me to have to make such "pointless" configuration entries just to cover all channels. Our Charge SOM supports up to 4 channels, meaning—in the worst-case scenario—I would have three entries.
A possible (simple) compromise would be to link the warnings to a flag that simply suppresses them. The default behavior could be to display the warning, allowing users to simply change the setting if they wish.
We can also wait for the opinions of other reviewers, but this point should not stand in the way of a merge.

mhei:
The module recently received support for the EM340 meter. I not sure whether the temperature register is available there (at least, it is not in the spec) - I'll try to test. Maybe it is necessary to also place this query in the guard for transaction support...

mhei:
I can confirm, that the Modbus query for the temp raises a Modbus Illegal Data Address exception on a CG EM340, so please guard it.

modules/EnergyManagement/AcTemperatureDerating/manifest.yaml
florinmihut: Hi, thanks for the update. I agree about the "pointless" curves - annoying and stupid. 
I think I like more your initial idea of a ignore list. If in the future you add a sensor but forgets to add a curve, it will complain reminding you to decide what do you want to do with that value. 
A general flag to ignore sensor identities is maybe too dangerous in this case. 
I will go with the ignore_list if that is ok with you too.

interfaces/temperature_sensor.yaml
florinmihut: Ok, let me propose something that I think would help: what if I add the generic errors list?
That contains: CommunicationFailure, VendorError and VendorWarning. 
A SefltestFailed is a VendorError or VendorWarning (at least) in that case. 
We would need to define how the derating component should behave. If we do a fallback on the derating value as defined in the config? A 0A as derating would mean pause (since we set the energy to 0) and a different value would mean charge but with minimal supported energy,
Would that be enough as flexibility?

modules/EnergyManagement/AcTemperatureDerating/manifest.yaml
mhei: Ok for me, thanks!

mhei (dismissed):
Needs rebase, but otherwise LGTM.

modules/EnergyManagement/AcTemperatureDerating/derating/derating_curve.hpp
```diff
@@ -0,0 +1,72 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Pionix GmbH and Contributors to EVerest
+
+#ifndef AC_TEMPERATURE_DERATING_DERATING_CURVE_HPP
+#define AC_TEMPERATURE_DERATING_DERATING_CURVE_HPP
```
Pietfried: Prefered is `#pragma once`

modules/EnergyManagement/AcTemperatureDerating/derating/derating_curve.hpp
```diff
@@ -0,0 +1,72 @@
...
+#include <map>
+#include <optional>
+#include <set>
+#include <string>
+#include <vector>
+
+namespace ac_temperature_derating {
+
+struct DeratingPoint {
+    double temp_C{0.0};
```
Pietfried: Please apply the [c++ coding guidelines](https://everest.github.io/nightly/how-to-guides/c%2B%2B-coding-guidelines.html) for naming of member variables.

modules/Simulation/TemperatureSensorSimulator/temperature_sensor/temperature_sensorImpl.cpp
```diff
@@ -0,0 +1,136 @@
...
+}
+
+} // namespace
+
+types::temperature::Temperature temperature_sensorImpl::current_reading() const {
+    return {
+        .temperature = temperature_C,
+        .identification = identification,
+        .location = location,
+    };
```
Pietfried: designated initializers is a c++20 feature

modules/Simulation/YetiSimulator/YetiSimulator.cpp
```diff
@@ -10,15 +10,21 @@ namespace module {
 
 namespace {
 
+std::vector<types::temperature::Temperature>
+temperatures_from_powermeter(const state::PowermeterData& powermeter_data) {
+    return {{.temperature = static_cast<float>(powermeter_data.tempL1),
+             .identification = "Powermeter",
+             .location = "Powermeter"}};
+}
```
Pietfried, line 20: See https://github.com/EVerest/EVerest/pull/2325/changes#r3623206373

modules/EnergyManagement/AcTemperatureDerating/AcTemperatureDerating.cpp
```diff
@@ -0,0 +1,224 @@
...
+        ac_temperature_derating::validate_curves_for_providers(derating_curves_, provider_module_ids);
+        ac_temperature_derating::validate_ignore_list_vs_curves(derating_curves_, ignore_list_);
+    } catch (const std::exception& e) {
+        EVLOG_error << "AcTemperatureDerating configuration error: " << e.what();
+        throw;
+    }
+
+    provider_states_.reserve(r_temperature.size());
+    for (const auto& sensor : r_temperature) {
+        provider_states_.push_back(ProviderState{.module_id = sensor->module_id});
```
Pietfried: Also use of designated initializers

modules/EnergyManagement/AcTemperatureDerating/AcTemperatureDerating.cpp
```diff
@@ -0,0 +1,224 @@
...
+    {
+        std::scoped_lock lock(state_mutex_);
+        for (auto& provider : provider_states_) {
+            bool any_non_ignored_reading_added = false;
+
+            if (provider.has_reading_without_identification) {
+                const auto& state = provider.reading_without_identification;
+                if (state.ever_received && (now - state.last_update) <= stale_timeout) {
+                    readings.push_back(ac_temperature_derating::SensorReadingInput{
+                        .module_id = provider.module_id,
```
Pietfried: Also use of designated initializers

modules/Simulation/YetiSimulator/YetiSimulator.cpp
```diff
@@ -10,15 +10,21 @@ namespace module {
 
 namespace {
 
+std::vector<types::temperature::Temperature>
+temperatures_from_powermeter(const state::PowermeterData& powermeter_data) {
+    return {{.temperature = static_cast<float>(powermeter_data.tempL1),
```
Pietfried: Also use of designated initializers

modules/EnergyManagement/AcTemperatureDerating/derating/derating_curve.cpp
```diff
@@ -0,0 +1,217 @@
...
+            effective_limit = std::min(effective_limit, sensor_limit);
+            continue;
+        }
+
+        sensor_limit = interpolate_max_current_A(*curve, reading.temperature_C.value());
+        any_limit_applied = true;
+        effective_limit = std::min(effective_limit, sensor_limit);
+    }
+
+    if (!any_limit_applied || !std::isfinite(effective_limit)) {
```
Pietfried, line 248: I'd recommend to make the check for `isfinite` earlier in the function for incoming temperatures

Pietfried (changes requested):
lgtm, please see a few comments in line that need to be addressed

modules/EnergyManagement/AcTemperatureDerating/derating/derating_curve.hpp
florinmihut: Addressed in the last commit!

modules/EnergyManagement/AcTemperatureDerating/derating/derating_curve.hpp
florinmihut: Addressed in the last commit!

modules/Simulation/TemperatureSensorSimulator/temperature_sensor/temperature_sensorImpl.cpp
florinmihut: Addressed in the last commit!

modules/Simulation/YetiSimulator/YetiSimulator.cpp
florinmihut: Addressed in the last commit!

modules/EnergyManagement/AcTemperatureDerating/AcTemperatureDerating.cpp
florinmihut: Addressed in the last commit!

modules/EnergyManagement/AcTemperatureDerating/AcTemperatureDerating.cpp
florinmihut: Addressed in the last commit!

modules/Simulation/YetiSimulator/YetiSimulator.cpp
florinmihut: Addressed in the last commit!

modules/EnergyManagement/AcTemperatureDerating/derating/derating_curve.cpp
florinmihut: Addressed in the last commit!

modules/EnergyManagement/AcTemperatureDerating/derating/derating_curve.hpp
```diff
@@ -0,0 +1,69 @@
...
+    std::string module_id;
+    std::optional<std::string> identification;
+    std::optional<double> temperature_C;
+};
+
+struct ComputeLimitResult {
+    std::optional<double> effective_limit_A;
+    std::vector<std::string> missing_curve_keys;
+    std::vector<std::string> missing_identification_curve_keys;
+};
```
Pietfried, line 65: Please apply the naming convention for all structs that are introduced in this PR

Pietfried (changes requested):
LGTM except the one comment in line for consistency.

modules/EnergyManagement/AcTemperatureDerating/derating/derating_curve.hpp
```diff
@@ -0,0 +1,69 @@
...
+
+namespace ac_temperature_derating {
+
+struct DeratingPoint {
+    double m_temp_c{0.0};
+    double m_max_current_a{0.0};
+};
+
+using DeratingCurve = std::vector<DeratingPoint>;
+using DeratingCurveMap = std::map<std::string, DeratingCurve>;
```
dorezyuk: unordered_map and unordered_set maybe?

modules/EnergyManagement/AcTemperatureDerating/manifest.yaml
```diff
@@ -0,0 +1,70 @@
+description: >-
+  Applies AC current limits to an EnergyNode based on temperature sensor readings
+  and configurable derating curves using linear interpolation between points.
+config:
+  derating_curves_json:
+    description: >-
+      JSON object: each key is module_id.identification, each value is an array of
+      derating points {"temp_C": number, "max_current_A": number}. Points must be
+      sorted by ascending temp_C; the module linearly interpolates between them and
```
dorezyuk: why do points have to be sorted by temp_C - we're sorting them inside derating_curve.cpp

modules/EnergyManagement/AcTemperatureDerating/derating/derating_curve.cpp
```diff
@@ -0,0 +1,222 @@
...
+
+namespace ac_temperature_derating {
+
+namespace {
+
+DeratingCurve parse_curve_points(const nlohmann::json& points_json) {
+    if (!points_json.is_array()) {
+        throw std::invalid_argument("derating curve must be a JSON array");
+    }
+    if (points_json.empty()) {
```
dorezyuk, line 22: how does the system work if there is only one point? might be also worth warning or throwing?

modules/EnergyManagement/AcTemperatureDerating/derating/derating_curve.cpp
```diff
@@ -0,0 +1,222 @@
...
+        DeratingPoint point;
+        point.m_temp_c = point_json.at("temp_C").get<double>();
+        point.m_max_current_a = point_json.at("max_current_A").get<double>();
+        curve.push_back(point);
+    }
+
+    std::sort(curve.begin(), curve.end(),
+              [](const DeratingPoint& lhs, const DeratingPoint& rhs) { return lhs.m_temp_c < rhs.m_temp_c; });
+
+    for (std::size_t i = 1; i < curve.size(); ++i) {
```
dorezyuk: https://en.cppreference.com/cpp/algorithm/adjacent_find might be also an option

modules/EnergyManagement/AcTemperatureDerating/derating/derating_curve.cpp
```diff
@@ -0,0 +1,222 @@
...
+    }
+    return curves;
+}
+
+void validate_curves_for_providers(const DeratingCurveMap& curves, const std::vector<std::string>& module_ids) {
+    if (curves.empty()) {
+        throw std::invalid_argument("derating_curves_json must define at least one curve");
+    }
+
+    for (const auto& module_id : module_ids) {
```
dorezyuk, line 112: is it ok if module-id's is empty?

modules/EnergyManagement/AcTemperatureDerating/derating/derating_curve.cpp
```diff
@@ -0,0 +1,222 @@
...
+    }
+    return nullptr;
+}
+
+ComputeLimitResult compute_effective_limit_A(const DeratingCurveMap& curves,
+                                             const std::vector<SensorReadingInput>& readings,
+                                             double fallback_max_current_A) {
+    ComputeLimitResult result;
+    if (readings.empty()) {
+        return result;
```
dorezyuk, line 202: not fallback?

modules/EnergyManagement/AcTemperatureDerating/AcTemperatureDerating.hpp
```diff
@@ -0,0 +1,137 @@
...
+    // ev@211cfdbe-f69a-4cd6-a4ec-f8aaa3d1b6c8:v1
+    struct ReadingState {
+        std::optional<double> temperature_C;
+        std::chrono::steady_clock::time_point last_update{};
+        bool ever_received{false};
+    };
+
+    struct ProviderState {
+        std::string module_id;
+        bool has_reading_without_identification{false};
```
dorezyuk: this would be a great use for `std::optional<ReadingState>` so you don't need two fields here

modules/EnergyManagement/AcTemperatureDerating/AcTemperatureDerating.hpp
```diff
@@ -0,0 +1,137 @@
...
+private:
+    friend class LdEverest;
+    void init();
+    void ready();
+
+    // ev@211cfdbe-f69a-4cd6-a4ec-f8aaa3d1b6c8:v1
+    struct ReadingState {
+        std::optional<double> temperature_C;
+        std::chrono::steady_clock::time_point last_update{};
+        bool ever_received{false};
```
dorezyuk: not sure if you need this - the default constructor fo time_point inits it to zero 
https://en.cppreference.com/cpp/chrono/time_point/time_point
so every check (now - last_update) <= stale will return false

modules/EnergyManagement/AcTemperatureDerating/docs/index.rst
```diff
@@ -0,0 +1,73 @@
...
+---------------------
+
+Place a dedicated ``EnergyNode`` for temperature derating between the grid connection and downstream nodes.
+The EnergyManager merges limits from all nodes; OCPP/API external limits on other nodes continue to work independently.
+
+Configuration
+-------------
+
+``derating_curves_json`` maps ``module_id.identification`` to an array of ``{"temp_C", "max_current_A"}`` points.
+
```
dorezyuk, line 26: just as an idea but it might be easier to maintain and understand if we would have the json format already express the structure of module id and identification instead of cramping both into the key 
```json
{
"module_id": ["identification_1": [{"temp_c...},..], "identification_2": [...]]
}
````

modules/EnergyManagement/AcTemperatureDerating/derating/derating_curve.hpp
florinmihut: Since the lists are small, there is no gain in speed. 
Moreover ordering is a "feature".

modules/EnergyManagement/AcTemperatureDerating/manifest.yaml
florinmihut: You are right. No need to be sorted. Maybe would be more intuitive but otherwise not necessary. I am updating the doc.

modules/EnergyManagement/AcTemperatureDerating/derating/derating_curve.cpp
florinmihut: Theoretically you could have a curve with one point. Sounds strange but it works. Under that temp, you do not de-rate, at that temp and above, you derate as given by the point. 
I will make the necessary changes in the code to go with this intuitive model, thank you for your comment!!!

modules/EnergyManagement/AcTemperatureDerating/derating/derating_curve.cpp
florinmihut: Makes sense, thank you. Changed.

modules/EnergyManagement/AcTemperatureDerating/derating/derating_curve.cpp
florinmihut: Yes, it is checked in multiple cases via validate_curve_key_format and at this line the module_ids are coming from the framework and must be non empty.

modules/EnergyManagement/AcTemperatureDerating/derating/derating_curve.cpp
florinmihut: Very good question but no fallback here. The only case we can get here is we have one provider and it is on the ignore list. If that is the case, no fallback is the right decision because the definition of the fallback is:  limit applied when a temperature reading is missing, stale, or has no matching curve. 
The temp is ignored, not missing, not stale since it is ignored and it has not a missing curve.

modules/EnergyManagement/AcTemperatureDerating/AcTemperatureDerating.hpp
florinmihut: Valid point, thank you! Changed!

modules/EnergyManagement/AcTemperatureDerating/AcTemperatureDerating.hpp
florinmihut: Generally, true but not always. After a reboot that is not the case it seems on all kind of systems.
To be honest it is just a bool flag and the code is clearer with it. 
Valid point but I will keep it for now.

modules/EnergyManagement/AcTemperatureDerating/docs/index.rst
florinmihut: I think this is a VERY good idea, thank you!
It is probabl better than moduleid.identification  that might be difficult to understand. Now it is this module has 2 possible identification readings and here is the curve for each of them. 
The initial idea was to have the config as simple as possible but the JSON already contain an array ... so the path to an array or arrays it not that long. The parsing is cleaner now, less ambiguity, so all in all better. 

Only disadvantage: kind of changes stuff.

modules/EnergyManagement/AcTemperatureDerating/derating/derating_curve.hpp
florinmihut: Fixed!

Pietfried (approved):

PR #2326, opened 2026-06-17
BuildKit: Use corepack to provide a newer version of npm in the CI containers

nzbr:
## Describe your changes

Extends the buildkit dockerfile with a step that replaces the ubuntu-provided npm with one from corepack so that we can manually specify the version. If npm is used with package.json files that contain a package-manager field it should now always use the pinned version for those as well.
Fixes the CI failure when building the AsyncAPI documentation.

### DIFF
--- a/.ci/build-kit/docker/Dockerfile
+++ b/.ci/build-kit/docker/Dockerfile
@@ -14,6 +14,7 @@ FROM ghcr.io/everest/everest-ci/build-kit-base:${BASE_IMAGE_TAG}
 # FIXME(kai): Install libsystemd-dev as a dependency of sdbus-c++
 RUN apt-get update && apt-get -y install libsystemd-dev
 
+# FIXME: move this snippet to everest-ci Dockerfiles
 # Make sure we use a recent enough npm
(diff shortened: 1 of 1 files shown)


.ci/build-kit/docker/Dockerfile
```diff
@@ -13,3 +13,9 @@ FROM ghcr.io/everest/everest-ci/build-kit-base:${BASE_IMAGE_TAG}
 
 # FIXME(kai): Install libsystemd-dev as a dependency of sdbus-c++
 RUN apt-get update && apt-get -y install libsystemd-dev
+
+# Make sure we use a recent enough npm
```
andistorm, line 18:
```suggestion
# FIXME: move this snippet to everest-ci Dockerfiles
# Make sure we use a recent enough npm
```

andistorm (changes requested):
LGTM in general, but I would like to have this comment added, because this should not be the final location, but makes sense for now to quickly have a solution

andistorm (approved):
🐧

andistorm (approved):
🐧

PR #2327, opened 2026-06-17
system, libocpp: Allow connectors to stay enabled during firmware updates

nzbr:
## Describe your changes

This is useful when the firmware update mechanism is used for trivial changes to the system that do not require a restart of EVerest / its modules / the whole computer.
This is accomplished by extending the system API with a new FirmwareUpdateMetadata type that can be attached to FirmwareUpdateStatus messages. Currently, this type can only carry a single value (`skip_making_connectors_unavailable`).

---
This is **part 1 of 3 in a stack** made with GitButler:
- <kbd>&nbsp;3&nbsp;</kbd> #2337 
- <kbd>&nbsp;2&nbsp;</kbd> #2336 
- <kbd>&nbsp;1&nbsp;</kbd> #2327 👈


lib/everest/ocpp/lib/ocpp/v16/charge_point_impl.cpp
Pietfried, line 1: A TriggerMessage(FirmwareStatusNotification) uses the default `skip_making_connectors_unavailable = false` and potentially overrides back to false. This should use the existing stored value for `skip_making_connectors_unavailable` https://github.com/EVerest/EVerest/blob/95d43b8eb2f454553764da148af9681d996484e2/lib/everest/ocpp/lib/ocpp/v16/charge_point_impl.cpp#L2677

types/system.yaml
```diff
@@ -189,6 +189,14 @@ types:
       - InstallVerificationFailed
       - InvalidSignature
       - SignatureVerified
+  FirmwareUpdateMetadata:
+    description: Type carrying metadata about a firmware update
+    type: object
+    properties:
+      skip_making_connectors_unavailable:
```
Pietfried: rather a nit but I would prefer a name like `disable_connectors` or `disable_connectors_during_install`

Pietfried (changes requested):

types/system.yaml
Pietfried: Changing it would of course have implications on https://github.com/EVerest/EVerest/pull/2336 and https://github.com/EVerest/EVerest/pull/2337 as well

nzbr: Yeah, the variable name isn't great. I wanted to have something where I could have it default to `false`, because I figured that would be more intuitive in the code, but after having done the implementation, I don't think it matters too much, so I'll change it

lib/everest/ocpp/lib/ocpp/v16/charge_point_impl.cpp
nzbr: I added a guard so that whether the connectors should be disabled is not overwritten if `initiated_by_trigger_message` is true

types/system.yaml
nzbr: Done!

lib/everest/ocpp/lib/ocpp/v16/charge_point_impl.cpp
Pietfried: Looks like you only added it for the non signed part, but the signed part needs the same handling.

Alternatively you could avoid the guard by invoking:

```cpp
this->firmware_status_notification(this->firmware_status, true, this->disable_connectors_during_install);
``` 
and
```cpp
this->signed_firmware_update_status_notification(this->signed_firmware_status,
                                                         this->signed_firmware_status_request_id, true, this->disable_connectors_during_install);
```

lib/everest/ocpp/include/ocpp/v16/charge_point.hpp
```diff
@@ -348,8 +348,11 @@ class ChargePoint {
     /// \param request_id A \p request_id of -1 indicates a FirmwareStatusNotification.req, else a
     /// SignedFirmwareUpdateStatusNotification.req .
     /// \param firmware_update_status The \p firmware_update_status
+    /// \param disable_connectors_during_install By default, all connectors will be disabled before installing the
+    /// firmware update. Settting this parameter to false will keep the connectors available during the update.
```
Pietfried:
```suggestion
    /// firmware update. Setting this parameter to false will keep the connectors available during the update.
```

lib/everest/ocpp/include/ocpp/v16/charge_point_impl.hpp
```diff
@@ -675,8 +678,11 @@ class ChargePointImpl : ocpp::ChargingStationBase {
     /// \param request_id A \p request_id of -1 indicates a FirmwareStatusNotification.req, else a
     /// SignedFirmwareUpdateStatusNotification.req .
     /// \param firmware_update_status The \p firmware_update_status
+    /// \param disable_connectors_during_install By default, all connectors will be disabled before installing the
+    /// firmware update. Settting this parameter to false will keep the connectors available during the update.
```
Pietfried:
```suggestion
    /// firmware update. Setting this parameter to false will keep the connectors available during the update.
```

lib/everest/ocpp/include/ocpp/v2/charge_point.hpp
```diff
@@ -131,8 +131,11 @@ class ChargePointInterface {
     ///        called during a Firmware Update to indicate the current firmware_update_status.
     /// \param request_id   The request_id. When it is -1, it will not be included in the request.
     /// \param firmware_update_status The firmware_update_status
+    /// \param disable_connectors_during_install By default, all connectors will be disabled before installing the
+    /// firmware update. Settting this parameter to false will keep the connectors available during the update.
```
Pietfried:
```suggestion
    /// firmware update. Setting this parameter to false will keep the connectors available during the update.
```

Pietfried (changes requested):

lib/everest/ocpp/lib/ocpp/v16/charge_point_impl.cpp
nzbr: Ah, I missed that. Should be fixed now

Pietfried (approved):

PR #2328, opened 2026-06-17
feat(ocpp21): V2X operationMode profile validation

mlitre:
Reject operationMode absent from V2XChargingCtrlr. SupportedOperationModes (Q09.FR.01, K01.FR.115). Fix K01.FR.126: reject evseSleep only when operationMode is not Idle (Q10.FR.05: unsupported EvseSleep is accepted).

### DIFF
--- a/lib/everest/ocpp/lib/ocpp/v2/functional_blocks/smart_charging.cpp
+++ b/lib/everest/ocpp/lib/ocpp/v2/functional_blocks/smart_charging.cpp
@@ -197,6 +197,10 @@ std::string profile_validation_result_to_string(ProfileValidationResultEnum e) {
         return "ChargingSchedulePeriodPriorityChargingNotChargingOnly";
     case ProfileValidationResultEnum::ChargingSchedulePeriodUnsupportedOperationMode:
         return "ChargingSchedulePeriodUnsupportedOperationMode";
+    case ProfileValidationResultEnum::ChargingSchedulePeriodOperationModeNotInSupportedList:
+        return "ChargingSchedulePeriodOperationModeNotInSupportedList";
+    case ProfileValidationResultEnum::ChargingSchedulePeriodLocalLoadBalancingNotSupported:
+        return "ChargingSchedulePeriodLocalLoadBalancingNotSupported";
     case ProfileValidationResultEnum::ChargingSchedulePeriodUnsupportedLimitSetpoint:
         return "ChargingSchedulePeriodUnsupportedLimitSetpoint";
     case ProfileValidationResultEnum::ChargingSchedulePeriodNoPhaseForDC:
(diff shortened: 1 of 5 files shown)


lib/everest/ocpp/lib/ocpp/v2/functional_blocks/smart_charging.cpp
```diff
@@ -927,6 +935,20 @@ ProfileValidationResultEnum SmartCharging::validate_priority_charging_profile(co
...
 
+bool SmartCharging::is_operation_mode_supported_by_evse(const OperationModeEnum operation_mode,
+                                                        const std::int32_t evse_id) const {
+    if (operation_mode == OperationModeEnum::ChargingOnly) {
+        return true;
+    }
+    const auto supported_operation_modes = this->context.device_model.get_optional_value<std::string>(
+        V2xComponentVariables::get_component_variable(evse_id, V2xComponentVariables::SupportedOperationModes));
+    if (!supported_operation_modes.has_value()) {
+        return false;
```
Pietfried, line 946: The doc comment says returns "true when supported, or when the EVSE does not expose the variable (no V2X constraint to apply)", so we probably want to return true here. Afaics this variable is always set in the EVerest DM though.

lib/everest/ocpp/tests/lib/ocpp/v2/device_model_test_helper.cpp
```diff
@@ -212,7 +212,29 @@ bool DeviceModelTestHelper::set_variable_attribute_value_null(const std::string&
...
+        "LocalLoadBalancing,Idle";
+    supported_operation_modes_attribute.variable_attribute.type = AttributeEnum::Actual;
+    supported_operation_modes.attributes = std::vector<DbVariableAttribute>{supported_operation_modes_attribute};
+    const std::vector<DeviceModelVariable> v2x_variables{supported_operation_modes};
+    for (const std::int32_t evse_id : {1, 2}) {
+        ComponentKey v2x_ctrl;
+        v2x_ctrl.name = "V2XChargingCtrlr";
+        v2x_ctrl.evse_id = evse_id;
+        component_configs.insert(std::make_pair(v2x_ctrl, v2x_variables));
+    }
```
Pietfried, line 236: nit: This could go into a separate function like `build_example_v2x_charging_ctrlr` because The V2XChargingCtrlr likely needs extensions like this in the future

Pietfried (changes requested):
See just some minor comments in line

lib/everest/ocpp/tests/lib/ocpp/v2/device_model_test_helper.cpp
mlitre: I have an EVerest DM model builder / improvement branch... I can fold that in here, but that adds more to review

Pietfried: Ok, just keep it for now then

lib/everest/ocpp/lib/ocpp/v2/functional_blocks/smart_charging.cpp
mlitre: That's a doc mistake it's easier to read and handle if we return false when it is missing. As you said though we always set it in the EVerest DM, this is just a precaution.

Pietfried (approved):

PR #2332, opened 2026-06-18
[OCPP] Create a new EVerest module that supports all OCPP versions

james-ctc:
## Describe your changes

Adds the new **OCPPmulti** module (`modules/EVSE/OCPPmulti`): a single OCPP module supporting both OCPP 1.6 and OCPP 2.0.1/2.1, selected at runtime via the `Mode` config option. It is intended to deprecate and eventually replace the separate `OCPP` and `OCPP201` modules, which remain unchanged in this PR. The module is marked **experimental** (startup warning): configuration parameters and EVerest integration may still change.

Design:

* A version-agnostic core (`generic_ocpp.cpp`) implements all EVerest interface handling once and talks to a `GenericChargePointInterface`; thin version wrappers (`v16_chargepoint.cpp`, `v2_chargepoint.cpp`) adapt it to the libocpp 1.6 and 2.x charge points, using OCPP 2.x types as the common vocabulary.
* Shared logic reused from `lib/everest/ocpp_module_common` (transaction handler, conversions, error mapping, device model storage).
* OCPP 1.6 mode runs on the 2.x device model; an optional one-time migration (`EnableLegacyConfigMigration`) seeds it from an existing legacy 1.6 JSON config.
* The legacy `ocpp_1_6_charge_point` (`main`) interface is not provided anymore; consumers use the shared `ocpp_generic` interface.

Testing:

* Unit tests for the module (~80 tests) covering the provided/required interfaces, the v16 helpers, and the v2 transaction lifecycle (`modules/EVSE/OCPPmulti/tests`).
* The existing OCPP 1.6/2.0.1/2.1 integration suites are parametrized to run against both the legacy modules and OCPPmulti (`--ocpp-impl both|legacy|multi` in `tests/run-tests.sh`).
* New SIL config `config/config-sil-ocpp-multi.yaml`.

Planned follow up PR that documents all changes: https://github.com/EVerest/EVerest/pull/2426 . Note that this PR already documents a planned change that the `mode` parameter from `OCPPmulti` will be completely replaced by the `InternalCtrlr.SupportedProtocolVersions` variable. This change will be added in a subsequent PR.


modules/EVSE/OCPPmulti/generic_ocpp.cpp
```diff
@@ -0,0 +1,2243 @@
...
+    m_everest_device_model_storage = std::make_shared<module::device_model::EverestDeviceModelStorage>(
+        m_requires.evse_manager, m_requires.extensions_15118, m_evse_hardware_capabilities_map,
+        m_evse_supported_energy_transfer_modes, m_evse_service_renegotiation_supported,
+        everest_device_model_database_path, device_model_database_migration_path, client);
+
+    // if charger information interface is connected, override only these specific
+    // properties which were loaded from configuration file(s)
+    if (!m_requires.charger_information.empty()) {
+        auto info = m_requires.charger_information.at(0)->call_get_charger_information();
+        m_charge_point.update_chargepoint_information(info.vendor, info.model, info.chargepoint_serial,
```
Pietfried: This should be moved to after `m_charge_point.init(args);` otherwise `m_charge_point` is nullptr

modules/EVSE/OCPPmulti/generic_chargepoint.cpp
```diff
@@ -0,0 +1,279 @@
...
+    m_charge_point_v2 = std::make_unique<ChargePointV2>(m_charge_point_callbacks, m_evse_security_interface);
+
+    switch (m_mode) {
+    case modes_t::ocpp_1_6_only:
+    case modes_t::prefer_ocpp_1_6:
+        m_active_ptr = m_charge_point_v16.get();
+        break;
+    case modes_t::ocpp_2_only:
+    case modes_t::prefer_ocpp_2:
+        m_active_ptr = m_charge_point_v2.get();
```
Pietfried:
```suggestion
        m_active_ptr = m_charge_point_v2.get();
        break;
```

modules/EVSE/OCPPmulti/v16_chargepoint.cpp
```diff
@@ -0,0 +1,1316 @@
...
+        if (send) {
+            data.attributeValue = static_cast<std::string>(key_value.value.value());
+            m_callbacks_ptr->cb_variable_set(data);
+        }
+    }
+}
+
+void ChargePointV16::cb_get_15118_ev_certificate_response(
+    const std::int32_t connector, const ocpp::v2::Get15118EVCertificateResponse& certificate_response,
+    const ocpp::v2::CertificateActionEnum& certificate_action) {
```
Pietfried: not implemented

modules/EVSE/OCPPmulti/common/conversions_v16.hpp
```diff
@@ -0,0 +1,119 @@
...
+/// \brief Converts a given types::system::UpdateFirmwareResponse \p response to a
+/// ocpp::v16::UpdateFirmwareStatusEnumType.
+ocpp::v16::UpdateFirmwareStatusEnumType
+to_ocpp_update_firmware_status_enum_type(const types::system::UpdateFirmwareResponse response);
+
+/// \brief Converts a given types::iso15118::HashAlgorithm \p hash_algorithm to a
+/// ocpp::v16::HashAlgorithmEnumType.
+ocpp::v16::HashAlgorithmEnumType to_ocpp_hash_algorithm_enum_type(const types::iso15118::HashAlgorithm hash_algorithm);
+
+/// \brief  Converts a given types::iso15118::Status \p status to a ocpp::v2::Iso15118EVCertificateStatusEnum.
```
Pietfried:
```suggestion
/// \brief  Converts a given types::iso15118::Status \p status to a ocpp::v16__BootReasonEnum.
```

Pietfried (commented):
The general approach looks good to me!

Some observations in addition to the in line comments:
- Using v2 for the generic interface is a good and pragmatic choice because it is the richer superset and seems to map well to v16. Inventing a third generic type dialect would be more code for little gain
- - While OCPP2.x is mostly reusing existing code, for OCPP1.6 quite some conversions as well as the state machine (process_session_event) changed. This becomes the critical path to verify
- The symlinks are currently pointing the wrong way, ownership of linked files should move into OCPPmulti eventually
- We can freeze and mark the other OCPP modules as deprecated as soon as the new module is merged
- The `Mode` parameter does currently (probably intentionally) only support either 1.6 or 2.x.
- https://github.com/EVerest/EVerest/pull/2276 allows to provide the `ConnectivityManager` externally to the `ChargePoint` constructors. This allows us to actually initiate the websocket handshake to identify the protocol version and then we construct a charge point. @james-ctc do you plan to integrate this rather in a subsequent PR ?

modules/EVSE/OCPPmulti/common/conversions_v16.hpp
james-ctc: other than the namespace this should be an exact copy from the OCPP module.

modules/EVSE/OCPPmulti/generic_chargepoint.cpp
james-ctc: it is the default: that is in the wrong place. The intent is to default to OCPP 2.x

modules/EVSE/OCPPmulti/OCPPmulti.hpp
```diff
@@ -0,0 +1,191 @@
...
+
+    ConfigAccess m_config{config};
+    ocpp_multi::GenericChargePoint m_charge_point{m_ocpp, *r_security};
+    ocpp_multi::GenericOcpp m_ocpp{
+        m_charge_point,
+        info,
+        m_config,
+        {*p_auth_validator, *p_auth_provider, *p_data_transfer, *p_ocpp_generic, *p_session_cost},
+        {*r_auth, r_charger_information, r_data_transfer, r_display_message, r_evse_energy_sink, r_evse_manager,
+         r_extensions_15118, r_reservation, *r_security, *r_system}};
```
mlitre, line 167: Can we make this cleaner where instead of passing references we actually move the unique ptrs so that we have more explicitly shown ownership of the interfaces? Especially if we are only using them in generic_ocpp.
Also I believe defining m_ocpp after m_charge_point is UB

modules/EVSE/OCPPmulti/OCPPmulti.cpp
```diff
@@ -0,0 +1,126 @@
...
+void OCPPmulti::init() {
+    EVLOG_warning << "This OCPPmulti module is currently experimental! Configuration parameters and the integration in "
+                     "EVerest may change without further notice";
+    // no code should be in the init methods
+    // invoke_init(*p_ocpp16);
+    // invoke_init(*p_auth_validator);
+    // invoke_init(*p_auth_provider);
+    // invoke_init(*p_data_transfer);
+    // invoke_init(*p_ocpp_generic);
+    // invoke_init(*p_session_cost);
```
mlitre, line 86: This is quite a departure from the standard way of doing modules, it's not wrong, but it is a little strange

modules/EVSE/OCPPmulti/generic_ocpp.hpp
mlitre, line 1: In general this seems to have just moved the "giant god class" from a module's generated template to a new class. I am wondering if we actually need to have all the interfaces reincluded here in this class and potentially have separated things where this generic_ocpp class could just take a pointer to the module to call on these interfaces

modules/EVSE/OCPPmulti/generic_ocpp.hpp
```diff
@@ -0,0 +1,434 @@
...
+    // protected by m_member_mux - used in device model storage and this object
+    std::map<std::int32_t, types::evse_board_support::HardwareCapabilities> m_evse_hardware_capabilities_map;
+    std::map<std::int32_t, std::vector<types::iso15118::EnergyTransferMode>> m_evse_supported_energy_transfer_modes;
+    std::map<std::int32_t, bool> m_evse_service_renegotiation_supported;
+
+    Everest::SteadyTimer m_charging_schedules_timer;
+
+    std::mutex m_chargepoint_state_mutex; // mutex used for start/stop operations
+    std::mutex m_session_event_mutex;
+    std::mutex m_member_mux;
```
mlitre, line 182: Any particular reason why we use mutexes instead of monitor here since the monitor is the pattern that is used everywhere else in this file?

modules/EVSE/OCPPmulti/generic_chargepoint_interface.hpp
```diff
@@ -0,0 +1,339 @@
...
+#include <everest/ocpp_module_common/device_model/everest_device_model_storage.hpp>
+#include <everest/ocpp_module_common/error_handling.hpp>
+#include <everest/ocpp_module_common/transaction_handler.hpp>
+
+#include <generated/interfaces/charger_information/Interface.hpp>
+#include <generated/interfaces/evse_manager/Interface.hpp>
+#include <generated/interfaces/iso15118_extensions/Interface.hpp>
+#include <generated/interfaces/reservation/Interface.hpp>
+#include <generated/interfaces/session_cost/Interface.hpp>
+#include <generated/interfaces/system/Interface.hpp>
```
mlitre: Shouldn't we include types rather than interfaces here?

modules/EVSE/OCPPmulti/generic_chargepoint.hpp
```diff
@@ -0,0 +1,165 @@
...
+class GenericChargePoint : public GenericChargePointInterface {
+public:
+    enum class state_t : std::uint8_t {
+        v2_active,
+        v16_active,
+        idle,
+    };
+
+private:
+    GenericChargePointCallbacks& m_charge_point_callbacks;
```
mlitre: Is there a particular reason why the GenericChargePoint shouldn't hold the callbacks instead of just having a reference?

modules/EVSE/OCPPmulti/generic_ocpp.cpp
```diff
@@ -0,0 +1,1979 @@
...
+// Copyright Pionix GmbH and Contributors to EVerest
+
+#include "generic_ocpp.hpp"
+#include "everest/logging.hpp"
+#include "ocpp/common/types.hpp"
+
+#include <everest/conversions/ocpp/ocpp_conversions.hpp>
+#include <everest/external_energy_limits/external_energy_limits.hpp>
+#include <everest/ocpp_module_common/conversions.hpp>
+#include <ld-ev.hpp>
```
mlitre, line 12: This looks a bit strange to me

modules/EVSE/OCPPmulti/generic_ocpp.hpp
```diff
@@ -0,0 +1,434 @@
...
+    requires_t mv_requires;
+    module::MREC_ERROR_MAP_TYPE mv_mrec_error_map;
+
+    std::atomic<std::int32_t> mv_event_id_counter{0};
+    std::atomic<ocpp::OcppProtocolVersion> mv_ocpp_protocol_version{ocpp::OcppProtocolVersion::Unknown};
+    std::atomic_bool mv_started{false};
+
+    everest::lib::util::monitor<std::map<std::int32_t, std::string>> mv_evse_evcc_id;
+    everest::lib::util::monitor<std::map<std::int32_t, bool>> mv_evse_ready_map;
+    everest::lib::util::monitor<std::map<std::int32_t, std::optional<float>>> mv_evse_soc_map;
```
mlitre, line 165: Let's stay consistent and use the "m_" prefix everywhere instead of "mv_"

modules/EVSE/OCPPmulti/generic_ocpp.cpp
```diff
@@ -0,0 +1,1979 @@
...
+        break;
+    case ocpp::v2::CertificateActionEnum::Update:
+    default:
+        result = types::iso15118::CertificateActionEnum::Update;
+        break;
+    }
+    return result;
+}
+
+auto convert(ocpp::v2::MessageFormatEnum value) {
```
mlitre: Might be worth moving to the conversions lib?

modules/EVSE/OCPPmulti/generic_ocpp.cpp
```diff
@@ -0,0 +1,1979 @@
...
+        break;
+    case ocpp::v2::MessageFormatEnum::UTF8:
+    default:
+        result = types::text_message::MessageFormat::UTF8;
+        break;
+    }
+    return result;
+}
+
+auto convert(const std::optional<ocpp::v2::Tariff>& value) {
```
mlitre: Same here

modules/EVSE/OCPPmulti/generic_ocpp.cpp
```diff
@@ -0,0 +1,1979 @@
...
+            tx_start_stop_points.insert(TxStartStopPoint::PowerPathClosed);
+        } else if (tx_start_stop_point == "EnergyTransfer") {
+            tx_start_stop_points.insert(TxStartStopPoint::EnergyTransfer);
+        } else if (tx_start_stop_point == "DataSigned") {
+            tx_start_stop_points.insert(TxStartStopPoint::DataSigned);
+        } else {
+            // default to PowerPathClosed for now
+            tx_start_stop_points.insert(TxStartStopPoint::PowerPathClosed);
+        }
+    }
```
mlitre, line 85: In ocpp utils we have a split string method we can reuse to do a lot of the handing here

modules/EVSE/OCPPmulti/generic_ocpp.cpp
```diff
@@ -0,0 +1,1979 @@
...
+namespace ocpp_multi {
+
+// ----------------------------------------------------------------------------
+// GenericOcpp
+
+types::authorization::ValidationResult
+GenericOcpp::handle_validate_token(const types::authorization::ProvidedIdToken& provided_token) {
+    types::authorization::ValidationResult validation_result;
+
+    if (mv_started) {
```
mlitre: Applicable to whole PR: prefer using load, store and exchange for accessing std atomic bool

modules/EVSE/OCPPmulti/v16_conversions.cpp
```diff
@@ -0,0 +1,341 @@
...
+        return ocpp::v2::DataTransferStatusEnum::Rejected;
+    }
+}
+
+ocpp::v2::EnhancedChargingSchedulePeriod convert(const ocpp::v16::EnhancedChargingSchedulePeriod& value) {
+    ocpp::v2::EnhancedChargingSchedulePeriod result;
+    result.startPeriod = value.startPeriod;
+    result.limit = value.limit;
+    result.stackLevel = value.stackLevel;
+    // TODO(james-ctc): periodTransformed need handling
```
mlitre, line 148: We dropped numberPhases accidentally I think here

modules/EVSE/OCPPmulti/v16_chargepoint.cpp
```diff
@@ -0,0 +1,1247 @@
...
+    EVLOG_info << "Transaction stopped at OCPP1.6 connector: " << ocpp_connector_id << ", session_id: " << session_id;
+    const auto evse = evse_from_ocpp_connector(ocpp_connector_id);
+    if (!evse) {
+        return;
+    }
+    ocpp::v2::TransactionEventRequest event;
+    event.eventType = ocpp::v2::TransactionEventEnum::Ended;
+    event.evse = {evse->evse_id, evse->connector_id};
+    event.transactionInfo.transactionId = session_id;
+    event.seqNo = transaction_id; // TODO(james-ctc): this mapping is incorrect
```
mlitre: I think this probably needs fixing before merging

modules/EVSE/OCPPmulti/v16_chargepoint.cpp
```diff
@@ -0,0 +1,1247 @@
...
+    ocpp::Measurement measurement;
+    measurement.power_meter = conversions_v16::to_ocpp_power_meter(power_meter);
+    if (soc) {
+        // soc is present, so add this to the measurement
+        measurement.soc_Percent = ocpp::StateOfCharge{soc.value()};
+    }
+    if (power_meter.temperatures.has_value()) {
+        measurement.temperature_C = conversions_v16::to_ocpp_temperatures(power_meter.temperatures.value());
+    }
+    m_charge_point->on_meter_values(evse_id, measurement);
```
mlitre, line 996: Shouldn't we use m_connector_mapping to get the correct connectorId?

modules/EVSE/OCPPmulti/v16_chargepoint.cpp
```diff
@@ -0,0 +1,1247 @@
...
+        }
+    }
+
+    config_info_t config{
+        args.v16_chargepoint_config_path,
+        args.message_log_path,
+        ocpp_share_path,
+        sql_init_path,
+        args.v16_user_config_path,
+        static_cast<std::uint32_t>(args.evse_connector_structure.size()), // number of connectors;
```
mlitre, line 591: Won't this cause issues if number of connectors != number evseId?

modules/EVSE/OCPPmulti/generic_chargepoint.hpp
```diff
@@ -0,0 +1,165 @@
...
+    evse_securityIntf& m_evse_security_interface;
+
+    std::atomic<modes_t> m_mode{modes_t::prefer_ocpp_2};
+    state_t m_state{state_t::idle};
+    std::unique_ptr<ChargePointV16> m_charge_point_v16;
+    std::unique_ptr<ChargePointV2> m_charge_point_v2;
+    GenericChargePointInterface* m_active_ptr{nullptr};
```
mlitre, line 31: Should probably be atomic as you have a potential race between callbacks and shutdown

modules/EVSE/OCPPmulti/generic_ocpp.cpp
```diff
@@ -0,0 +1,1979 @@
...
+
+    if (id_token.evse_id.has_value()) {
+        provided_token.connectors = std::vector<std::int32_t>{id_token.evse_id.value()};
+    } else if (!id_token.connectors.empty()) {
+        provided_token.connectors = id_token.connectors;
+    }
+    mv_provides.auth_provider.publish_provided_token(provided_token);
+}
+
+void GenericOcpp::cb_reservation_update(types::reservation::ReservationUpdateStatus status) {
```
mlitre, line 1161: This method needs to enqueue since we can potentially get an update on startup

modules/EVSE/OCPPmulti/generic_ocpp.cpp
```diff
@@ -0,0 +1,1979 @@
...
+        }
+    }
+}
+
+void GenericOcpp::cb_error_handler(const Everest::error::Error& error) {
+    using namespace module;
+
+    // handled by specific evse_manager error handler
+    if (error.type != EVSE_MANAGER_INOPERATIVE_ERROR) {
+        auto event_data = convert_error(error);
```
mlitre, line 887: Since we subscribe early in the init this convert might fail before the mrec error map is initialized

modules/EVSE/OCPPmulti/manifest.yaml
```diff
@@ -0,0 +1,240 @@
...
+    type: boolean
+    default: false
+  EverestDeviceModelDatabasePath:
+    description: >-
+      OCPP 2.0.1
+      Path to the SQLite databse for the EVerest device model. This database stores components and variables
+      like EVSE and Connector that are closely related to EVerest and therefore not owned and managed by libocpp.
+    type: string
+    default: everest_device_model_storage.db
+  EnableExternalWebsocketControl:
```
mlitre: Duplicate configuration

modules/EVSE/OCPPmulti/charge_point_config_factory_v16.cpp
```diff
@@ -0,0 +1,273 @@
...
+                  << " in component configs; NumberOfConnectors may not match the number of connected EVSEs.";
+}
+
+void initialize_device_model_direct(const DeviceModelInitializationContext& context, int32_t n_evse) {
+    EVLOG_info << "Updating device model database from component configs.";
+    auto component_configs = ocpp::v2::get_all_component_configs(context.component_config_path);
+    ocpp::v2::ensure_ocpp16_legacy_ctrlr(component_configs);
+    patch_number_of_connectors(component_configs, n_evse);
+    ocpp::v2::InitDeviceModelDb init_device_model_db(context.database_path, context.migration_files_path);
+    init_device_model_db.initialize_database(component_configs, false);
```
mlitre, line 140: Not sure how to solve this but theoretically you could have the program crash between the schame initialization and the initial insertion of components into the DB. This would mean that on the following boot your DB would be initialized but without any values. This might be a thing to fix in a follow up

modules/EVSE/OCPPmulti/generic_ocpp.hpp
Pietfried: I agree that the class is very broad, but it's a deliberate 1:1 carry over of the existing OCPP/201 modules  to maintain the existing state as close as possible and make the review not overly complicated. Extracting into the `GenericOcpp` class makes the EVerest integration unit testable, which is not possible with `OCPPmulti`. We can think about splitting GenericOcpp up more into individual parts, but it's out of scope for this PR.

modules/EVSE/OCPPmulti/generic_ocpp.hpp
Pietfried: The reason is that those were also used in the existing OCPP201 and OCPP modules. I agree  that the monitor utilities should be prefered and will change it.

modules/EVSE/OCPPmulti/generic_chargepoint_interface.hpp
Pietfried: yep, will change it

modules/EVSE/OCPPmulti/generic_chargepoint.hpp
Pietfried: Changed to a pointer to address https://github.com/EVerest/EVerest/pull/2332#discussion_r3545828805

modules/EVSE/OCPPmulti/generic_ocpp.cpp
Pietfried: Removed the include

modules/EVSE/OCPPmulti/v16_chargepoint.cpp
Pietfried: A seqNo does not exist in OCPP1.6 adn there is no seqNo in the EVerest OcppTransactionEvent type anyways. I will change the comment.

modules/EVSE/OCPPmulti/v16_conversions.cpp
Pietfried: Added it

modules/EVSE/OCPPmulti/manifest.yaml
Pietfried: Removed duplicate

modules/EVSE/OCPPmulti/generic_ocpp.cpp
Pietfried: Applied

modules/EVSE/OCPPmulti/generic_ocpp.cpp
Pietfried: Conversions actually already existed, used the existing once and removed from the module code

modules/EVSE/OCPPmulti/generic_ocpp.hpp
Pietfried: Applied

modules/EVSE/OCPPmulti/generic_ocpp.hpp
Pietfried: Turned out to be more complex than initially expected, mainly because certain mutexes do not protect single data structures but larger blocks of contending logic. I'd prefer to keep this as is for now

modules/EVSE/OCPPmulti/generic_ocpp.cpp
Pietfried: It was also just copied from the OCPP201 module, changed it to use the split_string functions from ocpp utils.

modules/EVSE/OCPPmulti/generic_ocpp.hpp
james-ctc: The v was to indicate which variables are don't require the mutex for protection.
Please do not remove useful information that improves code maintenance

modules/EVSE/OCPPmulti/generic_ocpp.hpp
james-ctc: Aims:
- don't change libocpp
- have a single interface for 1.6, 2.0.1, 2.1 (generic chargepoint)
- extract business logic into a single task (generic ocpp)
- make it unit testable

I'm concerned that these changes are considered to make OCPPmulti less testable

modules/EVSE/OCPPmulti/generic_ocpp.hpp
james-ctc: The monitors are an overhead

modules/EVSE/OCPPmulti/OCPPmulti.hpp
james-ctc: The whole aim is to move away from pointers because of the need to check against null for every use.
For things that never change a reference is a better solution.
Please provide the reference for undefined behaviour. The address (i.e. reference) doesn't change.
There could be an issue if the reference is used during construction - but it isn't

modules/EVSE/OCPPmulti/OCPPmulti.cpp
james-ctc: The aim was to make the code unit testable. Modules are very hard to unit test.
Hence all functions/methods are in generic ocpp

mlitre: That's fine for me I was just making a note of it as it is I believe is the first time it is done

modules/EVSE/OCPPmulti/generic_ocpp.hpp
mlitre: Totally understand the purpose. I just believe as a follow up we could probably divide up some of the tasks of generic ocpp. That would still allow it to be unit testable. I was imagining something with composition so that tests of generic ocpp can remain pretty much the same it is just the logic is in a different  class/file

modules/EVSE/OCPPmulti/generic_ocpp.hpp
mlitre: I think it depends, the queued event map could be a monitor imo. I think in general I did see as well that these mutexes protected several data structures which I think shows there is a problem in how it is handled currently.

modules/EVSE/OCPPmulti/OCPPmulti.hpp
mlitre: I see your point, I think it makes sense to keep references then.
Regarding the UB I just thought it was strange because aren't objects constructed in order of definition so having m_charge_point constructed using m_ocpp before m_ocpp is defined would be considered UB?

modules/EVSE/OCPPmulti/generic_ocpp.hpp
mlitre: That is indeed useful to keep then. However, those that are wrapped in monitors or std atomic I think it is unnecessary to have them. Also maybe we should think of adding that to the coding guidelines as that was not necessarily clear here when reviewing.

Pietfried: I decided to keep the names `mv_` and `m_` as is for now, although I also find it is a little confusing that the ones wrapped in monitors use `mv_` although they are internally protected by mutexes.

modules/EVSE/OCPPmulti/generic_ocpp.cpp
Pietfried: adressed it

modules/EVSE/OCPPmulti/generic_ocpp.cpp
Pietfried: adressed it

modules/EVSE/OCPPmulti/v16_chargepoint.cpp
Pietfried: The terms are confusing, had to wrap my head around it... OCPP1.6 uses a 2-tier model (Charging Station, Connector) while OCPP 2.x and EVerest use a 3-tier model (Charging Station, EVSE, Connector). An OCPP1.6 Connector plays the role of a 2.x/EVerest EVSE.

Speaking in the 3-tier model: EVerest interfaces technically allow multiple Connectors per EVSE, but the EvseManager doesn't implement it - it always publishes SessionEvents for EVSE with id n and Connector with id 1. The EVerest OCPP1.6 stack and it's EVerest integration currently also does not support https://openchargealliance.org/ocpp-info-whitepapers/multiple-connectors-per-evse/ ,  which recommends to implement each connector as if it has its own EVSE.

It's easier to describe using an example:
- EVSE 1
  - Connector 1
- EVSE 2
  - Connector 1
  - Connector 2 

The whitepaper would require 3 OCPP 1.6 connectors with individual states; EVerest can't transport this at the moment and represents this as 2 (one per EVSE), because EvseManager collapses EVSE 2's connectors into one. Changing this is out of scope for this PR, which
only ports the existing OCPP/OCPP201 behavior. I'd open up an issue for the implementation of the whitepaper mentioned above.

modules/EVSE/OCPPmulti/v16_chargepoint.cpp
Pietfried: See https://github.com/EVerest/EVerest/pull/2332#discussion_r3553909043

modules/EVSE/OCPPmulti/generic_ocpp.hpp
james-ctc: however it means you don't need to apply an additional mutex to use them - which is the point about the name. It doesn't have to be a v it could be expanded m_no_mutex_required_<> 
However it is important to know what does and does not need protection adding

modules/EVSE/OCPPmulti/OCPPmulti.hpp
james-ctc: how is that any different to using a forward reference (or a pointer)?
A class/struct has a known size and the items within it have locations that don't change.

modules/EVSE/OCPPmulti/generic_ocpp.hpp
james-ctc: As soon as you use more than one mutex you can cause a deadlock. The order of acquiring matters.
A single mutex protecting member variables doesn't have that issue

modules/EVSE/OCPPmulti/generic_chargepoint.hpp
james-ctc: references can't be nullptr and hence are preferred over pointers where possible

modules/EVSE/OCPPmulti/generic_ocpp.hpp
mlitre: True, but you could just have a monitor wrapping a struct with all those variables. It is essentially the same thing. I just think the monitor makes it more explicit / visible / avoids mistakes that you need to be careful to properly lock / protect.

james-ctc: which doesn't work when pointers/references have been passed to external code.
there are many problems that need addressing.

modules/EVSE/OCPPmulti/OCPPmulti.hpp
mlitre: True, and you're right that getting the address of `m_ocpp` is fine on its own. The real issue is that `m_charge_point{m_ocpp, ...}` isn't just taking a pointer, it's converting GenericOcpp to its GenericChargePointCallbacks base, and that conversion adds the base's offset within the object. With the current layout that's a plain multiple inheritance offset (a compile-time constant), so the address comes out right and it works, even though it's formally UB (https://eel.is/c++draft/class.cdtor#3, https://en.cppreference.com/w/cpp/language/lifetime#Access_outside_of_lifetime). It only actually causes an issue if the bases get reordered or one is made virtual, since then the offset isn't a constant and the adjustment reads data that isn't set up yet. So I think it is preferable to fix it now and avoid issues down the line

mlitre (dismissed):
This looks good to me now!

modules/EVSE/OCPPmulti/generic_chargepoint_interface.hpp
```diff
@@ -0,0 +1,322 @@
...
+#include <ocpp/v2/types.hpp>
+
+#include <cstdint>
+#include <set>
+#include <stdexcept>
+#include <string>
+
+namespace module {
+
+// Shared OCPP module support code lives in lib/everest/ocpp_module_common;
```
james-ctc: None of these are used in this header file
The purpose of this header file is to define an abstract interface. These should be moved elsewhere as they do not form part of the interface

modules/EVSE/OCPPmulti/generic_chargepoint.hpp
```diff
@@ -0,0 +1,146 @@
...
+class GenericChargePoint : public GenericChargePointInterface {
+public:
+    enum class state_t : std::uint8_t {
+        v2_active,
+        v16_active,
+        idle,
+    };
+
+private:
+    GenericChargePointCallbacks* m_charge_point_callbacks{nullptr};
```
james-ctc, line 24: please move this back to being a reference.

modules/EVSE/OCPPmulti/generic_ocpp.hpp
```diff
@@ -0,0 +1,447 @@
...
+    // session ids of sessions that have been resumed by the EvseManager (e.g. after power loss);
+    // passed to the charge point on start() so matching transactions are not stopped (protected by m_member_mux)
+    std::set<std::string> m_resuming_session_ids;
+
+    // Serialize + coalesce the charging-schedule recompute: it fires concurrently from the interval timer,
+    // the libocpp message thread, and the K28 on_deadline/reaper callbacks.
+    std::mutex recompute_mutex;
+    std::atomic_bool recompute_pending{false};
+
+    // Queue the event if OCPP hasn't started yet. Gate is re-checked under m_member_mux so a
```
james-ctc: this implementation detail should be in the .cpp file

the following is much clearer about intent
```
    std::lock_guard lock(m_member_mux);
    if (mv_started) {
        mv_charge_point.on_meter_value(evse_id, meter.state_of_charge, meter.meter.value());
        if (power_meter.power_W) {
            m_everest_device_model_storage->update_power(evse_id, power_meter.power_W->total);
        }
    } else {
        m_event_queue[evse_id].emplace(std::move(meter));
    }
}
```

Or alternatively add a double check when processing the queue that no events were added

modules/EVSE/OCPPmulti/generic_ocpp.cpp
```diff
@@ -0,0 +1,1934 @@
...
+
+        extensions_id++;
+    }
+}
+
+void GenericOcpp::ready_event_queue() {
+    using namespace module;
+    using namespace module::conversions;
+
+    EventQueue to_process;
```
james-ctc: put in a loop to catch events added rather than the odd multiple checks in a templated function

modules/EVSE/OCPPmulti/generic_ocpp.cpp
```diff
@@ -0,0 +1,1934 @@
...
+        resuming_session_ids = m_resuming_session_ids;
+    }
+    // we can now initialise the charge point's state machine. It reads the connector availability from the internal
+    // database and potentially triggers enable/disable callbacks at the evse.
+    mv_charge_point.start(boot_reason, resuming_session_ids, false);
+    {
+        // flip under m_member_mux: pairs with enqueue_if_not_started() so no event can be
+        // queued after this point (ready_event_queue() below drains the final queue state)
+        std::lock_guard lock(m_member_mux);
+        mv_started.store(true);
```
james-ctc: why not `mv_started = true`?

modules/EVSE/OCPPmulti/generic_ocpp.cpp
```diff
@@ -0,0 +1,1934 @@
...
+    }
+    return result;
+}
+
+bool GenericOcpp::handle_restart() {
+    // Connects the websocket and enables OCPP communication after a previous
+    // stop call.
+
+    bool result{false};
+    if (mv_started.load()) {
```
james-ctc, line 216: is the implication the `if (mv_started) {` isn't atomic? There needs to be a clear reference to the spec if that is indeed the case. I can't find any comment from https://cppreference.com/cpp/header/atomic that this is the case

modules/EVSE/OCPPmulti/generic_ocpp.cpp
```diff
@@ -0,0 +1,1934 @@
...
+    // was originally in ready()
+    const auto log_path = mv_config.getMessageLogPath();
+    if (!fs::exists(log_path)) {
+        try {
+            fs::create_directory(log_path);
+        } catch (const fs::filesystem_error& ex) {
+            EVLOG_AND_THROW(ex);
+        }
+    }
+
```
james-ctc, line 370: Where has 
```
    {
        std::lock_guard lock(m_member_mux);
        const auto map_path = mv_config.getCustomMrecErrorMapPath();
        mv_mrec_error_map =
            (map_path.empty()) ? module::MREC_ERROR_MAP : module::load_mrec_error_map_overrides(map_path);
    }
```
gone?

modules/EVSE/OCPPmulti/generic_ocpp.cpp
```diff
@@ -0,0 +1,1934 @@
...
+    {
+        std::lock_guard lock(m_member_mux);
+        to_process = std::move(m_event_queue);
+    }
+
+    // process event queue
+    for (auto& [evse_id, evse_event_queue] : to_process) {
+        while (!evse_event_queue.empty()) {
+            auto& queued_event = evse_event_queue.front();
+            try {
```
james-ctc: under what situation is an exception expected?
each of the visit_functions should be the same as how the event was handled when not on the queue.
i.e. if they throw then handling would be needed there. e.g. see line 734 where this is already done.

The else{} should handle the case where a queue type has been accidentally omitted

modules/EVSE/OCPPmulti/generic_ocpp.cpp
```diff
@@ -0,0 +1,1934 @@
...
+        mv_ocpp_protocol_version = ocpp::OcppProtocolVersion::Unknown;
+    }
+    mv_provides.ocpp_generic.publish_is_connected(is_connected);
+}
+
+ocpp::v2::DataTransferResponse GenericOcpp::cb_data_transfer(const ocpp::v2::DataTransferRequest& request) {
+    ocpp::v2::DataTransferResponse response{};
+    if (mv_requires.data_transfer.empty()) {
+        EVLOG_error << "data_transfer called with no configured connections";
+        response.status = ocpp::v2::DataTransferStatusEnum::UnknownVendorId;
```
james-ctc, line 853: why was the comment explaining why rejected isn't used removed?
UnknownVendorId when there are no connections is not an obvious result - hence the comment

modules/EVSE/OCPPmulti/generic_ocpp.cpp
```diff
@@ -0,0 +1,1934 @@
...
+}
+
+void GenericOcpp::cb_error_cleared_handler(const Everest::error::Error& error) {
+    using namespace module;
+
+    // handled by specific evse_manager error handler
+    if (error.type != EVSE_MANAGER_INOPERATIVE_ERROR) {
+        auto event_data = convert_error(error);
+        event_data.event_cleared = true;
+        if (!enqueue_if_not_started(event_data.evse_id, event_data)) {
```
james-ctc, line 876: I don't like the change in approach and how the check for `if (mv_started) {` is now no longer consistent.

modules/EVSE/OCPPmulti/generic_ocpp.cpp
```diff
@@ -0,0 +1,1934 @@
...
+            EVLOG_warning << "Could not convert OCPP ResetEnum to EVerest ResetType while executing reset_callack. No "
+                             "reset will be executed.";
+            do_reset = false;
+            break;
+        }
+
+        if (do_reset) {
+            // small delay before stopping the charge point to make sure all responses are received
+            std::this_thread::sleep_for(std::chrono::seconds(mv_config.getResetStopDelay()));
+            this->mv_charge_point.stop();
```
james-ctc: `this->` shouldn't be there

modules/EVSE/OCPPmulti/generic_ocpp.cpp
```diff
@@ -0,0 +1,1934 @@
...
+        m_charging_schedules_timer.interval([this]() { cb_set_charging_profiles(); }, std::chrono::seconds(interval));
+    }
+}
+
+void GenericOcpp::charging_schedules_timer_stop() {
+    m_charging_schedules_timer.stop();
+}
+
+void GenericOcpp::shutdown() {
+    // Order matters. Set the gates first: mv_started stops command handlers and event enqueue,
```
james-ctc: `mv_started` doesn't stop all the command handlers - if that is what is needed then the code needs updating so that nothing happens on `mv_started` being false other than queuing the event.
Some responses are given when it is false.
Alternatively the comment could be reworded to explain what actually happens

modules/EVSE/OCPPmulti/generic_ocpp.cpp
```diff
@@ -0,0 +1,1934 @@
...
+        evse_id++;
+    }
+
+    return {evse_connector_structure, connector_mapping};
+}
+
+void GenericOcpp::process_session_event(std::int32_t evse_id, const types::evse_manager::SessionEvent& session_event) {
+    const auto connector_id = session_event.connector_id.value_or(1);
+    std::lock_guard lock(m_session_event_mutex);
+    try {
```
james-ctc, line 1774: just add the try/catch to process_session_event_impl. It prevents process_session_event_impl from being called without locking the mutex

modules/EVSE/OCPPmulti/manifest.yaml
```diff
@@ -0,0 +1,233 @@
...
+    description: >-
+      OCPP 1.6
+      Optional path to a YAML file with custom OCPP1.6 to device model component-config mappings.
+      Relative paths are resolved against this module's share path.
+      Applied both when migrating the legacy OCPP 1.6 JSON config into the device model and when
+      serving the device-model-backed configuration at runtime.
+      If empty, only built-in mappings are used.
+    type: string
+    default: ""
+  Ocpp16NetworkConfigSlot:
```
james-ctc: please keep alphabetical ordering to make finding items easier

modules/EVSE/OCPPmulti/v2_chargepoint.cpp
```diff
@@ -0,0 +1,1012 @@
...
+    const ocpp::v2::VariableCharacteristics& characteristics, const ocpp::v2::VariableAttribute& attribute,
+    const std::string& value_previous, const std::string& value_current) {
+    // copy under lock, invoke outside
+    const listener_t listener = *m_variable_listener.handle();
+    if (listener != nullptr) {
+        listener(component, variable, value_current);
+    }
+}
+
+namespace {
```
james-ctc: 1) why a separate anonymous namespace
2) why replace something that already works rather than adding another templated function?

The following is using a defined method called on the libocpp object and is just making sure optional is nullopt. The replacement code is a copy from somewhere else that will need long term maintenance and it isn't as clear what it is doing.

```
    const auto result = m_charge_point->request_value<std::int32_t>(component_id, variable_id, attribute_enum);
    std::optional<std::int32_t> res;
    if (result.status == ocpp::v2::GetVariableStatusEnum::Accepted) {
        res = result.value;
    }
    return res;
```

modules/EVSE/OCPPmulti/v2_chargepoint.hpp
```diff
@@ -0,0 +1,163 @@
...
+class ChargePointV2 : public GenericChargePointInterface {
+private:
+    using ResetType = ocpp_multi::GenericChargePointCallbacks::ResetType;
+
+    std::shared_ptr<EvseSecurity> m_evse_security;
+    ocpp_multi::GenericChargePointCallbacks* m_callbacks_ptr{nullptr};
+    // interface type so unit tests can inject a mock; init() creates the real ocpp::v2::ChargePoint
+    std::unique_ptr<ocpp::v2::ChargePointInterface> m_charge_point;
+    // written from command threads (register_variable_listener), read from libocpp callback thread
+    everest::lib::util::monitor<listener_t> m_variable_listener;
```
james-ctc, line 24: Still think a mutex is simpler and more obvious and uses less resource

modules/EVSE/OCPPmulti/v16_chargepoint.cpp
```diff
@@ -0,0 +1,1132 @@
...
+    m_callbacks_ptr->cb_time_sync(ocpp::DateTime{system_time});
+}
+
+ocpp::v16::UpdateFirmwareStatusEnumType
+ChargePointV16::cb_signed_update_firmware(const ocpp::v16::SignedUpdateFirmwareRequest msg) {
+    return convert(m_callbacks_ptr->cb_update_firmware_request(convert(msg)));
+}
+
+bool ChargePointV16::cb_stop_transaction(std::int32_t ocpp_connector_id, ocpp::v16::Reason reason) {
+    const auto evse = evse_from_ocpp_connector(ocpp_connector_id);
```
james-ctc: please follow coding guidelines and the approach used in this module. Only a single return statement
```
    auto result = evse_from_ocpp_connector(ocpp_connector_id);
    if (result) {
        const auto stop_reason = conversions_v16::to_everest_stop_transaction_reason(reason);
        const auto res = m_callbacks_ptr->cb_stop_transaction(evse->evse_id, stop_reason);
        result = res == ocpp::v2::RequestStartStopStatusEnum::Accepted;
    }
    return result;
```

modules/EVSE/OCPPmulti/v16_chargepoint.cpp
```diff
@@ -0,0 +1,1132 @@
...
+    const auto msg = ocpp_conversions::to_everest_tariff_message(message);
+    m_callbacks_ptr->cb_tariff_message(msg);
+    ocpp::v16::DataTransferResponse response;
+    response.status = ocpp::v16::DataTransferStatus::Accepted;
+    return response;
+}
+
+void ChargePointV16::cb_transaction_started(const std::int32_t ocpp_connector_id, const std::string& session_id) {
+    const auto evse = evse_from_ocpp_connector(ocpp_connector_id);
+    if (!evse) {
```
james-ctc: see cb_stop_transaction

modules/EVSE/OCPPmulti/v16_chargepoint.cpp
```diff
@@ -0,0 +1,1132 @@
...
+    event.transactionInfo.transactionId = session_id;
+    // the numeric OCPP1.6 transaction id is not assigned yet at this point
+    m_callbacks_ptr->cb_transaction_event(event, std::nullopt);
+}
+
+void ChargePointV16::cb_transaction_stopped(const std::int32_t ocpp_connector_id, const std::string& session_id,
+                                            const std::int32_t transaction_id) {
+    EVLOG_info << "Transaction stopped at OCPP1.6 connector: " << ocpp_connector_id << ", session_id: " << session_id;
+    const auto evse = evse_from_ocpp_connector(ocpp_connector_id);
+    if (!evse) {
```
james-ctc: see cb_stop_transaction

modules/EVSE/OCPPmulti/v16_chargepoint.cpp
```diff
@@ -0,0 +1,1132 @@
...
+    event.transactionInfo.transactionId = session_id;
+    // event.seqNo = 0; seqNo does not exist in OCPP1.6
+    m_callbacks_ptr->cb_transaction_event(event, std::to_string(transaction_id));
+}
+
+void ChargePointV16::cb_transaction_updated(const std::int32_t ocpp_connector_id, const std::string& session_id,
+                                            const std::int32_t transaction_id,
+                                            const ocpp::v16::IdTagInfo& id_tag_info) {
+    const auto evse = evse_from_ocpp_connector(ocpp_connector_id);
+    if (!evse) {
```
james-ctc: see cb_stop_transaction

modules/EVSE/OCPPmulti/v16_chargepoint.cpp
```diff
@@ -0,0 +1,1132 @@
...
+        event_response.idTokenInfo = token;
+    }
+
+    m_callbacks_ptr->cb_transaction_event_response(event, event_response, std::to_string(transaction_id));
+}
+
+ocpp::v16::UnlockStatus ChargePointV16::cb_unlock_connector(std::int32_t ocpp_connector_id) {
+    const auto evse = evse_from_ocpp_connector(ocpp_connector_id);
+    if (!evse) {
+        return ocpp::v16::UnlockStatus::NotSupported;
```
james-ctc: see cb_stop_transaction

modules/EVSE/OCPPmulti/generic_ocpp.cpp
james-ctc: Further checking - there is a deliberate thow of runtime error that is now no longer actioned
```
throw std::runtime_error("SessionEvent TransactionFinished does not contain transaction_finished context");
```

modules/EVSE/OCPPmulti/v16_chargepoint.cpp
```diff
@@ -0,0 +1,1132 @@
...
+    check_configured("on_unavailable");
+    m_charge_point->on_disabled(evse_id);
+}
+
+void ChargePointV16::register_variable_listener(const ocpp::v2::Component& component,
+                                                const ocpp::v2::Variable& variable, listener_t listener) {
+    check_configured("register_variable_listener");
+    const std::string key = variable.name;
+    if (key.empty()) {
+        return;
```
james-ctc: please remove early returns and match existing code style and guidelines

james-ctc (commented):
starting set of comment - more still to review

modules/EVSE/OCPPmulti/charge_point_config_factory_v16.cpp
```diff
@@ -0,0 +1,273 @@
...
+    return context;
+}
+
+/// \brief Overrides the OCPP 1.6 NumberOfConnectors device-model value with the actual number of connected EVSEs.
+void patch_number_of_connectors(
+    std::map<ocpp::v2::ComponentKey, std::vector<ocpp::v2::DeviceModelVariable>>& component_configs, int32_t n_evse) {
+    constexpr auto legacy_ctrlr_name = "OCPP16LegacyCtrlr";
+    constexpr auto number_of_connectors_var = "NumberOfConnectors";
+
+    for (auto& [component_key, variables] : component_configs) {
```
james-ctc, line 113: rather than have continue the tests can be updated:
```
    for (auto& [component_key, variables] : component_configs) {
        if (component_key.name == legacy_ctrlr_name) {
            for (auto& variable : variables) {
                if (variable.name == number_of_connectors_var) {
                    for (auto& attribute : variable.attributes) {
                        if (attribute.variable_attribute.type == ocpp::v2::AttributeEnum::Actual) {
                            attribute.value_source = "OCPP16Config";
                            attribute.variable_attribute.value = std::to_string(n_evse);
                        }
                    }
                    variable.default_actual_value = std::to_string(n_evse);
                    return;
                }
            }
        }
    }
```

modules/EVSE/OCPPmulti/charge_point_config_factory_v16.cpp
```diff
@@ -0,0 +1,273 @@
...
+    }
+    return std::nullopt;
+}
+
+struct DeviceModelInitializationContext {
+    fs::path database_path;
+    fs::path migration_files_path;
+    fs::path component_config_path;
+    std::optional<fs::path> custom_mappings_path;
+    int32_t ocpp16_network_config_slot{1};
```
james-ctc: std::int32_t

modules/EVSE/OCPPmulti/charge_point_config_factory_v16.cpp
```diff
@@ -0,0 +1,273 @@
...
+    }
+
+    context.ocpp16_network_config_slot = static_cast<int32_t>(config.Ocpp16NetworkConfigSlot);
+
+    return context;
+}
+
+/// \brief Overrides the OCPP 1.6 NumberOfConnectors device-model value with the actual number of connected EVSEs.
+void patch_number_of_connectors(
+    std::map<ocpp::v2::ComponentKey, std::vector<ocpp::v2::DeviceModelVariable>>& component_configs, int32_t n_evse) {
```
james-ctc: std::int32_t

modules/EVSE/OCPPmulti/charge_point_config_factory_v16.cpp
```diff
@@ -0,0 +1,273 @@
...
+            variable.default_actual_value = std::to_string(n_evse);
+            return;
+        }
+    }
+
+    EVLOG_warning << "Could not find " << legacy_ctrlr_name << "/" << number_of_connectors_var
+                  << " in component configs; NumberOfConnectors may not match the number of connected EVSEs.";
+}
+
+void initialize_device_model_direct(const DeviceModelInitializationContext& context, int32_t n_evse) {
```
james-ctc: std::int32_t

modules/EVSE/OCPPmulti/charge_point_config_factory_v16.cpp
```diff
@@ -0,0 +1,273 @@
...
+    auto component_configs = ocpp::v2::get_all_component_configs(context.component_config_path);
+    ocpp::v2::ensure_ocpp16_legacy_ctrlr(component_configs);
+    patch_number_of_connectors(component_configs, n_evse);
+    ocpp::v2::InitDeviceModelDb init_device_model_db(context.database_path, context.migration_files_path);
+    init_device_model_db.initialize_database(component_configs, false);
+}
+
+void initialize_device_model_with_migration(const fs::path& ocpp_share_path, const fs::path& user_config_path,
+                                            const std::string& charge_point_config_json,
+                                            const DeviceModelInitializationContext& context, int32_t n_evse) {
```
james-ctc: std::int32_t

modules/EVSE/OCPPmulti/charge_point_config_factory_v16.cpp
```diff
@@ -0,0 +1,273 @@
...
+    return std::make_unique<ocpp::v16::ChargePointConfigurationDeviceModel>(
+        ocpp_share_path.string(), std::move(device_model), std::move(custom_mappings));
+}
+
+/// \brief Loads the legacy OCPP 1.6 JSON config used as the one-time device-model migration source.
+///
+/// Requires \p configured_config_path to exist, applies \p n_evse to the connector count, and overlays the
+/// user config (creating an empty one if missing).
+std::string load_charge_point_config_json(const fs::path& configured_config_path, const fs::path& user_config_path,
+                                          int32_t n_evse) {
```
james-ctc: std::int32_t

modules/EVSE/OCPPmulti/charge_point_config_factory_v16.cpp
```diff
@@ -0,0 +1,273 @@
...
+        create_empty_user_config(user_config_path);
+    }
+    return json_config.dump();
+}
+
+} // namespace
+
+std::unique_ptr<ocpp::v16::ChargePointConfigurationInterface>
+create_charge_point_configuration(const fs::path& ocpp_share_path, const Ocpp16DeviceModelParams& config,
+                                  int32_t n_evse) {
```
james-ctc: std::int32_t

modules/EVSE/OCPPmulti/charge_point_config_factory_v16.cpp
```diff
@@ -0,0 +1,273 @@
...
+#include <ocpp/v2/init_device_model_db.hpp>
+#include <ocpp/v2/ocpp16_component_config_patcher.hpp>
+#include <ocpp/v2/ocpp16_custom_config_mappings.hpp>
+
+namespace module::config_factory_v16 {
+
+namespace fs = std::filesystem;
+using json = nlohmann::json;
+
+namespace {
```
james-ctc, line 28: it is clearer to have the anonymous namespace on its own rather than a sub-namespace.

```
} // namespace

std::unique_ptr<ocpp::v16::ChargePointConfigurationInterface>
create_charge_point_configuration(const fs::path& ocpp_share_path, const Ocpp16DeviceModelParams& config,
                                  int32_t n_evse) {
```
makes it look like `create_charge_point_configuration` is in the global namespace

modules/EVSE/OCPPmulti/v16_conversions.cpp
```diff
@@ -0,0 +1,342 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Pionix GmbH and Contributors to EVerest
+
+#include "v16_conversions.hpp"
+
+#include <utility>
```
james-ctc: header file not used

modules/EVSE/OCPPmulti/generic_ocpp.cpp
Pietfried: It's just about being explicit and was requested by https://github.com/EVerest/EVerest/pull/2332#discussion_r3546055540

modules/EVSE/OCPPmulti/generic_ocpp.cpp
Pietfried: See https://github.com/EVerest/EVerest/pull/2332#discussion_r3569721243

modules/EVSE/OCPPmulti/generic_ocpp.cpp
Pietfried: Moved into the constructor of `GenericOcpp`. See

```// loaded before error subscriptions exist so cb_error_handler never sees an empty map```

File: modules/EVSE/OCPPmulti/generic_ocpp.hpp (hunk 27 above)
Comment (Pietfried) [reply modules/EVSE/OCPPmulti/generic_ocpp.hpp]: As far as I saw it `m_member_mux` is used to protect the module's own maps/queues. 
The idea is to not hold `m_member_mux` when calling into libocpp to protect against it calling back on the same thread, which could happen in case of e.g. `set_variables` -> `variable_listener` -> `cb_variable_listener` -> `cb_variable_monitor`

Agreed on the header point though, I'll make enqueue_if_not_started a non-template

File: modules/EVSE/OCPPmulti/generic_ocpp.cpp (hunk 28 above)
Comment (Pietfried) [reply modules/EVSE/OCPPmulti/generic_ocpp.cpp]: `mv_started` was flipped before, so no events could be added to the queue at this point. During the review I noticed that there was a window between flipping mv_started and ready_event_queue in which events could enter the live path before the event_queue was drained, so I moved flipping `mv_started` into `ready_event_queue` and synchronized with `enqueue_if_not_started`. (Will link commit here when I push)

File: modules/EVSE/OCPPmulti/generic_ocpp.cpp (hunk 32 above)
Comment (Pietfried) [reply modules/EVSE/OCPPmulti/generic_ocpp.cpp]: Good catch, I removed the try/catch block here

File: modules/EVSE/OCPPmulti/generic_ocpp.cpp (hunk 33 above)
Comment (Pietfried) [reply modules/EVSE/OCPPmulti/generic_ocpp.cpp:853]: I didnt like to leak test case specifics into this file

File: modules/EVSE/OCPPmulti/generic_ocpp.cpp (hunk 36 above)
Comment (Pietfried) [reply modules/EVSE/OCPPmulti/generic_ocpp.cpp]: Updated the comment

File: modules/EVSE/OCPPmulti/v2_chargepoint.cpp (hunk 39 above)
Comment (Pietfried) [reply modules/EVSE/OCPPmulti/v2_chargepoint.cpp]: 1. moved the function into the existing namespace
2. `m_charge_point` is no longer the concrete `ocpp::v2::ChargePoint` but it was changed to `ocpp::v2::ChargePointInterface` in d8964eae so unit tests can inject a mock.

File: modules/EVSE/OCPPmulti/generic_ocpp.cpp [hunk 56]
```diff
@@ -0,0 +1,1927 @@
...
+}
+
+bool GenericOcpp::enqueue_if_not_started(std::int32_t evse_id, Event event) {
+    if (mv_started.load()) {
+        return false;
+    }
+    // Re-check under m_member_mux: ready_event_queue() only flips mv_started in the critical
+    // section that observes the queue empty, so a concurrent start can't strand the event in a
+    // queue that is never drained again.
+    std::lock_guard lock(m_member_mux);
```
james-ctc, line 602: Now that the queue is drained again this is no longer needed

modules/EVSE/OCPPmulti/generic_ocpp.cpp
james-ctc: Being explicit can add to confusion
e.g. OCPP CiString string where get() is used which makes it look like a pointer type.
std::string from a CiString should just work.

It's not worth holding this review up - but I would not like to see this become the default approach in the code.

modules/EVSE/OCPPmulti/OCPPmulti.hpp
james-ctc: it is a compile time constant as you state. I still prefer references over pointers, so if you insist on using pointers then every use of that pointer needs an explicit nullptr check and appropriate error handling.

modules/EVSE/OCPPmulti/generic_ocpp.cpp
Pietfried: The race it guards against still exists:
1. enqueue_if_not_started see `mv_started == false`
2. `ready_event_queue` takes the lock with `m_member_mux`, see event_queue is empty and flips`mv_started = true`, breaks out of the loop and releases the lock
3. enqueue_if_not_started takes the lock with `m_member_mux`, and puts the event on the queue, nothing will consume the event afterwards

modules/EVSE/OCPPmulti/OCPPmulti.hpp
Pietfried: The pointer is derefenced in exactly one place `GenericChargePoint::init()` guarded by a NotConfigured throw. ChargePointV16/ChargePointV2 receive *m_charge_point_callbacks as plain references, so all downstream code keeps the reference semantics

modules/EVSE/OCPPmulti/generic_ocpp.cpp
james-ctc: the original code solved that:
```
    EventQueue to_process;
    {
        std::lock_guard lock(m_member_mux);
        to_process = std::move(m_event_queue);
    }
    ```
    allowing addition to the queue because the mutex was no longer being held

File: modules/EVSE/OCPPmulti/generic_ocpp.hpp (hunk 27 above)
Comment (james-ctc) [reply modules/EVSE/OCPPmulti/generic_ocpp.hpp]: The original solution was 
```
        if (mv_started) {
            mv_charge_point.on_event(event_data);
        } else {
            std::lock_guard lock(m_member_mux);
            m_event_queue[event_data.evse_id].emplace(event_data);
        }

...
mv_started = true;
...
    EventQueue to_process;
    {
        std::lock_guard lock(m_member_mux);
        to_process = std::move(m_event_queue);
    }
```
The addition of the while loop means that events added after `mv_started = true;` due to thread scheduling are still processed. 
i.e. something like
```
    bool finished{false};
    while (!finished) {
        EventQueue to_process;
        {
            std::lock_guard lock(m_member_mux);
            mv_started = true;
            to_process = std::move(m_event_queue);
        }

        for (auto& [evse_id, evse_event_queue] : to_process) {
            while (!evse_event_queue.empty()) {
                // dispatch via the visit_impl overload set
                std::visit([this, evse_id](const auto& arg) { visit_impl(evse_id, arg); }, evse_event_queue.front());
                evse_event_queue.pop();
            }
        }

        {
            std::lock_guard lock(m_member_mux);
            finished = m_event_queue.empty();
        }
    }

```

note `m_event_queue.clear();` isn't needed since the preceding move has cleared the queue.

Regardless of the above, if enqueue_if_not_started() is to remain then it should be modified
```
bool GenericOcpp::enqueue_if_not_started(std::int32_t evse_id, Event event) {
    if (mv_started.load()) {
        return false;
    }
    // Re-check under m_member_mux: ready_event_queue() only flips mv_started in the critical
    // section that observes the queue empty, so a concurrent start can't strand the event in a
    // queue that is never drained again.
    std::lock_guard lock(m_member_mux);
    if (mv_started.load()) {
        return false;
    }
    m_event_queue[evse_id].emplace(std::move(event));
    return true;
}
```

simplifies to

```
bool GenericOcpp::enqueue_if_not_started(std::int32_t evse_id, const Event &event) {
    std::lock_guard lock(m_member_mux);
    bool result{false};
    if (!mv_started) {
        m_event_queue[evse_id].emplace(event);
        result = true;
    }
    return result;
}
```

The move was removed because the original code was already making a copy anyway.

Pietfried: Right! I cut it down to the simplified version

Pietfried (approved):

PR #2333, opened 2026-06-18
Auth: Fix the ordering of the PluginEvent algorithm

dorezyuk:
## Describe your changes

I'm actually not certain from the behavior since it's not clearly documented but the name `get_latest_plugin` implies that we want to get the most recent plugin and not the oldest. This seems also to be nicer with the customers since if we have 2 customers connecting it seems to make sense to assume that the most recent one is the one authorizing

### DIFF
--- a/modules/EVSE/Auth/BUILD.bazel
+++ b/modules/EVSE/Auth/BUILD.bazel
@@ -1,21 +1,12 @@
-load("@rules_cc//cc:defs.bzl", "cc_library")
+load("@rules_cc//cc:defs.bzl", "cc_library", "cc_test")
 load("//modules:module.bzl", "cc_everest_module")
+load("//third-party/bazel/toolchains:defs.bzl", "CROSS_TEST_INCOMPATIBLE")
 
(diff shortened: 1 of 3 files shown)


Pietfried:
The documentation of the Auth module was indeed not very explicit about the selection behavior (and additionally somewhat outdated) but the current implementation represents the intended behavior. I took the liberty of reworking the Auth module documentation and added the requested `PlugEventsLIFO` selection mechanism. This avoids potential breaking changes for existing integrations: https://github.com/EVerest/EVerest/pull/2344 It would be great if you can have a look at the PR.

dorezyuk:
> The documentation of the Auth module was indeed not very explicit about the selection behavior (and additionally somewhat outdated) but the current implementation represents the intended behavior. I took the liberty of reworking the Auth module documentation and added the requested `PlugEventsLIFO` selection mechanism. This avoids potential breaking changes for existing integrations: #2344 It would be great if you can have a look at the PR.

love it - closing this one then

PR #2337, opened 2026-06-19
system: Add a reference implementation and a test for updates that do not disable the connectors

nzbr:
## Describe your changes

Extends the system module with a metadata parser script that can read metadata from the downloaded firmware which is then passed on via the new firmware_update_metadata type.
Adds a test that uses this new functionality to verify that during regular updates, all connectors are disable and that this does not happen when the corrsponding metadata flag is set


modules/Misc/System/main/systemImpl.cpp
```diff
@@ -2,7 +2,12 @@
 // Copyright Pionix GmbH and Contributors to EVerest
 
 #include "systemImpl.hpp"
+#include "everest/logging.hpp"
+#include "generated/types/system.hpp"
 
+#include <boost/algorithm/string/classification.hpp>
+#include <boost/algorithm/string/predicate.hpp>
+#include <boost/algorithm/string/split.hpp>
```
james-ctc: observation - there is slow move away from using boost.

james-ctc (dismissed):
lgtm

modules/Misc/System/main/systemImpl.cpp
nzbr: I just saw that it was available and had the functions that I need. Is there another dependency that has similar things or are we migrating to doing those things in EVerest itself? I am absolutely open to changing this

barsnick:
Some of the commits in the commit chain aren't signed off.

modules/Misc/System/main/systemImpl.cpp
nzbr: I switched to an implementation without boost

james-ctc (approved):
LGTM

PR #2338, opened 2026-06-19
Only snapshot the repo itself (instead of parent directory) when building the docs

nzbr:
## Describe your changes

The documentation build process creates a snapshot of the parent directory of the EVerest directory. This is a remnant of when EVerest was not yet a monorepo and can cause a lot of unneccesary files to be copied into that snapshot


hikinggrass (approved):

andistorm (changes requested):
The snapshot yaml should include at least all repos that contribute content to the documentation. The only repos that could do this would be out-of-tree repos that have modules in it with some documentation.

Did we ever deprecate this out-of-tree feature? Did we remove it?
I would suggest to couple this snapshot of the everest workspace feature to the out-of-tree feature. Means depracting both and removing both at the same time

hikinggrass:
> The snapshot yaml should include at least all repos that contribute content to the documentation. The only repos that could do this would be out-of-tree repos that have modules in it with some documentation.
> 
> Did we ever deprecate this out-of-tree feature? Did we remove it? I would suggest to couple this snapshot of the everest workspace feature to the out-of-tree feature. Means depracting both and removing both at the same time

Ah right, as far as I'm aware this is still a possibility, so we should probably rather close than merge this PR :sweat_smile:

PR #2347, opened 2026-06-22
Fixes a segmentation fault in EvseSlac

djchhp:
## Describe your change
Check pointer before usage

### DIFF
--- a/modules/EVSE/EvseSlac/main/slacImpl.cpp
+++ b/modules/EVSE/EvseSlac/main/slacImpl.cpp
@@ -13,10 +13,6 @@
 
 #include "fsm_controller.hpp"
 
-static std::promise<void> module_ready;
-// FIXME (aw): this is ugly, but due to the design of the auto-generated module skeleton ..
-static std::unique_ptr<FSMController> fsm_ctrl{nullptr};
-
 namespace module {
 namespace main {
 
@@ -28,19 +24,6 @@ static std::string mac_to_ascii(const std::string& mac_binary) {
 }
 
 void slacImpl::init() {
-    // setup evse fsm thread
-    std::thread(&slacImpl::run, this).detach();
-}
-
-void slacImpl::ready() {
-    // let the waiting run thread go
-    module_ready.set_value();
-}
-
-void slacImpl::run() {
-    // wait until ready
-    module_ready.get_future().get();
-
     if (config.startup_delay_ms > 0) {
         EVLOG_info << "Delaying SLAC startup by " << config.startup_delay_ms << "ms";
         std::this_thread::sleep_for(std::chrono::milliseconds(config.startup_delay_ms));
@@ -48,7 +31,6 @@ void slacImpl::run() {
     }
 
     // initialize slac i/o
-    SlacIO slac_io;
     try {
         slac_io.init(config.device);
     } catch (const std::exception& e) {
@@ -60,8 +42,7 @@ void slacImpl::run() {
     }
 
     // setup callbacks
-    slac::fsm::evse::ContextCallbacks callbacks;
-    callbacks.send_raw_slac = [&slac_io](slac::messages::HomeplugMessage& msg) { slac_io.send(msg); };
+    callbacks.send_raw_slac = [this](slac::messages::HomeplugMessage& msg) { slac_io.send(msg); };
 
     callbacks.signal_dlink_ready = [this](bool value) { publish_dlink_ready(value); };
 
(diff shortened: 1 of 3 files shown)


modules/EVSE/EvseSlac/main/slacImpl.cpp
```diff
@@ -118,17 +99,29 @@ void slacImpl::run() {
...
         if (msg.get_mmtype() == slac::defs::qualcomm::MMTYPE_VS_ATTENUATION_CHARACTERISTICS) {
             return;
         }
         fsm_ctrl->signal_new_slac_message(msg);
     });
+}
 
-    fsm_ctrl->run();
+void slacImpl::ready() {
+    if (fsm_ctrl) {
```
SebaLukas, line 118: `fsm_ctrl` is set in line 95. Is there any possibility that this ready() function is called before the init function is finished. I guess not

modules/EVSE/EvseSlac/main/slacImpl.cpp
```diff
@@ -118,17 +99,29 @@ void slacImpl::run() {
     // MMTYPE and logs "Received non-expected SLAC message of type 0xA14E" per frame, which
     // adds RX/log load. Drop it pre-FSM. Other MMTYPEs (incl. CM_SET_KEY.CNF, CM_ATTEN_PROFILE.IND)
     // pass through unchanged.
-    slac_io.run([](slac::messages::HomeplugMessage& msg) {
+    slac_io.run([this](slac::messages::HomeplugMessage& msg) {
+        if (not fsm_ctrl) {
```
SebaLukas, line 106: Same question here: Is there any possibility that fsm_ctrl is resetted and points to null? EvseSlac has no prober shutdown mechanism and runs forever

modules/EVSE/EvseSlac/main/slacImpl.cpp
```diff
@@ -140,14 +133,25 @@ void slacImpl::handle_reset(bool& enable) {
 };
 
 void slacImpl::handle_enter_bcd() {
+    if (not fsm_ctrl) {
```
SebaLukas, line 139: Same question: The framework should prevent that the cmds are called before every init() is finished. So I guess the checks are not necessary?

SebaLukas (changes requested):

modules/EVSE/EvseSlac/main/slacImpl.cpp
hikinggrass: Good point, before the changes in this PR this could actually happen quite easily because the `slacImpl::run()` was called in a detached thread originating in `slacImpl::init()` ( https://github.com/EVerest/EVerest/pull/2347/changes#diff-8bcd464c62e1875820db3b50897abac934816a3030ac954773ca1ede8bfa640fL32 ). But might not be possible anymore

modules/EVSE/EvseSlac/main/slacImpl.cpp
hikinggrass, line 41: @SebaLukas because of this line the following nullptr checks are still needed, because this will lead to fsm_ctrl not being initialized

modules/EVSE/EvseSlac/main/slacImpl.cpp
hikinggrass: still needed, see my other comment

modules/EVSE/EvseSlac/main/slacImpl.cpp
hikinggrass: still needed, see my other comment

modules/EVSE/EvseSlac/main/slacImpl.cpp
hikinggrass: still needed, see my other comment

modules/EVSE/EvseSlac/main/slacImpl.cpp
SebaLukas: But you return early the init function, so EVerest stops or starts without problems even with the raised error?

SebaLukas: If an error is here raised, then the module should be not ready and the other modules should wait or shutdown

hikinggrass: EVerest will continue its startup when you return from the init, the only way to prevent this is to raise an exception or exit here. But then the whole startup is aborted. Since we already raise a CommunicationFault here this already gives the other modules downstream the information that SLAC will not be possible on this EVSE. But since there can be other EVSEs, OCPP etc. we definitely do not want to shutdown/crash the whole EVerest in this scenario.

Retries for when the device comes back could be implemented, but this is probably best done in the re-written slac-neo implementation :D

SebaLukas (approved):
I haven't found a better alternative, so go ahead 👍

PR #2351, opened 2026-06-22
feat(EM580): monitor OCMF transaction state during active sessions

florinmihut:
Add optional monitor_transaction_state to detect device-side transaction closure (e.g. brief power loss) and raise VendorError so EvseManager can stop charging. Rename device_state_read_interval_ms to device_and_transaction_state_read_interval_ms to reflect shared polling.

Fixes as well an error when if a device gets powered off but it has an open transaction, a new transaction that it is issued without a close transaction (to use the OCMF data) would fail. Now it will trigger an WARNING of spurious transaction but it will work. 
This is a theoretical case since the upper layers should close first existing/pending transactions before starting a new one.


modules/HardwareDrivers/PowerMeters/CarloGavazzi_EM580/manifest.yaml
```diff
@@ -52,8 +52,12 @@ provides:
         minimum: 500
         maximum: 60000
         default: 1000
-      device_state_read_interval_ms:
-        description: Interval in milliseconds between reading the device state bitfield (used for VendorError reporting). Default 10000 ms (once per 10 seconds). Allowed range 500-60000 ms.
+      device_and_transaction_state_read_interval_ms:
```
Pietfried: That is a breaking change according to https://everest.github.io/latest/project/releases/breaking-changes.html#id2 `Renaming a configuration option`. I think we can simply avoid the renaming and simply adjust the description.

modules/HardwareDrivers/PowerMeters/CarloGavazzi_EM580/main/powermeterImpl.cpp
```diff
@@ -432,6 +440,37 @@ void powermeterImpl::read_transaction_state_and_id() {
...
+        return;
+    }
+    if (!m_transaction_active.load()) {
+        return;
+    }
+    if (ocmf_state != MODBUS_OCMF_STATE_READY) {
+        return;
+    }
+
+    m_pending_closed_transaction = true;
```
Pietfried: `m_pending_closed_transaction` is now written from the `live_measure_thread_` and the command handlers `handle_start_transaction` and `handle_stop_transaction`. Should therefore be at least an atomic_bool .

modules/HardwareDrivers/PowerMeters/CarloGavazzi_EM580/main/powermeterImpl.cpp
```diff
@@ -432,6 +440,37 @@ void powermeterImpl::read_transaction_state_and_id() {
...
+    const std::string error_description =
+        fmt::format("OCMF transaction closed unexpectedly on device (state READY) while transaction {} is still active",
+                    m_transaction_id);
```
Pietfried: Also `m_transaction_id` requires a mutex/guard. It is read from `live_measure_thread_` and written from the command handlers

modules/HardwareDrivers/PowerMeters/CarloGavazzi_EM580/main/powermeterImpl.cpp
Pietfried: Looks like if `wait_for_ready` returns false or anything throws after this statement, we end up not calling `clear_ocmf_transaction_closed_error` so the `OcmfTransactionClosed` error won't be cleared.

Pietfried (changes requested):

modules/HardwareDrivers/PowerMeters/CarloGavazzi_EM580/manifest.yaml
```diff
@@ -65,6 +69,14 @@ provides:
           - binary
           - der
         default: binary
+      monitor_transaction_state:
+        description: >-
+          When enabled, reads the OCMF state at device_and_transaction_state_read_interval_ms
+          while a transaction is active. If the device reports READY (transaction closed on
+          the meter side, e.g. after a brief power loss), raises powermeter/VendorError so
+          the charging session can be stopped when fail_on_powermeter_errors is enabled.
```
Pietfried: Not ideal that we need to cross-reference the EvseManager configuration here, but I also dont see a better way at the moment.
```suggestion
          the charging session can be stopped when fail_on_powermeter_errors is enabled in the EvseManager module.
```

modules/HardwareDrivers/PowerMeters/CarloGavazzi_EM580/main/powermeterImpl.cpp
florinmihut: True, but the VendorError should remain active if I am not able to close the transaction. 
I mean, imagine that I clear the VendorError but the device is still not usable since I can't run a stop transaction to completion (not sure why). With the VendorError there is at least an indication that something is wrong after a power outage for example. Clearing it would be here wrong.
A power outage ONLY on the Powermeter is already something veeeery strange and there should be an indication of it.

modules/HardwareDrivers/PowerMeters/CarloGavazzi_EM580/manifest.yaml
florinmihut: Ok, point taken, I reverted the parameter change, the name is okish enough and can be kept.

modules/HardwareDrivers/PowerMeters/CarloGavazzi_EM580/main/powermeterImpl.cpp
florinmihut: Goo catch, thank you, fixed!

modules/HardwareDrivers/PowerMeters/CarloGavazzi_EM580/main/powermeterImpl.cpp
florinmihut: Same, good catch, fixed.

modules/HardwareDrivers/PowerMeters/CarloGavazzi_EM580/README.md
Pietfried, line 1: Should we just merge index.rst and README.md and remove README.md in a subsequent PR?

modules/HardwareDrivers/PowerMeters/CarloGavazzi_EM580/manifest.yaml
```diff
@@ -14,14 +14,14 @@ provides:
         description: Number of retries for communication operations before giving up.
         type: integer
         minimum: 1
-        maximum: 100
-        default: 3
+        maximum: 3
```
Pietfried: [Breaking change](https://everest.github.io/latest/project/releases/breaking-changes.html#id2): "Narrowing acceptable value ranges" and "Changing default values that affect runtime behavior"

modules/HardwareDrivers/PowerMeters/CarloGavazzi_EM580/manifest.yaml
```diff
@@ -14,14 +14,14 @@ provides:
...
+        default: 2
       communication_retry_delay_ms:
         description: Delay in milliseconds between retry attempts.
         type: integer
         minimum: 10
-        maximum: 10000
-        default: 500
+        maximum: 500
```
Pietfried: Same here

Pietfried (changes requested):

modules/HardwareDrivers/PowerMeters/CarloGavazzi_EM580/manifest.yaml
florinmihut: Sorry, but we will have to live with those even if those are breaking. 
Allowing large values in the retry mechanism is causing issues and it is very difficult in fact to recover unless the device is going in CommunicationFailure. This mechanism was initially designed in fact to detect small interruptions in the communication. Larger than 2-3 seconds in the communication gap would indicate more serious problems (like power outrages for example). 
So I am afraid we have to stick to the new limitations otherwise a badly configured device will misbehave.

modules/HardwareDrivers/PowerMeters/CarloGavazzi_EM580/manifest.yaml
florinmihut: Sorry, but we will have to live with those even if those are breaking. 
Allowing large values in the retry mechanism is causing issues and it is very difficult in fact to recover unless the device is going in CommunicationFailure. This mechanism was initially designed in fact to detect small interruptions in the communication. Larger than 2-3 seconds in the communication gap would indicate more serious problems (like power outrages for example). 
So I am afraid we have to stick to the new limitations otherwise a badly configured device will misbehave.

modules/HardwareDrivers/PowerMeters/CarloGavazzi_EM580/manifest.yaml
florinmihut: Reverted the changes, I am issuing an warning now if the module is misconfigured.

modules/HardwareDrivers/PowerMeters/CarloGavazzi_EM580/manifest.yaml
florinmihut: Reverted the changes, I am issuing an warning now if the module is misconfigured.

modules/HardwareDrivers/PowerMeters/CarloGavazzi_EM580/README.md
florinmihut: Good idea, only one source of truth.

modules/HardwareDrivers/PowerMeters/CarloGavazzi_EM580/main/powermeterImpl.cpp
```diff
@@ -667,21 +758,23 @@ types::powermeter::TransactionStopResponse powermeterImpl::handle_stop_transacti
...
+            m_pending_closed_transaction.store(false);
             clear_transaction_states();
         } catch (const std::exception& e) {
             EVLOG_error << e.what();
         }
-        m_pending_closed_transaction = false;
+        m_pending_closed_transaction.store(false);
+        m_transaction_active.store(false);
+        clear_ocmf_transaction_closed_error();
         return make_transaction_stop_response(types::powermeter::TransactionRequestStatus::OK);
```
Pietfried, line 774: We still return `OK` and reset the flags in case of an exception at L769. Should we return `UNEXPECTED_ERROR`and only clear the flags and error if no exception was thrown?

florinmihut: True, we still return OK. 
However we are in the following case: somebody requrested to close a transaction with the id empty which indicates something went terrible wrong. It might havbe been a power loss. Or an everet coredump. Somehow we are out of sync.
The code that tries to clear the transaction open flags, should not return an error (except the case we try to close a transaction that was already closed for example). 

Getting an exception above could mean that the driver is totally out of sync with the device. Remember that the upper layer already tries to stop the transaction with a given id and probably that failed already.
So, the best idea here to clean up things as much as we can and hope for the best because we have no idea what the device does. Maybe it is not ready, maybe is corrupted, it can be anything.

Now, we have another cleaning up mechanism in start_transaction. So, I think here it is fine to "lie" and when the start transaction will be triggered, will try again to bring the device in a clean state. If that will succeed, all good, if not, it will fail there too.
Returning an UNEXPECTED_ERROR does nothing in EvseManager, the response is anyway ignored.

So, in fact makes no difference what we do return here.

Pietfried (approved):

PR #2355, opened 2026-06-23
feat(libiso15118): Adding AC DER IEC state machine

SebaLukas:
## Describe your changes
See title

Adding missing AC DER IEC state machine


docs/source/reference/EVerest_API/auth_token_validator_API.yaml
```diff
@@ -221,7 +221,8 @@ components:
         - DC
         - AC_BPT
         - AC_BPT_DER
-        - AC_DER
+        - AC_DER_IEC
+        - AC_DER_SAE
```
mlitre, line 226: This applies to all others as well, should this be AC_DER or AC_BPT_DER?

lib/everest/iso15118/include/iso15118/d20/config.hpp
```diff
@@ -27,6 +30,13 @@ struct BptSetupConfig {
     std::optional<message_20::datatypes::GridCodeIslandingDetectionMethod> grid_code_detection_method;
 };
 
+struct DerSetupConfig {
+    std::bitset<11> control_functions;
```
mlitre: Do we really need a bitset? I know in the messages it uses a bitset, but for our internal storage can't we just use a vector or set?
Also shouldn't this be 12 since we have 12 enum values?

lib/everest/iso15118/include/iso15118/d20/config.hpp
```diff
@@ -35,10 +45,12 @@ struct EvseSetupConfig {
     bool enable_certificate_install_service;
     d20::DcTransferLimits dc_limits;
     d20::AcTransferLimits ac_limits;
+    std::optional<d20::IecDerTransferLimits> der_limits;
     std::vector<ControlMobilityNeedsModes> control_mobility_modes;
     std::optional<std::string> custom_protocol{std::nullopt};
     std::optional<AcSetupConfig> ac_setup_config{std::nullopt};
     std::optional<BptSetupConfig> bpt_setup_config{std::nullopt};
+    std::optional<DerSetupConfig> der_setup_config{std::nullopt};
```
mlitre, line 52: Can we easily change this config? It could change runtime

lib/everest/iso15118/include/iso15118/d20/der_functions.hpp
```diff
@@ -0,0 +1,173 @@
...
+    GridForming,
+};
+
+enum class GridConnectionMode : std::uint8_t {
+    GridConnected,
+    GridIslanded
+};
+
+enum class DERControlName {
+    OverFrequencyWattMode = 0,
```
mlitre, line 24: Can we stay consistent and either they all have = 0 for the first element or they don't

lib/everest/iso15118/include/iso15118/d20/der_functions.hpp
```diff
@@ -0,0 +1,173 @@
...
+    DSOCosPhiSetpointProvision,
+    DCInjectionRestriction,
+    ZeroCurrentMode,
+    OverVoltageFaultRideThroughMode,
+    UnderVoltageFaultRideThroughMode
+};
+
+using MaximumLevelDCInjection = float;
+
+enum class PowerReference : uint8_t {
```
mlitre: Same can we stay consistent and use std::uint8_t instead of uint8_t

lib/everest/iso15118/src/iso15118/d20/state/ac_der_iec_charge_loop.cpp
```diff
@@ -0,0 +1,313 @@
...
+
+    if (not validate_and_setup_header(res.header, session, req.header.session_id)) {
+        return response_with_code(res, dt::ResponseCode::FAILED_UnknownSession);
+    }
+
+    const auto& selected_services = session.get_selected_services();
+    const auto selected_control_mode = selected_services.selected_control_mode;
+    const auto selected_mobility_needs_mode = selected_services.selected_mobility_needs_mode;
+
+    // TODO(SL): How to handle grid_event_condition?
```
mlitre, line 154: Probably as a feedback / callback

lib/everest/iso15118/src/iso15118/tbd_controller.cpp
```diff
@@ -159,6 +158,26 @@ void TbdController::set_dlink_ready(bool ready) {
     }
 }
 
+// TODO(SL): For both functions the update does only work before a session is started.
+// During a session, the updated der functions will be considered in the next session
```
mlitre, line 183: This is kind of a big issue, we could get changed directives for more urgent issues

mlitre (changes requested):

docs/source/reference/EVerest_API/auth_token_validator_API.yaml
SebaLukas: Thats the big question

modules/EVSE/Evse15118D20/charger/ISO15118_chargerImpl.cpp
```diff
@@ -926,9 +928,12 @@ void ISO15118_chargerImpl::handle_update_energy_transfer_modes(
...
             services.push_back(dt::ServiceCategory::AC_BPT);
             break;
-        case types::iso15118::EnergyTransferMode::AC_DER:
-            services.push_back(dt::ServiceCategory::AC_DER);
+        case types::iso15118::EnergyTransferMode::AC_DER_IEC:
+            services.push_back(dt::ServiceCategory::AC_DER_IEC);
             break;
+        case types::iso15118::EnergyTransferMode::AC_DER_SAE:
+            services.push_back(dt::ServiceCategory::AC_DER_SAE);
+            return;
```
mlitre:
```suggestion
            break;
```

lib/everest/iso15118/src/iso15118/d20/state/ac_der_iec_charge_loop.cpp
```diff
@@ -0,0 +1,313 @@
...
+        auto& res_mode = res.control_mode.emplace<Dynamic_DER_Res>();
+        fill(res_mode, target_powers, present_powers, ac_limits, der_limits.value(), dso_q_setpoint,
+             dso_cos_phi_setpoint);
+
+        if (selected_mobility_needs_mode == dt::MobilityNeedsMode::ProvidedBySecc) {
+            set_dynamic_parameters_in_res(res_mode, dynamic_parameters, res.header.timestamp);
+        }
+    }
+
+    res.target_frequency = dt::from_float(target_frequency);
```
mlitre: This violates V2G20-3147, we should only send this if target != nominal

lib/everest/iso15118/src/iso15118/d20/state/ac_der_iec_charge_loop.cpp
```diff
@@ -0,0 +1,313 @@
...
+
+    // TODO(sl): Setting EvseStatus, MeterInfo, Receipt
+
+    if (stop) {
+        res.status = {0, dt::EvseNotification::Terminate};
+    } else if (pause) {
+        const uint16_t notification_max_delay =
+            (selected_control_mode == dt::ControlMode::Dynamic) ? 60 : 0; // [V2G20-1850]
```
mlitre: This contradicts V2G20-3308, notification max delay shouldn't be 0

lib/everest/iso15118/src/iso15118/d20/state/ac_der_iec_charge_loop.cpp
```diff
@@ -0,0 +1,313 @@
...
+        const auto res = handle_request(*req, m_ctx.session, false);
+
+        m_ctx.respond(res);
+
+        if (res.response_code >= dt::ResponseCode::FAILED) {
+            m_ctx.session_stopped = true;
+            return {};
+        }
+
+        // V2G20-1623 -> state machine direct transition (skipped PowerDelivery)
```
mlitre: We can replace this citation of the requirement as the amd 1 replaces it: V2G20-3210

lib/everest/iso15118/src/iso15118/d20/state/ac_der_iec_charge_parameter_discovery.cpp
```diff
@@ -0,0 +1,374 @@
...
+                           m_ctx.session_config.der_limits, operating_mode, grid_connection_mode, der_control);
+
+        m_ctx.respond(res);
+
+        if (res.response_code >= message_20::datatypes::ResponseCode::FAILED) {
+            m_ctx.session_stopped = true;
+            return {};
+        }
+
+        // TODO(SL): Check [V2G20-3154]
```
mlitre: EVCC requirement we just need to allow coming back to service discovery (later once renegotiation is allowed)

lib/everest/iso15118/include/iso15118/d20/config.hpp
mlitre: Going through it again it seems like we can just derive the bitset from `supported_der_control_functions` so this bitset seems redundant

lib/everest/iso15118/src/iso15118/d20/state/ac_der_iec_charge_loop.cpp
```diff
@@ -0,0 +1,313 @@
...
+void set_dynamic_parameters_in_res(T& res_mode, const UpdateDynamicModeParameters& parameters,
+                                   uint64_t header_timestamp) {
+    if (parameters.departure_time) {
+        const auto departure_time = static_cast<uint64_t>(parameters.departure_time.value());
+        if (departure_time > header_timestamp) {
+            res_mode.departure_time = static_cast<uint32_t>(departure_time - header_timestamp);
+        }
+    }
+    res_mode.target_soc = parameters.target_soc;
+    res_mode.minimum_soc = parameters.min_soc;
```
mlitre: Probably need to check for V2G20-3155 simple `?` operator would suffice

lib/everest/iso15118/include/iso15118/d20/config.hpp
SebaLukas: Change will be done via `TbdController::update_supported_der_functions()`. Changing during a session I will answer in the other conversation

lib/everest/iso15118/include/iso15118/d20/config.hpp
SebaLukas: My bad. I changed the bitset to 12. I really like the bitset. With the `DERControlName` enum it is really easy to define the supported der functions

SebaLukas: > Going through it again it seems like we can just derive the bitset from `supported_der_control_functions` so this bitset seems redundant

I dont get your sentence. In this function `control_functions` is being used

lib/everest/iso15118/include/iso15118/d20/der_functions.hpp
SebaLukas: Done

lib/everest/iso15118/include/iso15118/d20/der_functions.hpp
SebaLukas: done

lib/everest/iso15118/src/iso15118/d20/state/ac_der_iec_charge_loop.cpp
SebaLukas: Yeah thats a todo after this PR

lib/everest/iso15118/src/iso15118/tbd_controller.cpp
SebaLukas: During an active session I see only one possibility to change the supported der functions via service renegotation. But right we dont have this feature. Because in the ServiceDetailReq/Res the car is selecting the service which has the supported der functions. After that selecting is done the charger or the ev can not change the der functions so easily.

lib/everest/iso15118/src/iso15118/d20/state/ac_der_iec_charge_loop.cpp
SebaLukas: Thanks for checking that. For D20 [V2G20-1555] applies here as well

lib/everest/iso15118/src/iso15118/d20/state/ac_der_iec_charge_parameter_discovery.cpp
SebaLukas: Yeah thats for later after the PR

lib/everest/iso15118/src/iso15118/d20/state/ac_der_iec_charge_loop.cpp
SebaLukas: Thanks for finding that. I changed that to 60 seconds for every charge loop state.
There is another req from amd1 [V2G20-3318] we need to have another look later.

lib/everest/iso15118/src/iso15118/d20/state/ac_der_iec_charge_loop.cpp
SebaLukas: :+1:

lib/everest/iso15118/src/iso15118/d20/state/ac_der_iec_charge_loop.cpp
SebaLukas: Thanks. There were the same req for ac and dc charge loop. So I changed the states as well

lib/everest/iso15118/include/iso15118/d20/config.hpp
mlitre: Basically we already have other structs that track the control functions so we don't really need to save an extra bitset since we could just construct it when we need it.

docs/source/reference/EVerest_API/auth_token_validator_API.yaml
mlitre: Maybe we should have a combinatorial with like if you support DER then you can do AC DER if you also support BPT then you can do AC_BPT_DER?

lib/everest/iso15118/src/iso15118/d20/state/ac_charge_loop.cpp
```diff
@@ -75,16 +75,20 @@ void set_dynamic_parameters_in_res(T& res_mode, const UpdateDynamicModeParameter
         }
     }
     res_mode.target_soc = parameters.target_soc;
-    res_mode.minimum_soc = parameters.min_soc;
+
+    // [V2G20-1366]
+    if (parameters.min_soc.has_value() and parameters.target_soc.has_value() and
+        parameters.min_soc.value() <= parameters.target_soc.value()) {
+        res_mode.minimum_soc = parameters.min_soc;
```
mlitre, line 82: If min_soc has value then we should either use its value or clamp it to the target_soc instead of just dropping it. Applies to the other parts

lib/everest/ocpp/include/ocpp/v2/ocpp_enums.hpp
mlitre, line 1: This applies to all ocpp files, but these should remain unchanged because they are part of the OCPP spec and changing it would violate that.

modules/EVSE/OCPP201/device_model/definitions.cpp
```diff
@@ -130,8 +130,8 @@ const VariableCharacteristics SupportedEnergyTransferModes = [] {
     VariableCharacteristics var;
     var.dataType = DataEnum::MemberList;
     var.supportsMonitoring = false;
-    var.valuesList =
-        "AC_single_phase,AC_two_phase,AC_three_phase,DC,AC_BPT,AC_BPT_DER,AC_DER,DC_BPT,DC_ACDP,DC_ACDP_BPT,WPT";
+    var.valuesList = "AC_single_phase,AC_two_phase,AC_three_phase,DC,AC_BPT,AC_BPT_DER,AC_DER_IEC,AC_DER_SAE,DC_BPT,DC_"
```
mlitre: Same here

modules/EVSE/OCPP201/conversions.cpp
mlitre, line 1: I guess the real question is how do we do this conversion here...

lib/everest/iso15118/src/iso15118/d20/state/ac_der_iec_charge_parameter_discovery.cpp
```diff
@@ -0,0 +1,375 @@
...
+
+        case iec::DERControlName::WattCosPhiMode:
+            if (not std::holds_alternative<iec::DERCurve>(function)) {
+                logf_warning("Unexpected variant type for WattCosPhiMode");
+                break;
+            }
+            {
+                auto& reactive_power_support = get_or_emplace(control.reactive_power_support);
+                reactive_power_support.name = dt::ReactivePowerSupport::ReactivePowerSupportName::WattCosPhi;
+                convert(reactive_power_support.curve, std::get<iec::DERCurve>(function));
```
mlitre, line 191: These all use the same control.reactive_power_support is that normal? This seems like only the last der control function would be set correctly

lib/everest/iso15118/src/iso15118/d20/state/ac_der_iec_charge_loop.cpp
mlitre, line 1: Probably for a future follow up PR, but we should look into how we could just inject the DER behavior into the base charge_loop state as they are like 90% identical. This would avoid some duplication and potential issues in the future

lib/everest/iso15118/include/iso15118/d20/config.hpp
```diff
@@ -27,6 +30,12 @@ struct BptSetupConfig {
     std::optional<message_20::datatypes::GridCodeIslandingDetectionMethod> grid_code_detection_method;
 };
 
+struct DerSetupConfig {
+    std::map<iec::DERControlName, iec::DERControlFunction> supported_der_control_functions;
```
mlitre, line 34: I am wondering if it is even worth using a map or just using the struct with optionals? This would avoid having to do `std::holds_alternative` guards in `create_der_control`.

lib/everest/iso15118/src/iso15118/d20/state/ac_der_iec_charge_parameter_discovery.cpp
SebaLukas: Yeah thats normal. Only one ReactivePower should be supported (VoltVar, WattVar or WattCosPhi). In the xsd its a choice.

lib/everest/iso15118/src/iso15118/d20/state/ac_der_iec_charge_loop.cpp
SebaLukas: Yeah thats a good point. I thought that for myself. Right now it was easier to copy the existing charge loop behaivour in another state. In the future we only have one dc and ac charge loop file with the different states or two charge loop states (ac and dc). But that needs additional guards and checks if the correct messages are be sent

lib/everest/iso15118/include/iso15118/d20/config.hpp
SebaLukas: Right now I like it. But we can change that in the future if its difficult to maintain

lib/everest/ocpp/include/ocpp/v2/ocpp_enums.hpp
SebaLukas: Done

modules/EVSE/OCPP201/conversions.cpp
SebaLukas: I tried something :sweat_smile:

mlitre: I think that's fine for now

mlitre (dismissed):
Lgtm

mlitre (dismissed):
lgtm

lib/everest/iso15118/include/iso15118/d20/state/ac_der_iec_charge_loop.hpp
```diff
@@ -0,0 +1,31 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2025 Pionix GmbH and Contributors to EVerest
```
hikinggrass, line 2:
```suggestion
// Copyright 2026 Pionix GmbH and Contributors to EVerest
```

lib/everest/iso15118/include/iso15118/d20/config.hpp
```diff
@@ -2,10 +2,13 @@
 // Copyright 2023 Pionix GmbH and Contributors to EVerest
```
hikinggrass, line 2:
```suggestion
// Copyright 2023-2026 Pionix GmbH and Contributors to EVerest
```

lib/everest/iso15118/include/iso15118/message/common_types.hpp
hikinggrass, line 2:
```suggestion
// Copyright 2023-2026 Pionix GmbH and Contributors to EVerest
```

lib/everest/iso15118/src/iso15118/d20/state/power_delivery.cpp
```diff
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 // Copyright 2023 Pionix GmbH and Contributors to EVerest
```
hikinggrass, line 2:
```suggestion
// Copyright 2023-2026 Pionix GmbH and Contributors to EVerest
```

lib/everest/iso15118/src/iso15118/d20/state/service_selection.cpp
```diff
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 // Copyright 2023 Pionix GmbH and Contributors to EVerest
```
hikinggrass, line 2:
```suggestion
// Copyright 2023-2026 Pionix GmbH and Contributors to EVerest
```

lib/everest/iso15118/src/iso15118/d20/state/session_setup.cpp
hikinggrass, line 2:
```suggestion
// Copyright 2023-2026 Pionix GmbH and Contributors to EVerest
```

hikinggrass (dismissed):
just a couple nits concerning license header copyright dates. can be addressed in a future PR though (there are very likely more anyway)

docs/source/reference/EVerest_API/auth_token_validator_API.yaml
cburandt: Maybe add to the description that "AC_DER" means the same as "AC_DER_IEC"?

cburandt (dismissed):
This change is no longer a breaking change - LGTM

It would be nice, though, if the comment (see below) were added.

docs/source/reference/EVerest_API/auth_token_validator_API.yaml
SebaLukas: Done

lib/everest/iso15118/include/iso15118/d20/state/ac_der_iec_charge_loop.hpp
SebaLukas: Will do that in a later PR :+1:

hikinggrass (dismissed):

hikinggrass (approved):

mlitre (approved):

modules/API/RpcApi/types/json_rpc_api/json_rpc_api.hpp
```diff
@@ -546,6 +552,12 @@ inline EnergyTransferModeEnum string_to_energy_transfer_mode_enum(const std::str
     if (s == "AC_DER") {
         return EnergyTransferModeEnum::AC_DER;
     }
+    if (s == "AC_DER_IEC") {
+        return EnergyTransferModeEnum::AC_DER_IEC;
+    }
+    if (s == "AC_DER_SAE") {
+        return EnergyTransferModeEnum::AC_DER_SAE;
+    }
```
barsnick, line 560: I see you didn't create the code using the code generation feature, but this is fine for now.

(The generator and its templates were refactored, I'll need to check that myself.)

barsnick (approved):
Probably okay from RpcApi view.

PR #2361, opened 2026-06-23
feat(CarloGavazzi_EM580): reduce log verbosity

lategoodbye:
## Describe your changes

Currently the module is too chatty for production. So drop or reduce the log level for messages which are not relevant for users.
I don't have a strong opinion about the individual changes, but in sum the log messages are too many and sometimes only relevant for developer.

### DIFF
--- a/modules/HardwareDrivers/PowerMeters/CarloGavazzi_EM580/main/powermeterImpl.cpp
+++ b/modules/HardwareDrivers/PowerMeters/CarloGavazzi_EM580/main/powermeterImpl.cpp
@@ -235,7 +235,7 @@ void powermeterImpl::init() {
 }
 
 void powermeterImpl::read_signature_config() {
-    EVLOG_info << "Read the signature public key...";
+    EVLOG_debug << "Read the signature public key...";
 
     enum SignatureType {
         SIGNATURE_256_BIT,
@@ -323,7 +323,7 @@ void powermeterImpl::read_identification() {
 }
 
 void powermeterImpl::read_firmware_versions() {
-    EVLOG_info << "Read the firmware versions...";
+    EVLOG_debug << "Read the firmware versions...";
 
     // Read measure module firmware version/revision (register 300771)
     transport::DataVector measure_fw_data = p_modbus_transport->fetch(MODBUS_FIRMWARE_MEASURE_MODULE_ADDRESS, 1);
@@ -355,7 +355,8 @@ void powermeterImpl::read_firmware_versions() {
 }
 
 void powermeterImpl::read_serial_number() {
-    EVLOG_info << "Read the serial number...";
+    EVLOG_debug << "Read the serial number...";
+
     // Read serial number (registers 320481-320487, 7 UINT16 registers = 14 bytes)
     transport::DataVector serial_data =
         p_modbus_transport->fetch(MODBUS_SERIAL_NUMBER_START_ADDRESS, MODBUS_SERIAL_NUMBER_REGISTER_COUNT);
@@ -435,7 +436,7 @@ void powermeterImpl::read_transaction_state_and_id() {
 }
 
 void powermeterImpl::configure_device() {
-    EVLOG_info << "Configure the device...";
+    EVLOG_debug << "Configure the device...";
     read_identification();
     read_firmware_versions();
     read_serial_number();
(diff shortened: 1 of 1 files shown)


modules/HardwareDrivers/PowerMeters/CarloGavazzi_EM580/main/powermeterImpl.cpp
```diff
@@ -584,8 +578,7 @@ void powermeterImpl::clear_transaction_states() {
     std::uint16_t ocmf_state = modbus_utils::to_uint16(state_data, modbus_utils::ByteOffset{0});
 
     if (ocmf_state == MODBUS_OCMF_STATE_READY) {
-        EVLOG_info << "Current OCMF state: " << ocmf_state_to_string(ocmf_state) << "(" << ocmf_state << ")";
-        EVLOG_info << "Cleanup necessary ...";
```
mhei, line 588: I agree, that two messages are not necessary here.

modules/HardwareDrivers/PowerMeters/CarloGavazzi_EM580/main/powermeterImpl.cpp
```diff
@@ -685,8 +678,7 @@ types::powermeter::TransactionStopResponse powermeterImpl::handle_stop_transacti
             // id since there is a pending closed transaction, I assume a power loss
             // occurred we need to check if the transaction id is equal to the
             // transaction id from OCMF file
-            EVLOG_info << "Power loss occurred, checking if the transaction id == "
-                          "transaction id from OCMF file";
```
mhei, line 689: Converting to debug fine with me, but I'd keep the original text.

modules/HardwareDrivers/PowerMeters/CarloGavazzi_EM580/main/powermeterImpl.cpp
```diff
@@ -981,8 +973,6 @@ void powermeterImpl::synchronize_time() {
 }
 
 void powermeterImpl::set_timezone(int offset_minutes) {
-    EVLOG_info << "Try to set the timezone ... ";
-
```
mhei, line 985: I think from regulatory perspective, it should be obvious whether setting the timezone succeeded or failed. With this message removed, I'm unsure whether it is still so transparent as it should be. I assume that `write_multiple_registers` raises an exception in case of failure? And is it verbose then already, or should be log it then here?

mhei (changes requested):
Except for the lines with dedicated comment, I would recommend to just switch all touched lines to DEBUG.

modules/HardwareDrivers/PowerMeters/CarloGavazzi_EM580/main/powermeterImpl.cpp
lategoodbye: So you suggest to remove both?

modules/HardwareDrivers/PowerMeters/CarloGavazzi_EM580/main/powermeterImpl.cpp
lategoodbye: I think it's worth to mention that a powerloss occurred during charging before, but I don't like that the log explains the code.

Maybe split the first part into info and the second part into debug?

modules/HardwareDrivers/PowerMeters/CarloGavazzi_EM580/main/powermeterImpl.cpp
lategoodbye: Since we already log the good case, I prefer to log the error case separately. In case `write_multiple_registers` doesn't raise a proper exception, we have a general problem.

So we should catch the exception, log the error case and throw it again?

modules/HardwareDrivers/PowerMeters/CarloGavazzi_EM580/main/powermeterImpl.cpp
mhei: keep first, but convert to debug, drop the second would be my favorite; no strong opinion though

modules/HardwareDrivers/PowerMeters/CarloGavazzi_EM580/main/powermeterImpl.cpp
mhei: the log should explain from the customer perspective what's happening and why - without that the customer knows the code. not sure whether splitting would help here then

modules/HardwareDrivers/PowerMeters/CarloGavazzi_EM580/main/powermeterImpl.cpp
mhei: I checked this, and yes, exceptions are raised, the call chain already handles exceptions in a generic way and also relies on this. So yes, catching it, log a proper error so that the context is visible and re-throw it would make sense.

modules/HardwareDrivers/PowerMeters/CarloGavazzi_EM580/main/powermeterImpl.cpp
lategoodbye: I kept the second as debug for context and consistency.

modules/HardwareDrivers/PowerMeters/CarloGavazzi_EM580/main/powermeterImpl.cpp
lategoodbye: Since after a powerloss comparing the transcation id is just a temporary step, I will keep this and the following log message.

mhei (approved):

Pietfried (approved):

PR #2368, opened 2026-06-24
Update EvseManager fault handling

djchhp:
## Describe your changes

  Updates EvseManager error handling so SLAC and HLC requirement errors can make the EVSE inoperative when they represent charging-relevant faults.

### Details

  - Subscribes EvseManager error handling to SLAC requirement errors.
  - Subscribes EvseManager error handling to HLC requirement errors.
  - Treats SLAC communication and vendor errors as fatal for charging.
  - Treats HLC communication and vendor errors as fatal for charging.
  - Keeps vendor warnings ignored for SLAC and HLC, matching the existing ignore-list pattern used for other requirements.
  - Makes session log path setup handle filesystem errors without terminating the process.
  - Adds test coverage for SLAC/HLC fatal and ignored error behavior.
  - Adds test coverage for session log filesystem failure handling.

### Related PRs
#2369 Raises comm fault error for EvseV2G
#1935 Raises comm fault for SLAC

## Checklist before requesting a review
- [x] I have performed a self-review of my code
- [ ] I have made corresponding changes to the documentation
- [x] I read the [contribution documentation](https://everest.github.io/nightly/project/contributing.html) and made sure that my changes meet its requirements

PR #2369, opened 2026-06-24
Improve EvseV2G startup procedure

djchhp:
### Summary
 Makes EvseV2G tolerate unavailable PLC/network interfaces during startup by retrying network setup instead of aborting module initialization. This fixes a start up race for certain architectures.
 ### Details
  - Replaces fatal startup failure with a retrying network lifecycle.
  - Reports network startup failures through `generic/CommunicationFault`.
  - Logs and raises repeated startup failures only when the failure reason changes.
  - Clears the startup communication fault once network setup succeeds.
  - Preserves the configured device name across retries, including `auto`.
  - Moves retry handling from low-level connection setup into the module startup lifecycle.
  - Cleans up partially initialized TCP, TLS, and SDP state before retrying.
  - Allows the ISO15118 charger interface to raise generic errors.

### Related PRs
#2368    This PR implements the changes that make the EvseManager aware of Fault status.

### DIFF
--- a/modules/EVSE/EvseV2G/EvseV2G.cpp
+++ b/modules/EVSE/EvseV2G/EvseV2G.cpp
@@ -9,8 +9,10 @@
 #include <everest/logging.hpp>
 
 #include <csignal>
+#include <cstdlib>
 #include <everest/tls/openssl_util.hpp>
 #include <stdexcept>
+#include <unistd.h>
 namespace {
 void log_handler(openssl::log_level_t level, const std::string& str) {
     switch (level) {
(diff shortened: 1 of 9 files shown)


modules/EVSE/EvseV2G/EvseV2G.cpp
```diff
@@ -79,46 +81,113 @@ void EvseV2G::init() {
...
+        sleep(1);
+    }
+}
+
+void EvseV2G::report_network_startup_failure(const std::string& reason) {
+    const auto message = "Device " + config.device + ": " + reason;
+
+    if (startup_fault_message == message) {
+        if (!startup_fault_raised && p_charger->error_factory && p_charger->error_manager) {
+            p_charger->raise_error(p_charger->error_factory->create_error("generic/CommunicationFault", "", message));
```
FaHaGit, line 177: If I'm understanding this correctly, the error that was raised could slip through because it happens before `invoke_ready` is called. Am I right about that? Is this a critical issue?

FaHaGit: Second minor point: For device: auto, the fault/recovery message uses config.device, so the user will see Device auto instead of the actual selected interface. It might be useful to include the resolved v2g_ctx->if_name as soon as interface selection succeeded.

djchhp: I have updated the code to print the actual interface name in case of `auto`, when it has already been resolved.

djchhp: > If I'm understanding this correctly, the error that was raised could slip through because it happens before `invoke_ready` is called. Am I right about that? Is this a critical issue?

I think what is done here is safe.
* `error_manger` and `error_factory` are populated in the constructor of `ImplBase`, long before `init`/`ready`.
* The modules `ready` only run's after the frameworks global ready signal, when all modules have finished `init`. They have all subscribed to the relevant MQTT topics at that stage.
* `invoke_ready` here only triggers the chargers own `ready`, which is independent of the error infrastructure, so raising here is no problem.

barsnick (dismissed):
Giving an approval after update, also on behalt of @FaHaGit .

Untested.

modules/EVSE/EvseV2G/EvseV2G.cpp
```diff
@@ -83,48 +85,127 @@ void EvseV2G::init() {
...
+                    dlog(DLOG_LEVEL_ERROR, "sdp_listen() failed");
+                }
+            } else {
+                wait_for_shutdown();
+            }
+            return;
         }
+
+        if (!v2g_ctx->shutdown) {
+            sleep(1);
```
SebaLukas: Normally we use `std::this_thread::wait_for(1s)`. But I dont have a preference. Both methods block the current thread.

SebaLukas (commented):
I found one small thing, except that looking good!

modules/EVSE/EvseV2G/EvseV2G.cpp
```diff
@@ -83,48 +85,127 @@ void EvseV2G::init() {
...
+        v2g_ctx->sdp_socket = -1;
+    }
+}
 
-    if (rv == -1) {
-        dlog(DLOG_LEVEL_ERROR, "sdp_listen() failed");
-        goto err_out;
+void EvseV2G::wait_for_shutdown() {
+    while (!v2g_ctx->shutdown) {
+        sleep(1);
```
SebaLukas: Here is also a sleep 😅

SebaLukas (dismissed):

modules/EVSE/EvseV2G/EvseV2G.cpp
djchhp: fixed

SebaLukas (approved):

Issue #2197, opened 2026-05-13
Certificate Trust List Usage


xandergot:
### Describe the problem

To enable EVerest to consume the new SAE EVPKI Certificate Trust List mechanism as part of EV charging trust management. SAE ITC positions the EVPKI CTL as the basis for onboarding PKI providers and creating an interoperable PKI solution for the EV ecosystem. Supporting CTLs in EVerest therefore helps align the project with emerging EVPKI trust-distribution practices and improves the ability to work across multiple approved roots within the charging ecosystem, rather than relying only on manually managed individual trust anchors. 

### EVerest Domain

ISO 15118-2, ISO 15118-20, Security

### Affected Component

lib/everest/evse_security 

### Describe your solution

Functionality has been added to the trust-store construction path used by EVerest so that it can load trusted roots from SAE-style EV CTLs in addition to loose root certificates. A configured directory is scanned for CTL files and loose root certificates. CTL contents are parsed with OpenSSL, the individual certificates contained in each CTL are extracted, and those certificates are then added to the trust store using the module’s existing trusted-root handling path. Loose PEM or DER roots in the same directory are also loaded through that same trust-store path. The CTL-loading logic is intended for the SAE EV charging CTL format and does not target Microsoft’s proprietary CTL format. The directory location is configurable at build time.

PR #2371, opened 2026-06-24
CTL-verification v2

xandergot:
## Describe your changes
Functionality has been added to the trust-store construction path used by EVerest so that it can load trusted roots from SAE-style EV CTLs in addition to loose root certificates. A configured directory is scanned for CTL files and loose root certificates. CTL contents are parsed with OpenSSL, the individual certificates contained in each CTL are extracted, and those certificates are then added to the trust store using the module’s existing trusted-root handling path. Loose PEM or DER roots in the same directory are also loaded through that same trust-store path. The CTL-loading logic is intended for the SAE EV charging CTL format and does not target Microsoft’s proprietary CTL format. The directory location is configurable at build time.

This new pull request has been opened at the request of  [SebaLukas](https://github.com/SebaLukas)

https://github.com/EVerest/EVerest/issues/2197


barsnick:
The PR is cluttered with many style changes I do not understand.

I see no need to go from
```
    if (condition) {
        one_line()
    }
```
to
```
    if (condition)
        one_line()
```
and it does not conform to the coding style.

Neither do I understand for example why this use of initializer list is dropped:
```diff
   X509CertificateBundle::X509CertificateBundle(const fs::path& path, const EncodingFormat encoding) :
-      hierarchy_invalidated(true), path(path) {
+      hierarchy_invalidated(true) {
+      this->path = path;
```

There are many other such (seemingly) unnecessary changes which make the diff much larger than required.

SebaLukas:
In the PR there are alot of changes that I think are not necessary. Changing C++ style to c (for example changing namespace to typedef or std::array changed to c array [])

We have a C++ coding guideline, that could be a good starting point to apply the guidelines on both PRs https://everest.github.io/nightly/how-to-guides/c%2B%2B-coding-guidelines.html

xandergot:
Removed as many of the unneeded changes as I could find, both in this and the other PR, but please let me know if you believe it needs another pass. I believe the issue arose in large part from the refactor at the end of the last year, in addition to me not being as careful with avoiding unneeded reformatting as I should have been.

I also realize that the version information is not correct, what is the projects convention for version naming?

SebaLukas:
@xandergot Thank you for updating both PRs. Now its very clear what are the actual changes. We're now trying to figure out internally who is best suited to review both PRs.

lib/everest/evse_security/lib/evse_security/utils/load_ctl.cpp
```diff
@@ -0,0 +1,168 @@
...
+#include <openssl/pkcs7.h>
+#include <openssl/x509.h>
+#include <vector>
+
+namespace fs = std::filesystem;
+namespace evse_security {
+void load_ctl(const fs::path& cert_dir, const std::map<CaCertificateType, fs::path>& ca_bundle_path_map) {
+
+    // DC to CA type mapping
+    static const std::map<std::string, CaCertificateType> dc_to_ca_type = {{"V2G", CaCertificateType::V2G},
```
AssemblyJohn, line 22: Could here be used the X509 bundle for easier loading handling?

lib/everest/evse_security/lib/evse_security/utils/load_ctl.cpp
```diff
@@ -0,0 +1,168 @@
...
+                                                                           {"CSMS", CaCertificateType::CSMS},
+                                                                           {"MO", CaCertificateType::MO},
+                                                                           {"MF", CaCertificateType::MF}};
+
+    auto getDC = [](X509_NAME* name) -> std::string {
+        int idx = X509_NAME_get_index_by_NID(name, NID_domainComponent, 0);
+        if (idx < 0)
+            return {};
+        ASN1_STRING* val = X509_NAME_ENTRY_get_data(X509_NAME_get_entry(name, idx));
```
AssemblyJohn, line 31: Could be moved as utilities in the X509Wrapper, that's why we have it there.

lib/everest/evse_security/lib/evse_security/utils/load_ctl.cpp
```diff
@@ -0,0 +1,168 @@
...
+        if (!f) {
+            EVLOG_error << "Failed to open bundle file: " << bundle_path;
+            continue;
+        }
+        bundle_files[ca_type] = f;
+        bundle_bios[ca_type] = BIO_new_fp(f, BIO_NOCLOSE);
+    }
+
+    // Iterate CTL directory
+    for (const auto& entry : fs::directory_iterator(cert_dir)) {
```
AssemblyJohn, line 49: X509Bundle does this iteration automatically.

lib/everest/evse_security/lib/evse_security/utils/load_ctl.cpp
```diff
@@ -0,0 +1,168 @@
...
+        BIO* bio = BIO_new_file(file_path.c_str(), "rb");
+        if (!bio) {
+            EVLOG_error << "Error opening file: " << file_path;
+            continue;
+        }
+
+        std::vector<X509*> extracted_certs;
+
+        PKCS7* pkcs7 = nullptr;
+        if (ext == ".der") {
```
AssemblyJohn, line 62: If .der support is missing should be added to the X509Bundle.

lib/everest/evse_security/lib/evse_security/evse_security.cpp
```diff
@@ -331,6 +332,15 @@ EvseSecurity::EvseSecurity(const FilePaths& file_paths, const std::optional<std:
         }
     }
 
+    fs::path ctl_dir =
```
AssemblyJohn, line 335: Should make sure the ctl_dir default is empty, since we don't want to break existing functionality.

lib/everest/evse_security/lib/evse_security/utils/load_ctl.cpp
xandergot: I don't full understand what is being requested here. This entire section is just .der handling. Is this in reference to the above comment of moving the CTL loading functionality into the X509 wrapper? Or do we want better error handling?

lib/everest/evse_security/lib/evse_security/evse_security.cpp
xandergot: Do we want this a a conditional check for when the module isn't loaded? I'm not sure how the module is supposed to function if the directory is empty. This directory is populated at build time, the the extension being enabled changes whether the directory is ever used, not whether it is populated.

lib/everest/evse_security/lib/evse_security/utils/load_ctl.cpp
xandergot: For clarification, is the suggestion to move the bulk of the CTL loading code into the X509Wrapper as a utility? If so, I would appreciate elaboration on why this change would be preferable so that I know what should be moved and what should be kept separate.

lib/everest/evse_security/lib/evse_security/utils/load_ctl.cpp
xandergot: see below

lib/everest/evse_security/lib/evse_security/utils/load_ctl.cpp
xandergot: I am unclear on how this would be leveraged. Are you saying that if We move the code as addressed above this iteration would be unnecessary? Currently this is just how we find the certificates and Certificate Trust Lists in the directory and allow the code to access it.

lib/everest/evse_security/lib/evse_security/utils/load_ctl.cpp
AssemblyJohn: What i've meant is to do something like:  X509CertificateBundle existing_certs(ca_bundle_path, EncodingFormat::PEM);

That parses the whole directory for certificates and returns lists of X509Wrappers. That's what I'd use instead of manually  fopen and reading the certs into X509s. 

See: https://github.com/EVerest/EVerest/blob/ebcd36d976526b661b306241b2aad397bfeb5db2/lib/everest/evse_security/lib/evse_security/evse_security.cpp#L377

lib/everest/evse_security/lib/evse_security/utils/load_ctl.cpp
```diff
@@ -0,0 +1,168 @@
...
+                BIO_seek(bio, 0);
+                pkcs7 = PEM_read_bio_PKCS7(bio, nullptr, nullptr, nullptr);
+            }
+        } else if (ext == ".ctl") {
+            pkcs7 = d2i_PKCS7_bio(bio, nullptr);
+            if (!pkcs7) {
+                BIO_seek(bio, 0);
+                pkcs7 = PEM_read_bio_PKCS7(bio, nullptr, nullptr, nullptr);
+            }
+        } else if (ext == ".p7b" || ext == ".p7c") {
```
AssemblyJohn, line 75: If this is required, as I've said, it's better to be added to the X509CertificateBundle class:

https://github.com/EVerest/EVerest/blob/ebcd36d976526b661b306241b2aad397bfeb5db2/lib/everest/evse_security/lib/evse_security/evse_security.cpp#L377

lib/everest/evse_security/lib/evse_security/utils/load_ctl.cpp
```diff
@@ -0,0 +1,168 @@
...
+            // Uppercase for case-insensitive comparison
+            std::transform(dc.begin(), dc.end(), dc.begin(), ::toupper);
+
+            auto it = dc_to_ca_type.find(dc);
+            if (it != dc_to_ca_type.end()) {
+                auto bio_it = bundle_bios.find(it->second);
+                if (bio_it != bundle_bios.end()) {
+                    EVLOG_info << "Adding cert with DC=" << dc << " to "
+                               << conversions::ca_certificate_type_to_string(it->second) << " bundle";
+                    PEM_write_bio_X509(bio_it->second, cert);
```
AssemblyJohn, line 145: I need a bit of explanation here, what are we doing with the certificates after we're writing them to this bio file?

lib/everest/evse_security/lib/evse_security/evse_security.cpp
AssemblyJohn: If this is meant to be given as a compile-time definition and not a 'compile_if' it should be removed. The value will be received via the standard everest mechanism of configuration. If it is empty it should be checked in the constructor and NOT be used.

See: https://github.com/EVerest/EVerest/blob/ebcd36d976526b661b306241b2aad397bfeb5db2/lib/everest/evse_security/lib/evse_security/evse_security.cpp#L286

And: https://github.com/xandergot/EVerest/blob/f4b36ea436aae31c4793330646c97fe299be4dc0/modules/EVSE/EvseSecurity/manifest.yaml

For example here: 
https://github.com/EVerest/EVerest/blob/main/config/config-sil-dc.yaml

that value could be theoretically over-written. If it is empty, then the code should not execute. We don't need conditional compilation.

lib/everest/evse_security/lib/evse_security/utils/load_ctl.cpp
xandergot: We are passing the certificates to Everest’s preexisting certificate management system. The BIO usage is how this is accomplished. This framing allows EVerest's preexisting certificate management system to modify the certificates using its already instated functions. Additionally integrating it will cause the certificates to be centralized, prevent any potential issues from having two parallel certificate management systems. The point is to process the CTL and properly sort the certificates, not necessarily determine what is then done with them.

lib/everest/evse_security/lib/evse_security/utils/load_ctl.cpp
xandergot: I agree, I will move PKCS#7/P7B/P7C handling into X509CertificateBundle as a supported input format/helper, assuming maintainers agree with that class responsibility.

lib/everest/evse_security/lib/evse_security/evse_security.cpp
xandergot: If I’m understanding correctly, I need to: remove compile-time definition, define be Directory in the EvseSecurity manifest, have it be validated in the constructor, and have the constructor skip CTL processing when the Directory is unset or empty.

xandergot:
The code has not yet been modified, as of yet we have only addressed the comments in writing.

lib/everest/evse_security/lib/evse_security/evse_security.cpp
AssemblyJohn: Correct!

lib/everest/evse_security/lib/evse_security/utils/load_ctl.cpp
AssemblyJohn: Yes, it would make sense.

lib/everest/evse_security/lib/evse_security/utils/load_ctl.cpp
AssemblyJohn: See the comment: https://github.com/EVerest/EVerest/pull/2371#discussion_r3807192268

Maybe that approach would make most sense?

lib/everest/evse_security/lib/evse_security/utils/load_ctl.cpp
AssemblyJohn: See: https://github.com/EVerest/EVerest/pull/2371#discussion_r3807192268

If it makes sense we can use `X509CertificateBundle` that can receive a directory path and automatically load everything in it.

lib/everest/evse_security/lib/evse_security/utils/load_ctl.cpp
AssemblyJohn: I am not entirely sure how this is supposed to work, I guess you know better. One thing I'd require is to have proper unit tests for this so that I can also run and try to see where/how those ordered certificates are written and that it doesn't break existing certificate handling (since of existing X509CertificateBundle parsing).

Issue #2196, opened 2026-05-13
Iso15118-2 and Iso15118-20 Requirement Enforcement


xandergot:
### Describe the problem

To improve security and interoperability by ensuring that certificate validation in EVerest checks not only generic X.509 chain validity, but also whether a certificate conforms to the role-specific certificate profile requirements defined by ISO 15118-2 and ISO 15118-20. This is especially relevant for ISO 15118-20, which requires TLS 1.3 and certificate validation against the appropriate certificate profiles, including items such as key usage, certificate extensions, and chain/profile conformance. For EVerest, that means more standards-aligned validation behavior and fewer cases where a certificate is technically parseable but not appropriate for its intended V2G role. 

### EVerest Domain

ISO 15118-2, ISO 15118-20, Security

### Affected Component

lib/everest/evse_security

### Describe your solution

Functionality has been added so that, during runtime certificate validation, EVerest performs profile-based checks in addition to the existing OpenSSL X.509 verification path. Certificate requirements are read from YAML profile definitions organized by standard version (iso15118-2 and iso15118-20). The enforcement logic then evaluates relevant certificate properties, including key usage and related required or prohibited fields, against the selected profile. For each definitive requirement, the corresponding certificate data is read via OpenSSL and compared to the profile rule. Any mismatch is recorded as a validation failure and logged using EVerest’s existing error reporting mechanisms, while successful checks proceed to the next requirement. The profile directory location is configurable at build time and defaults to the security module root. This matches the implementation split already described for the proposed PR.

PR #2372, opened 2026-06-24
profile-verification v2

xandergot:
## Describe your changes
Functionality has been added so that, during runtime certificate validation, EVerest performs profile-based checks in addition to the existing OpenSSL X.509 verification path. Certificate requirements are read from YAML profile definitions organized by standard version (iso15118-2 and iso15118-20). The enforcement logic then evaluates relevant certificate properties, including key usage and related required or prohibited fields, against the selected profile. For each definitive requirement, the corresponding certificate data is read via OpenSSL and compared to the profile rule. Any mismatch is recorded as a validation failure and logged using EVerest’s existing error reporting mechanisms, while successful checks proceed to the next requirement. The profile directory location is configurable at build time and defaults to being in the security module root.

This new pull request has been opened at the request of [SebaLukas](https://github.com/SebaLukas)

https://github.com/EVerest/EVerest/issues/2196


lib/everest/evse_security/include/evse_security/certificate/x509_wrapper.hpp
```diff
@@ -5,7 +5,7 @@
 #include <memory>
 #include <stdexcept>
 #include <string>
-
+#include <openssl/x509.h>
```
AssemblyJohn, line 8: Importing openssl here would break the initial design consideration, that is to not make dependent a header on openssl code. 

If a user imports this header, we don't want it to be dependent on the openssl header.

Should be removed and used through casting if required.

lib/everest/evse_security/include/evse_security/certificate/x509_wrapper.hpp
```diff
@@ -45,7 +45,7 @@ class X509Wrapper {
     inline X509Handle* get() const {
         return x509.get();
     }
-
+    X509* get_x509_raw() const;
```
AssemblyJohn, line 48: See comment above, should be removed and we should use the 'X509HandleOpenSSL' in .cpp files only.

lib/everest/evse_security/include/evse_security/utils/enforce_certificate_rules.hpp
```diff
@@ -0,0 +1,26 @@
+#include <vector>
```
AssemblyJohn: There are way too many dependencies here, must be removed. Only import what you need.

lib/everest/evse_security/lib/evse_security/certificate/x509_wrapper.cpp
```diff
@@ -142,6 +142,11 @@ X509CertificateSource X509Wrapper::get_source() const {
     return X509CertificateSource::STRING;
 }
 
+X509* X509Wrapper::get_x509_raw() const {
+    auto* ssl_handle = dynamic_cast<X509HandleOpenSSL*>(x509.get());
```
AssemblyJohn, line 146: See previous comments. Bad cast here, we can't presume it's always a x5098 handle.

lib/everest/evse_security/lib/evse_security/utils/enforce_certificate_rules.cpp
```diff
@@ -0,0 +1,307 @@
...
+                                         " section=" + section + " content='" + ruleContent + "' error=" + e.what());
+            }
+            if (!line.empty() && line[0] == '-')
+                goto process_rule;
+        }
+    }
+    return rules;
+}
+
+static std::string get_file_name(X509* cert) {
```
AssemblyJohn, line 123: We should not access raw X509 certs, that's why we have handles and wrappers.

lib/everest/evse_security/lib/evse_security/utils/enforce_certificate_rules.cpp
```diff
@@ -0,0 +1,307 @@
...
+    X509_NAME* subjectName = X509_get_subject_name(cert);
```
AssemblyJohn, line 124: Should be added to the x509_wrapper, extra geter should be used. That is why we use the wrapper, to get a layer between raw openssl and accessing a certificate's data.

lib/everest/evse_security/lib/evse_security/utils/enforce_certificate_rules.cpp
```diff
@@ -0,0 +1,307 @@
...
+    X509_NAME* issuerName = X509_get_issuer_name(cert);
+
+    if (!subjectName) {
+        EVLOG_warning << "Could not get subject name from certificate.";
+        return "";
+    }
+
+    // Extract Domain Component (DC)
+    char dcBuf[256];
+    int dcLen = X509_NAME_get_text_by_NID(subjectName, NID_domainComponent, dcBuf, sizeof(dcBuf));
```
AssemblyJohn, line 134: Same as above comment.

lib/everest/evse_security/lib/evse_security/utils/enforce_certificate_rules.cpp
```diff
@@ -0,0 +1,307 @@
...
+    if (dcLen <= 0) {
+        EVLOG_warning << "No Domain Component (DC) found in the certificate.";
+        return "";
+    }
+    std::string dc(dcBuf, dcLen);
+    dc.erase(std::remove(dc.begin(), dc.end(), ' '), dc.end());
+
+    BASIC_CONSTRAINTS* bc_name = (BASIC_CONSTRAINTS*)X509_get_ext_d2i(cert, NID_basic_constraints, nullptr, nullptr);
+    bool isCA = false;
+    bool isSelfSigned = (X509_NAME_cmp(subjectName, issuerName) == 0);
```
AssemblyJohn, line 144: Should use the X509Wrapper utility functions for this.

lib/everest/evse_security/lib/evse_security/utils/enforce_certificate_rules.cpp
```diff
@@ -0,0 +1,307 @@
...
+        if (bc_name->pathlen) {
+            long pathLen = ASN1_INTEGER_get(bc_name->pathlen);
+            certType = "_SubCA" + std::to_string(pathLen);
+        } else {
+            certType = "_SubCA";
+        }
+    }
+
+    if (bc_name) {
+        OPENSSL_free(bc_name);
```
AssemblyJohn, line 164: See above comments, also better to use unique pointers with auto-cleanup.

lib/everest/evse_security/lib/evse_security/utils/enforce_certificate_rules.cpp
```diff
@@ -0,0 +1,307 @@
...
+#include <evse_security/certificate/x509_wrapper.hpp>
+#include <fstream>
+#include <openssl/ssl.h>
+#include <openssl/x509.h>
+#include <openssl/x509v3.h>
+#include <sstream>
+#include <string>
+#include <vector>
+
+namespace yaml {
```
AssemblyJohn, line 16: We already have some yaml parsing in everest, could that lib be leveraged gere?

lib/everest/evse_security/lib/evse_security/utils/enforce_certificate_rules.cpp
```diff
@@ -0,0 +1,307 @@
...
+
+    if (bc_name) {
+        OPENSSL_free(bc_name);
+    }
+
+    std::string fileName;
+
+    if (dc == "CPO") {
+        if (certType == "_leaf") {
+            fileName = std::string(CERT_PROFILES_DIR) + "/SECC Leaf.yaml";
```
AssemblyJohn, line 171: Everything seems to be hard-coded to those standard yamls, is that a good idea? Can't we presume we might need more or less sometime?

lib/everest/evse_security/lib/evse_security/utils/enforce_certificate_rules.cpp
```diff
@@ -0,0 +1,307 @@
...
+        EVLOG_warning << "No matching profile for DC: " << dc;
+        return "";
+    }
+    return fileName;
+}
+
+int enforce_certificate_rules(evse_security::X509Handle* ctx) {
+    int is_valid = 1;
+    if (ENFORCE_CERT_PROFILES) {
+        evse_security::X509Wrapper wrapper(evse_security::OpenSSLSupplier::x509_duplicate_unique(ctx));
```
AssemblyJohn, line 223: This code is intricate. We can extract the raw directly from the handle.

lib/everest/evse_security/lib/evse_security/utils/enforce_certificate_rules.cpp
```diff
@@ -0,0 +1,307 @@
...
+                EVLOG_warning << msg;
+        };
+
+        static const std::vector<int> extensions{83, 87};
+        int is_valid = 0;
+        char buf[256];
+
+        for (auto& rule : fields) {
+            X509_NAME* name =
+                (rule.target == CertPart::Subject) ? X509_get_subject_name(cert) : X509_get_issuer_name(cert);
```
AssemblyJohn, line 260: Leverage the X509Wrapper.

lib/everest/evse_security/lib/evse_security/utils/enforce_certificate_rules.cpp
```diff
@@ -0,0 +1,307 @@
...
+
+            if (!std::binary_search(extensions.begin(), extensions.end(), rule.nid)) {
+                int len = X509_NAME_get_text_by_NID(name, rule.nid, buf, sizeof(buf));
+                if ((rule.mustExist && len <= 0) || (!rule.mustExist && len > 0)) {
+                    log(rule.critical, "NID " + std::to_string(rule.nid) + " does not comply (expected " +
+                                           std::to_string(rule.mustExist) + ")");
+                    is_valid = 1;
+                }
+            } else if (rule.nid == 83 && rule.mustExist) {
```
AssemblyJohn: What do these hard-coded NID's mean? Can we have some defines at least related to the usage?

lib/everest/evse_security/lib/evse_security/utils/enforce_certificate_rules.cpp
```diff
@@ -0,0 +1,307 @@
...
+                    if (br.val == "path_length") {
+                        if ((br.mustExist && !bc->pathlen) || (!br.mustExist && bc->pathlen)) {
+                            EVLOG_warning << "Path length presence does not comply";
+                            is_valid = 1;
+                        } else if (br.mustExist && bc->pathlen && br.data != 0 &&
+                                   ASN1_INTEGER_get(bc->pathlen) != br.data) {
+                            log(br.critical, "Path length value does not match expected " + std::to_string(br.data));
+                            is_valid = 1;
+                        }
+                    } else if (br.val == "CA" && ((br.mustExist && bc->ca != 1) || (!br.mustExist && bc->ca == 1))) {
```
AssemblyJohn: Again "CA" hardcoded values?

lib/everest/evse_security/lib/evse_security/evse_security.cpp
```diff
@@ -1448,6 +1450,22 @@ EvseSecurity::get_full_leaf_certificate_info_internal(const CertificateQueryPara
                 if (is_valid) {
                     any_valid_certificate = true;
 
+                    for (size_t i = 0; i < chain.size(); ++i) {
```
AssemblyJohn: At this moment this is generated each time `get_ful_leaf_certificate_info` is called. That can break existing users of the library.

A must have should be adding an extra bool inside the 'CertificateQueryParams' that should enable this options, maybe `bool enforce_certificate_rules{false}`.

lib/everest/evse_security/lib/evse_security/utils/enforce_certificate_rules.cpp
xandergot: We absolutely could use EVerest’s yaml handling library. This is a placeholder as I was not able to determine which handler EVerest uses for its C++ yaml handling and did not want to add a superfluous library. If you point me to what library is used for this, I would be happy to switch the code over.

lib/everest/evse_security/lib/evse_security/utils/enforce_certificate_rules.cpp
xandergot: I will go through and add comments saying what the specific hardcoded NID’s (Openssl's internal identification system for reffering to certificate parts) reference to. That said the reason they are hard coded is because they require unique handling and data values and as such are not cannot be handled by the automated loop above. If we would rather, I could move the values to a config, but they would need to remain as fixed values.

That said, if what you are requesting is that we use variables with more descriptive names this could be done. But as openssl already uses this NID system, renaming each element seems as though it would be needlessly extraneous code.

lib/everest/evse_security/include/evse_security/certificate/x509_wrapper.hpp
xandergot: Understood, it will be moved

lib/everest/evse_security/include/evse_security/certificate/x509_wrapper.hpp
xandergot: Requires clarification of two comments bellow

lib/everest/evse_security/lib/evse_security/utils/enforce_certificate_rules.cpp
xandergot: CA is a flag indicating weather a certificate can issue child certificates and weather those children can issue their own.

lib/everest/evse_security/lib/evse_security/evse_security.cpp
xandergot: Understood, should the testing be handled the same way?

lib/everest/evse_security/lib/evse_security/utils/enforce_certificate_rules.cpp
xandergot: To clarify, dose this mean that any and all component of the certificate should only be accessed within the wrapper and returned via a geter rather than being directly accessed to the verification code?

lib/everest/evse_security/lib/evse_security/utils/enforce_certificate_rules.cpp
xandergot: See above

lib/everest/evse_security/lib/evse_security/utils/enforce_certificate_rules.cpp
xandergot: See above

lib/everest/evse_security/lib/evse_security/utils/enforce_certificate_rules.cpp
xandergot: Please clarify the question, as is I can't see any preferable method of selecting which profile is to be used. The point of the module is to compare a certificate to the iso15118 requirements for that certificate type, as such quantity doesn't seem particularly relevant. Further, while the yaml naming convention is hard coded the idea behind using the yamls is that, without needing to change the code, we can update the requirements should they change of if a different require spec is desired.

lib/everest/evse_security/lib/evse_security/utils/enforce_certificate_rules.cpp
xandergot: See above

lib/everest/evse_security/lib/evse_security/utils/enforce_certificate_rules.cpp
xandergot: See above

lib/everest/evse_security/lib/evse_security/utils/enforce_certificate_rules.cpp
xandergot: Is the redundancy in passing it to the wrapper or are you saying that I should not need to x509_duplicate_unique in the first place?

lib/everest/evse_security/lib/evse_security/utils/enforce_certificate_rules.cpp
xandergot: See other comment related to Wrapper usage.

lib/everest/evse_security/lib/evse_security/certificate/x509_wrapper.cpp
xandergot: See above

lib/everest/evse_security/include/evse_security/utils/enforce_certificate_rules.hpp
```diff
@@ -0,0 +1,26 @@
...
+#include <openssl/ssl.h>
+#include <openssl/pem.h>
+#include <openssl/x509.h>
```
AssemblyJohn, line 4: This has been resolved while the issue persists:

#include <openssl/ssl.h>
#include <openssl/pem.h>
#include <openssl/x509.h>

Should go, the headers should not contain OpenSSL dependencies, since we might also want to support another provider (like WolfSSL) in the future.

lib/everest/evse_security/lib/evse_security/certificate/x509_wrapper.cpp
AssemblyJohn: As i've said, this should be removed from here completely, since X509 is a OpenSSL dependency. The cast should be done in whomsoever uses the wrapper/handle.

I don't know to which 'above' comment you're referring.

lib/everest/evse_security/lib/evse_security/utils/enforce_certificate_rules.cpp
AssemblyJohn: Please see here: https://github.com/EVerest/EVerest/tree/main/lib/everest/yaml

ryml

lib/everest/evse_security/include/evse_security/certificate/x509_wrapper.hpp
AssemblyJohn, line 8: Importing openssl here would break the initial design consideration, that is to not make dependent a header on openssl code. 

If a user imports this header, we don't want it to be dependent on the openssl header.

Should be removed and used through casting if required.

lib/everest/evse_security/lib/evse_security/utils/enforce_certificate_rules.cpp
AssemblyJohn: Like we have here: https://github.com/xandergot/EVerest/blob/13da171934b9ab37a59f6283760b9faa7a93deca/lib/everest/evse_security/lib/evse_security/certificate/x509_wrapper.cpp#L155

Extra functionaltiy like get_subject_name should be added to the X509Wrapper.cpp/hpp

lib/everest/evse_security/lib/evse_security/utils/enforce_certificate_rules.cpp
AssemblyJohn: Above where?

lib/everest/evse_security/lib/evse_security/utils/enforce_certificate_rules.cpp
AssemblyJohn: Above where?

lib/everest/evse_security/lib/evse_security/utils/enforce_certificate_rules.cpp
AssemblyJohn: What I mean is that we're hard-coding requirements presuming that all ISO providers will have 1 leaf with 2 intermediates and 1 V2GRootCA which is not always a correct assumption. Actually we're working in some environment where the provider has a 3-chain certificate, 1 leaf, 1 intermediate and one root. Sometimes 2-chains are used.

So my question is, hard-coding this to this specific provider is a good idea in a library that tries to be as generic as possible?

Wouldn't it be better if we have a single file that might provide for the mappings for a certain chain one might use?

For example CertProfiles.yaml that contains all the info required instead of hard-coding requirements to the files 'SECC Leaf.yaml' and 'CSO_CPO Sum-CA 1.yaml'?

Or at least a directory iterator that is smart enough to look through all the directory and know which certificate file maps to what? Let's take: https://github.com/xandergot/EVerest/blob/13da171934b9ab37a59f6283760b9faa7a93deca/lib/everest/evse_security/Security_Profiles/iso15118-2/Contract%20Leaf.yaml

It should have some possible identification that we know that it's mapped to the contract leaf but without hard-coding it in code?

As you see the library as-is is not really dependent on specific certificate names that can change and I'd want to keep that.

lib/everest/evse_security/lib/evse_security/evse_security.cpp
AssemblyJohn: For testing, yes it would be ideal.

lib/everest/evse_security/lib/evse_security/utils/enforce_certificate_rules.cpp
AssemblyJohn: Just pass in 'chain.at(i)' instead of 'chain.at(i).get()'? You'll get a X509Wrapper that you can do whatever you want with, including subject name and stuff.

lib/everest/evse_security/lib/evse_security/utils/enforce_certificate_rules.cpp
AssemblyJohn: Just instead of nid==83 is is CommonName, use it from openssl if you can, else define it with somewhere in this file.
`static consteval/constexpr auto NID_CommonName = 83`  or use the openssl NID_commonName version.

Even beter, in the spirit of the library, it would be even better if we could abstract completely from OpenSSL usage since, and abstract this from it.

lib/everest/evse_security/lib/evse_security/utils/enforce_certificate_rules.cpp
AssemblyJohn: If it's a Certificate Authority or something like that, see the comment: https://github.com/EVerest/EVerest/pull/2372/changes/BASE..13da171934b9ab37a59f6283760b9faa7a93deca#r3764858267

Place a constexpr somewhere near the header, let's try not to use hard-coded values. 

As a general rule, this hardcoding should be avoided everywhere in all project not just ours.

lib/everest/evse_security/lib/evse_security/certificate/x509_wrapper.cpp
xandergot: Understood, I will remove any public/raw X509 access from the generic wrapper and keep OpenSSL-specific casting inside the OpenSSL supplier.

lib/everest/evse_security/lib/evse_security/utils/enforce_certificate_rules.cpp
xandergot: Understood, I will use the x509 wrapper to access certificate components, any missing getters, such as for subject-name and subject-components will be added. Any OpenSSL-specific casting will be kept inside the OpenSSL supplier.

lib/everest/evse_security/lib/evse_security/utils/enforce_certificate_rules.cpp
xandergot: Sorry, the order seems to have been broken on my end. I was referring to [https://github.com/EVerest/EVerest/pull/2372#discussion_r3667845954]. That said my confusion surrounding the casting and wrappers has been resolved by your clarifications. I will be using the wrappers to remove any direct OpenSSL parsing from this layer. OpenSSL-specific casting will be kept inside the OpenSSL supplier and x509 handling will be kept in its wrapper as well.

lib/everest/evse_security/lib/evse_security/utils/enforce_certificate_rules.cpp
xandergot: Sorry, the order seems to have been broken on my end. I was referring to [https://github.com/EVerest/EVerest/pull/2372#discussion_r3667845954]. That said my confusion surrounding the casting and wrappers has been resolved by your clarifications. I will be using the wrappers to remove any direct OpenSSL parsing from this layer. OpenSSL-specific casting will be kept inside the OpenSSL supplier and x509 handling will be kept in its wrapper as well.

lib/everest/evse_security/lib/evse_security/utils/enforce_certificate_rules.cpp
xandergot: Fantastic, I will remove the custom yaml parsing and use EVerest’s existing ryml-based YAML utilities instead.

lib/everest/evse_security/lib/evse_security/utils/enforce_certificate_rules.cpp
xandergot: So, there is a level of automatic identification, the domain component of the certificate tells us what the cert is (V2G, MO, etc.). Then the CA component is used to identify if it is a root or an authority. We check if it is a root or an intermediary authority based on whether it is self-signed or not. As for the length of sub-CA, those sub-CAs can potentially be allowed to issue further sub-CAs, or only issue leaves. 

There is potential for the standards to be different based on the allowed length of the chain of CAs. For the sake of having a more generic library it may be a good idea to have a fallback if the path length is more than 2. We could also move the certificate identification to a yaml or json by making said file into a decision tree (ie. dc->ca->profile location). 

The idea behind having a different yaml for each standard is that it is easier to find and modify specific standards rather than having a single large file to dig through, but this is not essential. If a single yaml is more in line with the Everests conventions the profiles could be consolidated.

lib/everest/evse_security/lib/evse_security/utils/enforce_certificate_rules.cpp
xandergot: Makes sense, I will pass the X509Wrapper through the wrapper and then leverage said wrapper for any component access. Any missing access methods will be built in the x509Wraper.

lib/everest/evse_security/lib/evse_security/utils/enforce_certificate_rules.cpp
xandergot: Understood, I will move to using the names of the components and creating getters within the X509 wrapper. These getters will in turn utilize the crypto supplier’s retrieval methods, any missing getters will be created added to the OpenSSL wrapper.

lib/everest/evse_security/lib/evse_security/utils/enforce_certificate_rules.cpp
xandergot: I will remove hard-coded numeric values and expose values like CA or pathLen through wrapper and supplier methods.

lib/everest/evse_security/include/evse_security/utils/enforce_certificate_rules.hpp
xandergot: With the proposed changes relating to Component access these includes will no longer be needed and will be removed. The OpenSSL includes will be moved to the OpenSSL wrapper as will anything relying on them.

lib/everest/evse_security/include/evse_security/certificate/x509_wrapper.hpp
xandergot: With the proposed changes relating to Component access these includes will no longer be needed and will be removed. The OpenSSL includes will be moved to the OpenSSL wrapper as will anything relying on them.

xandergot:
The code has not yet been modified, as of yet we have only addressed the comments in writing.

lib/everest/evse_security/lib/evse_security/utils/enforce_certificate_rules.cpp
AssemblyJohn: >  If a single yaml is more in line with the Everests conventions the profiles could be consolidated.

I am not entirely sure on this one, now giving it another thought. Not sure this centralization in one single yaml that links to different files is the best way to go. Will have to talk with the internal team.

@Pietfried any ideas?

One idea that would be maybe faster/easier to implement: instead of the hard-coded paths could you have some sort of directory iteration that can detect to which certificate the .yaml file is linked? Maybe that would be the most logical initial step.

`for (entry : cert_rules_directory) if(our_cert.type == entry.type) ....`

Pietfried: I'd also prefer a directory-iteration approach combined with an additional `matches` block in the profile yamls. It could look like this:

sub-ca1.yaml

```
matches:
  iso_version: iso15118-2
  domain_component: CPO
  role: sub_ca
  path_length: 1
```

sub-ca2.yaml

```
matches:
  iso_version: iso15118-2
  domain_component: CPO
  role: sub_ca
  path_length: 0
```

v2g-root.yaml

```
matches:
  iso_version: iso15118-2
  domain_component: V2G
  role: root
```

We would not rely on the filenames in this case and apply a matching algorithm for the certificate to check against these matching rules. We would still have the per-profile files and it allows different number of sub ca  + leaf combinations.

Issue #1368, opened 2025-09-16
Draft: Base EvseManager Error Handling purely on severity


Pietfried:
### Describe the problem

The EvseManager module currently handles external and/or internal errors based on their type and based on their severity.

The error type currently defines if an error should cause the EvseManager to become inoperative (not ready for charging / energy transfer). For errors that cause the EvseManager to become inoperative, the error severity currently decides if the error should trigger an error shutdown or emergency shutdown according to IEC61851-23 .

### EVerest Domain

Charge Control

### Affected EVerest Module

* EvseManager
* Potentially all modules that raise errors 

### Describe your solution

The error handling can be implemented purely based on the severity of an error. There is not necessarily the need to differentiate between the error type and the error severity. This would also make the development within the EvseManager and other modules that raise errors easier and more transparent.

### Additional context

_This ticket is still WIP and the solution needs to be defined more precisely_

Pietfried:
One issue I see with extending the use of Severity to influence session shutdown behavior is that the EvseManager module does not receive all errors from the entire EVerest system by default.

Currently, the EvseManager only subscribes to errors from its direct dependencies — such as the BSP, DC power supply, AC RCD, etc. As a result, it does not have visibility into errors raised by unrelated modules elsewhere in the system.

This means that if we decide to make Severity a deciding factor for whether a session should be stopped/shutdown, this logic would only apply to errors originating from modules that the EvseManager directly depends on. Errors outside of that scope would not be observed by the EvseManager and therefore could not influence session shutdown behavior.

Iirc for that reason we decided to not provide any meaning to the Severity for now.

knowack1:
What kind of problems should we expect when extending EvseManager to subscribe to all errors (subscribe_global_all_errors)?
I also see that `evse_manager/Inoperative` error has special treatment (e.g. in OCPP modules). This does not seem to be an error at all, but rather a state of the EVSE. Taking this into account, I would consider notifying the system of this state via a dedicated EvseManager API rather than through the error framework. For example, in evse_manager.yaml:
```yaml
inoperative:
    description: Signals that the EVSE/connector is inoperative
    type: boolean
```

I have also a question regarding EVerest's policy on backward compatibility, particularly for vendors maintaining external modules:
1. What is the general policy on maintaining backward compatibility for both module APIs and their behavior?
[...]

PR #2374, opened 2026-06-25
fix(EvseManager): refresh Inoperative error when the active fatal cause set changes

biubiuzhang:
## Describe your changes

`EvseManager` raises a single aggregate `evse_manager/Inoperative` error whenever any non-ignored sub-module error is active. Today its type / description / vendor_id / uuid / timestamp are captured from the **first** active cause and never updated: `raise_inoperative_error()` early-returns while `Inoperative` is already raised, and `errors_prevent_charging()` only ever returns the first fatal error. After the original cause clears but other fatal errors remain, `Inoperative` stays raised yet still describes the now-cleared first cause — misleading for technicians.

### Observed

On a bench DC unit with an E-Stop pressed at boot **plus** a lost-DAB-heartbeat communication fault and a cable-over-temp stop all active, `Inoperative` was raised with description `MREC8EmergencyStop`. After **releasing** the E-Stop, `evse_board_support/MREC8EmergencyStop` cleared but `Inoperative` remained — still described as `MREC8EmergencyStop`, with its original boot timestamp/uuid — even though the actual remaining causes were `power_supply_DC/CommunicationFault` and `evse_board_support/MREC19CableOverTempStop`.

### What this changes

- `errors_prevent_charging()` now returns **every** active non-ignored error (in detection order) instead of only the first.
- `raise_inoperative_error()` takes the full set and:
  - builds the description from **all** currently-active fatal causes, so technicians see everything keeping the EVSE inoperative (not just the first);
  - tracks the `(type, sub_type)` set it last raised and **refreshes** the error (clear + re-raise, since the framework has no in-place error update) whenever that set changes, so the description follows the current causes instead of freezing at the first;
  - keeps the **first** cause driving `message`/`vendor_id` (preserving the behaviour added in #1175), and uses the **maximum** severity across all active causes to decide emergency vs. error shutdown.
- The refresh deliberately does **not** emit `AllErrorsPreventingChargingCleared`, so the charger stays shut down across a refresh.
- `process_error()` is serialized with a mutex: it is invoked from every requirement's error callbacks, and now mutates shared state (`inoperative_causes`), so the read-compare-clear-raise sequence must be atomic.

Downstream consumers with shorter limits (e.g. the OCPP 1.6 `StatusNotification` `vendorErrorCode`, a `CiString<50>`) already truncate the description themselves, so no length handling is added here.

### Notes / related

- This intentionally keeps the single "primary cause" semantics introduced in #1175 for `message`/`vendor_id`, while fixing the staleness and adding the full-cause aggregation in the description.
- Related but separate: #1368 (proposal to base error handling purely on severity) and #2368 (subscribing `EvseManager` to SLAC/HLC errors) — both also touch this area, so there may be a light merge to coordinate.

N/A (reported internally; happy to file a tracking issue if preferred).

## Test plan

- Updated/added `modules/EVSE/EvseManager/tests/ErrorHandlingTest.cpp`:
  - `IsActive` now asserts the error **refreshes** to the new cause instead of staying frozen.
  - `SameCausesDoNotRefresh` — re-evaluating the same cause set is a no-op (no clear/re-raise; uuid/timestamp preserved).
  - `DescriptionAggregatesAllCauses` — description lists every active cause; `message`/`vendor_id` come from the first.
  - `RefreshWhenACauseClears` — when one of several causes clears, the error refreshes to the remaining cause(s).
- `EvseManager` unit tests build and pass (16/16) locally.

### DIFF
--- a/modules/EVSE/EvseManager/ErrorHandling.cpp
+++ b/modules/EVSE/EvseManager/ErrorHandling.cpp
@@ -158,13 +158,17 @@ void ErrorHandling::clear_over_voltage_error() {
 
 // Find out if the current error set is fatal to charging or not
 void ErrorHandling::process_error() {
+    // Called from every requirement's error callbacks (potentially on different threads). Holding the monitor handle
+    // serializes the whole inoperative_causes read-compare-clear-raise sequence below so it cannot interleave.
+    auto causes = inoperative_causes.handle();
+
     const auto fatal = errors_prevent_charging();
-    if (fatal) {
-        // signal to charger a new error has been set that prevents charging
-        raise_inoperative_error(*fatal);
+    if (not fatal.empty()) {
+        // signal to charger that errors are active that prevent charging
+        raise_inoperative_error(fatal, *causes);
     } else {
         // signal an error that does not prevent charging
-        clear_inoperative_error();
+        clear_inoperative_error(*causes);
     }
 
     // All errors cleared signal is for OCPP 1.6. It is triggered when there are no errors anymore,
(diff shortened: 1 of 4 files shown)


modules/EVSE/EvseManager/ErrorHandling.cpp
```diff
@@ -173,103 +177,99 @@ void ErrorHandling::process_error() {
...
 
-    // raise externally
+    std::vector<std::pair<std::string, std::string>> cause_ids;
+    cause_ids.reserve(causes.size());
+    for (const auto& cause : causes) {
+        cause_ids.emplace_back(cause.type, cause.sub_type);
+    }
+
+    if (p_evse->error_state_monitor->is_error_active("evse_manager/Inoperative", "")) {
+        if (cause_ids == inoperative_causes) {
```
djchhp: This comparison of vectors relies on the order of the elements. Using a `std::set` would help.

modules/EVSE/EvseManager/ErrorHandling.cpp
```diff
@@ -173,103 +177,99 @@ void ErrorHandling::process_error() {
...
+    if (causes.empty()) {
         return;
     }
 
-    // raise externally
+    std::vector<std::pair<std::string, std::string>> cause_ids;
+    cause_ids.reserve(causes.size());
+    for (const auto& cause : causes) {
+        cause_ids.emplace_back(cause.type, cause.sub_type);
+    }
```
djchhp:
```suggestion
const std::set<Everest::error::Error, ErrorCauseLess> cause_ids(causes.begin(), causes.end());

```
`ErrorCauseLess` can become private member of `ErrorHandling` . That is easier to maintain then the opaque type used here.

modules/EVSE/EvseManager/ErrorHandling.hpp
```diff
@@ -118,6 +119,14 @@ class ErrorHandling {
     const std::vector<std::unique_ptr<powermeterIntf>>& r_powermeter;
     const std::vector<std::unique_ptr<over_voltage_monitorIntf>>& r_over_voltage_monitor;
     const bool inoperative_error_use_vendor_id;
+
+    // (type, sub_type) of the fatal causes the raised Inoperative error currently describes, in detection order.
+    // Used to refresh that error when the active fatal set changes. Empty while Inoperative is not raised.
+    std::vector<std::pair<std::string, std::string>> inoperative_causes;
```
djchhp:
```suggestion
      struct ErrorCauseLess {
          bool operator()(const Everest::error::Error& a, const Everest::error::Error& b) const {
              return std::tie(a.type, a.sub_type) < std::tie(b.type, b.sub_type);
          }
      };

      std::set<Everest::error::Error, ErrorCauseLess> inoperative_causes;

```

As mentioned above

modules/EVSE/EvseManager/ErrorHandling.hpp
```diff
@@ -118,6 +119,14 @@ class ErrorHandling {
...
+
+    // Serializes process_error(): it is invoked from every requirement's error callbacks (potentially on
+    // different threads) and mutates inoperative_causes, so the read-compare-clear-raise sequence must be atomic.
+    std::mutex process_error_mutex;
```
djchhp:
```suggestion
   monitor<std::set<Everest::error::Error, ErrorCauseLess>> inoperative_causes
```
This could replace inoperative_causes and the mutex, which bundles the mutex with the to-be-protected state. `process_error()` would look something like this:

```
void ErrorHandling::process_error() {
      auto handle = inoperative_causes.handle();
      const auto fatal = errors_prevent_charging();
      if (not fatal.empty()) {
          raise_inoperative_error(fatal, *handle);
      } else {
          clear_inoperative_error(*handle);
      }
      // ...
  }
```

This automatically circumvents the problem, that `raise/clear_inoperative_error` are `protected` and could be called from inheriting classes without holding the lock.

modules/EVSE/EvseManager/ErrorHandling.cpp
```diff
@@ -173,103 +177,99 @@ void ErrorHandling::process_error() {
...
-        signal_error(ErrorHandlingEvents::ForceErrorShutdown);
-    }
+    // Track what we just raised, so the next call detects a changed cause set.
+    inoperative_causes = std::move(cause_ids);
+
+    // Emergency shutdown if any active cause is high severity, otherwise error shutdown.
+    const bool emergency = std::any_of(causes.begin(), causes.end(), [](const Everest::error::Error& cause) {
+        return cause.severity == Everest::error::Severity::High;
+    });
+    signal_error(emergency ? ErrorHandlingEvents::ForceEmergencyShutdown : ErrorHandlingEvents::ForceErrorShutdown);
```
djchhp: As a side effect this calls `cancel_reservation(true)` in   `EvseManager.cpp:1289–1291` on every refresh. Please clarify if this is intended.

djchhp (changes requested):
Overall this PR describes concreted problems and addresses them correctly. The requested changes are improvements on maintainability and intend to prevent future bugs. The comparison of the `std::vector` relies on the order of the elements. This might be correct here, switching to a `std::set` removes all doubts.

modules/EVSE/EvseManager/ErrorHandling.cpp
biubiuzhang: Done in aaab133f. `inoperative_causes` is now a `std::set<Everest::error::Error, ErrorCauseLess>` keyed on `(type, sub_type)`, so change-detection no longer depends on detection order. One note: `Everest::error::Error` has no `operator==` in the framework, so I compare the two sets via the `ErrorCauseLess` comparator (a `same_causes` helper) rather than `==`.

modules/EVSE/EvseManager/ErrorHandling.cpp
biubiuzhang: Adopted in aaab133f — the manual `vector<pair>` build is gone; current causes are now `const InoperativeCauses current(causes.begin(), causes.end())` and `ErrorCauseLess` is a member type. I kept `ErrorCauseLess`/`InoperativeCauses` `protected` rather than fully `private` because the unit test constructs an `InoperativeCauses` to drive `raise_inoperative_error` directly.

modules/EVSE/EvseManager/ErrorHandling.hpp
biubiuzhang: Added in aaab133f: `struct ErrorCauseLess { return std::tie(a.type, a.sub_type) < std::tie(b.type, b.sub_type); }` and `using InoperativeCauses = std::set<Everest::error::Error, ErrorCauseLess>`. `inoperative_causes` is now that set (wrapped in the monitor, per your next comment).

modules/EVSE/EvseManager/ErrorHandling.hpp
biubiuzhang: Done as suggested in aaab133f. `inoperative_causes` is now `everest::lib::util::monitor<InoperativeCauses>`; `process_error()` acquires the handle once and passes `*handle` into `raise_inoperative_error`/`clear_inoperative_error`, which now take the set by reference. This bundles the mutex with the state it guards and makes it impossible to mutate the causes without holding the lock. The standalone `process_error_mutex` is removed.

modules/EVSE/EvseManager/ErrorHandling.cpp
biubiuzhang: Good catch — not intended. Re-emitting `Force*Shutdown` on every description refresh was redundant (the charger already latches the shutdown via `last_shutdown_type`) and had the side effect of re-running `EvseManager::cancel_reservation(true)` each time. In aaab133f I only emit the shutdown signal on a genuine transition — the first raise, or when the shutdown type actually changes (error↔emergency). A pure description refresh emits nothing, so the charger stays shut down and reservations aren't re-cancelled.

biubiuzhang:
@djchhp thanks a lot for the thorough review! 🙏 I've addressed all the comments in aaab133f — switched to `std::set` with an `ErrorCauseLess` comparator, wrapped the state in `monitor<>` so it can't be touched without the lock, and stopped the shutdown signal (and reservation cancel) from re-firing on a pure refresh. Whenever you have a moment, would you mind taking another look?

djchhp (dismissed):
All comments have been properly addressed. @biubiuzhang This branch needs a rebase before merge. @hikinggrass, @Pietfried  can anyone please take over, I cannot approve changes in these files.

biubiuzhang:
Done — rebased onto latest `main`, conflicts resolved. The only red check is a timeout in `EvseManagerCharger_tests`, which doesn't build any of the files this PR touches, so it looks like an unrelated flake. Thanks @djchhp! @hikinggrass @Pietfried — whenever one of you has a moment. 🙏

hikinggrass (approved):

PR #2384, opened 2026-06-30
feat(run_application): cancellable stop and output-accumulation opt-out

mlitre:
## Describe your changes

Adds a `run_application(name, args, RunOptions)` overload for supervising long-lived children (e.g. a gRPC sidecar). The existing `run_application(name, args, callback)` is unchanged, so all current callers are unaffected.

`RunOptions` bundles:

- **`callback`** — per-line stdout callback (as before).
- **`accumulation`** — an `AccumulationPolicy` variant selecting how stdout is retained: `RetainAll`, `RetainNone`, or bounded `RetainHead` / `RetainTail` / `RetainHeadTail` with line and byte caps. Bounded policies keep whole lines only.
- **`stop_requested`** (`shared_ptr<atomic_bool>`) — cancellation flag; a watcher sends `SIGTERM`, then `SIGKILL` after `terminate_grace`, unblocking a silent child so the call returns. On cancellation `exit_code` reflects the terminating signal.
- **`kill_child_on_parent_death`** — sets `PR_SET_PDEATHSIG(SIGKILL)` so the kernel reaps the child if the spawning thread dies (Linux only).

Covered by `lib/everest/run_application/tests/run_application_test.cpp`.

N/A

### DIFF
--- a/lib/everest/run_application/include/everest/run_application/run_application.hpp
+++ b/lib/everest/run_application/include/everest/run_application/run_application.hpp
@@ -2,8 +2,13 @@
 // Copyright Pionix GmbH and Contributors to EVerest
 #pragma once
 
+#include <atomic>
+#include <chrono>
+#include <cstddef>
 #include <functional>
+#include <memory>
 #include <string>
+#include <variant>
 #include <vector>
 
 namespace everest::run_application {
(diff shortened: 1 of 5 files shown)


lib/everest/run_application/include/everest/run_application/run_application.hpp
```diff
@@ -27,7 +29,12 @@ struct CmdOutput {
...
 /// callback. The callback must return a CmdContol of either Continue or Terminate, which decide if the application will
 /// continue running or terminated early
+/// \p stop_requested is an optional cancellation flag. When non-null, a watcher thread terminates the child (SIGTERM,
+/// then SIGKILL after a 2 s grace) once it is set; exit_code then reflects the terminating signal.
+/// \p accumulate_output controls whether CmdOutput retains stdout. When false, output and split_output are left empty
+/// (the output_callback still fires per line); use this for long-lived children to avoid unbounded memory growth.
 CmdOutput run_application(const std::string& name, std::vector<std::string> args,
-                          const std::function<CmdControl(const std::string& output_line)> output_callback = nullptr);
+                          const std::function<CmdControl(const std::string& output_line)> output_callback = nullptr,
+                          std::shared_ptr<std::atomic_bool> stop_requested = nullptr, bool accumulate_output = true);
```
hikinggrass: What do you think about turning the `accumulate_output` into an enum or even a `Output` struct that includes the output callback and a accumulate-output-policy where you can select full accumulation, a bounded queue or nothing. Maybe even a output buffer size etc.

That would have to be an overload of the existing function, but would give us a bit more flexibility in the future

hikinggrass (approved):

PR #2385, opened 2026-06-30
feat(EvseManager): advertise AC DER via set_der_available

mlitre:
Add a boolean set_der_available command: EvseManager folds the matching ISO 15118-20 DER energy transfer mode (AC_BPT_DER when also BPT-capable, otherwise AC_DER) into the set it advertises, gated on whether DER directive support is wired for the EVSE rather than the EV's runtime capability. OCPP201 asserts availability from the presence of a grid_support connection.


interfaces/evse_manager.yaml
```diff
@@ -136,6 +136,20 @@ cmds:
       description: Returns an enum indicating whether the update was successful or not.
       type: string
       $ref: /evse_manager#/UpdateAllowedEnergyTransferModesResult
+  set_der_available:
+    description: >-
+      Declare whether DER directive support (a grid_support provider) is wired for this EVSE.
+      When true and the EVSE is export-capable, EvseManager advertises the AC_DER_IEC energy
```
Pietfried: Isnt DER also possible without BPT support?

modules/EVSE/EvseManager/docs/index.rst
```diff
@@ -348,6 +348,22 @@ Take care especially with the power(watt) and time based hysteresis settings. Th
 actual use case to avoid relays wearing due too a lot of switching cycles. Consider also to limit the maximum
 number of switching cycles per charging session.
 
+DER (grid support) advertising
+==============================
+
+EvseManager exposes a ``set_der_available`` command that records, per EVSE, whether DER directive support
+(a ``grid_support`` provider) is wired for that EVSE. This is a boot-time fact asserted by the OCPP
+module from the presence of a ``grid_support`` connection; the EV's runtime DER capability is never sent here.
```
Pietfried: I dont think its requried to leak OCPP module specifics in this EvseManager documentation, especially because there could be other providers of `grid_support` that call `set_der_available`. I'd recommend to adjust the wording and maybe name OCPP as one example that activates it.

interfaces/evse_manager.yaml
mlitre: Technically yes, but it is unclear whether DER is BPT specific or not from amendment 1. So as a precaution I think you are right let's make it not tied to BPT

modules/EVSE/EvseManager/docs/index.rst
mlitre: Done

mlitre:
Replaced by #2472

PR #2393, opened 2026-07-02
Fix/iso15118 disconnect teardown

### LINKED ISSUE #2010: Ignoring SDP request after reconnection
### Describe the bug

The libiso reports a warning "Ignoring sdp request message because a session is already created and running". This occurs when the EV was re-connected before the session was finished. This leads to a warning loop for 40s until session is forced to be finished. We were able to recognize this issue even the CP state was changed from B to A for a while and a new reconnect of EV occurs.
See attached log file.

[journalctl_300326_0819.log](https://github.com/user-attachments/files/26346360/journalctl_300326_0819.log)

The expected behaviour is: after state change to A the current charging session should be finished. After reconnection a new session will be created.
 
EVerest release 2026.02.0-rc3
MCS charging with (chargebyte products CCY - CMY, Siemens SPS to control CE, ID connection)

### EVerest Domain

ISO15118

### Affected EVerest Module

EVSE15118D20

### To Reproduce

- Connect an EV and wait for charging session. Terminate the session on EV side. Disconnect EV directly after CP-state change from C to B. 
- Reconnect... Warning should be come up when EV requests SDP

### Anything else?

_No response_

mlitre:
## Describe your changes

> **Note:** this PR is stacked on #2225 (branch `refactor/tls-iso15118-adapter`); please review and merge that one first. The diff shown here includes #2225's changes until it lands.

When the EV side of an ISO 15118-20 session disappears without a clean shutdown (contactor drop, cable pull, EV-side crash — the hard-disconnect case from #2010), the SECC kept the session object alive: nothing on the wire signaled the loss, the poll loop kept spinning on the dead socket, and every reconnect SDP request from the returning EVCC was rejected with "session is already created and running" until the sequence timeout finally reaped the session. The EV could not start a new charging session for up to a minute after replugging. This PR makes the SECC tear the session down promptly through two independent paths — the data-link-ready signal from SLAC, and dead-peer detection on the TCP transport itself — so a reconnecting EV's SDP request is accepted within one poll cycle.

**iso15118 library — session teardown on data-link loss**

- `TbdController::set_dlink_ready(false)` now requests termination of the active session: `tick()` consumes an atomic `terminate_session_requested` flag and closes whatever session exists, whether it completed the V2G handshake or never got past SDP. Previously the dlink signal only reset the communication-setup timeout, so an established-but-dead session lingered until the 60 s sequence timeout. Behavior is the same for AC and DC.
- The flag is deliberately *not* cleared when `handle_sdp_server_input` creates a fresh session: a data-link loss racing session creation must win, so `tick()` reaps the fresh session too. This is safe because the EVCC retransmits its SDP request (~250 ms) and recovers; swallowing the terminate would instead strand a session on a dead link. The flag is consumed unconditionally each tick so a request raised while no session exists cannot tear down a later one.
- `Session::close()` drops a rate-limiter-deferred response: the transport is gone, the response is undeliverable, and keeping it queued held `is_finished()` false so the controller never reaped the session.
- The body of `loop()` is factored into a `tick()` method (plus `has_active_session()`), and the session connection now comes from an injectable `ConnectionFactory` (defaulting to `ConnectionPlain` as before). This is what lets the new teardown tests drive the controller through its public API without a network; production behavior is unchanged.

**iso15118 library — dead-peer detection on the transport**

- Accepted sockets (both plain TCP and TLS) now enable TCP keepalive (10 s idle, 3 s interval, 3 probes ≈ 19 s dead-peer detection), so a peer that vanished without a FIN/RST is detected at the transport layer even if the dlink signal never arrives.
- `ConnectionPlain::read` mapped *every* `read()` error to `would_block` — including `ECONNRESET` and the `ETIMEDOUT` raised by the keepalive — so the level-triggered poll spun on the dead socket until the sequence timeout. Only `EAGAIN`/`EWOULDBLOCK`/`EINTR` are treated as would-block now; any other errno reports the connection closed, which routes into the canonical session close. `ReadResult::connection_closed` is documented to cover fatal socket errors in addition to EOF / TLS `close_notify`.
- `ConnectionSSL::close()` handles the never-connected case: if teardown happens before a peer ever connected, the listening socket is now unregistered from the poll manager, closed, and a single `CLOSED` event is emitted — previously the listener stayed registered as a dangling poll entry.

**IsoMux module**

- `handle_dlink_ready(false)` is forwarded to *both* children (ISO 15118-2 and -20) instead of only the currently selected one. Data-link loss is a broadcast teardown: the session may be owned by the non-selected child, and it must be torn down regardless of which stack the mux currently points at. `dlink_ready(true)` still goes only to the selected child. (Keepalive on IsoMux's own EV-facing socket and teardown of its proxy connection are noted in-code as follow-ups.)

**Tests**

New GTest suites under `lib/everest/iso15118/test/`:

- `session/tbd_controller_teardown.cpp` — drives `TbdController` through `set_dlink_ready` / `tick()` via the injectable connection factory: dlink loss terminates an active session, a terminate racing a fresh session wins, and a stale terminate does not affect a later session.
- `session/session_teardown.cpp` (with `mock_connection.hpp`) — session close on connection-closed read results and the deferred-response drop.
- `io/connection_plain.cpp` — errno classification in `read()` (would-block vs. fatal-error-as-closed).
- `io/socket_keepalive.cpp` — keepalive options are applied to accepted sockets.
- `io/connection_openssl.cpp` — never-connected `ConnectionSSL::close()` releases the listener and emits exactly one `CLOSED`.

Fixes #2010

### DIFF
--- a/lib/everest/iso15118/src/iso15118/session/iso.cpp
+++ b/lib/everest/iso15118/src/iso15118/session/iso.cpp
@@ -215,7 +215,7 @@ TimePoint const& Session::poll() {
 
         for (const auto& timeout : reached) {
             if (timeout == d20::TimeoutType::SEQUENCE) {
-                logf_error("Sequence Timeout 40secs is reached. Stopping the session");
+                logf_error("Sequence timeout (60s) reached. Stopping the session");
                 ctx.session_stopped = true;
                 break;
             } else {
@@ -348,6 +348,8 @@ void Session::handle_connection_event(io::ConnectionEvent event) {
 
 void Session::close() {
     connection->close();
+    // transport gone; a rate-limiter-deferred response is undeliverable
+    [[maybe_unused]] auto res = message_exchange.check_and_clear_response();
     ctx.feedback.signal(session::feedback::Signal::DLINK_TERMINATE);
     ctx.session_stopped = true;
 }
--- a/lib/everest/iso15118/test/iso15118/io/connection_plain.cpp
+++ b/lib/everest/iso15118/test/iso15118/io/connection_plain.cpp
@@ -1,0 +1,147 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2026 Pionix GmbH and Contributors to EVerest
+#include <catch2/catch_test_macros.hpp>
+
+#include <array>
+#include <atomic>
+#include <cerrno>
+#include <chrono>
+#include <cstring>
+#include <future>
+#include <thread>
+
+#include <arpa/inet.h>
+#include <netinet/in.h>
+#include <sys/socket.h>
+#include <unistd.h>
+
+#include <iso15118/io/connection_plain.hpp>
+#include <iso15118/io/logging.hpp>
+#include <iso15118/io/poll_manager.hpp>
+
+using namespace std::chrono_literals;
+
+namespace {
+
+constexpr auto LOOPBACK_IFACE = "lo";
+constexpr auto SERVER_PORT = 50000;
+
+// Drive poll_manager until predicate returns true or the deadline expires.
+template <typename Predicate>
+bool poll_until(iso15118::io::PollManager& pm, Predicate done, std::chrono::milliseconds timeout) {
+    const auto deadline = std::chrono::steady_clock::now() + timeout;
+    while (std::chrono::steady_clock::now() < deadline) {
+        pm.poll(50);
+        if (done()) {
+            return true;
+        }
+    }
+    return false;
+}
+
+// Connect to the server on loopback, send data the server will NOT read, then
+// abort with SO_LINGER {1,0}. Because unread data is still queued at the peer,
+// the close sends a RST, so the server's next read() fails with ECONNRESET
+// rather than a clean EOF. Signals `reset_done` once the abort has been issued.
+bool run_resetting_client(std::atomic<bool>& reset_done) {
+    sockaddr_in6 addr{};
+    addr.sin6_family = AF_INET6;
+    addr.sin6_port = htons(SERVER_PORT);
+    if (inet_pton(AF_INET6, "::1", &addr.sin6_addr) != 1) {
+        return false;
+    }
+
+    const int fd = ::socket(AF_INET6, SOCK_STREAM, 0);
+    if (fd < 0) {
+        return false;
+    }
+
+    int connected = -1;
+    for (int i = 0; i < 50; ++i) {
+        connected = ::connect(fd, reinterpret_cast<sockaddr*>(&addr), sizeof(addr));
+        if (connected == 0) {
+            break;
+        }
+        std::this_thread::sleep_for(20ms);
+    }
+    if (connected != 0) {
+        ::close(fd);
+        return false;
+    }
+
+    const std::array<uint8_t, 8> payload{};
+    [[maybe_unused]] auto res = ::write(fd, payload.data(), payload.size());
+    // Give the server kernel time to buffer the bytes so they are still unread
+    // when the abortive close arrives.
+    std::this_thread::sleep_for(100ms);
+
+    struct linger lin {};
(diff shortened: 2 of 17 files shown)


lib/everest/iso15118/test/iso15118/io/connection_plain.cpp
```diff
@@ -0,0 +1,147 @@
...
+        }
+        std::this_thread::sleep_for(20ms);
+    }
+    if (connected != 0) {
+        ::close(fd);
+        return false;
+    }
+
+    const std::array<uint8_t, 8> payload{};
+    (void)::write(fd, payload.data(), payload.size());
```
barsnick: g++-11.4 (Ubuntu 22.04) warns despite the void type-cast. (And warnings are handled as errors.) This fixes it with C++17:
```suggestion
    [[maybe_unused]] auto res = ::write(fd, payload.data(), payload.size());
```

barsnick (dismissed):

lib/everest/iso15118/test/iso15118/io/connection_plain.cpp
mlitre: Done

mlitre: Fixed

lib/everest/iso15118/test/iso15118/session/tbd_controller_teardown.cpp
```diff
@@ -0,0 +1,196 @@
...
+            ++calls;
+            return connection;
+        };
+    }
+};
+
+iso15118::TbdController make_controller(const iso15118::session::feedback::Callbacks& callbacks,
+                                        MockConnectionFactory& factory) {
+    return iso15118::TbdController{iso15118::TbdConfig{{},
+                                                       "lo",
```
SebaLukas, line 39: Is `lo` actually working? Typically `lo` has only ipv4 support and no ipv6 address

SebaLukas: But I guess with the mock_connection it worked because everything is mocked. ipv6 address is empty but that does not matter here in the test

lib/everest/iso15118/test/iso15118/io/connection_plain.cpp
```diff
@@ -0,0 +1,147 @@
...
+
+// Connect to the server on loopback, send data the server will NOT read, then
+// abort with SO_LINGER {1,0}. Because unread data is still queued at the peer,
+// the close sends a RST, so the server's next read() fails with ECONNRESET
+// rather than a clean EOF. Signals `reset_done` once the abort has been issued.
+bool run_resetting_client(std::atomic<bool>& reset_done) {
+    sockaddr_in6 addr{};
+    addr.sin6_family = AF_INET6;
+    addr.sin6_port = htons(SERVER_PORT);
+    if (inet_pton(AF_INET6, "::1", &addr.sin6_addr) != 1) {
```
SebaLukas, line 50: This works only in this test. In a real hlc session the ipv6 address should be not `::1` 😅

lib/everest/iso15118/src/iso15118/session/iso.cpp
```diff
@@ -348,6 +348,8 @@ void Session::handle_connection_event(io::ConnectionEvent event) {
 
 void Session::close() {
     connection->close();
+    // transport gone; a rate-limiter-deferred response is undeliverable
+    (void)message_exchange.check_and_clear_response();
```
SebaLukas: Please use `[[maybe_unused]] result = ` instead of `(void)`

SebaLukas (changes requested):

SebaLukas (dismissed):

SebaLukas (approved):

PR #2395, opened 2026-07-02
fix(EvseV2G): report EVSE_Shutdown status after EV stop request

Pietfried:
## Describe your changes

- Mark the charge and welding phases as shut down when the EV sends PowerDeliveryReq with ReadyToChargeState = false, so that PowerDeliveryRes and the subsequent WeldingDetectionRes report EVSE_Shutdown instead of EVSE_Ready [IEC 61851-23:2023 CC.7.5.19].
- More specific status codes like EVSE_UtilityInterruptEvent or EVSE_Malfunction are preserved. Applies to both DIN 70121 and ISO 15118-2.

### DIFF
--- a/modules/EVSE/EvseV2G/tests/din_server_test.cpp
+++ b/modules/EVSE/EvseV2G/tests/din_server_test.cpp
@@ -560,4 +560,124 @@ TEST_F(DinServerTest, din_validate_response_code_V2G_DC_665) {
     EXPECT_GE(given_response_code, min_expected_response_code);
 }
 
+class DinServerPowerDeliveryTest : public DinServerTest {
+protected:
+    void SetUp() override {
+        DinServerTest::SetUp();
+
+        exi_in->V2G_Message.Body.PowerDeliveryReq_isUsed = true;
+        init_din_PowerDeliveryReqType(&exi_in->V2G_Message.Body.PowerDeliveryReq);
+
+        exi_out->V2G_Message.Body.PowerDeliveryRes_isUsed = 1u;
+        init_din_PowerDeliveryResType(&exi_out->V2G_Message.Body.PowerDeliveryRes);
+
+        ctx->is_dc_charger = true;
+        ctx->current_v2g_msg = V2G_POWER_DELIVERY_MSG;
+        ctx->last_v2g_msg = V2G_CURRENT_DEMAND_MSG;
+        ctx->state = WAIT_FOR_CURRENTDEMAND_POWERDELIVERY;
+        ctx->evse_v2g_data.session_id = 0;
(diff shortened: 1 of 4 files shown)


florinmihut (changes requested):
Regarding test coverage, a suggestion to make it more complete:

Test the WeldingDetectionRes directly. The main point of this fix is that WeldingDetectionRes reports EVSE_Shutdown, but the tests only assert the array slot evse_status_code[PHASE_WELDING] and trust that handle_din_welding_detection picks it up. Could you expose handle_din_welding_detection via the states:: namespace (same as done for the power delivery handler) and add a test that runs the stop PowerDelivery followed by a WeldingDetectionReq, asserting WeldingDetectionRes.DC_EVSEStatus.EVSEStatusCode == EVSE_Shutdown? That would cover the actual spec-visible behavior [IEC 61851-23 CC.7.5.19] end to end.

Pietfried:
> Regarding test coverage, a suggestion to make it more complete:
> 
> Test the WeldingDetectionRes directly. The main point of this fix is that WeldingDetectionRes reports EVSE_Shutdown, but the tests only assert the array slot evse_status_code[PHASE_WELDING] and trust that handle_din_welding_detection picks it up. Could you expose handle_din_welding_detection via the states:: namespace (same as done for the power delivery handler) and add a test that runs the stop PowerDelivery followed by a WeldingDetectionReq, asserting WeldingDetectionRes.DC_EVSEStatus.EVSEStatusCode == EVSE_Shutdown? That would cover the actual spec-visible behavior [IEC 61851-23 CC.7.5.19] end to end.

I added an aditional test and will resolve the remaining conflict after your review.

florinmihut (dismissed):
LGTM

modules/EVSE/EvseV2G/tests/din_server_test.cpp
SebaLukas, line 1: For the Req messages (PowerDelivery and WeldingDetection), all required (non-optional) elements should be set. I know the tests are working because using the init_* functions, but the messages should be created with "correct" values.

SebaLukas (changes requested):
Overall looking good!
Only one small required change from my side

SebaLukas:
I would also like to have some D2 tests, but I guess that would be too much work here in this PR

modules/EVSE/EvseV2G/tests/din_server_test.cpp
Pietfried: Changes are addressed, it's ready for re-review

SebaLukas (approved):

barsnick (approved):

PR #2405, opened 2026-07-03
Feature/external energy node api

marcnulens-pionix:
## Describe your changes

Adds two new API modules that let a multi-cluster EVerest deployment bridge energy management across separate EVerest processes over an external MQTT broker:

- **`external_energy_node_API`** (cluster side): aggregates `energy_flow_request`s from locally connected `EnergyNode`s and publishes the aggregate to external MQTT, so a remote site-level EnergyManager can participate in load management. It routes `enforce_limits` coming back from the external EnergyManager to all child `EnergyNode`s, taking priority over the internal EnergyManager. It also exposes an `energy_grid` interface so the internal EnergyManager can act as a fallback when the external EnergyManager has been silent longer than `timeout_s`.
- **`external_energy_node_client_API`** (site side, renamed from `external_energy_node_API_client` for consistent naming): subscribes to an `external_energy_node_API` server's aggregated `energy
_flow_request` over external MQTT and presents it as a local `energy_consumer` to the site-level `EnergyNode` (grid connection point), routing `enforce_limits` back to the remote cluster.


interfaces/energy_node_switch.yaml
cburandt, line 1: Looks like this file is not required, as none of the new modules uses it.

modules/API/EVerestAPI/external_energy_node_API/external_energy_node_API.cpp
cburandt, line 1: We tend to keep the init/ready functions as clean as possible. Have a look at the other API modules. They usually have some general setup in these functions and then call a generate_* function for each command and each variable the module is providing. Aligning with this scheme makes it easier to see what is happening.

modules/API/EVerestAPI/external_energy_node_client_API/energy_grid/energyImpl.cpp
```diff
@@ -0,0 +1,63 @@
...
+    //   everest_api/1/external_energy_node/{server_id}/e2m/energy_flow_request
+    //   everest_api/1/external_energy_node/{server_id}/m2e/enforce_limits
+    ev_API::Topics server_topics;
+    server_topics.setup(mod->config.server_id, "external_energy_node", 1);
+
+    // Subscribe to the server's published energy_flow_request (e2m = Everest to Machine).
+    // Republish on the local Everest bus so the site-level EnergyNode sees the cluster
+    // as a normal energy_consumer child.
+    // EnergyFlowRequest has no API-type wrapper, so we use EVerest's native JSON.
+    const auto flow_req_topic = server_topics.everest_to_extern("energy_flow_request");
```
cburandt: The missing conversion logic should become part of the lib/EVerest/everest_api_types/ library.

cburandt (changes requested):
The general approach of providing these two modules looks good.

Try to make the code for the API modules as similar to the existing API modules as possible (what's in the ready/init functions, how the variables and other member functions are called, etc).

Also add missing conversion/serialization routines to the everest_api_types lib. There the serialization/de-serialization automatically is tested via a round-trip test.

For future users providing the corresponding AsyncAPI specification is vital. It serves as the contract to the users (what is provided and how the datatypes look like) and is also being served in HTML format with the official EVerest documentation).

.gitignore
mlitre, line 1: I would double check some of these ignores to make sure we don't ignore some actually relevant files / directories

modules/API/EVerestAPI/external_energy_node_API/external_energy_node_API.cpp
```diff
@@ -0,0 +1,126 @@
...
+    comm_params.heartbeat_period_ms = config.cfg_heartbeat_interval_ms;
+    comm_params.communication_check_period_s = config.cfg_communication_check_to_s;
+    helper.init(comm_params);
+
+    // Subscribe to energy_flow_request from each local EnergyNode (EVSE nodes).
+    // On each update: merge the child, check external timeout, republish aggregate on
+    // the local Everest bus (for internal) and via ApiHelper topic (for external via bridge).
+    for (auto& entry : r_energy_consumer) {
+        entry->subscribe_energy_flow_request([this](types::energy::EnergyFlowRequest const& child) {
+            std::lock_guard<std::mutex> lock(aggregate_mutex);
```
mlitre: Please try to use everest util as much as possible, here we could use the monitor instead

mlitre (changes requested):

interfaces/energy_node_switch.yaml
marcnulens-pionix: Removed the unused interface

.gitignore
marcnulens-pionix: Removed al the changes from .gitignore.

modules/API/EVerestAPI/external_energy_node_API/external_energy_node_API.cpp
marcnulens-pionix: Rewrote it, using the monitor instead

modules/API/EVerestAPI/external_energy_node_API/external_energy_node_API.cpp
marcnulens-pionix: Cleaned the in init/read. Moved functionalty to separate genenerate_* functions.     
    void generate_api_var_energy_flow_request();
    void generate_api_cmd_enforce_limits();

modules/API/EVerestAPI/external_energy_node_client_API/energy_grid/energyImpl.cpp
marcnulens-pionix: Added all the conversion code.

docs/source/reference/EVerest_API/external_energy_node_client_API.yaml
```diff
@@ -0,0 +1,123 @@
...
+      $ref: '#/channels/receive_heartbeat'
+
+  send_communication_check:
+    title: 'Send communication check'
+    action: send
+    channel:
+      $ref: '#/channels/send_communication_check'
+
+components:
+  messages:
```
cburandt, line 102: The "messages" section is missing
- send_energy_flow_request 
- receive_enforce_limits
(but the associated types should exists)

lib/everest/everest_api_types/include/everest_api_types/energy/codec.hpp
```diff
@@ -23,6 +23,10 @@ std::string serialize(ScheduleSetpointEntry const& val) noexcept;
 std::string serialize(ExternalLimits const& val) noexcept;
 std::string serialize(EnforcedLimits const& val) noexcept;
 std::string serialize(CapabilityLimits const& val) noexcept;
+std::string serialize(NodeType const& val) noexcept;
+std::string serialize(EvseState const& val) noexcept;
```
cburandt: For the serialize functions, call by value for enums.

```suggestion
std::string serialize(NodeType val) noexcept;
std::string serialize(EvseState val) noexcept;
```

lib/everest/everest_api_types/tests/test_file_autogenerator/disable.csv
```diff
@@ -1 +1,2 @@
-generic/API.hpp,RequestReply
\ No newline at end of file
+generic/API.hpp,RequestReply
+energy/API.hpp,EnergyFlowRequest
```
cburandt:
```suggestion
```

This shouldn't be required. Did the automatic test creation fail otherwise?

modules/EnergyManagement/EnergyNode/EnergyNode.cpp
cburandt, line 1: Remove this change from this PR and set https://github.com/EVerest/EVerest/pull/2406 from "draft" to "open" instead

cburandt (changes requested):
This is a big step in the right direction.
Find some comments below. However, this time I did not look at the modules themselves.

lib/everest/everest_api_types/include/everest_api_types/energy/codec.hpp
mnulensihomer: Good point, applied the proposal.

modules/EnergyManagement/EnergyNode/EnergyNode.cpp
mnulensihomer: put the PR to open and remove the code from this one

lib/everest/everest_api_types/tests/test_file_autogenerator/disable.csv
mnulensihomer: When running with EnergyFlowRequest we run into infinite recursion. This is a test-generator issue.                                                                                                                                                                                                
The generator's rule is "for a std::vector<T> field, produce one element by calling generate<T>()." For std::vector<EnergyFlowRequest> children, that means generate<EnergyFlowRequest> calls itself unconditionally with no depth limit or termination condition → infinite recursion → stack overflow.                          
 In the real code the list of children will not have this issue. 

So this exclusion of the EnergyFlowReust is ncessary here.

docs/source/reference/EVerest_API/external_energy_node_client_API.yaml
marcnulens-pionix: Added the messages sections

lib/everest/everest_api_types/tests/test_file_autogenerator/disable.csv
cburandt: That seems to be the first time an API type is defined like this. In another PR I have a commit which should fix the test generator in that regard: https://github.com/EVerest/EVerest/pull/1869 (cherry-pick 58f4394daafff47fa5284d595ac183e708e6db36 into your branch to make this detail part of your PR)

modules/API/EVerestAPI/external_energy_node_API/energy_grid/energyImpl.cpp
```diff
@@ -0,0 +1,31 @@
...
+void energyImpl::handle_enforce_limits(types::energy::EnforcedLimits& value) {
+    // This is called by the internal EnergyManager via the Everest internal bus.
+    // Apply internal limits ONLY when the external EnergyManager is not currently active.
+    // When external is active, its enforce_limits (received via external MQTT in the module's
+    // ready() callback) take priority and internal limits are discarded here.
+    if (!mod->external_active.load()) {
+        for (auto& entry : mod->r_energy_consumer) {
+            entry->call_enforce_limits(value);
+        }
+    }
```
cburandt, line 38: I don't have the insight: Will this do as a fallback? Provided that external_active gets false, is it sufficient to wait for the next enforce_limits call to come in? Or should the last one be cached and immediately send out when the timeout occurs?

modules/API/EVerestAPI/external_energy_node_API/external_energy_node_API.cpp
```diff
@@ -0,0 +1,135 @@
...
+            if (config.timeout_s > 0 && external_active.load()) {
+                const auto elapsed = std::chrono::duration_cast<std::chrono::seconds>(std::chrono::steady_clock::now() -
+                                                                                      external_last_seen)
+                                         .count();
+                if (elapsed >= config.timeout_s) {
+                    external_active = false;
+                    EVLOG_info << info.id
+                               << ": external EnergyManager timed out — falling back to internal EnergyManager";
+                }
+            }
```
cburandt: Why is this checked here? As far as I can tell this should work, because the EvseManager at the end of the chain will publish flow requests periodically, but wouldn't a timeout be more explicit? (Everest::Timer[.timeout(...)] could probably be used)

modules/API/EVerestAPI/external_energy_node_API/manifest.yaml
cburandt, line 1: This reads like it was written with a very specific setup in mind (e.g. "cluster", "site-level", "the internal EnergyManager", ...). Keep it a bit more generic and maybe point out the other API module of this PR to be the natural counter-part.

modules/API/EVerestAPI/external_energy_node_API/manifest.yaml
```diff
@@ -0,0 +1,60 @@
...
+  external_energy_node_client_API runs in the external (site-level) EVerest process.
+config:
+  timeout_s:
+    description: >-
+      Seconds without an enforce_limits message from the external EnergyManager
+      before falling back to limits from the internal EnergyManager. Set to 0 to
+      disable the timeout (always prefer external limits, ignore internal).
+    type: integer
+    minimum: 0
+    default: 30
```
cburandt: "cfg_communication_check_to_s" is used to detect if the other side is still present. Does timeout_s interact with this setting? (e.g. does timeout_s need to be larger, or is it a fully independent mechanism? Just clarify.)

modules/API/EVerestAPI/external_energy_node_API/manifest.yaml
```diff
@@ -0,0 +1,60 @@
...
+      The internal EnergyManager connects here as its energy trunk. Limits from
+      internal are applied only when the external EnergyManager has been silent
+      for longer than timeout_s. When external is active, its MQTT limits take
+      priority.
+    interface: energy
+requires:
+  energy_consumer:
+    interface: energy
+    min_connections: 0
+    max_connections: 128
```
cburandt, line 86: Just wondering if this is really required to allow arbitrary many consumers (opposed to one optional)? If yes, it's fine but if not, why not avoid the added complexity?

modules/API/EVerestAPI/external_energy_node_client_API/CMakeLists.txt
cburandt, line 1: Please stick to the auto-generated format of the other API modules CMakeLists.txt files (e.g. copy from your other API module and substitute the target sources names)

modules/API/EVerestAPI/external_energy_node_client_API/manifest.yaml
cburandt, line 1: This reads like it was written with a very specific setup in mind (esp. "cluster", "site-level", "Run one instance per ..."). Keep it a bit more generic and maybe point out the other API module of this PR to be the natural counter-part.

cburandt (changes requested):
Some general thought: As the external_energy_node_client_API is a very special module (being an EVerestAPI module which incorporates a client at the same time) we should make sure it has a good name (maybe "...client_API" is already a good name, but since it should become a generic like the existing "...consumer_API" we should make sure it's a good name; "...bridge_API" could also work)

Also some additional documentation is indicated. Either as module documentation or in the docs/ folder of the EVerest repo. It should explain (and ideally contain a graphical scheme) how these modules are typically used.

lib/everest/everest_api_types/tests/test_file_autogenerator/disable.csv
marcnulens-pionix: Cherry Picked the change and removed the exclusion from the csv file.

modules/API/EVerestAPI/external_energy_node_API/energy_grid/energyImpl.cpp
marcnulens-pionix: Leave as is. The update_interval config param, default 1 second (manifest.yaml:8-11) is configurable. Every consumer, including the external_energy_node_API, gets a fresh enforce_limits call at this cadence regardless of activity.                                                                                                                                                                                                            
  - EnforcedLimits.valid_for is set to update_interval * 10 (10 s by default) — so even a "stale" limit is still considered valid for a while.

modules/API/EVerestAPI/external_energy_node_API/manifest.yaml
marcnulens-pionix: Removed the cluster and site-level form the description. The reference to the .._client module is already in there.

modules/API/EVerestAPI/external_energy_node_API/manifest.yaml
marcnulens-pionix: Update the discription of the timeout_s to clarify the difference between the two parameters. They are not related.

modules/API/EVerestAPI/external_energy_node_API/manifest.yaml
marcnulens-pionix: It can be more than one, so 128 is fine.

modules/API/EVerestAPI/external_energy_node_client_API/manifest.yaml
marcnulens-pionix: Removed the cluster and site-level references.

modules/API/EVerestAPI/external_energy_node_client_API/CMakeLists.txt
marcnulens-pionix: Alligned to auto generated format

modules/API/EVerestAPI/external_energy_node_API/external_energy_node_API.cpp
marcnulens-pionix: remove the check here an moved the code to an Everest:Timer as you proposed.

PR #2407, opened 2026-07-03
Reduce build dependencies for libs

cburandt:
## Describe your changes

Reducing build dependencies for building libs only (EVEREST_LIBS_ONLY CMake option).

EDM combines all dependencies.yaml files it can find, without regard of which part of the project is actually being build. There were to many non-dependencies being pulled in when building e.g. libiso15118.

This change does

- centralize the dependencies.yaml (no more per-lib dependencies.yaml)
- add missing cmake_condition properties
- remove local per-lib dependencies.yaml where possible
- fix EVerest lib inter-dependencies (ev-lib-dependencies.cmake)
- add some missing find_package() calls for non-EDM builds
- no longer install/use ev-cli when EVEREST_LIBS_ONLY is set
- no longer install/build nanopb when EVEREST_LIBS_ONLY is set


SebaLukas:
I have one problem with the changes :sweat_smile: When I try to build libiso15118 (clone `EVerest` and change the folder to `lib/iso15118/iso15118` `cmake` are not able to include `Catch2` anymore.
I guess because you removed the `dependencies.yaml` within `libiso15118`

cburandt:
@SebaLukas , please try again. It should work now. Also fixed for cbv2g and ieee2030_1_1.

SebaLukas:
Your change works :+1:
But now the CI had some problems finding `catch2` :sweat_smile:

cburandt:
> Your change works 👍 But now the CI had some problems finding `catch2` 😅

Should be fixed now.
(Tested locally: building the individual libs which use catch2 as standalone and using the EVEREST_LIBS_ONLY option; configuring the whole EVerest project works as well)

dependencies.yaml
```diff
@@ -138,3 +141,55 @@ grpc-extended-cpp-plugin:
...
+    ]
+pybind11:
+  git: https://github.com/pybind/pybind11.git
+  git_tag: v3.0.2
+  cmake_condition: "NOT EVEREST_LIBS_ONLY OR (EVEREST_ENABLE_PY_SUPPORT AND EVEREST_BUILD_LIB_framework)"
+  options: ["PYBIND11_TEST OFF", "PYBIND11_USE_CROSSCOMPILING ON"]
+pybind11_json:
+  git: https://github.com/pybind/pybind11_json.git
+  git_tag: 0.2.15
+  cmake_condition: "NOT EVEREST_LIBS_ONLY OR (EVEREST_ENABLE_PY_SUPPORT AND EVEREST_BUILD_LIB_framework)"
```
SebaLukas: Missing EOL

SebaLukas (dismissed):
Looking good! 👍

dependencies.yaml
cburandt: fixed

SebaLukas (dismissed):

CMakeLists.txt
```diff
@@ -181,21 +188,6 @@ if(NOT DISABLE_EDM)
...
-        find_package(CURL 7.84.0 REQUIRED)
-        find_package(ftxui 6 QUIET) # optional, if not available BringUp module will not be built
-        find_package(sdbus-c++ REQUIRED)
-    endif()
-
-    find_package(ryml REQUIRED)
-    find_package(nlohmann_json REQUIRED)
-    find_package(fmt REQUIRED)
-    find_package(date REQUIRED)
-    find_package(OpenSSL 3 REQUIRED)
```
hikinggrass, line 198: I think this was removed by accident, since it leads to configure failures in yocto builds. But we could make this a bit more flexible by moving this to the modules / applications where these dependencies are actually needed.

hikinggrass (changes requested):

CMakeLists.txt
cburandt: That certainly went well beyond the target ... Now re-introduced these find_packages-calls again, but execute them only if EVEREST_LIBS_ONLY is not set. The individual libs have the same find_packages commands now (executed if EDM_DISABLED is set).

hikinggrass (approved):
tested with yocto scarthgap (everest-core and everest-examples recipes) as well as everest-examples in EDM mode and EVerest as a FetchContent dependency. Nice work!

hikinggrass:
Now just a squash+rebase (should fix the DCO check as well) is missing, will re-approve afterwards

SebaLukas (approved):

PR #2409, opened 2026-07-03
feat(system): Expose network-configuration through the EVerest system interface

Pietfried:
## Describe your changes

- types/network.yaml: InterfaceClass, APN, VPN, request/response/status types for network interface preparation
- interfaces/system.yaml: configure_network cmd (direct or Processing-deferred result), configure_network_status and network_interfaces vars
- NotSupported stubs in all in-tree providers (Misc/System, system_API, Linux_Systemd_Rauc); no behavior change

This is part 1 of 3 from a stack:

1) https://github.com/EVerest/EVerest/pull/2409 👈
2) https://github.com/EVerest/EVerest/pull/2411
3) https://github.com/EVerest/EVerest/pull/2413


types/network.yaml
```diff
@@ -0,0 +1,189 @@
...
+    enum:
+      - Wired0
+      - Wired1
+      - Wired2
+      - Wired3
+      - Wireless0
+      - Wireless1
+      - Wireless2
+      - Wireless3
+      - Any
```
hikinggrass, line 17: Not sure if this mapping really makes sense for the internal types. Ok for OCPP compatibility, but seems weirdly restrictive

types/network.yaml
```diff
@@ -0,0 +1,189 @@
...
+  VPN:
+    description: VPN configuration
+    type: object
+    additionalProperties: false
+    required:
+      - server
+      - user
+      - password
+      - key
+      - type
```
hikinggrass, line 57: That seems a bit much, what if I have a vpn solution (eg wireguard) without a password?

types/network.yaml
```diff
@@ -0,0 +1,189 @@
...
+        description: VPN shared secret
+        type: string
+      type:
+        description: Type of VPN
+        type: string
+        enum:
+          - IKEv2
+          - IPSec
+          - L2TP
+          - PPTP
```
hikinggrass, line 81: Maybe needs at least a "Other" entry

types/network.yaml
Pietfried: For that reason I also added the `interface_name` property as an optional override, e.g. `eth0` . Providers of system would need a mapping to fullfil the OCPP requirements.

types/network.yaml
```diff
@@ -0,0 +1,187 @@
...
+      - request_id
+      - status
+    properties:
+      request_id:
+        description: Correlation id echoed from the ConfigureNetworkRequest
+        type: integer
+      status:
+        description: Final state — Ready or Failed
+        type: string
+        $ref: /network#/ConfigureNetworkStatusEnum
```
marcemmers: This also shows `Processing` as a valid response in the documentation. Not that big of a deal but also not really logical.

Pietfried: True. Instead of adding another type I documented that `Processing` shall not be used in the Async response and is treated like `Failed` by consumers.

cburandt: I would also go with a specialized, albeit very similar enum type here. But that may be a matter of taste. Biggest cost would probably be the additional conversion routines in the external API.

cburandt (dismissed):
Only one comment (which we don't need to argue about).

(and check if the types.yaml/interfaces.yaml hashes really match - looks like the corresponding test is failing)

cburandt (approved):
LGTM

PR #2412, opened 2026-07-03
Feature/framework configserviceextension

cburandt:
## Describe your changes
Extension of the internal configuration service.
Allows runtime parameter changes to be dispatched to the modules.
Prepares to attach an external API.


lib/everest/framework/everestpy/src/everest/module.cpp
```diff
@@ -143,3 +150,42 @@ std::shared_ptr<Everest::error::ErrorStateMonitor>
...
+    identifier.module_implementation_id = impl_id;
+
+    const pybind11::gil_scoped_release release;
+    return json(handle->get_config_service_client()->get_config_value(identifier));
+}
+
+void Module::register_config_change_handler(const std::string& impl_id, const std::string& param_name,
+                                            std::function<json(const std::string&)> handler) {
+    auto& stored = config_change_handlers.emplace_back(std::move(handler));
+    // TODO(CB): Not working for implementations, does it?
```
Pietfried: Afaics this should work and the comment can be removed

lib/everest/framework/include/utils/config/slot_manager.hpp
```diff
@@ -44,10 +43,32 @@ class SqliteConfigSlotManager {
...
+    /// \brief Updates an existing configuration slot
+    /// Optionals without a value will set the corresponding value to NULL in the DB
+    /// \param slot_id Id of the slot; must not exist yet
+    /// \param config_dump JSON dump of the config file that was used to create the configuration
+    /// \param config_file_path Path to the config file that was used to create the configuration
+    /// \param description Arbitrary text
+    GenericResponseStatus update_config_slot(int slot_id, const std::string& config_dump,
+                                             const std::optional<std::filesystem::path>& config_file_path,
+                                             const std::optional<std::string>& description);
+    /// \brief Updates an existing configuration slot'S description
```
Pietfried:
```suggestion
    /// \brief Updates an existing configuration slot's description
```

lib/everest/framework/include/utils/mqtt_config_service.hpp
```diff
@@ -158,13 +186,28 @@ class MqttConfigServiceHandler {
     ///                          set_config_parameters) are routed through this interface.
     MqttConfigServiceHandler(MQTTAbstraction& mqtt_abstraction, ConfigServiceInterface& config_svc);
 
+    /// \brief This class provides an MQTT interface to modules and uses the ConfigServiceInterface to fulfill incoming
+    /// requests This function allows to request a config parameter change from a module. This function skips validating
+    /// module_ids and characteristics on purpose. It should only be called with inputs known to work. \param
```
Pietfried:
```suggestion
    /// \brief This class provides an MQTT interface to modules and uses the ConfigServiceInterface to fulfill incoming
    /// requests. This function allows to request a config parameter change from a module. This function skips validating
    /// module_ids and characteristics on purpose. It should only be called with inputs known to work. \param
```

lib/everest/framework/lib/config/config_service_core.cpp
```diff
@@ -3,347 +3,676 @@
...
+// practice a healthy module replies well within it; the bounded wait is a
+// deliberate trade-off, and any future that does time out is parked in
+// orphaned_futures_ (reaped on the next call) so the actor still returns.
+//
+// Event delivery: publish_*_update() copy the handler list and, in threaded
+// mode, run handlers on a detached thread, so handlers may call back into the
+// service without deadlocking the actor.
+//
+// Shutdown: ~ConfigServiceCore() clears running_, stops command_queue_ and
+// joins the worker thread.
```
Pietfried: This comment does not seem to be up to date with the current implementation

lib/everest/framework/lib/config/config_service_core.cpp
```diff
@@ -3,347 +3,676 @@
...
+                                                                  const Origin& origin) {
+    return post_to_actor(
+        [this, slot_id, updates, origin]() { return internal_set_config_parameters(slot_id, updates, origin); });
+}
+SetConfigParameterResult
+ConfigServiceCore::internal_set_config_parameters(int slot_id, const std::vector<ConfigParameterUpdate>& updates,
+                                                  const Origin& origin) {
+    SetConfigParameterResult result;
+    result.status = SetConfigParameterStatus::Ok;
+    const bool modules_are_running = module_status_ == ActiveSlotStatus::Running;
```
Pietfried: unused variable

lib/everest/framework/schemas/migrations/1_up-initial.sql
```diff
@@ -4,7 +4,6 @@ CREATE TABLE IF NOT EXISTS DATATYPE (ID INT PRIMARY KEY, DATATYPE TEXT);
 CREATE TABLE CONFIG_META (
     ID INTEGER PRIMARY KEY,
     LAST_UPDATED TEXT NOT NULL,
-    VALID TEXT INTEGER NOT NULL,
```
Pietfried, line 7: No changes to existing sql files should be made or required

lib/everest/framework/lib/config/config_service_core.cpp
Pietfried, line 1: The serialization through a single thread for incoming requests includes MQTT roundtrips for setting config parameters at certain modules, which theoretically can take up to multiple seconds (based on the handler implementation in the module). For a follow up, I'd suggest to extract the MQTT roundtrip into a separate thread/threadpool.

lib/everest/framework/lib/config/sqlite_storage.cpp
```diff
@@ -478,6 +465,9 @@ GenericResponseStatus SqliteStorage::write_module_config_items(const ModuleConfi
                     GetSetResponseStatus::OK) {
                     EVLOG_error << "Failed to write configuration parameter for module: " << module_id
                                 << ", param: " << identifier.configuration_parameter_name;
+                } else {
+                    EVLOG_debug << "Written configuration parameter for modul: " << module_id
```
Pietfried:
```suggestion
                    EVLOG_debug << "Written configuration parameter for module: " << module_id
```

lib/everest/framework/lib/mqtt_config_service.cpp
```diff
@@ -268,80 +226,57 @@ Response handle_set_request(const SetRequest& set_request, const std::string& or
...
+        case Everest::config::SetConfigParameterResultEnum::AccessDenied:
+            response.status = ResponseStatus::AccessDenied;
+            response.status_info = fmt::format("Access to config item denied: {} cannot access {}", origin,
+                                               set_request.identifier.module_id);
+            break;
+        case Everest::config::SetConfigParameterResultEnum::DoesNotExist:
+            response.status = ResponseStatus::Error;
+            break;
+        case Everest::config::SetConfigParameterResultEnum::Rejected:
+            // TODO(CB) : commented, because the test failed otherwise - correct?
```
Pietfried: can be removed?
```suggestion
```

lib/everest/framework/include/utils/config/config_service_core.hpp
```diff
@@ -14,68 +21,155 @@
...
+
+    void process_queue();
+
+    template <typename Func> auto post_to_actor(Func&& f) {
+        using ReturnType = std::invoke_result_t<Func>;
+        auto promise = std::make_shared<std::promise<ReturnType>>();
+        auto future = promise->get_future();
+        command_queue_.push([promise, f = std::forward<Func>(f)]() mutable {
+            if constexpr (std::is_void_v<ReturnType>) {
+                f();
```
Pietfried: We should wrap task execution in try/catch and route exceptions into the promise so that the manager does not terminate on uncaught exceptions

lib/everest/framework/lib/config/config_service_core.cpp
```diff
@@ -3,347 +3,676 @@
...
+    if (worker_thread_.joinable()) {
+        worker_thread_.join();
+    }
+}
+
+void ConfigServiceCore::process_queue() {
+    while (running_) {
+        auto task = command_queue_.wait_and_pop();
+        if (task) {
+            (*task)();
```
Pietfried: Wrap in try/catch, the actor thread should never die

lib/everest/sqlite/lib/everest/database/sqlite/connection.cpp
```diff
@@ -22,31 +22,70 @@ class DatabaseTransaction : public TransactionInterface {
...
         }
     }
 
     void commit() override {
-        const auto retval = this->database.execute_statement("COMMIT TRANSACTION");
+        if (!is_active) {
+            return;
+        }
+        is_active = false;
+        const bool success = this->database.execute_statement("COMMIT TRANSACTION");
```
Pietfried: This sets `is_active = false` and unlocks the mutex before checking the result of `COMMIT TRANSACTION`. On a failed COMMIT the transaction can still be open in SQLITE and the destructor checking `is_active` does not roll back and a new transaction cant be initiated.

On COMMIT failure we should roll back if a transaction is still pending

lib/everest/framework/include/utils/config/types.hpp
```diff
@@ -191,8 +191,8 @@ struct ConfigurationParameterCharacteristics {
     Datatype datatype = Datatype::Unknown;
     Mutability mutability = Mutability::ReadOnly;
     std::optional<std::string> unit;
-    std::optional<int32_t> min_value;
-    std::optional<int32_t> max_value;
+    std::optional<float> min_value;
+    std::optional<float> max_value;
```
Pietfried, line 195: The adl_serializer in config/types.cpp needs to be adjusted, too. Currently its still using int32_t:

```cpp
if (j.contains("min_value")) {
        c.min_value = j.at("min_value").get<int32_t>();
    }
    if (j.contains("max_value")) {
        c.max_value = j.at("max_value").get<int32_t>();
    }
```

lib/everest/framework/lib/config/config_service_core.cpp
cburandt: removed

lib/everest/framework/schemas/migrations/1_up-initial.sql
cburandt: resolved in 9a0ea7b

lib/everest/framework/include/utils/config/config_service_core.hpp
cburandt: done

lib/everest/framework/lib/config/config_service_core.cpp
cburandt: done

lib/everest/framework/include/utils/config/types.hpp
cburandt: changed to float

lib/everest/framework/lib/config/config_service_core.cpp
cburandt: Correct. This should have been removed after a failed attempt to implement it.

lib/everest/framework/lib/config/config_service_core.cpp
cburandt: Tried to implement this. However, there seems to be a problem in MQTTAbstractionImpl::get_internal(...) that made this attempt fail.

lib/everest/sqlite/lib/everest/database/sqlite/connection.cpp
cburandt: Fixed this and reworked it a bit more

lib/everest/framework/lib/config/config_service_core.cpp
Pietfried: For me it's ok to work on this in a follow up PR.

lib/everest/framework/include/utils/config_service_interface.hpp
```diff
@@ -0,0 +1,202 @@
...
+    everest::config::ModuleConfigurations module_configurations;
+};
+
+struct GetConfigParametersResult {
+    GetConfigurationStatus status = GetConfigurationStatus::SlotDoesNotExist;
+    std::vector<std::optional<everest::config::ConfigurationParameter>> parameters;
+};
+
+// ---------------------------------------------------------------------------
+// ConfigServiceInterface
```
Pietfried, line 186: nit: This deserves a proper description

lib/everest/framework/include/utils/config/slot_manager.hpp
```diff
@@ -0,0 +1,92 @@
...
+struct SlotInfo {
+    int id;
+    std::string last_updated;
+    std::optional<std::string> config_file_path;
+    std::optional<std::string> description;
+};
+
+using StoredSlotInfo = SlotInfo;
+
+class SqliteConfigSlotManager {
```
Pietfried, line 48: nit: Deserves a brief description

lib/everest/framework/include/utils/mqtt_config_service.hpp
```diff
@@ -137,25 +164,55 @@ class ConfigServiceClient {
...
+    /// \param config_svc        All domain operations — reads (via get_configuration) and writes (via
+    ///                          set_config_parameters) are routed through this interface.
+    MqttConfigServiceHandler(MQTTAbstraction& mqtt_abstraction, ConfigServiceInterface& config_svc);
+
+    /// \brief This class provides an MQTT interface to modules and uses the ConfigServiceInterface to fulfill incoming
+    /// requests. This function allows to request a config parameter change from a module. This function skips
+    /// validating module_ids and characteristics on purpose. It should only be called with inputs known to work. \param
+    /// cfg_param_id  The identifier for the parameter to be changed \param value         The new configuration
+    /// parameter value as string
+    std::optional<Everest::config::SetResponse>
```
Pietfried: nit: formatting

lib/everest/framework/schemas/migrations/4_up-multi_config.sql
```diff
@@ -0,0 +1,161 @@
...
+);
+INSERT INTO MODULE_CONFIG_ACCESS_NEW
+    SELECT 0, MODULE_ID, OTHER_MODULE_ID, ALLOW_READ, ALLOW_WRITE, ALLOW_SET_READ_ONLY
+    FROM MODULE_CONFIG_ACCESS;
+DROP TABLE MODULE_CONFIG_ACCESS;
+ALTER TABLE MODULE_CONFIG_ACCESS_NEW RENAME TO MODULE_CONFIG_ACCESS;
+
+-- 9. BOOT_CONFIG: single-row table storing next-boot slot selection.
+CREATE TABLE BOOT_CONFIG (
+    NEXT_BOOT_SLOT_ID INTEGER NOT NULL DEFAULT 0
```
Pietfried: This should have an FK to CONFIG(ID)

lib/everest/framework/src/manager.cpp
```diff
@@ -514,11 +506,19 @@ void Manager::handle_restart_modules_after_shutdown(RuntimeContext& ctx) {
...
+        shutdown_cause_ = ShutdownCause::None;
+        shutdown_start_time_ = std::nullopt;
+        force_terminate_start_time_ = std::nullopt;
+        force_kill_sent_ = false;
+        shutdown_info_.clear();
+        EVLOG_info << "Modules restart initiated with reloaded configuration.";
+    } else {
+        EVLOG_error << "Failed to reload the configuration.";
+        config_service_core_->notice_cfg_validation_failed();
+        transition_to(ManagerState::Idle);
```
Pietfried: There is no transition from `FailedToStart` to `Idle`. The [registered transition handler](https://github.com/EVerest/EVerest/pull/2412/changes#diff-179fcfc02a1c98568ecc58607be67905160752c263f6517d80957f2a6e91bb48R831) has no branch `to == Idle` o nothing ever moves the status back. 

Consider adding this branch:
`else if (to == ManagerState::Idle) {
    config_service_core_->set_modules_stopped();
}`

Otherwise `internal_reinitialize_from_db` would return early and integrators could not fix by trying another config.

lib/everest/framework/docs/ConfigService.md
```diff
@@ -0,0 +1,263 @@
...
+Since the ConfigServiceCore is part of the manager process, if the manager crashes, the Configuration API is
+unavailable and modules are also down. On restart, the manager reinitializes the ConfigServiceCore, loads the
+boot_slot and starts the modules.
+
+### Deployment
+The manager is the single long-lived process. It is started by the system init (e.g. via systemd), but does not
+depend on systemd for module lifecycle management. Systemd only ensures the manager itself starts on boot.
+
+No special tooling for production and development deployments is required. Running `./manager --config my_config.yaml`
+starts the manager, imports the YAML as the active slot, and starts modules. There is no distinction between
```
Pietfried: This only works on the first boot and afterwards only with `--reset-from-yaml`

lib/everest/framework/lib/config/slot_manager.cpp
```diff
@@ -0,0 +1,365 @@
...
+    if (meta_stmt->step() != SQLITE_DONE) {
+        return GenericResponseStatus::Failed;
+    }
+
+    transaction->commit();
+    return GenericResponseStatus::OK;
+}
+
+GenericResponseStatus
+SqliteConfigSlotManager::update_config_slot(int slot_id, const std::string& config_dump,
```
Pietfried, line 115: This should also write the `LAST_UDPATED` column

lib/everest/framework/lib/mqtt_config_service.cpp
```diff
@@ -362,57 +436,213 @@ ConfigServiceClient::get_config_value(const everest::config::ConfigurationParame
...
+        auto change_token =
+            std::make_shared<TypedHandler>(HandlerType::ConfigurationRequest, std::make_shared<Handler>(mqtt_handler));
+        const auto topic =
+            fmt::format("{}modules/{}/config/set_request", mqtt_abstraction->get_everest_prefix(), origin);
+
+        mqtt_abstraction->register_handler(topic, change_token, QOS::QOS2);
+    }
+
+    // store handler
+    change_callbacks[impl_id].emplace(param_name, std::move(handler));
```
Pietfried: emplace silently keeps the first handler on reregistration, use insert_or_assign

lib/everest/framework/lib/mqtt_config_service.cpp
```diff
@@ -149,66 +170,119 @@ GetResponse handle_get_all_configs(const std::string& origin, const ManagerConfi
...
     }
 
     get_response.data = all_mappings;
     return get_response;
 }
 
-Response handle_set_request(const SetRequest& set_request, const std::string& origin, ManagerConfig& config) {
+Response handle_set_request(const SetRequest& set_request, const std::string& origin,
+                            const everest::config::ModuleConfigurations& module_configs,
+                            MQTTAbstraction& mqtt_abstraction, Everest::config::ConfigServiceInterface& config_svc) {
```
Pietfried: mqtt_abstraction is not used anymore in this function

lib/everest/framework/lib/config/config_service_core.cpp
```diff
@@ -0,0 +1,705 @@
...
+                                                  const Origin& origin) {
+    SetConfigParameterResult result;
+    result.status = SetConfigParameterStatus::Ok;
+
+    const int resolved_slot_id = (slot_id == ConfigServiceInterface::ACTIVE_SLOT) ? active_slot_id_ : slot_id;
+    const bool modifies_active_slot = resolved_slot_id == active_slot_id_;
+    const bool modules_in_transient =
+        module_status_ != ActiveSlotStatus::Running and module_status_ != ActiveSlotStatus::Stopped;
+
+    result.status = SetConfigParameterStatus::Ok;
```
Pietfried, line 456:
```suggestion
```

Pietfried (commented):
The changes to the command line arguments for the manager process introduce breaking changes against 2026.02 so we should add a migration guide/notice in a follow up.

I'd suggest to change the handling of `--db-init` in this PR.   `--db-init` was removed so any usage of this will cause EVerest not to start. We could keep it as a no-op and add a log line and add it to the 
https://github.com/EVerest/EVerest/blob/main/docs/source/project/releases/deprecation-index.rst .

lib/everest/framework/schemas/migrations/4_up-multi_config.sql
cburandt: Required a non-trivial change; done.

lib/everest/framework/docs/ConfigService.md
cburandt: Changed explanation

lib/everest/framework/lib/mqtt_config_service.cpp
cburandt: Now using insert_or_assign (last one is used, formerly registered handlers dismissed).

lib/everest/framework/lib/mqtt_config_service.cpp
cburandt: Removed from list of args.

lib/everest/framework/lib/config/config_service_core.cpp
cburandt: Removed (was a duplicate)

lib/everest/framework/src/manager.cpp
cburandt: Confirmed - fix is w.i.p

lib/everest/framework/src/manager.cpp
```diff
@@ -998,16 +1181,49 @@ bool Manager::is_idle() const {
...
+                          "Skipping transition to Running.";
+            goto_running_transition = false;
+        }
+    }
+    if (goto_running_transition) {
+        transition_to_unlocked(ManagerState::Running);
+        status_fifo.update(StatusFifo::ALL_MODULES_STARTED);
+        MqttMessagePayload payload{MqttMessageType::GlobalReady, nlohmann::json(true)};
+        mqtt_abstraction.publish(fmt::format("{}ready", mqtt_everest_prefix), payload);
+    }
```
Pietfried, line 1222: Shouldnt this be protected by `state_transition_mutex_` as well?

lib/everest/framework/src/manager.cpp
```diff
@@ -708,59 +786,144 @@ int Manager::run() {
...
+            EVLOG_error << "Cannot dump config: no valid module configuration is available.";
+            return EXIT_FAILURE;
+        }
+        dump_config_and_manifests(*runtime_ctx.config, fs::path(vm_["dump"].as<std::string>()));
+    }
+
+    auto config_service = std::make_unique<config::MqttConfigServiceHandler>(*mqtt_abstraction, *config_service_core_);
+
+    config_service_core_->register_set_runtime_parameter_handler(
+        [&config_service](const everest::config::ConfigurationParameterIdentifier& cfg_param_id,
```
Pietfried, line 860: this captures [&config_service] which is local to the `run` method. The callback might outlive the run() because the `config_service_core` is a Manager member

lib/everest/framework/schemas/migrations/4_up-multi_config.sql
```diff
@@ -0,0 +1,171 @@
...
+
+-- 2. CONFIG_META: update FK from SETTING to CONFIG; add DESCRIPTION column.
+CREATE TABLE CONFIG_META_NEW (
+    ID               INTEGER PRIMARY KEY REFERENCES CONFIG (ID) ON DELETE CASCADE,
+    LAST_UPDATED     TEXT    NOT NULL,
+    CONFIG_DUMP      TEXT    NOT NULL,
+    CONFIG_FILE_PATH TEXT,
+    DESCRIPTION      TEXT
+);
+INSERT INTO CONFIG_META_NEW SELECT ID, LAST_UPDATED, CONFIG_DUMP, CONFIG_FILE_PATH, NULL FROM CONFIG_META;
```
Pietfried, line 40: Migration drops the CONFIG_META.VALID flag and promotes every legacy CONFIG_META row to an existing slot, removing the old contains_valid_config() gate that made an invalid-marked database fall back to re-seeding from YAML.

Can be mitigated by purging invalid configs before step 1:

```
DELETE FROM MODULE          WHERE (SELECT VALID FROM CONFIG_META) = 0;
DELETE FROM CONFIG_META     WHERE VALID = 0;
DELETE FROM SETTING         WHERE NOT EXISTS (SELECT 1 FROM CONFIG_META);
```

lib/everest/framework/docs/ManagerLifecycle.md
Pietfried, line 1: For a follow up: Lets move this into the generated docs/ into explaination and potentially add a how to guide for the manager startup

Pietfried: Should probably be done for most of the documentation here.

lib/everest/framework/include/utils/config_service_interface.hpp
```diff
@@ -0,0 +1,279 @@
...
+
+/// \brief True while a module lifecycle transition is in flight.
+///
+/// Writes to the active slot cannot be resolved in these statuses - it is not yet decided whether a
+/// module will be alive to receive a runtime update - so they are rejected with
+/// SetConfigParameterStatus::ModulesInTransientState and the caller should retry once the modules
+/// have settled. This is the complement of "down or running".
+constexpr bool modules_in_transient_state(ActiveSlotStatus status) {
+    return not modules_are_down(status) and status != ActiveSlotStatus::Running;
+}
```
Pietfried, line 94: nit: The function names are quite self-explainatory. The comments mainly describe how these functions are consumed and what the consequences are, which is not the right place for this.

lib/everest/framework/lib/config/config_service_core.cpp
```diff
@@ -0,0 +1,744 @@
...
+        // (min/max) validation happens at this layer.
+        if (const auto validation_error =
+                ec::validate_config_value(in_memory_parameter->characteristics, update.value)) {
+            result_enum = SetConfigParameterResultEnum::Rejected;
+            per_result.status_info = validation_error.value();
+            continue;
+        }
+
+        // When a persistence mirror is configured (in-memory database deployments), the mirror -
+        // the user-config YAML - is the only persistence that survives a restart, so it is written
```
Pietfried: No need to leak the user config impl here
```suggestion
        // When a persistence mirror is configured (in-memory database deployments), the mirror
        // is the only persistence that survives a restart, so it is written
```

tests/core_tests/RESULTS
Pietfried, line 1: Should probably not be checked in

Pietfried (changes requested):

lib/everest/framework/src/manager.cpp
cburandt: done ("PR-discussion" commits from 2026-08-10)

lib/everest/framework/src/manager.cpp
cburandt: yes - done!

lib/everest/framework/src/manager.cpp
cburandt: now using an RAII guard now to "unregister" the callback when leaving run()

lib/everest/framework/schemas/migrations/4_up-multi_config.sql
cburandt: FKEYS are documented to be disabled during migration. Therefore manually cleaned all the other tables as well.

lib/everest/framework/include/utils/config_service_interface.hpp
cburandt: reduced to the "\brief" part of the comments

lib/everest/framework/lib/config/config_service_core.cpp
cburandt: removed

tests/core_tests/RESULTS
cburandt: committed by accident - removed.

lib/everest/framework/docs/ManagerLifecycle.md
cburandt: The framework docs themselves required an update to match the actual imlementation (done).
Moving to the generated docs in a follow-up.

cburandt:
One aspect of the "Reduce breaking changes" commits from above:

Generated run-scripts do no-longer contain --db and --reset-from-yaml by default, as the original draft of this PR did. Should this be re-introduced?

Pietfried:
> One aspect of the "Reduce breaking changes" commits from above:
> 
> Generated run-scripts do no-longer contain --db and --reset-from-yaml by default, as the original draft of this PR did. Should this be re-introduced?

The run configs should use `--reset-from-yaml` because they are mainly for development purposes where modifications of the yaml should be applied directly

Pietfried (approved):
LGTM

PR #2424, opened 2026-07-08
fix(iso15118): route AC_DER_IEC SECC to DER charge loop

mlitre:
## Describe your changes

SECC-side fixes so an ISO 15118-20 AC_DER_IEC session runs the DER charge loop.

- `power_delivery.cpp`: route AC_DER_IEC sessions to `AC_DER_IEC_ChargeLoop`. The prior `is_ac_charger()` branch matched first and sent DER sessions into the plain AC loop.
- `context_helper.cpp`: add `FAILED_SequenceError` handlers for `DER_AC_ChargeParameterDiscoveryReq` and `DER_AC_ChargeLoopReq`, matching the existing AC/DC entries.
- `types/iso15118.yaml` + `Evse15118D20/charger/utils.hpp`: add the AC DER V2gMessageId values and map DER message types to them for logging.

IEC only; SAE is out of scope.

N/A


lib/everest/iso15118/src/iso15118/d20/state/power_delivery.cpp
```diff
@@ -149,12 +153,12 @@ Result PowerDelivery::feed(Event ev) {
...
 
-        if (m_ctx.session.is_ac_charger()) {
-            return m_ctx.create_state<AC_ChargeLoop>();
-        }
         if (m_ctx.session.is_ac_der_iec_charger()) {
             return m_ctx.create_state<AC_DER_IEC_ChargeLoop>();
         }
+        if (m_ctx.session.is_ac_charger()) {
+            return m_ctx.create_state<AC_ChargeLoop>();
+        }
```
SebaLukas, line 158: Is moving this block really necessary?

SebaLukas (dismissed):
Looking good! I have only one question

lib/everest/iso15118/src/iso15118/d20/state/power_delivery.cpp
mlitre: Yes it is necessary because is_ac_charger will return true if a charger is ac der iec

SebaLukas: That should be not the case. `is_ac_charger()` should be only true if d20 ac is used, not if ac_der_iec or ac_der_sae is active

SebaLukas:
```
[[nodiscard]] bool is_ac_charger() const {
        return selected_services.selected_energy_service == dt::ServiceCategory::AC or
               selected_services.selected_energy_service == dt::ServiceCategory::AC_BPT;
    }
```

mlitre: Ah right then it's fine we don't need to move it

SebaLukas (approved):

Pietfried (approved):

PR #2002, opened 2026-03-25
fix(v16): use configured OcspRequestInterval instead of hardcoded 12h default

rishabhvaish:
## Describe your changes

Reopening https://github.com/EVerest/libocpp/pull/1177 in the monorepo as requested by @Pietfried in [this comment](https://github.com/EVerest/libocpp/pull/1177#issuecomment-4112701687).

The OCSP request timer callback in `ChargePointImpl` constructor uses a hardcoded 12-hour interval (`OCSP_REQUEST_TIMER_INTERVAL`) for recurring OCSP cache updates. This means the user-configured `OcspRequestInterval` value (default: 604800s / 7 days, minimum: 86400s / 1 day) is ignored at startup.

When the CSMS later updates the interval via `ChangeConfiguration`, the correct config-based value is used (line ~1973), but the initial timer setup never reads from configuration.

### Changes

- Replace `OCSP_REQUEST_TIMER_INTERVAL` with `std::chrono::seconds(this->configuration.getOcspRequestInterval())` in the timer callback (consistent with the pattern already used in the `ChangeConfiguration` handler)
- Remove the now-unused `OCSP_REQUEST_TIMER_INTERVAL` constant

## Issue ticket number and link

Fixes https://github.com/EVerest/EVerest/issues/1991

## Checklist

- [x] My changes follow the existing code style
- [x] I have performed a self-review of my code
- [x] New and existing tests pass locally with my changes

Issue #1991, opened 2025-08-12
Configured OCSPRequestInterval is not used at startup


Pietfried:
### OCPP Version

OCPP1.6

### Describe the bug

In OCPP1.6, the hardcoded default interval is used instead of using the configured value:
https://github.com/EVerest/libocpp/blob/main/lib/ocpp/v16/charge_point_impl.cpp#L206

The interval is only updated when explicity set by the CSMS.

### To Reproduce

The hardcoded `OCSP_REQUEST_TIMER_INTERVAL` can be removed. The value from the JSON configuration can be used instead.

Tarekchehahde:
Opened a PR for this: https://github.com/EVerest/EVerest/pull/2852

PR #2428, opened 2026-07-08
fix (ocpp) : OCSP_REQUEST_TIMER_INTERVAL hardcoded value

ThatsLucas:
Hi,

i saw that rishabhvaish began to fix it in https://github.com/EVerest/EVerest/pull/2002 but since he didn't answered since a few month, i thought i could do it by myself.

## Describe your changes
Removed the hardcoded value and use the already existing config value.

https://github.com/EVerest/EVerest/issues/1991

### DIFF
--- a/lib/everest/ocpp/config/v16/config-full.json
+++ b/lib/everest/ocpp/config/v16/config-full.json
@@ -138,20 +138,16 @@
     "CostAndPrice": {
         "CustomDisplayCostAndPrice": true,
         "NumberOfDecimalsForCostValues": 4,
-        "DefaultPrice":
-        {
+        "DefaultPrice": {
             "priceText": "This is the price",
             "priceTextOffline": "Show this price text when offline!",
-            "chargingPrice":
-            {
+            "chargingPrice": {
                 "kWhPrice": 3.14,
                 "hourPrice": 0.42
             }
         },
-        "DefaultPriceText":
-        {
-            "priceTexts":
-            [
+        "DefaultPriceText": {
+            "priceTexts": [
                 {
                     "priceText": "This is the price",
                     "priceTextOffline": "Show this price text when offline!",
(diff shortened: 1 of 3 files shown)


lib/everest/ocpp/lib/ocpp/v16/charge_point_impl.cpp
```diff
@@ -214,7 +213,7 @@ ChargePointImpl::ChargePointImpl(
             };
         this->ocsp_request_timer = std::make_unique<Everest::SteadyTimer>(&this->io_context, [this]() {
             this->update_ocsp_cache();
-            this->ocsp_request_timer->interval(OCSP_REQUEST_TIMER_INTERVAL);
+            this->ocsp_request_timer->interval(std::chrono::seconds(this->configuration.getOcspRequestInterval()));
```
mlitre: Since the configuration can be from the device model now, we access the database which can throw errors, could you maybe try catch the get method and provide a default in case of failure before setting the interval

mlitre (commented):
One comment except for that look good to me

lib/everest/ocpp/lib/ocpp/v16/charge_point_impl.cpp
ThatsLucas: Oh yea, absolutely, i didn't paid attention this was throwing error.

lib/everest/ocpp/config/v16/config-full.json
```diff
@@ -48,7 +48,7 @@
         "UseSslDefaultVerifyPaths": true,
         "VerifyCsmsCommonName": true,
         "VerifyCsmsAllowWildcards": true,
-        "OcspRequestInterval": 604800,
```
mlitre, line 51: Applicable to the entire PR was there a reason for changing the default? If not I think we can keep the current default

ThatsLucas: I tried without changing the ocpp tests and they were not successfull with the old value when i ran them locally.

ThatsLucas: And for this config, i kept the value used in the old hardcoded value.

mlitre: I am wondering then if changing the default is not just hiding another issue

ThatsLucas: I mean, it was using the hardcoded value anyway so i don't think the behavior is changed, from what i understand.
I can change it to 604800 instead of 12h if you think thats better

mlitre: Just thinking it's probably best to leave old default as to avoid "breaking" changes even if it was always hardcoded before

ThatsLucas: alright 👍​, like this ? The config is now the same as before, not 12hours

mlitre (approved):
Lgtm, thanks for your contribution! FYI for future contributions please only squash and rebase after approval. It's easier to follow changes as we review the PR by just adding commits on top of the original

PR #2430, opened 2026-07-08
feat(EnergyNode): add configurable phase rotation for powermeter read…

marcnulens-pionix:
## Describe your changes

Adds a new `phase_rotation` configuration option to the **EnergyNode** module to
correct for a physical L1/L2/L3 wiring rotation at a node's powermeter.

When set, the per-phase voltage/current/power/energy/VAR values of a powermeter
reading are remapped before they are used in the energy flow request, so that the
reported phases align with the actual grid phases:

- `RST` — no rotation (identity mapping); reported L1/L2/L3 already matches grid L1/L2/L3 *(default)*
- `STR` — reported L1 is grid L2, reported L2 is grid L3, reported L3 is grid L1
- `TRS` — reported L1 is grid L3, reported L2 is grid L1, reported L3 is grid L2

Grid frequency is intentionally left untouched, since it is not phase-specific.

Changes:
- New `phase_rotation` config key (enum `RST`/`STR`/`TRS`, default `RST`) in the manifest
- New `energy_grid/phase_rotation_utils.{hpp,cpp}` implementing `apply_phase_rotation()`
- `energyImpl.cpp` applies the rotation to powermeter readings before building the energy flow request
- Unit tests covering all three rotation mappings

Signed-off-by: Marc Nulens <marc.nulens@pionix.de>


modules/EnergyManagement/EnergyNode/EnergyNode.hpp
```diff
@@ -10,6 +10,8 @@
 
 #include "ld-ev.hpp"
 
+#include <string>
```
Pietfried: This should be moved into the custom include header section between L25 and L27

modules/EVSE/EvseManager/tests/CMakeLists.txt
```diff
@@ -81,3 +81,5 @@ target_link_libraries(${CHARGER_TEST_TARGET_NAME} PRIVATE
 
 add_test(${CHARGER_TEST_TARGET_NAME} ${CHARGER_TEST_TARGET_NAME})
 ev_register_test_target(${CHARGER_TEST_TARGET_NAME})
+
+# Phase rotation is now covered by lib/everest/phase_rotation/tests/phase_rotation_utils_test.cpp
```
Pietfried:
```suggestion
```

modules/EVSE/EvseManager/manifest.yaml
```diff
@@ -446,6 +446,20 @@ config:
...
+      by remapping its per-phase voltage/current/power/energy/VAR values before
+      they are used in energy_flow_request.energy_usage_root. 123 means no rotation,
+      i.e. reported L1/L2/L3 already matches grid L1/L2/L3. 312 means reported L1 is
+      grid L2, reported L2 is grid L3, reported L3 is grid L1. 231 means reported L1
+      is grid L3, reported L2 is grid L1, reported L3 is grid L2.
+    type: string
+    enum:
+      - '123'
+      - '312'
+      - '231'
```
Pietfried: OCPP uses a different notation, we have to make sure to add a proper mapping once we integrate it

Pietfried (changes requested):
LGTM, see a few comments in line.

modules/EVSE/EvseManager/manifest.yaml
mnulens: Good point. I switch to using the OCPP phase rotation notation (RST, STR, TRS) instead of (123, 231, 312)

Updated in the code, the tests and the docs.

modules/EnergyManagement/EnergyNode/EnergyNode.hpp
mnulens: Moved the include to the custom header section

modules/EVSE/EvseManager/tests/CMakeLists.txt
mnulens: Removed the line.

Pietfried (approved):

Pietfried (dismissed):

lib/everest/helpers/tests/helpers_test.cpp
```diff
@@ -44,4 +47,149 @@ TEST(HelpersTest, get_base64_id) {
...
+    pm.VAR = types::units::ReactivePower{0.0f, 1.0f, 2.0f, 3.0f};
+    pm.current_A = types::units::Current{std::nullopt, 1.0f, 2.0f, 3.0f, std::nullopt};
+
+    pm.energy_Wh_import_signed =
+        types::units_signed::Energy{std::nullopt, signed_value("L1"), signed_value("L2"), signed_value("L3")};
+    pm.energy_Wh_export_signed =
+        types::units_signed::Energy{std::nullopt, signed_value("L1"), signed_value("L2"), signed_value("L3")};
+    pm.power_W_signed =
+        types::units_signed::Power{std::nullopt, signed_value("L1"), signed_value("L2"), signed_value("L3")};
+    pm.voltage_V_signed =
```
a-w50: This test might be dropped, as it can't guarantee that new added members will also be rotated - because this test then also need to updated.

a-w50 (dismissed):

lib/everest/helpers/tests/helpers_test.cpp
mnulens: Wen adding a new member this requires thinking about the ability or need to rotate it. So the failing test will help to put attention to it. 
Will leave it in for now.

marcnulens-pionix: When adding a new member this requires thinking about the ability or need to rotate it. So the failing test will help to put attention to it.
Will keep it in for now.

a-w50: But how should the test fail?  It doesn't know that there is any new member variable in the powermeter struct, or am I missing something?

marcnulens-pionix: Dropped this test

a-w50 (approved):
already approved it

Pietfried (approved):

PR #2434, opened 2026-07-09
feat(EvseV2G/Evse15118D20): make ports configurable via manifest

mhei:
## Describe your changes

When IsoMux module is used, a common configuration variant is to bind EvseV2G and Evse15118D20 to loopback interface and only the IsoMux to the real CP powerline interface. 
On multiport chargers such a setup is currently not possible because multiple EvseV2G instance and Evse15118D20 instances would try to allocate the same ports on the loopback interface.
To circumvent this, allow configuration of ports via manifest for each module.
For backwards compatibility, the new config options have the previous static values as defaults.

none


mlitre (commented):
In my iso15118 teardown PR I created a connection factory that can do exactly what you are doing for the tcp port is it not worth it to integrate that?

SebaLukas:
What is the status here @mhei ?

PR #2444, opened 2026-07-09
Fix buffer overflows in EXI bitstream parser

lippit:
## Describe your changes
Full description in cbExiGen:
https://github.com/EVerest/cbexigen/pull/129

**This PR only regenerated auto-generated .c files from cbExiGen templates.**

## Fixed issues
exi_bitstream_has_overflow() only advanced and validated byte_pos when bit_count reached a full byte (EXI_BITSTREAM_MAX_BIT_COUNT). It never checked that the byte about to be accessed was within the stream before exi_bitstream_read_bit and exi_bitstream_write_bit dereference data[byte_pos]. This allowed out-of-bounds access.

V2GTP20_ReadHeader() exhibited undefined behavior. When left-shifting uint8_t bytes from stream_data, each operand is implicitly promoted to signed int. Shifting a value into the sign bit (e.g. stream_data[4] << 24) is undefined behavior in C, which can lead to incorrect payload length and payload id values depending on the compiler.

### DIFF
--- a/lib/everest/cbv2g/lib/cbv2g/common/exi_bitstream.c
+++ b/lib/everest/cbv2g/lib/cbv2g/common/exi_bitstream.c
@@ -35,10 +35,11 @@ static int exi_bitstream_has_overflow(exi_bitstream_t* stream)
             stream->byte_pos++;
             stream->bit_count = 0;
         }
-        else
-        {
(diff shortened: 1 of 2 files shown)


mlitre:
@barsnick should we merge this or regenerate from the latest cbexigen and open a new PR?

PR #2445, opened 2026-07-09
feat(Evse15118D20): wire AC_DER_IEC grid_support

mlitre:
## Describe your changes

Adds the device/producer side of DER grid support to Evse15118D20 for the AC_DER_IEC service.

- Provides the `grid_support` interface: accepts active DER directives and raises grid alarms from
  EV-reported grid-event conditions. A rising/falling edge detector emits exactly one `GridAlarm`
  per fault edge (deduplicating repeated conditions) and only advances its state after a successful
  publish, so a dropped alarm is retried rather than lost.
- Routes real AC_DER_IEC transfer limits to the EV instead of the previous hardcoded values. The IEC
  DER limits pass through from the module's AC transfer limits; nominal and max discharge are set
  equal so the advertised pair satisfies V2G20-3229, and the discharge sign convention
  (ISO 15118-20 8.3.5.2.1) is carried through from the AC limits.
- Surfaces the EV's DER charging parameters (reactive-power limits, session discharge energy) into
  `ChargingNeeds`.
- Surfaces the DER control functions the EV negotiates at service selection into
  `ev_supported_dercontrol`, so the backend can learn what the EV supports without the EVSE having
  to advertise a capability it cannot know up front in the AC_DER_IEC case.

This is the device/producer half; it pairs with the OCPP-consumer side on
`feature/grid-support-der-dynamic`.

Out of scope, tracked as a follow-up: relaying grid directives back down onto the EV's DER control
functions. That requires the EV-supported set (added here) plus a source for the nominal-voltage and
reactive-power bases needed to encode ISO 15118-20's absolute-unit control curves, which this module
does not currently hold. This module does not publish a DER capability; that is sourced elsewhere.

### Testing

- New module unit tests (`everest-core_Evse15118D20_tests`): DER transfer-limit sign, clamp and
  pass-through; DER charging-parameter mapping (reactive power, per-phase, session energy);
  grid-event edge detection (rising/falling/dedup/reset/retry); and the EV-supported-control-function
  mapping. `ctest -R Evse15118D20` passes.

### Config / migration impact

- None for existing setups. Adds a provided `grid_support` interface to the module manifest.

_n/a_

### DIFF
--- a/modules/EVSE/Evse15118D20/CMakeLists.txt
+++ b/modules/EVSE/Evse15118D20/CMakeLists.txt
@@ -10,6 +10,8 @@ ev_setup_cpp_module()
 # ev@bcc62523-e22b-41d7-ba2f-825b493a3c97:v1
 target_sources(${MODULE_NAME}
     PRIVATE
+        charger/der_setup.cpp
+        charger/grid_event.cpp
         charger/session_logger.cpp
         charger/utils.cpp
 )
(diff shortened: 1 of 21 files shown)


SebaLukas (approved):
Looking good!

PR #2446, opened 2026-07-10
feat(Evse15118D20): relay DER directives to EV

mlitre:
## Describe your changes

Stacked on #2445. Relays backend `grid_support` DER directives onto the EV's AC_DER_IEC control
functions, completing the follow-up called out there.

- Evse15118D20 converts curve-family DER directives (VoltVar, WattVar, WattPF) to AC_DER_IEC
  control functions: the percentage curves are converted to absolute values against the EVSE
  voltage, power, and reactive-power bases and pushed into the ISO 15118-20 stack. Directives take
  effect at the next V2G session and are re-applied whenever the bases change; non-curve directives
  are logged and skipped.
- EvseManager: new optional `ac_max_reactive_power` config, passed via `update_ac_parameters` as
  the reactive-power base for the conversion (new optional `evse_max_reactive_power` field on
  `AcParameters`).
- libiso: `evse_setup` is now guarded by a monitor, closing a pre-existing race between the config
  update commands and session construction.

### Testing

- Module unit tests green (`ctest -R Evse15118D20`): 18 directive-to-curve mapping cases covering
  denormalization per unit, missing-base and empty-curve skips, point truncation, and unmapped
  curve features.
- libiso suite green after the monitor change (47/47), including the AC_DER_IEC charge parameter
  discovery and charge-loop tests.

### Config / migration impact

- None for existing setups. Adds an optional `ac_max_reactive_power` config to EvseManager (absent
  means reactive-power curves are not relayed).

_n/a_


modules/EVSE/Evse15118D20/charger/ISO15118_chargerImpl.cpp
```diff
@@ -174,6 +174,54 @@ void ISO15118_chargerImpl::init() {
...
+    const auto directives = mod->get_active_der_directives();
+    if (not directives.has_value()) {
+        return;
+    }
+
+    // Snapshot the GEL-protected setup_config fields under the lock, then release it before touching the
+    // controller: update_*_der_functions reach into the evse_setup monitor (a second lock), and holding GEL
+    // across that would risk lock inversion. controller is set once in ready() and lives for the impl lifetime,
+    // so the snapshotted pointer stays valid after the lock is released.
+    iso15118::TbdController* controller_ptr = nullptr;
```
SebaLukas, line 194: I am not so happy with this approach. But I can't think of any other alternative. Maybe we should take a look at mutex and thread handling in general later on.

SebaLukas (dismissed):
Looking good!

modules/EVSE/Evse15118D20/charger/ISO15118_chargerImpl.cpp
mlitre: Yeah I looked into this and I couldn't find a better solution either. Definitely need to look into this further

SebaLukas (approved):

cburandt (approved):

PR #2451, opened 2026-07-13
feat(evse_manager): add StoppingTransaction session event

ManuelJanocha:
## Describe your changes

Emit StoppingTransaction right before stop_transaction(), once the contactor is confirmed open, so consumers can reliably tell that the transaction is really ending before the final meter value is taken.

This event is not Eichrecht-relevant by itself, but it provides the missing hook to implement Eichrecht-compliant (German calibration law) switch/measurement coordination checks ("Schaltmesskoordination"): StoppingTransaction does not guarantee that the relay/contactor is physically open — it marks the point at which this can be verified. Consumers can use this checkpoint to check the actual relay state before the final (calibrated) meter reading is taken, and to prevent the transaction stop if the relay is not confirmed open.


modules/EVSE/EvseManager/Charger.cpp
```diff
@@ -1028,6 +1028,12 @@ void Charger::run_state_machine() {
                 // Transaction may already be stopped when it was cancelled earlier.
                 // In that case, do not sent a second transactionFinished event.
                 if (shared_context.flag_transaction_active) {
+                    // We only ever reach Finished with an active transaction after StoppingCharging has
+                    // confirmed shared_context.contactor_open, so the relais/contactor is safely open (power
+                    // off) at this point and the transaction is really ending (not resumable). Signal this
+                    // before writing the stop to the power meter so external consumers (e.g.
+                    // evse_manager_consumer_API) can observe it.
```
Pietfried: I'd try to avoid leaking consumer specifics (evse_manager_consumer_API) in the comment, but rather highlight **why** the event is relevant for consumers. You could e.g. reuse the second section of the PR description (Schaltmesskoordination)

Pietfried (changes requested):
In general this looks good to me. Please adress the following + one comment in line: 
- In order to resolve the unit test issue, you need to update the hash of `types/evse_manager.yaml` in https://github.com/EVerest/EVerest/blob/main/lib/everest/everest_api_types/tests/expected_types_file_hashes.csv . See https://github.com/EVerest/EVerest/tree/main/lib/everest/everest_api_types#automatic-detection-of-changes-of-everest-types-and-interfaces for more info.
- https://github.com/EVerest/EVerest/blob/main/modules/EVSE/Auth/lib/AuthHandler.cpp#L885 : extend the switch by adding the new enum value
- Please consider extending https://github.com/EVerest/EVerest/blob/main/modules/EVSE/EvseManager/tests/ChargerTest.cpp to test if `StoppingTransaction` is emitted on entering `Finished` with an active transaction

docs/source/reference/EVerest_API/evse_manager_consumer_API.yaml
```diff
@@ -2574,6 +2574,7 @@ components:
         - ChargingPausedEV: Signaled when charging is paused by the EV
         - ChargingPausedEVSE: Signaled when charging is paused by the EVSE
         - StoppingCharging: EVSE has started to stop the charging process. DC: CurrentDemand has finished, now doing WeldingCheck etc, AC: Wait for car to return to state B or A
+        - StoppingTransaction: Signaled right before the transaction is stopped in the power meter. At this point the transaction is really ending (not resumable) and the contactor/relais is confirmed open (power is off).
```
florinmihut, line 2577: I am not sure if the relays are confirmed open when the signal is emitted.

modules/EVSE/EvseManager/Charger.cpp
```diff
@@ -1064,6 +1064,14 @@ void Charger::run_state_machine() {
...
                 // In that case, do not sent a second transactionFinished event.
                 if (shared_context.flag_transaction_active) {
+                    // We only ever reach Finished with an active transaction after StoppingCharging has
+                    // confirmed shared_context.contactor_open, so the relais/contactor is safely open (power
+                    // off) at this point and the transaction is really ending (not resumable). Signal this
+                    // before writing the stop to the power meter: StoppingTransaction is not Eichrecht-relevant
+                    // by itself, but it is the checkpoint at which consumers can verify the relay/contactor
+                    // state before the final (calibrated) meter reading is taken, enabling Eichrecht-compliant
+                    // switch/measurement coordination ("Schaltmesskoordination").
+                    signal_simple_event(types::evse_manager::SessionEventEnum::StoppingTransaction);
```
florinmihut, line 1074: I am not 100% sure that the relays are confirmed open when the signal is emitted.

It seems that at least theoretically it is possible to reach this point with the contactor_open false. 
We should at least test it and gate emitting the event. 
I seems that a AC HLC charge session with the following events could get in Finished with the contactors still closed: PWM 5%, contactor closed, D-LINK_ERROR is triggered (session will be restarted), T_step_EF can exit with an error, WaitingForAuthentication (exit with internal error), goes directly into Finished. If the contactor is really open or still closed it is a race. 

The PR just does an assumption, but it doesn't check it. I would recommend however to check (or wait for the contactor to be open) before sending the event to avoid race conditions.

florinmihut (changes requested):

modules/EVSE/EvseManager/Charger.cpp
florinmihut: I've saw conflicting information:
- the PR description says: "StoppingTransaction does not guarantee that the relay/contactor is physically open"
- the event description in the PR says: "StoppingTransaction: Signaled right before the transaction is stopped in the power meter. At this point the transaction is really ending (not resumable) and the contactor/relais is confirmed open (power is off)."

Which one is true?

PR #2453, opened 2026-07-13
Refactor(Evse15118D20): Removing unused session logger

SebaLukas:
## Describe your changes
- Removed SessionLogger in Evse15118D20 and libiso15118
- To debug exi messages we typically use wireshark + v2g plugin
- Switched to another log mechanism to log enter state events

`SessionLogger` hasn't been actively used for two or more years, so it will be removed.


mlitre:
This looks okay to me, we might need some wording on the config to let people know it is deprecated and probably follow the deprecation policy? @Pietfried ?

Pietfried:
> This looks okay to me, we might need some wording on the config to let people know it is deprecated and probably follow the deprecation policy? @Pietfried ?

I dont know the details here and maybe I miss something, but afaics no changes to the [public API](https://everest.github.io/latest/project/releases/releases-and-versioning.html#public-api-definition) are part of this, so I dont see there is any need for deprecation.

SebaLukas:
I intentionally did not remove a specific config option `logging_path` from the manifest in this PR. Otherwise, it could potentially break some configurations.
However, I will add a warning stating that the option should no longer be used and will be removed in the future.

Pietfried:
> I intentionally did not remove a specific config option `logging_path` from the manifest in this PR. Otherwise, it could potentially break some configurations. However, I will add a warning stating that the option should no longer be used and will be removed in the future.

Sounds good. We plan to introduce some tooling that automates this soon, e.g. in the manifest.yaml you can mark a config parameter with a property like `deprecated: true` or something like `status: deprecated/active/experimental`

mlitre:
> > I intentionally did not remove a specific config option `logging_path` from the manifest in this PR. Otherwise, it could potentially break some configurations. However, I will add a warning stating that the option should no longer be used and will be removed in the future.
> 
> Sounds good. We plan to introduce some tooling that automates this soon, e.g. in the manifest.yaml you can mark a config parameter with a property like `deprecated: true` or something like `status: deprecated/active/experimental`

I mainly wanted to know if it should be marked as so because we didn't remove the config, but the functionality behind is now gone

Pietfried:
> > > I intentionally did not remove a specific config option `logging_path` from the manifest in this PR. Otherwise, it could potentially break some configurations. However, I will add a warning stating that the option should no longer be used and will be removed in the future.
> > 
> > 
> > Sounds good. We plan to introduce some tooling that automates this soon, e.g. in the manifest.yaml you can mark a config parameter with a property like `deprecated: true` or something like `status: deprecated/active/experimental`
> 
> I mainly wanted to know if it should be marked as so because we didn't remove the config, but the functionality behind is now gone

Yes that was apparently what I missed while scanning the PR. Was this SessionLogger ever working with the respective config option enabled?

SebaLukas:
It is running in the background every time you started the d20 module. The config option is only the path in which the session file is saved. The default path is `.`

Pietfried (changes requested):
Please also see the compilation failed in the CI

SebaLukas:
@Pietfried I updated the branch. Also I removed the entry in the deprecation table

Pietfried (approved):

PR #2454, opened 2026-07-13
refactor(libiso15118): Refactored message_20::Variant to a std::unique_ptr

SebaLukas:
## Describe your changes
See title.
Changed void pointer and custom_deleter to a `std::unique_ptr<void, CustomDeleter>` in the `Variant` class.

The `Variant` class in `libiso15118` used raw pointers with manual memory management, which is error-prone and not exception-safe.

### DIFF
--- a/lib/everest/iso15118/include/iso15118/message/variant.hpp
+++ b/lib/everest/iso15118/include/iso15118/message/variant.hpp
@@ -2,8 +2,7 @@
 // Copyright 2023 Pionix GmbH and Contributors to EVerest
 #pragma once
 
-#include <cstddef>
-#include <cstdint>
+#include <functional>
 #include <memory>
 #include <stdexcept>
 #include <string>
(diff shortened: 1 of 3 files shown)


lib/everest/iso15118/include/iso15118/message/variant.hpp
```diff
@@ -18,29 +17,27 @@ namespace iso15118::message_20 {
...
-        data = new MessageType;
-        *static_cast<MessageType*>(data) = in;
-        custom_deleter = [](void* ptr) { delete static_cast<MessageType*>(ptr); };
-        type = message_20::TypeTrait<MessageType>::type;
+        *static_cast<MessageType*>(data.get()) = in_;
     }
-    ~Variant();
-
-    Type get_type() const;
+    ~Variant() = default;
```
mlitre: We should drop the default as it suppresses the implicit moves as we don't follow the rule of zero. The unique ptr generates the correct operations already

SebaLukas: Done!

mlitre (approved):

PR #2455, opened 2026-07-13
OCPP 1.6: add option to always report cleared errors to the CSMS

nzbr:
## Describe your changes

Previously, when multiple errors were active at the same time, clearing one of them did not result in any StatusNotification being sent as long as the charge point stayed Faulted, so the CSMS was never informed which errors had been resolved.
This PR changes the OCPP 1.6 state machine to send a StatusNotification for every cleared error, reporting the cleared error's `vendorId` and `vendorErrorCode` together with an `info` value of "Error <vendorErrorCode> resolved", while the `errorCode` still reflects the latest error that remains active. Also adds a test for clearing multiple errors out of order.

---

This PR is part of a stack

- `2` #2456
- `1` #2455 👈 you are here


nzbr:
I will fix the DCO check when squashing before this gets merged

lib/everest/ocpp/lib/ocpp/v16/charge_point_state_machine.cpp
```diff
@@ -246,20 +252,33 @@ bool ChargePointFSM::handle_error_cleared(const std::string uuid) {
...
     std::optional<CiString<50>> vendor_error_code;
 
-    // report the latest error if there are still errors active
+    if (this->report_cleared_errors && !node.empty()) {
+        // Report the cleared error as resolved
+        auto cleared_error = std::move(node.mapped());
+        if (cleared_error.vendor_error_code.has_value()) {
+            info =
+                std::string("Error ") + std::string(cleared_error.vendor_error_code.value()) + std::string(" resolved");
+        }
```
nzbr, line 246: This is _one_ option for how to report which error has been cleared, but I am open for different suggestions as well

lohralfen: I'm not sure if CSMS developers are eager to try and parse the info field to allow for automated processing of error codes for CPOs, as it is rather free form, maybe someone else knows? However the only alternative I can think of is to use a CSV list of active error codes in the vendorErrorCode field and to report the "remaining active error codes", so they always have the latest state of which error codes are active, and can deduce which one got solved. I'm not sure if this is desirable in terms of OCPP specifications, but that way from a single message you could see what the error state of the charger is, rather than needing the history of StatusNotification messages. What do you think?

Although when using MREC or EVerest-specific error codes/names, the 50 character limit could in some cases become a pain point.

lib/everest/ocpp/lib/ocpp/v16/charge_point_state_machine.cpp
```diff
@@ -246,20 +252,33 @@ bool ChargePointFSM::handle_error_cleared(const std::string uuid) {
     std::optional<CiString<255>> vendor_id;
     std::optional<CiString<50>> vendor_error_code;
 
-    // report the latest error if there are still errors active
+    if (this->report_cleared_errors && !node.empty()) {
+        // Report the cleared error as resolved
+        auto cleared_error = std::move(node.mapped());
+        if (cleared_error.vendor_error_code.has_value()) {
+            info =
+                std::string("Error ") + std::string(cleared_error.vendor_error_code.value()) + std::string(" resolved");
```
Pietfried: This can potentially throw if info exceeds 50 characters so you should use `StringTooLarge::Truncate` in the CiString ctor. I'd also recommend to drop "Error " to save some characters.

```
info = CiString<50>(cleared_error.vendor_error_code.value().get() + " resolved",
                    StringTooLarge::Truncate);

```

Please also use `StringTooLarge::Truncate` also for vendorId and vendorErrorCode to be on the safe side

lib/everest/ocpp/tests/lib/ocpp/v16/test_charge_point_state_machine.cpp
Pietfried, line 1: These tests use errors that do not define any vendorId or vendorErrorCode and so the ` ... resolved` info is never verified in the tests. Worth adding an explicit test to verify the behavior.

Pietfried (changes requested):

lib/everest/ocpp/lib/ocpp/v16/charge_point_state_machine.cpp
Pietfried: This is how [MREC](https://inl.gov/chargex/mrec/) actually recommends to implement multiple simultaneous faults:

```json
{
"connectorId": 1,
"errorCode": "OtherError",
"info": "50;105",
"status": "Faulted",
"timestamp": "2023-09-06T00-08-09Z",
"vendorId": "www.vendor.com;https://chargex.inl.gov",
"vendorErrorCode": "CustomError;CX019"
}
```

We decided against it because
- of the character limits of the properties
- it is clearer when a StatusNotification can be unambiguously associated with either an error event or the resolution of an error.

lib/everest/ocpp/tests/lib/ocpp/v16/test_charge_point_state_machine.cpp
nzbr: I added a test for that

Pietfried (approved):

PR #2456, opened 2026-07-13
Send `message` error property from the stable API to the CSMS

nzbr:
## Describe your changes

Sends the `message` that can be set when raising an error via the stable APIs to the CSMS via the `info` field on an MREC error or `vendorId` otherwise (this was the case before) for OCPP 1.6 and in the `techInfo` field for OCPP 2.x

---

This PR is part of a stack

- `2` #2456 👈 you are here
- `1` #2455

### DIFF
--- a/docs/source/reference/EVerest_API/ev_board_support_API.yaml
+++ b/docs/source/reference/EVerest_API/ev_board_support_API.yaml
@@ -449,7 +449,10 @@ components:
           description: The subtype of the error
           type: string
         message:
-          description: Addition information about the error
+          description: |-
+            Optional additional information about the error
+            OCPP 1.6: Sets `info` on MREC errors and `vendorId` otherwise
+            OCPP 2.x: Sets `techInfo`
           type: string
 
     CommunicationCheck:
--- a/docs/source/reference/EVerest_API/ev_board_support_API.yaml
+++ b/docs/source/reference/EVerest_API/ev_board_support_API.yaml
@@ -297,7 +297,7 @@ components:
       examples:
         - summary: "Report cable ampicity of 32A and some RCD current + PWM duty cycle"
           payload:
-            proximity_pilot:  
+            proximity_pilot:
               "ampacity": "A_32"
             rcd_current_mA: 5
             cp_pwm_duty_cycle: 53
@@ -411,7 +411,7 @@ components:
         - cp_pwm_duty_cycle
         - proximity_pilot
       properties:
-        proximity_pilot:  
+        proximity_pilot:
           description: Current capability of the cable
           type: object
           $ref: 'evse_board_support_API.yaml#/components/schemas/ProximityPilot'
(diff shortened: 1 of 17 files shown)


modules/EVSE/OCPP/OCPP.cpp
```diff
@@ -55,7 +55,11 @@ static ocpp::v16::ErrorInfo get_error_info(const Everest::error::Error& error) {
     if (mrec_it != MREC_ERROR_MAP.end()) {
         // lambda to create MREC error info
         auto make_mrec_error_info = [&](ocpp::v16::ChargePointErrorCode code, const std::string& vendor_error_code) {
-            return ocpp::v16::ErrorInfo{uuid, code, false, std::nullopt, CHARGE_X_MREC_VENDOR_ID, vendor_error_code};
+            std::optional<std::string> info = std::nullopt;
+            if (!error.message.empty()) {
+                info = error.message;
+            }
+            return ocpp::v16::ErrorInfo{uuid, code, false, info, CHARGE_X_MREC_VENDOR_ID, vendor_error_code};
```
Jaleant, line 66: Suppose we do not want to show the MREC error, but our own number. Is that possible now? Can we both overwrite the mrec error code and the info field?

lohralfen: And the same goes for non-MREC EVerest-specific errors. When using vendor-specific error codes, it would likely be needed to be able to 'override' all of these error codes/descriptions. But that's probably outside the scope of this specific work package.

nzbr: When this is merged you can set the `info` field freely. The `vendor_id` is always `https://chargex.inl.gov` for MREC errors and the `vendor_error_code` is the kind of MREC error that was raised. Which types you can raise depends on which of the stable APIs you are using to raise the error through. It is in theory also possible to supply a custom mapping from EVerest error codes to MREC error codes that replaces the one in lib/everest/ocpp_module_common/include/everest/ocpp_module_common/.v16/error_mapping.hpp

docs/source/reference/EVerest_API/ev_board_support_API.yaml
```diff
@@ -449,7 +449,10 @@ components:
           description: The subtype of the error
           type: string
         message:
-          description: Addition information about the error
+          description: |-
+            Optional additional information about the error
+            OCPP 1.6: Sets `info` on MREC errors and `vendorId` otherwise
+            OCPP 2.x: Sets `techInfo`
```
Pietfried: I stumbled a bit over this documentation and it wasn't quite clear to me how the message actually influences the StatusNotification/NotifyEvent. Heres a slightly more detailed suggestion, should be applied to all api docs:

```suggestion
description: |-
            Optional free-text detail about this specific occurrence of the error,
            e.g. a measured value or a diagnostic string from the hardware. Unlike
            the fixed per-error-type description, this is chosen by the caller.

            If an OCPP module is part of the EVerest configuration, the message is
            forwarded to the CSMS:

            * OCPP 1.6 (`StatusNotification.req`):
              * MREC errors: sent as `info`, truncated to 50 characters.
                Omitted entirely when the message is empty.
              * all other errors: sent as `vendorId`, truncated to 255 characters.
            * OCPP 2.x (`NotifyEvent.req`): sent as `techInfo`, truncated to 500
              characters. When the message is empty, the error description is sent
              instead.
```

modules/EVSE/OCPP/docs/index.rst
```diff
@@ -354,10 +354,12 @@ This module currently deviates from the MREC specification in the following poin
 * Simultaneous errors: MREC requires reporting simultaneous errors by reporting them in a single **StatusNotification.req** by separating 
   the information of the fields **vendorId** and **info** by a semicolon. This module sends one **StatusNotification.req** per individual error 
   because of the limited maximum characters of the **info** field.
-* MREC requires always using the value **Faulted** for the **status** field when reporting an error. The OCPP1.6 specification defines the 
-  **Faulted** value as follows: "When a Charge Point or connector has reported an error and is not available for energy delivery.  
+* MREC requires always using the value **Faulted** for the **status** field when reporting an error. The OCPP1.6 specification defines the
+  **Faulted** value as follows: "When a Charge Point or connector has reported an error and is not available for energy delivery.
```
Pietfried, line 358: A similar section is missing in OCPP201 and OCPPmulti

Pietfried (changes requested):

Pietfried (approved):

PR #2467, opened 2026-07-15
Bump project version to 2026.10.0 and pin everest-cmake to v0.7.0

hikinggrass:
## Describe your changes
This will automatically append "-dev" to the compiled-in project version if not on a "stable" or "release" branch


SebaLukas:
Should we use this PR to update the versions in libs, etc.?

hikinggrass:
> Should we use this PR to update the versions in libs, etc.?

Good idea, althought that could also be in a follow-up

SebaLukas:
> Good idea, althought that could also be in a follow-up

A follow-up is fine :+1:

Pietfried (approved):

andistorm (approved):
🐧

barsnick (commented):
Didn't we agree that the version should report itself as "-dev"?

hikinggrass:
> Didn't we agree that the version should report itself as "-dev"?

Yes that happens when everest-cmake 0.7.0+ is used: https://github.com/EVerest/everest-cmake/commit/29fdd49cb9db7c4867f8e6aac6d810e79ad5b2a6

The [CMake version format](https://cmake.org/cmake/help/v3.16/command/project.html) itself unfortunately doesn't allow such suffixes :/

barsnick (approved):

PR #2470, opened 2026-07-15
EvseManager: skip calculation of limits if no. of phases is unknown

### LINKED ISSUE #2469: EvseManager: crash / division by zero due to race between enforce_limits and capabilities
### Describe the bug

I believe there is a race condition between the setting of the number of supported phases.

## Summary
Delayed reception of number of phases from the BSP can lead to use of the original zero-initialized value when calculating limits, incl. a division by zero.

```
[ERRO] connector:EvseM  :: Energy manager requests switching #ph from 0 to 3, but switching phases during charging is not supported by HW.
[CRIT] manager          :: Module connector (pid: 17211) exited with status: 11. Terminating all modules.
```

## Analysis
The BSP sends the number of supported phases via `evse_board_support`'s var `capabilities::min_phase_count_export`/`capabilities::max_phase_count_export`.

EvseManager initializes its `ac_nr_phases_active` to zero:
https://github.com/EVerest/EVerest/blob/39fc544c594e1a9245c2b361297b0394a8706c49/modules/EVSE/EvseManager/EvseManager.hpp#L293
and then clamps it to the BSP's min/max from the capabilities, once the BSP sends them (usually in the module's `evse_board_support` interface's `ready()`):
https://github.com/EVerest/EVerest/blob/39fc544c594e1a9245c2b361297b0394a8706c49/modules/EVSE/EvseManager/EvseManager.cpp#L276-L282
In other words, `ac_nr_phases_active` remains zero until the BSP's capabilities are received.

At the same time, EvseManager's `energy` interface provides `handle_enforce_limits()`, usually fed from EnergyNode or EnergyManager. There, a (copy of) `ac_nr_phases_active` is used:
https://github.com/EVerest/EVerest/blob/39fc544c594e1a9245c2b361297b0394a8706c49/modules/EVSE/EvseManager/energy_grid/energyImpl.cpp#L391

Its value is checked:
https://github.com/EVerest/EVerest/blob/39fc544c594e1a9245c2b361297b0394a8706c49/modules/EVSE/EvseManager/energy_grid/energyImpl.cpp#L401-L402

Should the value still be zero, because EnergyManager has managed to start updating the limits, but the BSP not yet sent it capabilities, an error is emitted (because 0 happens to always be out of the phase switching range):
https://github.com/EVerest/EVerest/blob/39fc544c594e1a9245c2b361297b0394a8706c49/modules/EVSE/EvseManager/energy_grid/energyImpl.cpp#L415-L418
**BUT**: The value is used a few lines below in a division:
https://github.com/EVerest/EVerest/blob/39fc544c594e1a9245c2b361297b0394a8706c49/modules/EVSE/EvseManager/energy_grid/energyImpl.cpp#L427-L428
leading to a crash of the module (or best case undefined behavior).

Indeed, in certain setups, our BSP's `ready()` phase is delayed (waiting for something essential to happen), and we can get this crash:
```
[ERRO] connector:EvseM  :: Energy manager requests switching #ph from 0 to 3, but switching phases during charging is not supported by HW.
[CRIT] manager          :: Module connector (pid: 17211) exited with status: 11. Terminating all modules.
```

### Expected behavior

No crash.

### EVerest Domain

Energy Management

### Affected Component

EvseManager

### EVerest Version

2026.02.0, main/HEAD

### To Reproduce

Delay BSP's `publish_capabilities()`.

### Anything else?

I'm not sure what would be the best fix (and we need one quickly 😉):

1. As a missing `ac_nr_phases_active` gives no meaningful limit calculation, skip the complete `handle_enforce_limits()` by exiting early after a sanity check of `ac_nr_phases_active`.
2. Skip phase switch handling, Watt-to-Ampere conversion and `signalNrOfPhasesAvailable()` in `handle_enforce_limits()`.
3. Some fallback if the phase count is invalid (e.g. set max current to 0, or retain previous valid max current value - if there ever was one.

barsnick:
The number of phases needs to be set via the `evse_board_support` interface before using them.

Fixes a potential division by zero.

Fixes https://github.com/EVerest/EVerest/issues/2469 .

## Describe your changes

Exit hander early if no. of phases is not valid.

Tested with SIL and YetiSimulator, the latter with an additional forced delay before publishing its capabilities.

Fixes:
https://github.com/EVerest/EVerest/issues/2469

### DIFF
--- a/modules/EVSE/EvseManager/energy_grid/energyImpl.cpp
+++ b/modules/EVSE/EvseManager/energy_grid/energyImpl.cpp
@@ -387,8 +387,19 @@ void energyImpl::handle_enforce_limits(types::energy::EnforcedLimits& value) {
         // EVLOG_info << "Incoming enforce limits" << value;
 
         //   set hardware limit
+        const int active_phasecount = mod->ac_nr_phases_active;
+
+        static bool warning_shown{false};
+        if (active_phasecount == 0) {
+            if (not warning_shown) {
+                EVLOG_warning << "Number of active phases is still uninitialized, skipping limits calculation "
+                                 "while waiting for BSP";
+                warning_shown = true;
+            }
+            return;
+        }
+
         float limit = 0.;
-        int active_phasecount = mod->ac_nr_phases_active;
 
         // apply enforced limits
 


modules/EVSE/EvseManager/energy_grid/energyImpl.cpp
```diff
@@ -387,8 +387,13 @@ void energyImpl::handle_enforce_limits(types::energy::EnforcedLimits& value) {
         // EVLOG_info << "Incoming enforce limits" << value;
 
         //   set hardware limit
+        const int active_phasecount = mod->ac_nr_phases_active;
+        if (active_phasecount == 0) {
+            EVLOG_warning << "Number of activee phases is still uninitialized, skipping limits calculation "
```
Pietfried:
```suggestion
            EVLOG_warning << "Number of active phases is still uninitialized, skipping limits calculation "
```

barsnick:
I have fixed the typo and made the warning appear only once.

Pietfried (approved):

hikinggrass (approved):

PR #2471, opened 2026-07-15
fix(ocpp): make v2 DeviceModel thread-safe

Pietfried:
## Describe your changes

Serialize access to device_model_map and the device_model storage with an internal recursive mutex. Reads lock in the request_value_internal choke point; report/monitor/metadata readers, mutators and the register_* listener setters lock directly. Recursive because locked methods re-enter public methods (e.g. via DeviceModelInterface& composites).

Add a concurrency smoke test hammering the read/write/listener paths from multiple threads.

### DIFF
--- a/lib/everest/ocpp/lib/ocpp/v2/device_model.cpp
+++ b/lib/everest/ocpp/lib/ocpp/v2/device_model.cpp
@@ -290,6 +290,8 @@ GetVariableStatusEnum DeviceModel::get_variable(const Component& component_id, c
 GetVariableStatusEnum DeviceModel::request_value_internal(const Component& component_id, const Variable& variable_id,
                                                           const AttributeEnum& attribute_enum, std::string& value,
                                                           bool allow_write_only) const {
+    // Choke point for all value reads (get_variable and the get_*_value/request_value templates).
+    std::lock_guard<std::recursive_mutex> lock(this->mutex);
     const auto component_it = this->device_model_map.find(component_id);
     if (component_it == this->device_model_map.end()) {
         EVLOG_debug << "unknown component in " << component_id.name << "." << variable_id.name;
@@ -322,6 +324,7 @@ GetVariableStatusEnum DeviceModel::request_value_internal(const Component& compo
 
 std::optional<MutabilityEnum> DeviceModel::get_mutability(const Component& component, const Variable& variable,
                                                           const AttributeEnum& attribute_enum) {
+    std::lock_guard<std::recursive_mutex> lock(this->mutex);
     const auto attribute = this->device_model->get_variable_attribute(component, variable, attribute_enum);
     if (!attribute.has_value()) {
         return std::nullopt;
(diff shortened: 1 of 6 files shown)


lib/everest/ocpp/lib/ocpp/v2/device_model.cpp
```diff
@@ -333,6 +336,7 @@ std::optional<MutabilityEnum> DeviceModel::get_mutability(const Component& compo
 SetVariableStatusEnum DeviceModel::set_value(const Component& component, const Variable& variable,
                                              const AttributeEnum& attribute_enum, const std::string& value,
                                              const std::string& source, bool allow_read_only) {
+    std::lock_guard<std::recursive_mutex> lock(this->mutex);
```
mlitre: I think we should reduce the scope of this lock. There is a potential deadlock with consumer code from the variable_listner. I suggest we copy variable_listener under the lock and the release the lock and then call them. This also applies to clear_value and set_monitors.
With the ocpp module it should be a problem because ocppImpl::handle_monitor_variables takes monitor_list_mutex and then register_variable_listener takes DeviceModel::mutex and if we call this set_value code path then we take DeviceModel::mutex then the listener (ocppImpl::variable_changed) takes monitor_list_mutex.

lib/everest/ocpp/tests/lib/ocpp/v2/test_device_model_concurrency.cpp
mlitre, line 1: These are good tests, but it didn't test for the potential deadlock mentioned previously that's probably why it was missed

lib/everest/ocpp/lib/ocpp/v2/device_model.cpp
hikinggrass, line 456: You could store the `attribute.value()` here temporarily, so you don't have to access it by `.value()` multiple times later on

hikinggrass: const auto& attribute_value = attribute.value();

lib/everest/ocpp/lib/ocpp/v2/device_model.cpp
```diff
@@ -451,13 +463,17 @@ SetVariableStatusEnum DeviceModel::clear_value(const Component& component, const
...
+        const std::string value_previous = attribute.value().value.value_or(EMPTY_VALUE);
         if (value_previous != empty) {
-            for (const auto& listener : variable_listener) {
-                listener(monitors, component, variable, variable_map[variable].characteristics, attribute.value(),
-                         value_previous, empty);
+            // Copy listeners and payload, then release the lock before invoking: see set_value.
+            const auto listeners = variable_listener;
+            const auto monitors = variable_map[variable].monitors;
+            const auto characteristics = variable_map[variable].characteristics;
+            const auto attribute_value = attribute.value();
```
hikinggrass: If doing the previous suggestion you could do:
```suggestion
        const std::string value_previous = attribute_value.value.value_or(EMPTY_VALUE);
if (value_previous != empty) {
            // Copy listeners and payload, then release the lock before invoking: see set_value.
            const auto listeners = variable_listener;
            const auto monitors = variable_map[variable].monitors;
            const auto characteristics = variable_map[variable].characteristics;
```

hikinggrass (dismissed):

hikinggrass (approved):
flaky test?

PR #2475, opened 2026-07-16
OCPP websocket reconnect and reset fixes

mlitre:
## Describe your changes

OCPP 2.0.1 websocket reconnect and reset-window fixes from OCTT
certification, plus two teardown races found while testing them.

- Honor the section 5.3 RetryBackOff on every re-dial, not just inside
  the per-profile attempt loop (TC_B_49_CS).
- Fall back to the last-successful network profile once the priority
  list is exhausted; N connection attempts now means exactly N.
- Persist ActiveNetworkProfile on a confirmed connection (not on dial)
  and seed the post-reboot fallback from it.
- Suppress auto-reconnect during the pre-reset window of an accepted
  immediate Reset without closing the socket (TC_A_10, TC_B_45, TC_B_22).
- Slot-1 MessageTimeout default 60s; per-slot MessageTimeout now hot
  applies on SetVariables (TC_E_41/42).
- Wake the recv thread on interrupt and join self-exited worker threads
  on close, so a failed dial cannot stall or terminate the process.
- Disarm connection callbacks in ChargePoint::stop(), fixing a
  pre-existing teardown use-after-free.

### Verification
- Full libocpp unit test suite green (1376 passed); teardown race
  reproduced (2-core taskset) and clean over 50 repeats after the fix.

### DIFF
--- a/lib/everest/ocpp/lib/ocpp/v2/charge_point.cpp
+++ b/lib/everest/ocpp/lib/ocpp/v2/charge_point.cpp
@@ -162,6 +162,9 @@ void ChargePoint::stop() {
     this->availability->stop_heartbeat_timer();
     this->provisioning->stop_bootnotification_timer();
     this->connectivity_manager->disconnect();
+    // Disarm before tearing down the message queue: a late websocket-thread disconnect callback
+    // must not reach into members being destroyed.
+    this->connectivity_manager->disarm_connection_callbacks();
     this->security->stop_certificate_expiration_check_timers();
     this->diagnostics->stop_monitoring();
     this->message_queue->stop();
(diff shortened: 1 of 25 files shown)


lib/everest/ocpp/lib/ocpp/v2/charge_point.cpp
```diff
@@ -162,6 +162,9 @@ void ChargePoint::stop() {
     this->availability->stop_heartbeat_timer();
     this->provisioning->stop_bootnotification_timer();
     this->connectivity_manager->disconnect();
+    // Disarm before tearing down the message queue: a late websocket-thread disconnect callback
+    // must not reach into members being destroyed.
+    this->connectivity_manager->disarm_connection_callbacks();
```
Pietfried, line 167: Should probably also be done for v16.

mlitre: Done

Pietfried (approved):

PR #2477, opened 2026-07-16
fix(OCPP201): report connector availability on ChangeAvailability(Inoperative)

biubiuzhang:
## Describe your changes

`OCPP201`'s `connector_effective_operative_status_changed_callback` gated `charge_point->on_unavailable()` behind the return value of `call_enable_disable(..., Disable)`:

```cpp
} else {
    if (this->r_evse_manager.at(evse_id - 1)
            ->call_enable_disable(connector_id, {..., Enable_state::Disable, 5000})) {
        this->charge_point->on_unavailable(evse_id, connector_id);
    }
}
```

But `EvseManager::enable_disable()` returns the **resulting effective enabled state** of the connector (the combined decision across all enable/disable sources), not a success flag — see `Charger::enable_disable()` (`return is_enabled;`). A *successful* disable therefore returns `false`, the guard fails, and `on_unavailable()` is never called, so **no `StatusNotification` with `connectorStatus = Unavailable` is sent** in response to a `ChangeAvailability` request that sets an EVSE/connector `Inoperative`. The `Operative` branch happened to work because enable returns `true`.

This change reads `call_enable_disable()`'s return as what it is — the effective enabled state — and reports connector availability from it in both directions: a disable now emits `on_unavailable()` (Unavailable) and an enable emits `on_enabled()` (Available). As a side benefit, a CSMS `Operative` request that is overruled by a higher-priority disable source is now correctly reported as Unavailable rather than Available.

Per OCPP 2.0.1 an accepted `ChangeAvailability` must be reflected by a `StatusNotification` reporting the new connector status; today a CSMS setting a connector `Inoperative` receives `Accepted` but never sees it transition to `Unavailable`.

N/A — no pre-existing upstream issue. Found during OCPP 2.0.1 remote-control/availability verification against a CSMS.

- [ ] I have made corresponding changes to the documentation (no doc changes needed — behavioral bug fix)

## Testing

Verified in SIL (native build of this branch, `config-sil-ocpp201`, against an OCPP 2.0.1 CSMS, on a healthy simulated connector):

- **Before (stock main):** `ChangeAvailability(Inoperative)` for an EVSE → `Accepted`, but **no `StatusNotification`** — the connector never reports `Unavailable`.
- **After (this change):** `ChangeAvailability(Inoperative)` → `StatusNotification(connectorStatus = Unavailable)`; `ChangeAvailability(Operative)` → `StatusNotification(connectorStatus = Available)`.

### DIFF
--- a/modules/EVSE/OCPP201/OCPP201.cpp
+++ b/modules/EVSE/OCPP201/OCPP201.cpp
@@ -564,18 +564,22 @@ void OCPP201::ready() {
 
     callbacks.connector_effective_operative_status_changed_callback =
         [this](const int32_t evse_id, const int32_t connector_id, const ocpp::v2::OperationalStatusEnum new_status) {
-            if (new_status == ocpp::v2::OperationalStatusEnum::Operative) {
-                if (this->r_evse_manager.at(evse_id - 1)
-                        ->call_enable_disable(connector_id, {types::evse_manager::Enable_source::CSMS,
-                                                             types::evse_manager::Enable_state::Enable, 5000})) {
-                    this->charge_point->on_enabled(evse_id, connector_id);
-                }
+            const auto enable_state = (new_status == ocpp::v2::OperationalStatusEnum::Operative)
+                                          ? types::evse_manager::Enable_state::Enable
+                                          : types::evse_manager::Enable_state::Disable;
+            // call_enable_disable() returns the resulting effective enabled state of the connector
+            // (the combined decision across all enable/disable sources), not a success flag. Report the
+            // connector availability to the CSMS based on that effective state. Previously the Disable
+            // branch guarded on_unavailable() with this return value, which is false for a successful
+            // disable, so no StatusNotification (connectorStatus = Unavailable) was ever sent in response
+            // to a ChangeAvailability(Inoperative).
+            const bool is_enabled =
+                this->r_evse_manager.at(evse_id - 1)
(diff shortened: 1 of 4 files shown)


Pietfried (changes requested):
Hi @biubiuzhang , thanks for the late response to this one. The change looks reasonable. Could you please integrate this change also inside the new OCPPmulti module and add respective test coverage for this?

biubiuzhang:
Thanks @Pietfried — done. Rebased onto current `main`, and `OCPPmulti` now carries the same fix.

`GenericOcpp::cb_connector_effective_operative_status` had the same shape: it guarded `on_unavailable()` with the return value of `call_enable_disable()`, which `interfaces/evse_manager.yaml` defines as the connector's *effective* state after the command ("true if evse is enabled after the command … may not be the same value as the request"), not a success flag — so a successful disable returned `false` and never reported `connectorStatus = Unavailable`.

It now reports the effective state and returns whether the requested state was reached. That also fixes the v16 path, where `cb_disable_evse()` previously signalled a successful disable as a failure (libocpp ignores the value, so nothing else changes).

Tests in `evse_managerIntf_tests.cpp`: `callEnableDisable` updated (it asserted the old interpretation), plus new cases for a higher priority source overriding the request and for an invalid `evse_id`. Suite: 171 passed, 3 skipped (pre-existing), 0 failed.

PR #2479, opened 2026-07-17
OCPP: reject SignCertificate TriggerMessage when it cannot be fulfilled

mlitre:
## Describe your changes

Rebased onto `main` and rescoped to the CSR trigger fix only. The
`AuthorizeRemoteStart` change has been dropped, so this no longer depends on
#2475.

`TriggerMessage` for `SignChargingStationCertificate` was answered `Accepted`
unconditionally and then went silent when the device-model CSR inputs were
unset, so the CSMS waited for a `SignCertificate.req` that never came. F06.FR.10
requires that "Messages that the Charging Station marks as Accepted SHALL be
sent". Both signing uses are now gated on a shared check.
`SignV2GCertificate` keeps its `V2GCertificateInstallationEnabled` check, which
still short-circuits first.

The same violation was reachable a second way: a `TriggerMessage` arriving while
a `CertificateSigned.req` was awaited was accepted, but the send path drops a
request in that state. The gate now covers that too, so the trigger is rejected
instead. F06.FR.10 constrains only what is marked `Accepted`, so rejecting is
the spec-clean answer while a request is genuinely in flight.

Rejecting is only acceptable because the awaited state can no longer wedge.
`handle_sign_certificate_response` returned with
`awaited_certificate_signing_use_enum` still set when `CertSigningWaitMinimum`
or `CertSigningRepeatTimes` was unset. No retry timer is armed on those paths,
so nothing cleared it again and every later `SignCertificate.req` was dropped
for the lifetime of the process. Every path that stops awaiting a
`CertificateSigned.req` now goes through one `reset_certificate_signing_state()`.
With that fixed, the awaited state lives only for a bounded retry window.

CSR device-model reads are extracted into one `get_csr_inputs` helper shared by
the gate and the send path, so the two cannot disagree.

Every rejection now carries its reason to the CSMS in
`TriggerMessageResponse.statusInfo` rather than only to the log, using OCPP 2.1
Appendix 5 reason codes: `MissingDevModelInfo` with every missing device model
input listed in `additionalInfo`, `Unspecified` for an awaited
`CertificateSigned.req`, and `NotEnabled` for the pre-existing
`V2GCertificateInstallationEnabled` check.

### Dropped since the last review

@Pietfried was right that the spec does not require authorizing a remote start
through the CSMS when `AuthorizeRemoteStart` is true. `F01.FR.20/21` do not
concern authorization in edition 3 or 4, and the F01 postcondition permitting the
Local Authorization List and Authorization Cache is present verbatim in edition
3, so this was not an edition difference. The `TC_F_04_CS` reference was also a
misattribution; that test fails for an unrelated `EVConnectTimeout` stop-reason
issue on the EvseManager side.

Also dropped: the earlier revision superseded the in-flight request on a trigger
rather than rejecting it. Fixing the wedge at its root removes the reason that
existed, and superseding orphaned the previous request.

Separately, while removing the `AuthorizeRemoteStart` arm I found a pre-existing
data race on `TransactionBlock::remote_start_id_per_evse`: written from the
message-handler thread via `set_remote_start_id_for_evse`, read and erased from
the module thread via `ChargePoint::on_authorized` /
`on_charging_state_changed`. Unrelated to this PR, fixed separately.

### Verification

- `ctest -R libocpp`: 7/7 suites passed, including `libocpp_unit_tests`.
- `libocpp_test_security` 35/35, `libocpp_test_remote_transaction_control` 11/11,
  `libocpp_unit_tests` 1404 passed / 8 skipped.
- Making a device-model value genuinely absent is not expressible through
  `set_value`, which rejects an empty integer, so the two wedge tests build their
  device model with `CertSigningWaitMinimum` / `CertSigningRepeatTimes` deleted
  and assert that a later `SignCertificate.req` is still sent. Both fail against
  the previous revision.
- `get_csr_inputs` is behavior-preserving: the existing
  `sign_certificate_request_no_country` / `_no_organization_name` /
  `_no_serial_number` / `_v2g_no_*` tests are unmodified and still pass, as do
  `sign_certificate_request_twice` and `handle_sign_certificate_response_backoff_*`,
  confirming the non-triggered and retry paths are untouched.

### DIFF
--- a/lib/everest/ocpp/lib/ocpp/v2/functional_blocks/authorization.cpp
+++ b/lib/everest/ocpp/lib/ocpp/v2/functional_blocks/authorization.cpp
@@ -173,6 +173,19 @@ ocpp::v2::Authorization::validate_token(const IdToken id_token, const std::optio
         this->context.device_model.get_optional_value<bool>(ControllerComponentVariables::DisableRemoteAuthorization)
             .value_or(false);
 
+    // F01.FR.20 / F01.FR.21: with AuthorizeRemoteStart enabled a remote start must be authorized by
+    // the CSMS, so it must not be served from the Local Authorization List or Authorization Cache. Only bypass while
+    // online; offline the CSMS is unreachable and a cached-Accepted token must still be honored.
+    const bool bypass_local_authorization_for_remote_start =
+        is_online and
+        this->context.device_model.get_optional_value<bool>(ControllerComponentVariables::AuthorizeRemoteStart)
+            .value_or(false) and
+        this->is_token_awaiting_remote_start and this->is_token_awaiting_remote_start(id_token);
+    if (bypass_local_authorization_for_remote_start) {
+        EVLOG_info << "Remote start authorization with AuthorizeRemoteStart enabled: bypassing Local Authorization "
+                      "List and Authorization Cache and sending AuthorizeRequest";
+    }
+
     // C07: Authorization using contract certificates
     if (id_token.type == IdTokenEnumStringType::eMAID) {
         // Temporary variable that is set to true to avoid immediate response to allow the local auth list
@@ -292,7 +305,8 @@ ocpp::v2::Authorization::validate_token(const IdToken id_token, const std::optio
     const bool local_pre_authorize =
         this->context.device_model.get_value<bool>(ControllerComponentVariables::LocalPreAuthorize);
     const bool can_locally_check = ((is_online and local_pre_authorize) or (!is_online and local_authorize_offline));
-    if (this->context.device_model.get_optional_value<bool>(ControllerComponentVariables::LocalAuthListCtrlrEnabled)
+    if (!bypass_local_authorization_for_remote_start and
+        this->context.device_model.get_optional_value<bool>(ControllerComponentVariables::LocalAuthListCtrlrEnabled)
             .value_or(false) and
         can_locally_check) {
         std::optional<IdTokenInfo> id_token_info = std::nullopt;
@@ -329,7 +343,7 @@ ocpp::v2::Authorization::validate_token(const IdToken id_token, const std::optio
     const auto hashed_id_token = utils::generate_token_hash(id_token);
     const auto auth_cache_enabled = this->is_auth_cache_ctrlr_enabled();
 
-    if (auth_cache_enabled and can_locally_check) {
+    if (!bypass_local_authorization_for_remote_start and auth_cache_enabled and can_locally_check) {
         try {
             const auto cache_entry = this->authorization_cache_get_entry(hashed_id_token);
             if (cache_entry.has_value()) {
@@ -404,6 +418,11 @@ ocpp::v2::Authorization::validate_token(const IdToken id_token, const std::optio
 
     response.idTokenInfo.status = AuthorizationStatusEnum::Unknown;
     return response;
+}
+
+void ocpp::v2::Authorization::set_remote_start_pending_check(
+    std::function<bool(const IdToken& id_token)> is_token_awaiting_remote_start) {
+    this->is_token_awaiting_remote_start = std::move(is_token_awaiting_remote_start);
 }
 
 void ocpp::v2::Authorization::stop_auth_cache_cleanup_thread() {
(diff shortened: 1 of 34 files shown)


lib/everest/ocpp/lib/ocpp/v2/functional_blocks/authorization.cpp
```diff
@@ -173,6 +173,19 @@ ocpp::v2::Authorization::validate_token(const IdToken id_token, const std::optio
         this->context.device_model.get_optional_value<bool>(ControllerComponentVariables::DisableRemoteAuthorization)
             .value_or(false);
 
+    // F01.FR.20 / F01.FR.21: with AuthorizeRemoteStart enabled a remote start must be authorized by
```
Pietfried: These requirements do not seem related to the issue the code tries to solve.

Pietfried (changes requested):
The CSR handling looks good.

For AuthorizeRemoteStart, I can't find anything in the spec that requires to Authorize through the CSMS when AuthorizeRemoteStart is true, it is rather the opposite:

```
If the value of AuthorizeRemoteStart is true, the
Charging Station will only offer energy when it successfully authorized the IdToken, using Local
Authorization List, Authorization Cache and/or an AuthorizeRequest.
```

F01.FR.01 states a similar requirement.

mlitre:
> The CSR handling looks good.
> 
> For AuthorizeRemoteStart, I can't find anything in the spec that requires to Authorize through the CSMS when AuthorizeRemoteStart is true, it is rather the opposite:
> 
> ```
> If the value of AuthorizeRemoteStart is true, the
> Charging Station will only offer energy when it successfully authorized the IdToken, using Local
> Authorization List, Authorization Cache and/or an AuthorizeRequest.
> ```
> 
> F01.FR.01 states a similar requirement.

You are right regarding the AuthorizeRemoteStart, I'll remove that part. I think I found the underlying issue I was trying to solve and it is mainly an issue with the fact that I was looking at an older edition of the spec which was missing some clarifications

lib/everest/ocpp/lib/ocpp/v2/functional_blocks/security.cpp
```diff
@@ -161,61 +161,93 @@ void Security::security_event_notification_req(const CiString<50>& event_type,
...
+        if (!initiated_by_trigger_message) {
+            EVLOG_warning
+                << "Not sending new SignCertificate.req because still waiting for CertificateSigned.req from CSMS";
+            return;
+        }
+        // A TriggerMessage is an explicit CSMS instruction to send a SignCertificate.req now, and F06.FR.10 requires a
+        // message marked Accepted to actually be sent. The awaited state can go stale: when CertSigningWaitMinimum or
+        // CertSigningRepeatTimes is unset, handle_sign_certificate_response leaves it set with no retry timer, which
+        // would silence every later request. Supersede the in-flight request instead of dropping the trigger.
+        EVLOG_info << "TriggerMessage received while awaiting CertificateSigned.req: superseding in-flight "
```
Pietfried, line 220: I would prefer to just Reject the TriggerMessage when a certificate signing request is in flight.

We should still clear the wedge though in the early returns in handle_sign_certificate_response .

lib/everest/ocpp/lib/ocpp/v2/functional_blocks/security.cpp
```diff
@@ -161,61 +161,93 @@ void Security::security_event_notification_req(const CiString<50>& event_type,
...
+        this->awaited_certificate_signing_use_enum = std::nullopt;
+        // The superseded attempt chain is abandoned, so the trigger starts a fresh one. Without this a wedged-high
+        // counter would immediately terminate the new chain in handle_sign_certificate_response.
+        this->csr_attempt = 1;
+    }
+
+    const auto csr_inputs = this->get_csr_inputs(certificate_signing_use);
+    if (!csr_inputs.has_value()) {
         return;
     }
```
Pietfried, line 232: We already cleared awaited_certificate_signing_use_enum and bail out here, so any old request is now orphaned. As mentioned in the other comment, I'd just reject the TriggerMessage if a request is in flight.

Pietfried (changes requested):

Pietfried (approved):

PR #2480, opened 2026-07-17
OCPP opt-in firmware download deferral

mlitre:
## Describe your changes

Rebased onto `main`; no longer stacked on #2475 / #2479.

New opt-in `InternalCtrlr.DeferFirmwareDownloadDuringTransaction` (default
`false`). When enabled and a transaction is active, an accepted `UpdateFirmware`
is stashed, `DownloadScheduled` is sent with the requestId, and idle connectors
go Unavailable; the download starts when the last transaction ends (L01.FR.13,
TC_L_13_CS). The deferred request is held in memory only, so after a restart in
`DownloadScheduled` the CSMS must re-send `UpdateFirmware`.

It ships `ReadOnly`, making it a commissioning-time setting. A deployment that
wants the CSMS to control it declares the attribute `ReadWrite` in its own
component config; libocpp picks that up on the next boot.

Also fixes the `AllowNewSessionsPendingFirmwareUpdate` lookup, where a stray
copy-pasted instance name made every Get/SetVariables miss the variable (a
TC_L_13_CS precondition). Note this only makes the variable reachable, nothing
consumes it yet, so it does not gate new sessions. No device-model migration is
needed: variable identity includes the instance, so the stale row is deleted and
the correct one inserted on boot.

### Addressing @Pietfried's review

On bypassing the `update_firmware_request_callback`: keeping this in libocpp
matches existing precedent in the same function. `firmware_update.cpp:89-95`
already defers the **install** phase on transaction state, with a `FIXME(Kai)`
recording that the System module cannot track transactions, driven by TC_L_15_CS.
Forwarding download to System while install stays here would split one feature
across two owners. Agreed that the unification is the cleaner end state, so it is
recorded as follow-up rather than done here. Deferral being opt-in also means
chargers that can download during a transaction are unaffected, per your point
about TC_L_13_CS not applying to them.

Your comment did surface a real gap: a request superseding a deferred one was
answered `Accepted` and dropped the pending one silently, leaving the CSMS with a
requestId stuck in `DownloadScheduled`. L01.FR.24 requires `AcceptedCanceled`, so
that is now reported (second commit). The optional `FirmwareStatusNotification`
for the canceled request is deliberately not sent: `DownloadFailed` is an end
state and would restore connectors that a re-deferral immediately disables again.

### Verification

- `ctest` in a libocpp-only build: 158/158 passed.
- `FirmwareUpdate` suite 11/11, covering the gate on/off, deferral while another
  transaction is still active, release on last transaction end, a rejected release
  reporting `DownloadFailed`, an invalid-certificate request leaving a deferred
  one intact, and the `AcceptedCanceled` cases including that a first request is
  not reported as canceled and that a callback rejection is not upgraded.

### DIFF
--- a/lib/everest/ocpp/lib/ocpp/v2/functional_blocks/firmware_update.cpp
+++ b/lib/everest/ocpp/lib/ocpp/v2/functional_blocks/firmware_update.cpp
@@ -120,6 +120,33 @@ void FirmwareUpdate::on_firmware_status_notification_request() {
     this->context.message_dispatcher.dispatch_call(call, true);
 }
 
+void FirmwareUpdate::on_transaction_finished() {
+    std::optional<UpdateFirmwareRequest> request;
+    {
+        std::lock_guard<std::mutex> lock(this->deferred_update_firmware_request_mutex);
+        if (this->deferred_update_firmware_request.has_value() and
+            !this->context.evse_manager.any_transaction_active(std::nullopt)) {
+            request = this->deferred_update_firmware_request;
+            this->deferred_update_firmware_request.reset();
+        }
+    }
+    if (request.has_value()) {
+        const UpdateFirmwareResponse response = this->update_firmware_request_callback(request.value());
+        if (response.status != UpdateFirmwareStatusEnum::Accepted) {
+            EVLOG_warning << "Deferred firmware update was not accepted after transaction finished: "
+                          << conversions::update_firmware_status_enum_to_string(response.status);
+            this->on_firmware_update_status_notification(request.value().requestId, FirmwareStatusEnum::DownloadFailed);
+        }
+    }
+}
+
+bool FirmwareUpdate::is_download_deferral_required() const {
+    return this->context.device_model
+               .get_optional_value<bool>(ControllerComponentVariables::DeferFirmwareDownloadDuringTransaction)
+               .value_or(false) and
+           this->context.evse_manager.any_transaction_active(std::nullopt);
+}
+
 void FirmwareUpdate::handle_firmware_update_req(Call<UpdateFirmwareRequest> call) {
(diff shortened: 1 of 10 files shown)


lib/everest/ocpp/lib/ocpp/v2/functional_blocks/firmware_update.cpp
```diff
@@ -141,16 +168,41 @@ void FirmwareUpdate::handle_firmware_update_req(Call<UpdateFirmwareRequest> call
...
+                    deferral_required = false;
+                }
+            }
+        }
+        if (deferral_required) {
+            // L01.FR.13: busy charging, defer the download until the last transaction ends.
+            response.status = UpdateFirmwareStatusEnum::Accepted;
+            download_deferred = true;
+        } else {
+            response = update_firmware_request_callback(msg);
```
Pietfried, line 197: This callback and therefore the module implementing the actual firmware update is now completely bypassed here. This workaround feels a little awkward and I am not sure if we want to accept that.

`TC_L_13_CS` should not apply for chargers that a capable of downloading while a transaction is active.

mlitre: Maybe I should have implemented this as a flag for the system module? I just thought CPOs would appreciate it if this was configurable on their side to change user experience. It allows us to more easily test different edge cases by centralizing the logic and having other modules be "dumb"

Pietfried: Yes I understand the intent and since it is not the default behavior I guess we can accept it here. Forwarding to system would be cleaner, but you also have to make sure to signal to system then that transactions are now over so the Download can start - adds more complexity. I'd leave it up to you.

Pietfried (dismissed):

Pietfried (approved):

PR #2481, opened 2026-07-17
OCPP: no transient Available on inoperative connector

mlitre:
## Describe your changes

Derive the reported connector status from the effective operational
status, so a scheduled or applied Inoperative connector cannot report a
transient Available when a transaction ends (TC_G_11/14/17). The status
precedence is now stated in one place.

### Verification
- Full libocpp unit test suite green.

### DIFF
--- a/lib/everest/ocpp/lib/ocpp/v2/component_state_manager.cpp
+++ b/lib/everest/ocpp/lib/ocpp/v2/component_state_manager.cpp
@@ -6,6 +6,32 @@
 
 namespace ocpp::v2 {
 
+namespace {
+/// \brief Single definition of the connector-status precedence:
+/// faulted > inoperative > unavailable > occupied > reserved > available.
+/// \p effective_operational_status is the connector's effective operational status; pass Operative to skip the
+/// inoperative rung (used by the raw session-flag mapping which has no operational-status concept).
+ConnectorStatusEnum derive_connector_status(OperationalStatusEnum effective_operational_status,
+                                            const FullConnectorStatus& status) {
+    if (status.faulted) {
+        return ConnectorStatusEnum::Faulted;
+    }
+    if (effective_operational_status == OperationalStatusEnum::Inoperative) {
+        return ConnectorStatusEnum::Unavailable;
+    }
+    if (status.unavailable) {
+        return ConnectorStatusEnum::Unavailable;
+    }
+    if (status.occupied) {
+        return ConnectorStatusEnum::Occupied;
+    }
+    if (status.reserved) {
+        return ConnectorStatusEnum::Reserved;
+    }
+    return ConnectorStatusEnum::Available;
+}
+} // namespace
+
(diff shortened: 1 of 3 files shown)


lib/everest/ocpp/lib/ocpp/v2/component_state_manager.cpp
```diff
@@ -229,20 +254,8 @@ void ComponentStateManager::set_connector_individual_operational_status(std::int
 }
 
 ConnectorStatusEnum FullConnectorStatus::to_connector_status() const {
```
Pietfried: This function is not used anymore

lib/everest/ocpp/lib/ocpp/v2/component_state_manager.cpp
```diff
@@ -255,11 +268,17 @@ OperationalStatusEnum ComponentStateManager::get_evse_effective_operational_stat
 ConnectorStatusEnum ComponentStateManager::get_connector_effective_status(std::int32_t evse_id,
                                                                           std::int32_t connector_id) {
     this->check_evse_and_connector_id(evse_id, connector_id);
+    const FullConnectorStatus& connector = this->individual_connector_status(evse_id, connector_id);
+    // The EVSE or CS being Inoperative hides every other connector state, including faults.
```
Pietfried: Now Unavailable has precedence over Faulted, which changes the behavior that is currently implemented and I also can't find that this is clearly required by the spec: G03.FR.06:

> When the availability of an EVSE becomes Inoperative (Unavailable, Faulted) → All operative connectors (i.e. not Faulted) of that EVSE SHALL become Unavailable.

Scenario: CSMS sets EVSE 1 Inoperative -> StatusNotification(Unavailable) sent and cached. A ground fault then occurs -> set_connector_faulted(1,1,true) -> derived status is still Unavailable -> equals last_reported_status -> no notification, and update_dm_availability_state is skipped. Before this PR the CSMS got Faulted.

Afaics the spec is not really clear here, but I would prefer to keep the existing behavior of Faulted having precedence over Unavailable.

Fix is trivial: Delete the early return in L273.

Its worth adding tests for the Unavailable+Faulted combination.

Pietfried (changes requested):

lib/everest/ocpp/lib/ocpp/v2/component_state_manager.cpp
mlitre: Yeah that's a good idea, I made some changes in this direction. I didn't do your trivial fix because it would still cause the initial issue. I also added some tests

lib/everest/ocpp/lib/ocpp/v2/component_state_manager.cpp
mlitre: Changed

Pietfried (approved):

PR #2484, opened 2026-07-17
SIL modules and device-model config fixes

mlitre:
## Describe your changes

SIL module and device-model config fixes from OCTT certification.

- EvseManager: a granted authorization with no EV connection before the
  connection timeout now ends the transaction with
  StopTransactionReason::Timeout instead of a generic stop (TC_F_04_CS).
- EvManager: preserve the simulated plug state on reset completion so
  later commands act on the real line, and fail fast when waiting on a
  V2G session that never started.
- Config: trim advertised measurands (values and defaults) to what the
  SIL powermeter produces (TC_J_01/02/09); seed
  ISO15118Ctrlr.CountryName=DE (ReadWrite) to unblock CSR generation
  (TC_M_23_CS); align NetworkConnectionProfiles messageTimeout with the
  device model (TC_E_41/42).

### Verification
- EvseManagerCharger and EvManager ctest suites green; JSON validated.

### DIFF
--- a/lib/everest/ocpp/config/common/component_config/standardized/AlignedDataCtrlr.json
+++ b/lib/everest/ocpp/config/common/component_config/standardized/AlignedDataCtrlr.json
@@ -65,7 +65,7 @@
         {
           "type": "Actual",
           "mutability": "ReadWrite",
-          "value": "Energy.Active.Import.Register,Voltage,Frequency"
+          "value": "Energy.Active.Import.Register,Voltage"
         }
       ],
       "description": "Clock-aligned measurand(s) to be included in MeterValuesRequest, every AlignedDataInterval seconds.",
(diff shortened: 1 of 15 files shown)


Pietfried (changes requested):
**OCPP Changes:**
The changes to the OCPP configuration look line for me.

**EvManager Changes:**
Since those changes seem unrelated, can you split this into a separate PR?
@SebaLukas is probably best to have a look at this. 

**EvseManager Changes:**
The changes allow a `TransactionFinished` event to be published without any previous `TransactionStarted` event and they are bypassing the path through `stop_transaction()` which other code paths take to stop transactions. The issue that this is trying to solve is purely OCPP specific and afaics can also be handled inside the OCPP201/OCPPmulti transaction handling / process_deuathorized. My suggestion is to also move this into a separate PR using a different approach.

PR #2485, opened 2026-07-19
feat(iso15118): DIN SPEC 70121 and ISO 15118-2 SECC support in the C++ stack

corneliusclaussen:
## Describe your changes

Adds SECC-side **DIN SPEC 70121** and **ISO 15118-2** support to the C++ ISO 15118 stack (`lib/everest/iso15118`), so `Evse15118D20` can run all three protocol generations. Until now the C++ stack only spoke ISO 15118-20; DIN and -2 sessions required the C-based `EvseV2G` module.

### Library (`lib/everest/iso15118`)

- **Protocol-neutral plumbing** (457983ce3): a `V2gMessageType` variant over the -20, -2 and DIN message type enums so feedback reports real pre-20 message names, `ProtocolId` + namespace constants, `IConnection::is_secure()` to gate plaintext-only protocols, and transport robustness fixes found on the way (fatal socket errors reported as would-block, transient `accept4`/`poll`/`recvfrom` errors killing the controller, TLS server fd leak per failed connect).
- **DIN SPEC 70121 message layer and SECC states** (654471c0a): C++ message structs with cbv2g converters, EXI (de)serialization and a decode-by-type Variant; the SECC state machine from SessionSetup through CurrentDemand, WeldingDetection and SessionStop, including the sequence-error responder ([V2G-DC-539]). DIN is DC-only, EIM-only, plaintext-only, no pause/resume.
- **ISO 15118-2 message layer and SECC states** (f1cf00a0d): the same for the -2 message set, with both the DC branch (CableCheck, PreCharge, CurrentDemand, WeldingDetection) and the AC branch (ChargingStatus). Includes contract chain validation against the MO/V2G roots and the AuthorizationReq/MeteringReceipt signature checks for Plug-and-Charge; CertificateInstallation is relayed verbatim to the backend.
- **SECC session driver** (897cbc934): the Session now owns transport, buffers and timing only and delegates the protocol to an engine held in a `std::variant`. Every session starts on the `SapEngine` (SupportedAppProtocol handshake, moved out of the -20 FSM, enforcing [V2G2-170]/[V2G2-172]/[V2G2-098] and never offering DIN over TLS [V2G-DC-869]) and is handed over to the `DinSeccEngine`, `D2SeccEngine` or `D20SeccEngine`. The -20 session config becomes the protocol-neutral `session::SessionConfig`/`EvseSetupConfig`.

### Module / interface changes

- **`Evse15118D20`** (e6bd42789): `supported_DIN70121` / `supported_ISO15118_2` / `supported_ISO15118_20` config options select what the SAP handshake offers; the charger implementation feeds the pre-20 sessions (DC limits and present values, authorization, cable-check results, stop requests, EVSE error states) and relays ISO 15118-2 Plug-and-Charge.
- **`ISO15118_charger` interface** (38c9dd063): two additions the DIN timing rules need, implemented by all providers (`EvseV2G`, `IsoMux`, `DummyV2G`):
  - `cp_state_changed(CpState)` — EvseManager forwards every measured IEC 61851-1 CP state; DIN gates cable check on CP C/D ([V2G-DC-967]), welding detection on CP B ([V2G-DC-988]/[V2G-DC-556]), and terminates on CP A ([V2G-DC-962]) or unexpected B ([V2G-DC-668]).
  - `session_stop_res_sent(SessionStopAction)` — published when a session-ending response hits the wire, anchoring the CP oscillator retain time ([V2G-DC-968]) or, on failed termination, switching the oscillator off without delay ([V2G-DC-942]).

### Tests and configs

- Unit tests per layer: EXI round trips per message, per-state response tests over the pure `handle_request()` functions, FSM-level transition tests for both new state machines, and engine-level SAP handshake tests.
- SIL and OCPP test configurations run on the new stack; the previous `EvseV2G`-based variants are kept as `*-evsev2g` configs for comparison runs.


djchhp:
**Spurious `EVLOG_error` at every start-up on a DC-only Evse15118D20 (init-order race in `apply_supported_protocols()`)**

While integration-testing this PR (squash @ 580a7ea52) on a DC-only SIL config (`Evse15118D20` with only `device: auto`, i.e. `supported_ISO15118_20: true`, `-2` and DIN off; `EvseManager` `charge_mode: DC`), every manager start logs:

```
[ERRO] iso15118_charge  ISO15118_chargerImpl::apply_supported_protocols() :: No protocol left to offer: DIN SPEC 70121 is the only configured protocol but the charger offers AC energy transfer. Every SupportedAppProtocol handshake will fail
```

Neither statement is true for this setup: DIN is not configured, and the charger is DC. Sessions afterwards run fine and negotiate ISO 15118-20, so the effect is cosmetic, but the message sends an integrator hunting for a DIN/AC misconfiguration that does not exist.

**What happens**

1. `EvseManager::ready()` publishes its energy transfer modes and calls `update_energy_transfer_modes` on the HLC requirement.
2. That reaches `ISO15118_chargerImpl::handle_update_energy_transfer_modes()` *before* `ISO15118_chargerImpl::ready()` has run the block that fills `configured_protocol_offer` from the `supported_*` config flags (the two modules' `ready()` are not ordered).
3. `handle_update_energy_transfer_modes()` sets `ac_energy_transfer_mode` and calls `apply_supported_protocols()`. There `offered = configured_protocol_offer` is still empty, `setup_config.supported_protocols` still holds the library default `{ISO15118_20}`, so the "nothing changed" early return does not trigger, `offered.empty()` is true, and the DIN error is logged although `din_removed` is false. `setup_config.supported_protocols` is then written empty.
4. A few ms later `ready()` sets `configured_protocol_offer` and calls `apply_supported_protocols()` again, which repairs the offer. The `controller` does not exist yet at step 3, so nothing downstream sees the empty offer, which is why sessions are unaffected.

Timeline from the log (SIL, same host): `evse_manager ... initialized` 05.711 → ERRO 05.719 → `iso15118_charge` SDP server start (inside its `ready()`) 05.775.

**Suggested fix**

Treat "offer not configured yet" as a no-op in `apply_supported_protocols()`; `ready()` calls it again once the offer exists. The genuine DIN-only-plus-AC case still reaches the error, because there `configured_protocol_offer` is non-empty and `din_removed` is true.

```diff
 void ISO15118_chargerImpl::apply_supported_protocols() {
+    if (configured_protocol_offer.empty()) {
+        // Not configured yet: ready() builds the offer from the supported_* flags and calls us again.
+        // EvseManager's ready() can push its energy transfer modes before ours has run, and judging an
+        // empty offer against the library default would log the DIN/AC error on a DC-only charger.
+        return;
+    }
     auto offered = configured_protocol_offer;
```

Optionally tighten the error itself so it can only fire for the case its text describes:

```diff
-    if (offered.empty()) {
+    if (offered.empty() and din_removed) {
         EVLOG_error << "No protocol left to offer: DIN SPEC 70121 is the only configured protocol but the "
```

A unit test that calls `handle_update_energy_transfer_modes()` with an AC mode before the offer is configured and asserts `setup_config.supported_protocols` is left untouched (and no error is logged) would pin this.

lib/everest/iso15118/src/iso15118/session/iso.cpp
```diff
@@ -137,27 +162,176 @@ size_t setup_response_header(uint8_t* buffer, iso15118::io::v2gtp::PayloadType p
...
+    if (skip_app_protocol_negotiation) {
+        // The caller already ran the SupportedAppProtocol handshake (external SAP on a handed-over
+        // socket): replace the SapEngine before anything runs and start directly on the ISO 15118-20
+        // engine, which expects a SessionSetupReq as the first message. No offered/selected protocol
+        // data exists on this path (the caller keeps it), and the vehicle certificate hash comes with
+        // the connection instead of the TLS OPEN event.
+        vehicle_cert_hash = connection->get_vehicle_cert_hash();
+        engine.emplace<D20SeccEngine>(engine_output_view(), config, pause_ctx, callbacks, timeouts,
+                                      d20::EVSupportedAppProtocols{}, message_20::SupportedAppProtocol{},
+                                      vehicle_cert_hash, skip_app_protocol_negotiation);
```
Xerusial: Broken feature: This one starts a -20 session on `skip_app_protocol_negotiation`, regardless of the vehicle having negotiated a -2 or DIN session before

lib/everest/iso15118/include/iso15118/session/d20_secc_engine.hpp
```diff
@@ -0,0 +1,64 @@
...
+
+    bool has_outgoing() const;
+    std::optional<SeccOutgoing> take_outgoing();
+
+    bool is_finished() const;
+    bool is_paused() const;
+    std::optional<session::feedback::SessionStopAction> pop_session_stop_res_pending();
+
+    // ISO 15118-20 has no equivalent of the DIN [V2G-DC-962] / CP State A error termination handled
+    // outside a response, so the session always ends through the regular EV-first close linger.
```
Xerusial: Does not conform with contract in EvseManager.cpp:1100-1104. Either this has to follow proper state A handling like DIN and -2 or the documented contract in EvseManager.cpp needs to be updated

modules/EVSE/EvseManager/EvseManager.cpp
```diff
@@ -1082,6 +1097,42 @@ void EvseManager::ready() {
         }
     }
 
+    // Inform the HLC stack about every measured CP state change. It needs it for the SECC-side CP
+    // checks tied to the message sequence (DIN 70121 [V2G-DC-988]: CP State B after
+    // PowerDelivery(off)), and on unplug (CP State A, [V2G-DC-962]) it closes the V2G TCP
+    // connection. IECStateMachine emits this signal BEFORE the derived CPEvents, so the stack
+    // learns state A before the SLAC teardown below is triggered.
```
Xerusial: -2 and DIN conform with this, D20 does not. Either update this contract or update d20_secc_engine.hpp:44-48

lib/everest/iso15118/src/iso15118/session/iso.cpp
```diff
@@ -187,131 +363,285 @@ TimePoint const& Session::poll() {
...
+        return false;
+    }();
+
+    // A stop requested during the SupportedAppProtocol phase reached only the handshake engine, which
+    // ignores control events: hand it to the fresh protocol engine so the session it now runs tells the
+    // EV to stop from the first response on (EvseV2G reacts to stop_charging in any state).
+    if (created and pending_stop_charging) {
+        const d20::ControlEvent stop_event{d20::StopCharging{true}};
+        visit_engine([&stop_event](auto& e) { e.on_control_event(stop_event); });
+    }
```
Xerusial: Unclean: SAP drops all control events (SapEngine::on_control_event). Only StopCharging is latched and replayed after handover to 15118. Latch+replay CpStateChanged the same way so unplug during handshake FINs TCP before SLAC leaves, instead of waiting on later dlink.

lib/everest/iso15118/src/iso15118/session/d2_secc_engine.cpp
```diff
@@ -0,0 +1,480 @@
...
+
+    // The EV did not end the session within the grace period after the StopCharging request: enforce
+    // the stop -- the next response (the charge loop delivers one within a second) is answered FAILED
+    // and terminates the session (EvseV2G stop_hlc parity). An EV that sends nothing at all is bounded
+    // by the sequence timeout above.
+    // The EV sent nothing the emergency shutdown could be reported on: close anyway rather than hold
+    // the connection until the sequence timeout.
+    if (timeout == d20::TimeoutType::EMERGENCY_SHUTDOWN) {
+        if (not ctx.session_stopped) {
+            logf_warning("No request to answer within %%d ms of the emergency shutdown; closing the connection",
```
Xerusial: nit: %%d is wrong here

lib/everest/iso15118/src/iso15118/session/d20_secc_engine.cpp
```diff
@@ -0,0 +1,116 @@
...
+    // Save some control events. It can happen that these events are sent before the corresponding state. They are
+    // stored temporarily here.
+
+    [[maybe_unused]] const auto res = fsm.feed(d20::Event::CONTROL_MESSAGE);
+    active_control_event.reset();
+}
+
+void D20SeccEngine::on_timeout(d20::TimeoutType timeout) {
+    if (timeout == d20::TimeoutType::SEQUENCE) {
+        logf_error("Sequence Timeout 40secs is reached. Stopping the session");
```
Xerusial: nit: Outdated log message. Should be filled by `TIMEOUT_SEQUENCE`

lib/everest/iso15118/include/iso15118/session/secc_engine.hpp
```diff
@@ -0,0 +1,96 @@
...
+
+// Feed a decoded request into the FSM and re-feed it while a resting state defers the still-pending
+// request to the state it transitioned to (peek-and-divert without consuming, e.g. PreCharge ->
+// PowerDelivery or any resting state -> SessionStop). Shared by the d2, din and d20 SECC engines. The
+// loop exits as soon as a response has been staged (the safe exit condition) and is bounded to guard
+// against an accidental transition ping-pong.
+template <typename Fsm, typename Exchange, typename EventType>
+void drive_request(Fsm& fsm, Exchange& exchange, EventType message_event) {
+    auto res = fsm.feed(message_event);
+    for (int guard = 0; guard < detail::SECC_REFEED_BOUND and res.transitioned() and exchange.has_request() and
```
Xerusial: Bound is 8, then this returns even if the request is still pending and no response was staged. poll() has already reset the packet, so the request sits in the exchange until SEQUENCE timeout. Log/fail here when the loop ends with has_request() && !has_response()

lib/everest/iso15118/src/iso15118/session/d2_secc_engine.cpp
```diff
@@ -0,0 +1,437 @@
...
+        timeouts),
+    fsm(ctx.create_state<d2::state::SessionSetup>()) {
+}
+
+void D2SeccEngine::on_packet(io::v2gtp::PayloadType payload_type, const io::StreamInputView& view) {
+    // All ISO 15118-2 messages share the single SAP payload type (0x8001); any other V2GTP payload
+    // type is ignored (EvseV2G parity, libiso15118 finding F-001).
+    if (payload_type != io::v2gtp::PayloadType::SAP) {
+        return;
+    }
```
mlitre, line 237: If we go through this code path we can wait indefinitely since we never arm the sequence timeout timer again. We setup v2g session stop timer and then we go here and since we never sent a response we never timeout

lib/everest/iso15118/src/iso15118/session/d2_secc_engine.cpp
```diff
@@ -0,0 +1,437 @@
...
+                           d20::Timeouts& timeouts, bool tls_active) :
+    message_exchange(output_view),
+    ctx(std::move(callbacks), make_d2_config(config, tls_active), pause_ctx, active_control_event, message_exchange,
+        timeouts),
+    fsm(ctx.create_state<d2::state::SessionSetup>()) {
+}
+
+void D2SeccEngine::on_packet(io::v2gtp::PayloadType payload_type, const io::StreamInputView& view) {
+    // All ISO 15118-2 messages share the single SAP payload type (0x8001); any other V2GTP payload
+    // type is ignored (EvseV2G parity, libiso15118 finding F-001).
```
mlitre: There are a lot of plan labels leaking into the comments

lib/everest/iso15118/src/iso15118/d2/crypto.cpp
```diff
@@ -0,0 +1,1017 @@
...
+
+std::vector<uint8_t> serialize_signed(const message_2::CertificateInstallationRequest& req, const PrivateKey& key) {
+    iso2_exiDocument doc{};
+    message_2::convert(req.header, doc.V2G_Message.Header);
+    CB_SET_USED(doc.V2G_Message.Body.CertificateInstallationReq);
+    message_2::convert(req, doc.V2G_Message.Body.CertificateInstallationReq);
+    return finalize_signed(doc, req.id, key);
+}
+
+bool verify_certificate_installation_res(const std::vector<uint8_t>& res_exi, const std::string& v2g_root_path) {
```
mlitre, line 693: Is there a reason why we don't do the check on the reference digests?
On the evcc side it is done properly in verify_res_references

lib/everest/iso15118/src/iso15118/d2/crypto.cpp
```diff
@@ -0,0 +1,1017 @@
...
+    in.erase(std::remove(in.begin(), in.end(), '-'), in.end());
+    return in;
+}
+
+// xmldsig algorithm identifiers (ISO 15118-2 uses EXI canonicalization).
+constexpr char ALGO_CANONICAL_EXI[] = "http://www.w3.org/TR/canonical-exi/";
+constexpr char ALGO_ECDSA_SHA256[] = "http://www.w3.org/2001/04/xmldsig-more#ecdsa-sha256";
+constexpr char ALGO_SHA256[] = "http://www.w3.org/2001/04/xmlenc#sha256";
+
+template <typename CbStringField> void set_cb_string(CbStringField& field, const char* value) {
```
mlitre, line 135: No length check against the max exi field size

lib/everest/iso15118/src/iso15118/d2/crypto.cpp
```diff
@@ -0,0 +1,1017 @@
...
+
+    std::array<uint8_t, MAX_EXI_SIZE> exi_buffer{};
+    exi_bitstream_t stream;
+
+    // 1) digest over the signed AuthorizationReq EXI fragment; compare against Reference[0].DigestValue.
+    {
+        auto fragment = std::make_unique<iso2_exiFragment>();
+        init_iso2_exiFragment(fragment.get());
+        fragment->AuthorizationReq_isUsed = 1;
+        std::memcpy(&fragment->AuthorizationReq, &doc->V2G_Message.Body.AuthorizationReq,
```
mlitre, line 388: This memcpy between an iso2_exiFragment member and an iso2_BodyType member that happen to
share a type at the current version of cbv2g. A plain assignment costs nothing and turns a future divergence into a compile error rather than silently wrong bytes when sent to the EV.

eMAID already diverges in exactly this way, which is why the EVCC side needs a typed
conversion for it.

Same issue on lines: 490, 559, 562, 565.

lib/everest/iso15118/src/iso15118/d2/crypto.cpp
mlitre, line 1: General comment on this crypto file, I think only PnC specific things should stay in it. There are quite a lot of things that could go into libtls or should use things from libtls already.
Additionally, a lot of the actual PnC parts are untested here and it would be nice to have specific tests for xmldsig reference handling / emaid handling, etc.

lib/everest/iso15118/src/iso15118/d2/crypto.cpp
```diff
@@ -0,0 +1,1017 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2025 Pionix GmbH and Contributors to EVerest
```
mlitre, line 2: General comment on copyright year

lib/everest/iso15118/src/iso15118/d2/crypto.cpp
```diff
@@ -0,0 +1,1017 @@
...
+namespace {
+
+constexpr std::size_t MAX_EXI_SIZE = 8192;
+constexpr std::size_t SHA256_LEN = 32;
+constexpr std::size_t ECDSA_SIG_LEN = 64; // r (32) || s (32)
+
+using X509_ptr = std::unique_ptr<X509, decltype(&X509_free)>;
+using PKEY_ptr = std::unique_ptr<EVP_PKEY, decltype(&EVP_PKEY_free)>;
+
+X509_ptr der_to_x509(const std::vector<uint8_t>& der) {
```
mlitre, line 44: Could be replaced by everest/tls/openssl_util.hpp:der_to_certificate

lib/everest/iso15118/src/iso15118/d2/crypto.cpp
```diff
@@ -0,0 +1,1017 @@
...
+    const unsigned char* p = der.data();
+    return X509_ptr(d2i_X509(nullptr, &p, static_cast<long>(der.size())), &X509_free);
+}
+
+bool sha256(const uint8_t* data, std::size_t len, std::array<uint8_t, SHA256_LEN>& out) {
```
mlitre, line 49: Already a method in everest/tls/openssl_util.hpp that does exactly this

lib/everest/iso15118/src/iso15118/d2/crypto.cpp
```diff
@@ -0,0 +1,1017 @@
...
+    unsigned int md_len = 0;
+    return EVP_Digest(data, len, out.data(), &md_len, EVP_sha256(), nullptr) == 1 and md_len == SHA256_LEN;
+}
+
+bool ecdsa_verify(EVP_PKEY* pkey, const uint8_t* sig_rs, std::size_t sig_len,
```
mlitre, line 54: Parts of this are already in libtls we should extract to there

lib/everest/iso15118/src/iso15118/d2/crypto.cpp
```diff
@@ -0,0 +1,1017 @@
...
+    }
+    char buf[256] = {0};
+    const int len = X509_NAME_get_text_by_NID(name, NID_commonName, buf, sizeof(buf) - 1);
+    if (len > 0) {
+        cn.assign(buf, static_cast<std::size_t>(len));
+    }
+    return cn;
+}
+
+std::string cert_to_pem(X509* cert) {
```
mlitre, line 108: Already in everest/tls/openssl_util.hpp:certificate_to_pem

lib/everest/iso15118/src/iso15118/d2/crypto.cpp
```diff
@@ -0,0 +1,1017 @@
...
+    PKEY_ptr pkey(PEM_read_bio_PrivateKey(bio, nullptr, nullptr, pw), &EVP_PKEY_free);
+    BIO_free(bio);
+    if (pkey == nullptr) {
+        logf_error("PnC: failed to load EC private key from PEM");
+        ERR_clear_error();
+    }
+    return pkey;
+}
+
+std::vector<uint8_t> ecdsa_sign(EVP_PKEY* pkey, const std::array<uint8_t, SHA256_LEN>& digest) {
```
mlitre, line 156: Parts already in libtls we should extract

lib/everest/iso15118/src/iso15118/d2/crypto.cpp
```diff
@@ -0,0 +1,1017 @@
...
+constexpr char ALGO_ECDSA_SHA256[] = "http://www.w3.org/2001/04/xmldsig-more#ecdsa-sha256";
+constexpr char ALGO_SHA256[] = "http://www.w3.org/2001/04/xmlenc#sha256";
+
+template <typename CbStringField> void set_cb_string(CbStringField& field, const char* value) {
+    const std::size_t len = std::strlen(value);
+    std::memcpy(field.characters, value, len);
+    field.charactersLen = static_cast<uint16_t>(len);
+}
+
+PKEY_ptr load_private_key(const PrivateKey& key) {
```
mlitre, line 141: I suggest extending the libtls version which takes a file to take PEM bytes instead so we have only one method implementing this

lib/everest/iso15118/src/iso15118/d2/crypto.cpp
```diff
@@ -0,0 +1,1017 @@
...
+    }
+    if (out_len > hash.size()) {
+        return {};
+    }
+    return std::vector<uint8_t>(hash.begin(), hash.begin() + out_len);
+}
+
+} // namespace
+
+ContractValidationResult validate_contract_chain(const std::vector<uint8_t>& leaf_der,
```
mlitre, line 219: The x509 verify cert calls inside this block could use verify_chain from libtls no reason to reimplement here.

lib/everest/iso15118/src/iso15118/d2/crypto.cpp
```diff
@@ -0,0 +1,1017 @@
...
+    if (serial != nullptr) {
+        int64_t value = 0;
+        if (ASN1_INTEGER_get_int64(&value, serial) == 1) {
+            id.serial_number = value;
+        }
+    }
+    return id;
+}
+
+std::string der_chain_to_pem(const std::vector<uint8_t>& leaf_der, const std::vector<std::vector<uint8_t>>& subs_der) {
```
mlitre, line 966: Already a certificate_to_der in libtls

lib/everest/iso15118/src/iso15118/d2/crypto.cpp
```diff
@@ -0,0 +1,1017 @@
...
+    out.assign(ECDSA_SIG_LEN, 0);
+    if (BN_bn2binpad(r, out.data(), 32) != 32 or BN_bn2binpad(s, out.data() + 32, 32) != 32) {
+        out.clear();
+    }
+    ECDSA_SIG_free(sig);
+    return out;
+}
+
+// ConcatKDF (NIST SP 800-56A) with SHA-256, single block over counter(0x00000001) || Z || OtherInfo.
+std::vector<uint8_t> concat_kdf_sha256(const std::vector<uint8_t>& shared_secret,
```
mlitre, line 198: Seems generic enough that we can extract

lib/everest/iso15118/test/exi/cb/iso2/certificate_installation.cpp
mlitre, line 1: General comment, we only ever check exi encoding between our own stack round trip which is not very useful. I suggest having predefined EXI strings that we know are correct and test those against the stack or having a test harness that makes Josev or another encoder create EXI strings which we then use to validate ours

lib/everest/iso15118/src/iso15118/session/iso.cpp
```diff
@@ -187,131 +323,244 @@ TimePoint const& Session::poll() {
...
 
     // check for complete sdp packet
     if (packet.is_complete()) {
-        // FIXME (aw): this event loop only acts on new packets, seems to be enough for now ...
-
-        message_exchange.set_request(make_variant_from_packet(packet));
-
-        packet = {};            // reset the packet
-        state.new_data = false; // reset new_data flag
+        if (session_over()) {
```
mlitre, line 351: Comment on this section related to a previous finding. If an engine returns without staging it leaves the session with no timeout and no response which silently hangs the process rather than failing.

We should do a reset_timeout so that this issue can never occur.

lib/everest/iso15118/src/iso15118/din/state/sequence_error.cpp
```diff
@@ -0,0 +1,174 @@
...
+        ctx.respond(res);
+        return;
+    }
+    default:
+        logf_warning("cannot build a DIN response for received type id: %d", received.get_type());
+        return;
+    }
+}
+
+void respond_sequence_error(Context& ctx, const message_din::Variant& received) {
```
mlitre: Why don't we do something similar as the d2 sequence error and put into one helper this avoids having `ctx.session_stopped = true` in each state and avoid potential maintainability issues.

modules/EVSE/EvseManager/Charger.cpp
```diff
@@ -2194,19 +2246,67 @@ types::iso15118::DcEvseMinimumLimits Charger::get_evse_min_hlc_limits() {
...
+        cp_state_X1();
+    }
     shared_context.hlc_charging_terminate_pause = HlcTerminatePause::Terminate;
 }
 
+// A positive SessionStopRes was sent: start the CP oscillator retain timer [V2G-DC-968]. The PWM
+// stays on for V2G_SECC_CP_OSCILLATOR_RETAIN and is then switched off by the state machine,
+// independent of the EV's TCP close. dlink_terminate()/dlink_pause() follow later (after the TCP
+// connection is gone) and apply a harmless second X1.
+void Charger::notify_session_stop_res_sent(types::iso15118::SessionStopAction action) {
```
mlitre, line 2293: Could we not move this functionality into a different PR?

lib/everest/iso15118/src/iso15118/d2/crypto.cpp
```diff
@@ -0,0 +1,1017 @@
...
+            logf_error("PnC: failed to encode AuthorizationReq fragment");
+            return false;
+        }
+
+        std::array<uint8_t, SHA256_LEN> digest{};
+        if (not sha256(exi_buffer.data(), exi_bitstream_get_length(&stream), digest)) {
+            return false;
+        }
+
+        const iso2_ReferenceType& ref = signature.SignedInfo.Reference.array[0];
```
mlitre, line 402: No checks on size

lib/everest/iso15118/src/iso15118/d2/crypto.cpp
```diff
@@ -0,0 +1,1017 @@
...
+            logf_error("PnC: failed to encode MeteringReceiptReq fragment");
+            return false;
+        }
+
+        std::array<uint8_t, SHA256_LEN> digest{};
+        if (not sha256(exi_buffer.data(), exi_bitstream_get_length(&stream), digest)) {
+            return false;
+        }
+
+        const iso2_ReferenceType& ref = signature.SignedInfo.Reference.array[0];
```
mlitre, line 504: No checks on size

lib/everest/iso15118/src/iso15118/session/iso.cpp
```diff
@@ -187,131 +323,244 @@ TimePoint const& Session::poll() {
...
+            // packet for the engine that will own the protocol instead of failing it against the handshake
+            // engine; the response is due within RESPONSE_DELAY_AFTER_REQUEST_MS, so this defers by one poll.
+        } else {
+            const auto payload_type = packet.get_payload_type();
+            const io::StreamInputView view{packet.get_payload_buffer(), packet.get_payload_length()};
+
+            // Timestamp the request so -2/DIN and the handshake can pace their response after it.
+            last_request_rx_time = now;
+
+            if (not in_sap_phase()) {
```
mlitre, line 368: SEQUENCE is armed for 60 s at connection accept and poll() only stops it once out of the SAP
phase, so when send_response() re-arms it after the SupportedAppProtocolRes the slot is still
occupied and the call is refused:

  Timeout 0 already started

The window for SessionSetupReq is therefore 60 s minus everything before it, not the 60 s the
comment above intends. Practical impact is small since TbdController's 18 s setup timeout
bounds the phase more tightly, but the warning fires on every session.

mlitre (changes requested):
Not sure what the problem is exactly, need more time to investigate, but the following two tests fail reliably: test_iso15118_ac_session_no_energy_during_session_timeout_triggers and test_iso15118_dc_cable_check_failed.

Potential root cause:

The AC hypothesis — explicitly inferred, not established. The SIL scenarios were never run.

The symptom is the absence of a ChargingPausedEVSE event; the test gets ['StoppingCharging', 'AuthRequired', 'PrepareCharging'] instead.

Two new early-outs park in ChargingPausedEV rather than ChargingPausedEVSE:
- modules/EvseManager/Charger.cpp:759-766, the PrepareCharging path
- modules/EvseManager/Charger.cpp:1053-1060, the r.reasons.size() == 0 resume path

Both are guarded on hlc_charging_terminate_pause != HlcTerminatePause::Unknown. That flag is now set earlier than it used to be, by the new notify_session_stop_res_sent() at Charger.cpp:2287-2308, which fires when SessionStopRes reaches the wire rather than at dlink_*. If that guard fires during a no-energy pause, ChargingPausedEVSE never happens, and what remains is exactly the sequence the test reports.

a-w50:
@mlitre @corneliusclaussen @SebaLukas 

Thanks for the thorough review. Status of the items:

**Blocker – the two failing SIL tests (fixed)**
- `test_iso15118_ac_session_no_energy_during_session_timeout_triggers`: the ISO 15118-2 `SessionStop` state applied the CP State B gate of [V2G2-920]..[V2G2-922] to AC sessions. Those are DC-only requirements (8.7.4.4), so an AC EV still in state C got SessionStopRes FAILED, a SECC-side TCP close and D-LINK_ERROR, which sent EvseManager through its HLC error reinit. The gate is removed from the AC path; `PostCharge`/`WeldingDetection` keep it for DC.
- `test_iso15118_dc_cable_check_failed`: `hlc_session_failed` was suppressed when the module itself had requested a stop or emergency shutdown. That condition was ported from EvseV2G's `connection_teardown()`, where it is dead code (`v2g_handle_connection()` resets both flags before the teardown reads them). A failed cable check ends with CableCheckRes FAILED and D-LINK_ERROR ([V2G2-585], [V2G2-539], [V2G2-727]) and is reported now. EvseManager also reports the cable check result to the stack before raising the resistance fault, so the FAILED response carries the isolation fault instead of `EVSE_EmergencyShutdown`.
- The Charger.cpp hypothesis is ruled out: both tests pass with the same EvseManager when EvseV2G is selected.
- Full ISO 15118 smoke set: 25 passed, 5 skipped (pre-existing markers), 0 failed.

**Also done**
- DIN sequence-error termination centralised in one helper.
- Plan label removed from the `d2_secc_engine.cpp` comment.
- `Timeouts::check()` no longer drops a timeout whose deadline coincides with another one; `%%d` format strings fixed.
- The -2 `power_delivery_stopped` flag renamed to `cp_state_b_gate_armed`, since arming the DC gate is its only role.

**crypto.cpp**
- The EV-side half (`verify_certificate_installation_res`, the signing path, `set_cb_string`) has no callers and no tests in this repo; it will be removed. That takes the missing reference-digest check and the unbounded `memcpy` with it.
- `Reference.arrayLen` guards and `memcpy` → right now ok by definition (cbexigen)
- libtls duplication: on hold on purpose. The plan is a link-time crypto interface, and these sites get revisited once it exists.

**Agreed, separate changes**
- Dropped packet disarming the sequence watchdog (`d2_secc_engine.cpp`, `iso.cpp`): real, needs a `poll()` refactor with an explicit invalid-packet strategy rather than a spot fix.
- SEQUENCE timer double-arm warning every session: real, needs a decision on the `Timeouts` contract.
- Copyright years: repo-wide sweep, not per file here.

**Not done**
- Splitting the EvseManager changes into a separate PR: assessed and not feasible, the three parts depend on each other and splitting would not have unblocked the tests. Details on request.

SebaLukas:
I will start with my review later today and hopefully be finished tomorrow afternoon

lib/everest/iso15118/test/exi/cb/din/cable_check.cpp
SebaLukas: Where are the exi streams to test encoding, decoding and converting c <-> cpp?
For every D20 message we have at least one exi stream for the req and res message.
https://github.com/EVerest/EVerest/tree/feat/adding-iso2-support and https://github.com/EVerest/EVerest/tree/feat/libiso-adding-din-support contain some unit tests with exi streams

lib/everest/iso15118/test/exi/cb/app_hand/app_hand.cpp
```diff
@@ -37,7 +37,30 @@ SCENARIO("App Protocol Ser/Des") {
         }
     }
 
-    // Todo(sl): Missing Decode message
-    // 80400040
-    // {"supportedAppProtocolRes": {"ResponseCode": "OK_SuccessfulNegotiation", "SchemaID": 1}}
+    GIVEN("Round-trip of a supportedAppProtocolRes document") {
```
SebaLukas: Whre is the exi stream in the test? In the comment I added the exi stream and the xml output already

lib/everest/iso15118/test/exi/cb/iso2/certificate_installation.cpp
SebaLukas: This applies to all exi/cb d2 and din tests

lib/everest/iso15118/test/iso15118/fsm/d20_transitions.cpp
SebaLukas: Please look at the changes. I think the whole fsm tests are removed and only `handle_request` is now tested. Before the whole state was tested.

lib/everest/iso15118/test/iso15118/fsm/d2_transitions.cpp
SebaLukas: First I dont get this tests. Are the tests are really testing the state or only some part of it?

A combination of states/iso2 tests and this fsm tests would be nice. That only the whole state is tested.
For D20 I already started to migrate the tests from states to fsm: https://github.com/EVerest/EVerest/tree/feat/adding-more-libiso-tests

Also the original d20_transistions.cpp + dc_pre_charge.cpp + service_detail.cpp + session_setup.cpp in test/iso1511/fsm are good example.

lib/everest/iso15118/test/iso15118/io/CMakeLists.txt
```diff
@@ -18,20 +18,21 @@ target_link_libraries(socket_helper_test
 )
 catch_discover_tests(socket_helper_test)
 
-add_executable(test_sdp_server sdp_server_test.cpp)
-target_link_libraries(test_sdp_server PRIVATE iso15118 Catch2::Catch2WithMain)
-catch_discover_tests(test_sdp_server)
-
```
SebaLukas: Is there a reason to move these lines to line 32?

PR #2487, opened 2026-07-20
feat(powermeter): add powermeter min capabilities and apply DC min current limits

Pietfried:
## Describe your changes

Add an optional capabilities var to the powermeter interface reporting the minimum export/import current the meter can measure within its accuracy class (e.g. required by the German Calibration Law).

- types/interfaces: new powermeter Capabilities type and var
- everest_api_types + powermeter_API: expose capabilities via external API
- EvseManager: merge meter minimum currents into the DC power supply capabilities pushed to the HLC layer, clamping to supply maximums
- DCSupplySimulator: optional config to publish simulated capabilities


florinmihut (changes requested):
Looks good but I think it is missing 2 things:
1. A minimum of 0 for both of the caps fields. For example, a bogus negative value — e.g. injected via the external m2e/capabilities MQTT topic, which is an unvalidated external input — flows through max_optional(std::nullopt, meter_min) and would set nominal_min_export_current_A to a negative value where it was previously absent, which then goes to the HLC layer. How damaging this could be ... tbd but setting a min would improve the robustness.
2. apply_powermeter_limits of the EVSEManager would be great to have a test since it is not trivial and future changes might cause issues - better to have a test for it.

Pietfried:
> Looks good but I think it is missing 2 things:
> 
> 1. A minimum of 0 for both of the caps fields. For example, a bogus negative value — e.g. injected via the external m2e/capabilities MQTT topic, which is an unvalidated external input — flows through max_optional(std::nullopt, meter_min) and would set nominal_min_export_current_A to a negative value where it was previously absent, which then goes to the HLC layer. How damaging this could be ... tbd but setting a min would improve the robustness.
> 2. apply_powermeter_limits of the EVSEManager would be great to have a test since it is not trivial and future changes might cause issues - better to have a test for it.

Addressed requested changes, please re-review.

florinmihut (dismissed):
LGTM

modules/EVSE/EvseManager/EvseManager.cpp
```diff
@@ -2695,9 +2703,85 @@ types::power_supply_DC::Capabilities EvseManager::get_powersupply_capabilities()
     caps.max_export_power_W = min_optional(caps.max_export_power_W, derate.max_export_power_W);
     caps.max_import_power_W = min_optional(caps.max_import_power_W, derate.max_import_power_W);
 
+    // Raise minimum current limits according to power meter capabilities
+    caps = apply_powermeter_limits(caps);
```
djchhp: As a side effect a power meter with capabilities above 2A silently raises the isolation-test limit. This is likely unintended, since we specifically have the 2A limit to limit the fault current.

modules/EVSE/EvseManager/EvseManager.cpp
```diff
@@ -2695,9 +2703,85 @@ types::power_supply_DC::Capabilities EvseManager::get_powersupply_capabilities()
...
+    }
+
+    return module::apply_powermeter_limits(std::move(caps), meter, log_warnings);
+}
+
+void EvseManager::update_powersupply_capabilities(types::power_supply_DC::Capabilities caps) {
+    {
+        std::scoped_lock lock(powersupply_capabilities_mutex);
+        if (powersupply_capabilities == caps) {
+            return;
```
djchhp, line 2736: Here is an early return, which changes the original behavior. I don't see a problem with this but it is certainly worth double checking this is intended.

modules/EVSE/EvseManager/powermeter_limits.cpp
```diff
@@ -0,0 +1,87 @@
...
+    }
+    // if a has a value, return that one.
+    if (a.has_value()) {
+        return a;
+    }
+
+    // else return b. It is either the only value or empty.
+    return b;
+}
+
```
djchhp: util/math/comparision.hpp implements exactly this.

modules/EVSE/EvseManager/EvseManager.cpp
djchhp: The derating happens before applying the power meter limits. Compare to line 2757 where those limits are not taken into account. So both functions can yield different values. 

I think because of this difference, we could reach min above max, with let's say meter min 40A and derate max 30A.

modules/EVSE/EvseManager/EvseManager.cpp
```diff
@@ -2695,9 +2703,85 @@ types::power_supply_DC::Capabilities EvseManager::get_powersupply_capabilities()
...
+    r_hlc[0]->call_update_dc_minimum_limits(evse_min_limits);
+
+    // HLC layer will also get new maximum current/voltage/watt limits etc, but those will need to run through
+    // energy management first. Those limits will be applied in energy_grid implementation when requesting
+    // energy, so it is enough to set the powersupply_capabilities here.
+    // FIXME: this is not implemented yet: enforce_limits uses the enforced limits to tell HLC, but capabilities
+    // limits are not yet included in request.
+}
+
 void EvseManager::set_external_derating(types::dc_external_derate::ExternalDerating d) {
```
djchhp, line 2796: I guess we need to publish capabilities here as well. Updates are otherwise only published on PSU and power meter values changes.

djchhp (changes requested):
Different code paths can result in conflicting values. See inline comments.

modules/EVSE/EvseManager/EvseManager.cpp
Pietfried: Adressed with 2db26368f7de8aedaf210903ea2826dd70abc0ca

modules/EVSE/EvseManager/powermeter_limits.cpp
Pietfried: Adressed with ade224eb6aefc4099664402d701539cfe2824b3b

modules/EVSE/EvseManager/EvseManager.cpp
Pietfried: Adressed with 1591dda7bce0679bcead86493673ddbd16f33700

modules/EVSE/EvseManager/EvseManager.cpp
Pietfried: This could skip the initial push, addressed with 59a7b5c0d3fcaf43bda05b0f34e0d3b0f99d5d39

modules/EVSE/EvseManager/EvseManager.cpp
Pietfried: Adressed with 2db26368f7de8aedaf210903ea2826dd70abc0ca

modules/EVSE/EvseManager/EvseManager.cpp
```diff
@@ -2686,29 +2687,34 @@ bool EvseManager::session_is_iso_d20_dc_bpt() {
...
 types::power_supply_DC::Capabilities EvseManager::get_powersupply_capabilities() {
     types::power_supply_DC::Capabilities caps;
-    types::dc_external_derate::ExternalDerating derate;
 
     {
         std::scoped_lock lock(powersupply_capabilities_mutex);
         caps = powersupply_capabilities;
     }
+
+    return apply_external_derating(std::move(caps));
```
djchhp, line 2698: This function is called in read(). With a derate already active at that stage it is included into the raw PSU capabilities.

modules/EVSE/EvseManager/EvseManager.cpp
```diff
@@ -2752,9 +2759,11 @@ void EvseManager::update_powermeter_capabilities(const types::powermeter::Capabi
 }
 
 void EvseManager::push_powersupply_capabilities_to_hlc() {
-    // Hold the lock across merge and send so concurrent pushes cannot reach HLC out of order
+    // Hold the lock across merge and send so concurrent pushes cannot reach HLC out of order.
+    // Derating is applied before the power meter limits so that a meter minimum above a derated
+    // maximum is clamped consistently with the enforce_limits path.
     std::scoped_lock lock(powersupply_capabilities_mutex);
-    const auto caps = apply_powermeter_limits(powersupply_capabilities, true);
+    const auto caps = apply_powermeter_limits(apply_external_derating(powersupply_capabilities), true);
```
djchhp, line 2768: feeding `apply_powermeter_limits` with derated capabilities can trigger misleading warning messages pointing to a power meter issue, where there is no issue.

djchhp (changes requested):
All raised issues have been correctly addressed. Two new ones were discovered though, see inline comments for details.

One suggestion. EvseManager.cpp:2790 has a FIXME comment. Maybe now is a good time to fix it.

modules/EVSE/EvseManager/EvseManager.cpp
Pietfried: This was pre-existing, but I changed it by keeping the external derating out of the raw PSU caps store (only updated by subscribe_capabilities).

modules/EVSE/EvseManager/EvseManager.cpp
Pietfried: I updated the warning message

Pietfried:
> All raised issues have been correctly addressed. Two new ones were discovered though, see inline comments for details.
> 
> One suggestion. EvseManager.cpp:2790 has a FIXME comment. Maybe now is a good time to fix it.

This looks not very trivial and I'd like to keep it out of this PR's scope.

djchhp (approved):
All discovered issues have been addressed. Looking good.

mlitre (approved):

SebaLukas (approved):

PR #2490, opened 2026-07-21
fix(Huawei_V100R023C10): initialize externally provided connector data

Pietfried:
## Describe your changes

Members of external_provided_data were default-constructed so the dispenser could potentially report a garbage contactor status and voltage/current readings to the PSU between module start and the first BSP PowerOn/PowerOff event. This could fail the power unit's boot self-check with Gun Contactor Abnormal. Initialize to contactors off / 0 V / 0 A.

### DIFF
--- a/modules/HardwareDrivers/PowerSupplies/Huawei_V100R023C10/connector_base/base.hpp
+++ b/modules/HardwareDrivers/PowerSupplies/Huawei_V100R023C10/connector_base/base.hpp
@@ -102,7 +102,7 @@ private:
     std::string log_prefix;
     std::string telemetry_subtopic;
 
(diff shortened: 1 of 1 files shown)


florinmihut (changes requested):
The garbage in the std::atomic constrtucts could potentially cause problems in module_placeholder_allocation_failure_raised. A garbage true value would suppress raising the allocation failed error. 
Moreover, there is a best practive to always initialize an enum variables - types::power_supply_DC::Mode last_mode, types::power_supply_DC::ChargingPhase last_phase - i recommend to initialize them - not sure if this is valid use case but since we are about to fix errors due to missing initializations, let's do it (Mode::Off/ChargingPhase::Other)

florinmihut (approved):
LGTM

PR #2492, opened 2026-07-21
Fix dev tooling bugs: devrd portability/purge, OCPP user_config install

hirokiri:
## Describe your changes

Two independent bug fixes in the development tooling, found while setting up the devcontainer workflow on a macOS host:

1. **devrd script**:
   - `flow` used GNU-only `sed -i`, which fails on BSD/macOS sed; replaced with a portable temp-file rewrite.
   - `purge` ran `docker compose down` without `--profile all` — since every service is profile-gated, no containers were ever stopped or removed.
   - `purge` matched images/volumes case-sensitively against the workspace folder name (`EVerest`), which never matches the lowercased docker resource names (`everest_devcontainer-*`).
   - `purge` blocked forever on the CPM-cache prompt when stdin is not a TTY; it now defaults to keeping the cache. The pattern-based volume removal also explicitly skips the shared CPM cache volume so the (now case-insensitive) match cannot force-remove it before the prompt.

2. **OCPP v16 config install**: `if(EXISTS user_config/user_config.json)` uses a relative path, which CMake's `EXISTS` does not resolve — so a provided `user_config.json` was silently never installed. Now checked relative to `CMAKE_CURRENT_SOURCE_DIR`.

> Note: this PR originally also contained a fix for the devcontainer GID collision on macOS hosts; that was fixed upstream by #2431 in the meantime, so that part was dropped when resolving conflicts.

All changes are no-ops in the previously-working paths (Linux hosts with default UID/GID 1000, interactive shells).

N/A — found during local development setup.

- [x] I have made corresponding changes to the documentation (n/a)

### DIFF
--- a/lib/everest/ocpp/config/v16/CMakeLists.txt
+++ b/lib/everest/ocpp/config/v16/CMakeLists.txt
@@ -65,7 +65,7 @@ install(
 )
 
 
-if(EXISTS ${USER_CONFIG_FILE})
+if(EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/${USER_CONFIG_FILE})
      install(
           FILES ${USER_CONFIG_FILE}
           DESTINATION ${CMAKE_INSTALL_DATADIR}/everest/modules/OCPP)
--- a/.devcontainer/general-devcontainer/Dockerfile
+++ b/.devcontainer/general-devcontainer/Dockerfile
@@ -56,9 +56,16 @@ RUN ls -al /tmp/everest_dev_tool \
 ##################################################################
 
 # Modify the existing docker user and group to match the host's USER_UID and USER_GID
-RUN groupmod --gid ${USER_GID} ${USERNAME} && \
-    usermod --uid ${USER_UID} --gid ${USER_GID} ${USERNAME} && \
-    usermod -aG dialout ${USERNAME}
+# If a group with USER_GID already exists (e.g. GID 20 = dialout on macOS hosts),
+# reuse it as the user's primary group instead of renumbering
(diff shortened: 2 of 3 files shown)


hirokiri:
Conflicts are resolved. The conflict was in the devcontainer Dockerfile: the GID-collision fix in this PR was also fixed upstream by #2431 in the meantime, so I dropped that part and kept the upstream version — this PR is now only the devrd portability/purge fixes and the OCPP `user_config` install fix (title/description updated accordingly). Should be mergeable now.

hirokiri:
Squashed the branch into a single commit (847272e). As noted above, the devcontainer GID fix was dropped during the merge since it was fixed upstream by #2431 — the PR is now only the devrd flow/purge fixes and the OCPP v16 user_config install fix.

lib/everest/ocpp/config/v16/CMakeLists.txt
andistorm, line 2: Why not making this path absolute directly?

andistorm: Apart fromt this: Please export this fix in its own PR since it is bit out of scope and more ocpp related

andistorm (changes requested):
It looks good in general. I would love to merge this after the ocpp config thing is extracted to a separate PR

lib/everest/ocpp/config/v16/CMakeLists.txt
hirokiri: @andistorm 
Thank you for your review.
Dropped in d78ebf4 and moved to #2797 , with the path made absolute directly as suggested.

andistorm (approved):
LGTM 🐧

PR #2505, opened 2026-07-22
Add RsModDefMeter: a profile-driven generic powermeter module

### LINKED ISSUE #2504: Profile-driven powermeter module: describe Modbus meters (incl. OCMF signing) as data instead of per-model drivers
### Describe the problem

Adding support for a new Modbus power meter today means writing a new module, or extending an existing one, for each model. `GenericPowermeter` addresses part of this for plain metering by describing a device in a configuration file, but every additional meter still requires its register map to be expressed in a form specific to that module, and it does not cover signed metering at all.

Signed metering (German Eichrecht / OCMF) makes this sharper. Each signing meter has its own multi-step procedure — write metadata, arm a status register, poll until the signature is ready, read the signed document at a length that is only known at runtime. Today that logic is hand-written per meter (`RsIskraMeter`, `IsabellenhuetteIemDcr`, `LemDCBM400600`, `CarloGavazzi_EM580`, …), even though the *shape* of the procedure is very similar across devices and the differences are mostly register addresses, framing and encodings.

The result is a growing set of near-parallel drivers, each needing separate maintenance and hardware access to verify, and no shared way to describe "this meter's registers and this meter's signing procedure" as data.

### EVerest Domain

Hardware Drivers

### Affected Component

`modules/HardwareDrivers/PowerMeters` — a new module alongside `GenericPowermeter`.

### Describe your solution

A powermeter module driven by a declarative device profile, so that adding a meter means adding a profile rather than writing code.

We have been working on [ModDef](https://moddef.org/), an open specification and set of runtimes (Rust, C, C++, Go, Python, TypeScript) for describing Modbus devices: register maps, encodings, scaling, semantic measurand tags, and — relevant here — multi-step register procedures with typed parameters, poll-until-condition steps and runtime-resolved read lengths.

The proposal is a `RsModDefMeter` module that:

- publishes the `powermeter` variable for **any** profile by looking up measurand-tagged points (`base_quantity` plus phase / direction / aggregation) and rescaling into the units the EVerest `Powermeter` type expects;
- implements `start_transaction` / `stop_transaction` for meters that expose the OCMF fields as **typed registers**, by mapping `TransactionReq` onto the profile's declared command parameters and executing the profile's write/poll/read steps;
- performs capability discovery — a profile with no transaction commands advertises plain metering and returns `NOT_SUPPORTED` — so metering-only and signing meters share one module;
- reaches the bus through the existing `serial_communication_hub`, so it coexists with other devices on a shared RS485 line.

An implementation is ready and I would be glad to contribute it; a PR will follow this issue.

### Alternatives considered

- **Extending `GenericPowermeter`.** Its configuration format is not designed to express multi-step procedures, runtime-dependent read lengths, or the semantic tagging needed to map arbitrary registers onto the `Powermeter` fields. Growing it in that direction would effectively mean designing a device-description language inside the module.
- **Continuing with one driver per meter.** Straightforward and what exists today, but the maintenance cost scales linearly with the number of supported meters, and each driver needs its own hardware to verify.
- **A fully generic signed-transaction binding covering every meter.** We explored this and concluded it does not hold. Some meters (Bauer BSM, Iskra WM3M4C) accept a single opaque OCMF blob that the *caller* must serialise from configuration and runtime state; that composition cannot be expressed declaratively, so those meters are deliberately out of scope and keep their dedicated drivers. Meters that pre-decompose OCMF into typed registers (for example the Carlo Gavazzi EM580, which has a module in this repository) map cleanly and are what this module targets. The reasoning is written up in [ModDefOrg/moddef#3](https://github.com/ModDefOrg/moddef/issues/3).

### Additional context

- The multi-step command construct the module relies on: [ModDefOrg/moddef#2](https://github.com/ModDefOrg/moddef/issues/2).
- Scope limits worth stating up front: Modbus RTU via `serial_communication_hub` only (a Modbus TCP transport is a follow-up); discrete-input reads are not offered by the hub interface; and the implementation is currently verified against mocks rather than hardware.

suda:
## Describe your changes

`RsModDefMeter` is a new EVerest `powermeter` driver that is configured by a [ModDef](https://moddef.org/) device profile rather than written per meter model. One binary drives any conforming meter: the register map, framing, scaling, polling and — where the meter supports it — the signed-transaction procedure all come from a declarative profile instead of from code. It is the profile-driven counterpart to `GenericPowermeter`, extended to cover signed metering.

**Metering (any profile).** The `powermeter` variable is published by querying the profile's measurand-tagged points (`base_quantity` plus phase / direction / aggregation) and rescaling each value into the unit the EVerest field expects (Wh, W, var, V, A, Hz). Fields the profile does not provide are left absent. Where a measurand tuple matches more than one point (a live value and its signed-snapshot twin, say), the point whose unit already matches the target is preferred.

**Signed transactions (typed-register OCMF meters).** When the profile declares a start/stop command — matched by the `command_ref` role tag `ocmf:start_transaction` / `ocmf:stop_transaction`, otherwise by command id — the module maps the EVerest `TransactionReq` onto that command's typed parameters and executes it through ModDef's multi-step command construct (write / poll / read). Identification enums are resolved against the profile's own enum table (OCMF value names match), the OCMF identification flags are routed to their per-group registers, and the stop command's OCMF result is returned as a `SignedMeterValue`.

**Capability discovery.** A profile that declares no transaction commands advertises plain metering and returns `NOT_SUPPORTED` from `start_transaction` / `stop_transaction`, so metering-only and signing meters are served by the same module.

**Implementation.** Modbus RTU goes through the required `serial_communication_hub`: a small `HubBridge` implements ModDef's `Transport` trait on top of the hub commands. everestrs handlers are synchronous while ModDef's `Device` is async and `&mut`, so the device is driven with `pollster::block_on` behind a `Mutex<Device>`, which also serialises access to the shared RS485 bus. No async runtime is introduced — the hub calls and the poll delay are blocking, so every await completes without a reactor. `moddef-core` is consumed as a git dependency, and the module is registered with both `ev_add_module` (under `EVEREST_ENABLE_RS_SUPPORT`) and the `MODULE.bazel` crate index plus `modules/Cargo.lock`.

Configuration keys — `profile` (path to a `.moddef`, `.moddef.yaml` or `.moddef.json`), `device_id`, `powermeter_device_id`, `read_meter_values_interval_ms`, `communication_errors_threshold` — are documented in the module README.

**Scope and known limitations**

- Transactions require a *typed-register* OCMF meter, where each OCMF field is its own register and the meter composes the signed document itself; the reference profile is the Carlo Gavazzi EM580. Meters that expose a single opaque OCMF blob assembled by the caller (for example the Bauer BSM or Iskra WM3M4C) need payload composition that cannot be expressed declaratively — they keep their dedicated drivers and this module reports `NOT_SUPPORTED` for them.
- Modbus RTU via `serial_communication_hub` only. A config-selected Modbus TCP transport is a natural follow-up, since ModDef's `Transport` trait already has a tokio-modbus backend.
- Discrete-input reads are not offered by the hub interface, and `COMPOSED` (multi-register mantissa/exponent) measurand points are not yet readable through the ModDef facade; such fields publish as absent.
- **Not hardware-verified.** The register-level behaviour has so far been exercised against mocks only.

**Tests.** Eight unit tests run against a mocked `serial_communication_hub` publisher, following the approach used by `RsIskraMeter`: metering (measurand lookup, decode, and the kWh → Wh rescale); `start_transaction` (an identification enum resolved against the profile's enum table to its register value, the `'B'` trigger written, the result register read); capability discovery via the `command_ref` role tag and the `NOT_SUPPORTED` paths; and the helpers, including a test pinning the assumption that the EVerest OCMF enums' `Debug` representation equals the OCMF value name, which the enum mapping relies on. Run in the `everest-ci` build-kit container: 8 passed, 0 failed, with `cargo fmt` and `cargo clippy` clean.

**Bazel.** `bazelisk build --config=ci` on this target completes the crate-universe repin (the `moddef-core` git dependency resolves) and the analysis phase — `66 packages loaded, 8 targets configured` — but I was not able to run the execution phase to completion locally for want of disk space, so a full Bazel build on CI is the outstanding confirmation. I would welcome a maintainer's view on whether the crate index needs anything further for a git-sourced dependency.

Closes #2504 — *Profile-driven powermeter module: describe Modbus meters (incl. OCMF signing) as data instead of per-model drivers*.

Background from the ModDef project, which motivated the design (context only):

- `ModDefOrg/moddef#2` — the multi-step command construct (`commands` / `poll` / `length_ref`) that this module drives.
- `ModDefOrg/moddef#3` — why declarative payload composition was rejected for opaque-blob meters, and why typed-register meters such as the EM580 are the ones this module can serve.

All commits are DCO signed off, new files carry SPDX and copyright headers, tests and documentation are included, and the change is limited to a single logical addition. The repository's `clang-format` lint covers C++ only, so the Rust sources are kept `rustfmt`- and `clippy`-clean instead. For transparency: the branch was rebased once *before* this PR was opened, solely to add the DCO sign-offs; there will be no further force-pushes during review.

### DIFF
--- a/modules/HardwareDrivers/PowerMeters/RsModDefMeter/BUILD.bazel
+++ b/modules/HardwareDrivers/PowerMeters/RsModDefMeter/BUILD.bazel
@@ -1,0 +1,70 @@
+# NOTE: The Bazel build additionally requires `moddef-core` (git) and `pollster`
+# to be present in the EVerest crate universe so `all_crate_deps` can resolve
+# them. Until the crate index is updated, build via cargo:
+#   EVEREST_CORE_ROOT=<repo root> cargo build   (from this module directory)
+load("@rules_rust//rust:defs.bzl", "rust_binary", "rust_test")
+load("@everest_crate_index//:defs.bzl", "all_crate_deps")
+load("@rules_rust//cargo:defs.bzl", "cargo_build_script")
+
+cargo_build_script(
+    name = "build_script",
+    srcs = ["build.rs"],
+    edition = "2021",
+    build_script_env = {
+        "EVEREST_CORE_ROOT": "../../../..",
+    },
+    data = [
+        "manifest.yaml",
+        "@everest-core//errors",
+        "@everest-core//interfaces",
+        "@everest-core//types",
+    ],
+    deps = all_crate_deps(build = True) + [
+        "//lib/everest/framework/everestrs/everestrs-build",
+    ],
+)
+
+rust_binary(
+    name = "RsModDefMeterBinary",
+    srcs = glob(["src/*.rs"]),
+    edition = "2021",
+    proc_macro_deps = all_crate_deps(proc_macro = True),
+    visibility = ["//visibility:public"],
+    deps = all_crate_deps(normal = True) + [
+        ":build_script",
+        "//lib/everest/framework/everestrs/everestrs",
+        "//lib/everest/framework/everestrs/everestrs:everestrs_sys",
+        "//lib/everest/framework/everestrs/everestrs:everestrs_bridge",
+    ],
+)
+
+rust_test(
+    name = "RsModDefMeterTest",
+    edition = "2021",
+    srcs = [],
+    crate = ":RsModDefMeterBinary",
+)
+
+binary = ":RsModDefMeterBinary"
+manifest = ":manifest.yaml"
+name = "RsModDefMeter"
+
+genrule(
+    name = "copy_to_subdir",
+    srcs = [binary, manifest],
+    outs = [
+        "{}/manifest.yaml".format(name),
+        "{}/{}".format(name, name),
+    ],
+    cmd = "mkdir -p $(RULEDIR)/{} && ".format(name) +
+          "cp $(location {}) $(RULEDIR)/{}/{} && ".format(binary, name, name) +
+          "cp $(location {}) $(RULEDIR)/{}/".format(manifest, name),
+)
+
+filegroup(
+    name = name,
+    srcs = [
+        ":copy_to_subdir",
+    ],
+    visibility = ["//visibility:public"],
+)
--- a/modules/HardwareDrivers/PowerMeters/RsModDefMeter/BUILD.bazel
+++ b/modules/HardwareDrivers/PowerMeters/RsModDefMeter/BUILD.bazel
@@ -1,3 +1,6 @@
+# SPDX-License-Identifier: Apache-2.0
+# Copyright 2026 ModDefOrg and Contributors to EVerest
+#
 # NOTE: The Bazel build additionally requires `moddef-core` (git) and `pollster`
 # to be present in the EVerest crate universe so `all_crate_deps` can resolve
 # them. Until the crate index is updated, build via cargo:
--- a/modules/HardwareDrivers/PowerMeters/RsModDefMeter/Cargo.toml
+++ b/modules/HardwareDrivers/PowerMeters/RsModDefMeter/Cargo.toml
@@ -1,3 +1,6 @@
+# SPDX-License-Identifier: Apache-2.0
+# Copyright 2026 ModDefOrg and Contributors to EVerest
+
 [package]
 name = "RsModDefMeter"
 version = "0.1.0"
--- a/modules/HardwareDrivers/PowerMeters/RsModDefMeter/Cargo.toml
+++ b/modules/HardwareDrivers/PowerMeters/RsModDefMeter/Cargo.toml
@@ -18,3 +18,11 @@ pollster = "0.3"
 anyhow = "1.0"
 log = "0.4"
 chrono = "0.4"
+# Optional, mirrors RsIskraMeter: the everestrs-generated code gates its mock
+# publisher behind these, so declaring them keeps `cfg(feature = "mockall")`
+# known (no warnings) and enables mock-based unit tests.
+mockall = { version = "0.12.1", optional = true }
+mockall_double = { version = "0.3.1", optional = true }
+
+[features]
+default = ["mockall", "mockall_double"]
--- a/modules/HardwareDrivers/PowerMeters/RsModDefMeter/Cargo.toml
+++ b/modules/HardwareDrivers/PowerMeters/RsModDefMeter/Cargo.toml
@@ -1,0 +1,20 @@
+[package]
+name = "RsModDefMeter"
+version = "0.1.0"
+edition = "2021"
+
+[build-dependencies]
+everestrs-build = { workspace = true }
+
+[dependencies]
+everestrs = { workspace = true }
+# ModDef runtime: profile loading, the Device facade, run_command and the
+# Transport trait we bridge onto the serial_communication_hub.
+moddef-core = { git = "https://github.com/ModDefOrg/moddef-rs" }
+# Minimal futures executor: drives moddef-core's async Device/Transport methods
+# to completion synchronously on the calling thread (all awaits complete without
+# a reactor because the transport and delay are blocking).
+pollster = "0.3"
+anyhow = "1.0"
+log = "0.4"
+chrono = "0.4"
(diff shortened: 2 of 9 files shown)


modules/HardwareDrivers/PowerMeters/RsModDefMeter/src/main.rs
```diff
@@ -0,0 +1,1107 @@
...
+}
+
+// Note: the everestrs-generated hub command arguments are ordered ALPHABETICALLY
+// by argument name, not in manifest order — e.g.
+// modbus_read_holding_registers(first_register_address, num_registers_to_read,
+// target_device_id). Matches the RsIskraMeter usage.
+impl moddef_core::Transport for HubBridge {
+    type Error = HubError;
+
+    async fn read_holding(&mut self, offset: u16, out: &mut [u16]) -> Result<(), HubError> {
```
dorezyuk, line 100: not sure about the trait but the call below is a blocking one - could we also use non-async here?

suda: Hi Dima! That's a good question. The async in this case is a formality as the runtime cost is one poll of an already-ready future per call, against a Modbus RTU round trip measured in milliseconds. So we can leave it as it is or:
1. We could write the impls as fn read_holding(..) -> impl Future<Output = ..>` returning `core::future::ready(..)` which would be more verbose way of clarifying this
2. We could add blocking `Transport` + blocking `Device` to `moddef-core` (just wrapping in `block_on`) and use those versions here. We didn't have any other consumer that would need it.

WDYT?

hikinggrass:
@suda I think you missed checking in the modules/HardwareDrivers/PowerMeters/RsModDefMeter/build.rs (that's unfortunately accidentally excluded from our .gitignore, we'll fix that in another PR) with `git add -f`. This can use the same code as in `modules/HardwareDrivers/PowerMeters/RsIskraMeter/build.rs`:
```rust
use everestrs_build::Builder;

pub fn main() {
    Builder::new(
        "manifest.yaml",
        vec![std::env::var("EVEREST_CORE_ROOT").unwrap()],
    )
    .generate()
    .unwrap();

    println!("cargo:rerun-if-changed=build.rs");
    println!("cargo:rerun-if-changed=manifest.yaml");
}

```

modules/HardwareDrivers/PowerMeters/RsModDefMeter/BUILD.bazel
```diff
@@ -0,0 +1,73 @@
...
+        "//lib/everest/framework/everestrs/everestrs",
+        "//lib/everest/framework/everestrs/everestrs:everestrs_sys",
+        "//lib/everest/framework/everestrs/everestrs:everestrs_bridge",
+    ],
+)
+
+rust_test(
+    name = "RsModDefMeterTest",
+    edition = "2021",
+    srcs = [],
```
hikinggrass, line 47:
```suggestion
    srcs = [],
    crate_features = ["mockall", "mockall_double"],
```

otherwise the build fails

hikinggrass (changes requested):

suda:
Thanks @hikinggrass! Both addressed. Branch is also rebased on current `main` (hence the force-push). Ready for another look.

modules/HardwareDrivers/PowerMeters/RsModDefMeter/Cargo.toml
```diff
@@ -0,0 +1,31 @@
...
+edition = "2021"
+
+[build-dependencies]
+everestrs-build = { workspace = true }
+
+[dependencies]
+everestrs = { workspace = true }
+# ModDef runtime: profile loading, the Device facade, run_command and the
+# Transport trait we bridge onto the serial_communication_hub.
+moddef-core = { git = "https://github.com/ModDefOrg/moddef-rs" }
```
hikinggrass, line 16: This should pin at least a specific commit (with `rev=`), or a version that is published on crates.io

hikinggrass (changes requested):

PR #2511, opened 2026-07-23
feat(YetiSimulator): make reported import current limits configurable

sjh37:
## Describe your changes

`YetiSimulator` reported a hard-coded 32 A / 6 A as its import current capability. Those two values are now
config keys, `max_current_A_import` and `min_current_A_import`, with the defaults left at 32.0 and 6.0 so every
shipped config behaves exactly as before.

**Why this matters beyond AC.** The energy manager derives the power available to an EVSE from

```
power = max_current_A_import * max_phase_count_import * nominal_voltage_V
```

so the board support figure caps a DC configuration at the AC-side number regardless of what the DC supply can
deliver. Using `YetiSimulator` in a high-power DC SIL setup silently limits the session to roughly 22 kW, with
no error and nothing in the logs pointing at the cause - the only way I found it was enabling
`energy_manager: debug` and reading which limit was binding each trading round.

Note for reviewers: `struct Conf` in `YetiSimulator.hpp` is checked in rather than generated, so it is edited by
hand here to match the manifest. Without that the build fails in the generated `ld-ev.cpp`.

None - happy to open one first if that is preferred.

Verified by taking a simulated CCS session from ~22 kW to 240 kW at 800 V / 300 A against a live CSMS, and by
confirming the shipped configs are unaffected because the defaults are unchanged.

### DIFF
--- a/modules/Simulation/YetiSimulator/YetiSimulator.hpp
+++ b/modules/Simulation/YetiSimulator/YetiSimulator.hpp
@@ -30,6 +30,8 @@ struct Conf {
     bool dummy_meter_value_send_on_transaction_start;
     std::string dummy_meter_value_blob_start;
     std::string dummy_meter_value_blob_stop;
+    double max_current_A_import;
+    double min_current_A_import;
 };
 
 class YetiSimulator : public Everest::ModuleBase {
--- a/modules/Simulation/YetiSimulator/YetiSimulator.hpp
+++ b/modules/Simulation/YetiSimulator/YetiSimulator.hpp
@@ -33,6 +33,8 @@ struct Conf {
     std::string dummy_meter_value_blob_stop;
     double ac_nominal_voltage;
     double ac_nominal_frequency;
+    double max_current_A_import;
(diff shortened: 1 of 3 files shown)


Pietfried:
Hi @sjh37 , thanks for your contribution! The changes look good to me! Could you do a rebase and resolve the conflicts?

sjh37:
Hi @Pietfried - rebased onto current `main` and the conflicts are resolved.

The conflicts came from #2523 and #2533 landing in the same two places this branch touches: the tail of `config:` in `manifest.yaml` and the tail of `struct Conf`. Both were add/add, so I kept both sides and appended `max_current_A_import` / `min_current_A_import` after the new `ac_nominal_*` keys, keeping the `Conf` field order in step with the manifest. `evse_board_supportImpl.cpp` applied unchanged.

The diff is otherwise identical to what you reviewed - defaults still 32.0 and 6.0, so shipped configs are unaffected.

Pietfried (approved):

PR #2512, opened 2026-07-23
feat(EvManager): allow the simulated EV's EVCCID to be set at runtime

sjh37:
## Describe your changes

Lets the simulated EV announce an EVCCID other than the MAC address of its network interface, so one SIL setup
can simulate a series of different vehicles arriving.

**Why.** josev derives the EVCCID from the NIC (`get_nic_mac_address`), which is fixed for the lifetime of the
process - in a container, for the lifetime of the container. Autocharge identifies a vehicle solely by that
address: `EvseManager` turns the EVCCID into the token `VID:<mac>` and sends it to the CSMS as the `Authorize`
idTag. So today a SIL deployment can only ever present one vehicle, and testing anything fleet-shaped means
recreating the container with a different MAC.

image

**What changed**

- `interfaces/ISO15118_ev.yaml` - new `set_evcc_id` command, returning the EVCCID that will actually be
  announced (12 uppercase hex digits) or an empty string if the argument was not a MAC address.
- `PyEvJosev` - new `evcc_id` config key, plus the runtime command. Implemented by subclassing josev's
  `SimEVController` and overriding `get_evcc_id`, so nothing in the fetched josev dependency is patched. The
  override is read fresh per session.
- `EvManager` - a `set_evcc_id` simulation command and an
  `everest_external/nodered/<n>/carsim/cmd/set_evcc_id` MQTT topic, echoing the accepted value back retained on
  `.../carsim/state/evcc_id` as `VID:...`, or `rejected`.
- `config/nodered/config-sil-dc-flow.json` - a "Vehicle MAC (VID)" input and an "Announced VID" readout on the
  Connector 1 card.

Input is accepted as `VID:0242AC110099`, `02:42:AC:11:00:99` or `0242ac110099`; blank restores the interface
MAC address.

**Notes for reviewers**

- This adds a command to a shared interface. `PyEvJosev` is the only implementer of `ISO15118_ev` today, so
  nothing else needs updating, but flagging it as an API change.
- The override applies from the **next** V2G session, since the EVCCID is fixed at `SessionSetupReq`.
- ISO 15118-20 is deliberately unaffected - it announces a manufacturer string rather than a MAC, so a MAC
  override is meaningless there and is skipped.
- Only the DC SIL flow gained the dashboard widget. Happy to add it to the other flows in `config/nodered/` if
  you would like that in the same PR.
- No automated tests added. `EvManager/tests/` would be the natural home for coverage of the command
  registration and the MAC normalisation - happy to add it if you want that before merge.

None - happy to open one first if that is preferred.

Verified end to end against a live CSMS: setting `VID:AABBCCDDEE01` produced
`EVCCID: AA:BB:CC:DD:EE:01` in `SessionSetupReq` and `"idTag": "VID:AABBCCDDEE01"` in the OCPP transcript,
where the container's own MAC would otherwise have been used. Blank input restored the interface MAC, and a
non-MAC input was rejected with the previous value left in place.

### DIFF
--- a/config/nodered/config-sil-dc-flow.json
+++ b/config/nodered/config-sil-dc-flow.json
@@ -1924,5 +1924,97 @@
                 "620b0d248a89ece0"
             ]
         ]
+    },
+    {
+        "id": "b1d0ae5efa11ce00",
+        "type": "comment",
+        "z": "ed603c51db9dcbb9",
+        "name": "Simulated vehicle identity (Autocharge)",
+        "info": "The EV announces its MAC address as the EVCCID in SessionSetupReq. EvseManager turns that into the Autocharge token \"VID:<mac>\" and sends it to the CSMS as the Authorize idTag.\n\nWithout an override the MAC comes from the container's network interface, so every simulated session looks like the same car. Setting a value here makes a different vehicle turn up on the next plug-in.\n\nAccepts VID:0242AC110099, 02:42:AC:11:00:99 or 0242AC110099. Leave blank to go back to the interface MAC address.",
+        "x": 240,
+        "y": 1240,
+        "wires": []
+    },
+    {
+        "id": "b1d0ae5efa11ce01",
+        "type": "ui_text_input",
+        "z": "ed603c51db9dcbb9",
+        "name": "Vehicle MAC (VID)",
+        "label": "Vehicle MAC (VID)",
+        "tooltip": "Applied on the next plug-in. Blank restores the network interface MAC address.",
+        "group": "b364f7eb4621082b",
+        "order": 14,
+        "width": 0,
+        "height": 0,
+        "passthru": false,
+        "mode": "text",
+        "delay": "0",
+        "topic": "everest_external/nodered/#/carsim/cmd/set_evcc_id",
+        "sendOnBlur": true,
+        "className": "",
+        "topicType": "str",
+        "x": 200,
+        "y": 1300,
+        "wires": [
+            [
+                "361b3d846c4e6673"
+            ]
+        ]
+    },
+    {
+        "id": "b1d0ae5efa11ce02",
+        "type": "mqtt in",
+        "z": "ed603c51db9dcbb9",
+        "name": "",
+        "topic": "everest_external/nodered/+/carsim/state/evcc_id",
+        "qos": "2",
+        "datatype": "auto",
+        "broker": "fc8686af.48d178",
+        "nl": false,
+        "rap": true,
+        "rh": 0,
+        "inputs": 0,
+        "x": 290,
+        "y": 620,
+        "wires": [
+            [
+                "b1d0ae5efa11ce03"
+            ]
+        ]
+    },
+    {
+        "id": "b1d0ae5efa11ce03",
+        "type": "function",
+        "z": "ed603c51db9dcbb9",
+        "name": "Filter connector number",
+        "func": "// The retained EVCCID is delivered the instant the broker connects, which can beat the inject node that\n// seeds connector_number, so fall back to this tab's connector rather than dropping the value silently.\nconst connector = flow.get('connector_number') || '1';\nif (msg.topic.indexOf(String(connector)) > -1) return msg;",
+        "outputs": 1,
+        "noerr": 0,
+        "initialize": "",
+        "finalize": "",
+        "libs": [],
+        "x": 710,
+        "y": 620,
+        "wires": [
+            [
+                "b1d0ae5efa11ce04"
+            ]
+        ]
+    },
+    {
+        "id": "b1d0ae5efa11ce04",
+        "type": "ui_text",
+        "z": "ed603c51db9dcbb9",
+        "group": "b364f7eb4621082b",
+        "order": 15,
+        "width": 0,
+        "height": 0,
+        "name": "",
+        "label": "Announced VID",
+        "format": "{{msg.payload}}",
+        "layout": "row-spread",
+        "x": 1000,
+        "y": 620,
+        "wires": []
     }
 ]
--- a/config/nodered/config-sil-dc-flow.json
+++ b/config/nodered/config-sil-dc-flow.json
@@ -1810,7 +1810,7 @@
         "type": "function",
         "z": "ed603c51db9dcbb9",
         "name": "Buffer sim commands",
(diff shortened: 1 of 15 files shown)


modules/EV/EvManager/main/car_simulatorImpl.cpp
```diff
@@ -415,6 +421,18 @@ void car_simulatorImpl::subscribe_to_external_mqtt() {
                        auto data_copy = data;
                        handle_modify_charging_session(data_copy);
                    });
+    mqtt.subscribe("everest_external/nodered/" + std::to_string(mod->config.connector_id) + "/carsim/cmd/set_evcc_id",
+                   [this](const std::string& data) { publish_evcc_id(car_simulation->set_evcc_id(data)); });
```
SebaLukas: I would remove these lines. To change the evcc id the external `execute_charging_session` or `modify_charging_session` should be used and should be enough.

config/nodered/config-sil-dc-flow.json
SebaLukas, line 1: It would be great if the other Node-RED flows were updated as well. That should be included in this PR.

SebaLukas (commented):
Thank you very much for your contribution.
Overall looking good but I have two comments. Please take a look at them

modules/EV/EvManager/main/car_simulatorImpl.cpp
sjh37: **On the dedicated MQTT topic** - agreed, removing it. `set_evcc_id` is already registered as a simulation command, so the topic was redundant surface area on a shared module. I've checked that `publish_evcc_id()` stays reachable through the registered command, so the retained `carsim/state/evcc_id` readout still works after the removal. Your route is also better semantically: putting `set_evcc_id <mac>` at the front of the session command string sets the identity as part of the session, which removes the "applies on the next plug-in" caveat I had to document.

Two things fall out of the removal that I'd like your steer on.

**1. Clearing the override.** The topic accepted an empty payload to fall back to the interface MAC address. The command path can't express that, since `register_command("set_evcc_id", 1, ...)` requires exactly one argument. `CommandRegistry` is a multimap that resolves by arity - `iso_start_v2g_session` already has 1/2/3/4-argument variants - so I'd add a 0-argument `set_evcc_id` overload that restores the interface MAC. Happy to drop the clear behaviour instead if you'd rather keep the command surface minimal.

**2. Registration guard.** `set_evcc_id` is currently registered inside the `if (!mod->r_ev.empty())` block with the `iso_*` commands, and `CommandRegistry::get_registered_command` throws `std::invalid_argument` for an unregistered name, with no `try`/`catch` around `parse_sim_commands` in `update_command_queue`. So on a connector with no `ISO15118_ev` connection, `set_evcc_id` in a command string raises rather than logging. This is pre-existing behaviour shared by all the `iso_*` commands, not something this PR introduces, but it becomes relevant to your second comment below.

`CarSimulation::set_evcc_id` already handles the missing-connection case gracefully - it logs `set_evcc_id requires an ISO15118_ev connection, none is configured` and returns an empty string. So I'd like to move just this one registration outside the guard. That makes the command safe on any connector and keeps this PR's change self-contained. Let me know if you'd prefer I leave the guard alone.

config/nodered/config-sil-dc-flow.json
sjh37: **On the other Node-RED flows** - will do, with one caveat worth flagging. `config-sil-two-evse-flow.json` and `config-sil-energy-management-flow.json` both have a connector 2 group labelled `[Basic Charging]` whose `ev_manager_2` has no `ISO15118_ev` requirement wired. If the registration guard stays as it is, a VID widget there would be misleading at best, so I'd add the widget only to the `[ISO15118]` connectors and leave connector 2 alone. If you're happy with the guard change in point 2 above, I'll add it uniformly to all connectors instead, which I think is the cleaner outcome.

I also need to rework the widget itself rather than repoint it. `handle_modify_charging_session` replaces the whole command queue via `update_command_queue`, so publishing a bare `set_evcc_id <mac>` mid-session would discard the remaining commands and trigger `reset_car_simulation_defaults()`. The text input is `sendOnBlur`, so simply clicking away from the field would end a running session. Instead I'll have the input store the value in a Node-RED flow variable and prepend `set_evcc_id <mac>;` to the existing session-start command strings, which is non-destructive and matches how the flows already compose commands.

Finally, there's currently no test coverage for this command. If it's going to be reachable only as a parsed simulator command, I'll add a `SimCommandTest` case covering the parse and the argument handling. Say if you'd rather that went in a separate PR.

sjh37:
Hi @SebaLukas Both comments are actioned. 

**1. The dedicated MQTT topic is gone.** `carsim/cmd/set_evcc_id` is removed. `set_evcc_id` is now reachable
only as a simulation command, via `execute_charging_session` / `modify_charging_session`, as you asked.

Two things followed from that:

- **Clearing the override.** The topic accepted an empty payload to fall back to the interface MAC address, and
  a 1-argument command cannot express that. I have added a 0-argument `set_evcc_id` overload that restores the
  interface MAC, so `set_evcc_id 0242AC110099` sets and a bare `set_evcc_id` clears. `CommandRegistry` already
  resolves by arity and `CommandRegistryTest` already covers that, so this is no new machinery. Without it the
  override would be unclearable at runtime, because the EV module holds it across sessions.
- **The registration guard stays as it is.** I raised moving `set_evcc_id` out of the `if (!mod->r_ev.empty())`
  block; it turns out not to be needed for this PR, so I have left it alone. See point 2 for why. The wider
  point that an unregistered command throws out of `parse_sim_commands` still stands for every `iso_*` command,
  but that is pre-existing and belongs in its own PR.

**2. All five Node-RED flows now have the widget**, on connector 1 only. `ev_manager_2` in
`config-sil-two-evse.yaml` and `config-sil-energy-management.yaml` has only an `ev_board_support` connection, so
those `[Basic Charging]` connectors have no `ISO15118_ev` to talk to and a VID field there would be misleading.
Each connector 2 has its own "Buffer sim commands" node on its own tab, so nothing leaks between them.

The widget is reworked rather than repointed. Publishing a bare `set_evcc_id <mac>` mid-session would have gone
through `update_command_queue` and discarded the rest of the queue, and the input is `sendOnBlur`, so clicking
away from the field would have ended a running session. Instead the input stores the value in a flow variable
and "Buffer sim commands" prepends `set_evcc_id <mac>;` to the session start string. That matches how the flows
already compose commands, and because the EVCCID is fixed at `SessionSetupReq` it now applies to the session
being started rather than the one after it. The "applies on the next plug-in" caveat is gone.

Input is normalised in the flow as well as in the EV module: `VID:0242AC110099`, `02:42:AC:11:00:99`,
`02-42-ac-11-00-99` and `0242ac110099` all work, anything else is rejected with a node warning and leaves the
previous value in place, and a blank field sends the bare `set_evcc_id` to restore the interface MAC.

**One thing I have kept, so please say if you would rather it went.** The retained
`carsim/state/evcc_id` readout is still there, now published only when a `set_evcc_id` command actually runs. I
read your comment as being about the command topic being redundant with `execute_charging_session`, which it
was. The state topic is not redundant with anything: it is how the dashboard confirms what the EV actually
accepted, since Node-RED cannot know that on its own. Happy to drop it and have the flow display only what was
typed if you would prefer no new MQTT surface at all.

**Tests.** Added a `SimCommandTest` scenario covering arity resolution through the parser, which the new
0/1-argument overload depends on and which was not covered before. Also updated `EvManager/docs/index.rst` and
the `ISO15118_ev` command description to match the new behaviour.

PR #2514, opened 2026-07-23
OCPP: Allow disabling connectors with InstallScheduled status message

nzbr:
## Describe your changes

Additionally allow a firmware status message with a status of InstallScheduled to carry a request to disable the connectors. Currently this is only possible with a status of `SignatureVerified` for signed updates and `Downloaded` for unsigned updates


lib/everest/ocpp/lib/ocpp/v16/charge_point_impl.cpp
```diff
@@ -4671,23 +4682,35 @@ void ChargePointImpl::on_log_status_notification(std::int32_t request_id, std::s
...
-void ChargePointImpl::on_firmware_update_status_notification(std::int32_t request_id,
-                                                             const FirmwareStatusNotification firmware_update_status,
-                                                             const bool disable_connectors_during_install) {
+void ChargePointImpl::on_firmware_update_status_notification(
+    std::int32_t request_id, const FirmwareStatusNotification firmware_update_status,
+    const std::optional<bool> disable_connectors_during_install) {
+    if (firmware_update_status == FirmwareStatusNotification::DownloadScheduled or
+        firmware_update_status == FirmwareStatusNotification::Downloading) {
+        this->all_connectors_unavailable_notified = false;
+    }
```
Pietfried: This is not reset e.g. when the fw update aborts/crashes and doesnt notify or reports Idle and then this guards stays latched. I think we should reset on every firmware update request and we can remove this part here . This also needs fixing in v2.

lib/everest/ocpp/lib/ocpp/v2/charge_point.cpp
```diff
@@ -645,8 +651,18 @@ void ChargePoint::initialize(const std::map<std::int32_t, std::int32_t>& evse_co
...
+        guarded_all_connectors_unavailable_callback = [this]() {
+            if (!this->all_connectors_unavailable_notified.exchange(true)) {
+                this->callbacks.all_connectors_unavailable_callback.value()();
+            }
+        };
+    }
+
     this->availability = std::make_unique<Availability>(*functional_block_context, this->callbacks.time_sync_callback,
-                                                        this->callbacks.all_connectors_unavailable_callback);
+                                                        guarded_all_connectors_unavailable_callback);
```
Pietfried, line 671: Shouldnt the `guarded_all_connectors_unavailable_callback` only be passed to this->firmware_update but not to this->availability? This could cause availability to fire this callback for events that have nothing to do with firmware and latch the `all_connectors_unavailable_notified` guard.

lib/everest/ocpp/lib/ocpp/v2/functional_blocks/firmware_update.cpp
```diff
@@ -44,62 +44,76 @@ void FirmwareUpdate::handle_message(const ocpp::EnhancedMessage<MessageType>& me
...
+            if (disable_connectors_during_install.value_or(true)) {
+                this->change_all_connectors_to_unavailable_for_firmware_update();
+            }
+        }
+    }
+
+    // Explicitly allow disabling the connectors when an update is scheduled
+    if (firmware_update_status == FirmwareStatusEnum::InstallScheduled and
+        disable_connectors_during_install.value_or(false)) {
+        this->change_all_connectors_to_unavailable_for_firmware_update();
```
Pietfried, line 116: Shouldnt this also go into the is_duplicate block?

lib/everest/ocpp/lib/ocpp/v16/charge_point_impl.cpp
```diff
@@ -4671,23 +4682,35 @@ void ChargePointImpl::on_log_status_notification(std::int32_t request_id, std::s
...
+    // Explicitly allow disabling the connectors when an update is scheduled
+    if (firmware_update_status == FirmwareStatusNotification::InstallScheduled and
+        disable_connectors_during_install.value_or(false)) {
+        this->set_firmware_install_pending(true);
+        this->change_all_connectors_to_unavailable_for_firmware_update();
+    }
+
     if (firmware_update_status == FirmwareStatusNotification::InstallationFailed or
         firmware_update_status == FirmwareStatusNotification::Installed or
         firmware_update_status == FirmwareStatusNotification::InstallVerificationFailed) {
```
Pietfried, line 4729: We should also treat `Idle` as a reset state and run the restore sequence

lib/everest/ocpp/include/ocpp/v16/charge_point_impl.hpp
```diff
@@ -265,6 +266,15 @@ class ChargePointImpl : ocpp::ChargingStationBase {
     /// update can proceed
     void change_all_connectors_to_unavailable_for_firmware_update();
 
+    /// \brief Marks a firmware install as pending; connectors will be disabled (now or once running transactions
+    /// end) when \p disable_connectors is true.
+    void set_firmware_install_pending(bool disable_connectors);
+
+    /// \brief Clears all pending-install state set by \ref set_firmware_install_pending, re-arms the
+    /// all_connectors_unavailable_callback single-fire guard and drops availability changes queued for the
+    /// firmware update that have not been executed yet.
```
Pietfried, line 275: This claim of `drops availability changes queued for the
    /// firmware update that have not been executed yet.` is not actually implemented in the cpp file. You need to erase every change_availability_queue entry whose persist == false . The v2 path has the same (pre-existing) issue.

Pietfried (changes requested):

lib/everest/ocpp/lib/ocpp/v16/charge_point_impl.cpp
nzbr: Makes sense, I moved the reset to when a new UpdateFirmwareRequest is received for both versions

lib/everest/ocpp/lib/ocpp/v2/charge_point.cpp
nzbr: The update doesn't need to know _why_ the connectors became unavailable, just _that_ they did, so I added the guarded callback here as well. But you are right that this could have made the guard latch too early. Instead of splitting the callback I switched to a tri-state variable instead of a boolean. That way the guard only latches if there is actually an update waiting

lib/everest/ocpp/lib/ocpp/v2/functional_blocks/firmware_update.cpp
nzbr: No, this explicitly needs to stay outside the is_duplicate block, because we are expecting duplicate InstallScheduled messages to arrive sometimes: If the connectors can't be disabled immediately, libocpp will itself issue an InstallScheduled status, but if the updater module issues another one that carries `disable_connectors_during_install`, this should still be honoured

lib/everest/ocpp/lib/ocpp/v16/charge_point_impl.cpp
nzbr: Done

lib/everest/ocpp/include/ocpp/v16/charge_point_impl.hpp
nzbr: Should be fixed for both now

PR #2517, opened 2026-07-24
fix(cmake): use SQLite3::SQLite3 instead of deprecated SQLite::SQLite3

### LINKED ISSUE #2516: Fix CMake warnings with CMake 4.3+
FindSQLite3.cmake in CMake 4.3 deprecated the `SQLite::SQLite3` target name in favor of `SQLite3::SQLite3`.
This leads to a lot of warnings on configure like:
```bash
CMake Warning (dev) at modules/Misc/PersistentStore/CMakeLists.txt:14 (target_link_libraries):
  The library that is being linked to, SQLite::SQLite3, is marked as being
  deprecated by the owner.  The message provided by the developer is:

  The target name SQLite::SQLite3 is deprecated.  Please use SQLite3::SQLite3
  instead.

This warning is for project developers.  Use -Wno-dev to suppress it.
```

We should use the new target name and [use the workaround stated in the commit that introduced the new target name](https://github.com/Kitware/CMake/commit/d4754c1a0c7110429db4854c13bc0baf4608a9e7):
```
If your project needs to support CMake < 4.3, consider adding the following
  to your project after calling ``find_package(SQLite3 ...)``:

  .. code-block:: cmake

    if(NOT TARGET SQLite3::SQLite3) # CMake < 4.3
      add_library(SQLite3::SQLite3 ALIAS SQLite::SQLite3)
    endif()
```

Shaurya2k06:
CMake 4.3 deprecates the FindSQLite3 target `SQLite::SQLite3` in favor of `SQLite3::SQLite3`, which produces developer warnings at configure time. Switch all link sites to the new target and add the documented CMake < 4.3 alias workaround after each `find_package(SQLite3)`.

### Describe your changes
- Replace `SQLite::SQLite3` with `SQLite3::SQLite3` in everest-sqlite, libocpp, ErrorHistory, and PersistentStore
- After each `find_package(SQLite3 REQUIRED)`, add an alias for CMake < 4.3 (with `IMPORTED_GLOBAL` so the alias works below CMake 3.18)

### Issue ticket number and link
Closes #2516 

### Checklist before requesting a review

### DIFF
--- a/lib/everest/ocpp/CMakeLists.txt
+++ b/lib/everest/ocpp/CMakeLists.txt
@@ -46,6 +46,10 @@ endif()
 # dependencies
 find_package(Boost COMPONENTS program_options thread REQUIRED)
 find_package(SQLite3 REQUIRED)
+if(NOT TARGET SQLite3::SQLite3) # CMake < 4.3
+  set_target_properties(SQLite::SQLite3 PROPERTIES IMPORTED_GLOBAL TRUE)
+  add_library(SQLite3::SQLite3 ALIAS SQLite::SQLite3)
+endif()
 find_package(OpenSSL 3 REQUIRED)
 
 if(NOT DISABLE_EDM)
--- a/lib/everest/ocpp/lib/CMakeLists.txt
+++ b/lib/everest/ocpp/lib/CMakeLists.txt
@@ -228,7 +228,7 @@ target_link_libraries(ocpp
     PRIVATE
         OpenSSL::SSL
         OpenSSL::Crypto
-        SQLite::SQLite3
+        SQLite3::SQLite3
         Threads::Threads
         nlohmann_json::nlohmann_json
         ryml::ryml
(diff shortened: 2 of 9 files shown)


tests/ocpp_tests/test_sets/ocpp16/ocpp_generic_interface_integration_tests.py
hikinggrass, line 1: This change has nothing to do with the issue and the changes should be reverted

tests/ocpp_tests/conftest.py
hikinggrass, line 1: This change has nothing to do with the issue and the changes should be reverted

hikinggrass (changes requested):

tests/ocpp_tests/test_sets/ocpp16/ocpp_generic_interface_integration_tests.py
Shaurya2k06: agreed, but seems like a complement to the issue. As mentioned in commit description, the OCPP
probe fixtures were not updated, so probe_module.start() fails missing that command. Happy to pick this up separately

hikinggrass (approved):

PR #2519, opened 2026-07-26
fix(EvseManager): stop pushing stale DC HLC limits after session stop

florinmihut:
## Describe your changes

After SessionStop the HLC stack sends D-LINK_TERMINATE while the car may stay plugged in for a long time (charger in `ChargingPausedEV`). In this phase `ev_info` still contains values from the previous session (e.g. an old `target_voltage`), which were pushed to the ISO 15118 stack and the power supply on every `enforce_limits` update.

This PR gates the DC branch of `energyImpl::handle_enforce_limits` on a new free helper `should_update_dc_hlc_limits(charger_state, hlc_link_status)`:

- returns `false` once the HLC link was terminated (`HlcTerminatePause::Terminate` latch, read under `state_machine_mutex` via a new `Charger::get_hlc_terminate_pause()`), since no limit update can reach the EV anymore until a new session starts
- returns `true` only in active-session states (`WaitingForAuthentication`, `PrepareCharging`, `Charging`, `ChargingPausedEV`, `ChargingPausedEVSE`)

The `Terminate` latch is reset to `Unknown` in the Idle init block and on entering `Charging`, so a new session is not blocked by a stale value.

Also fixes `SharedContext::hlc_charging_terminate_pause` being read uninitialized if `enforce_limits` arrives before the state machine's first Idle init.

The helper lives in its own `dc_hlc_limits.cpp/.hpp` (same pattern as `energy_transfer_modes.*`) so it can be unit tested; `energyImpl.cpp` is not part of any test target. Optionally, we could merge them in one charger_helpers.* if this is wanted.

n/a

## Tests

New unit test `DcHlcLimitsTest.cpp` (added to the Charger test target) covers:
- updates allowed in all active-session states while the link is not terminated
- regression: no updates in any state once D-LINK_TERMINATE was received
- no updates outside of a session (`Idle`, `Finished`, `StoppingCharging`, `Disabled`, `T_step_*`, `SwitchPhases`)

### DIFF
--- a/modules/EVSE/EvseManager/CMakeLists.txt
+++ b/modules/EVSE/EvseManager/CMakeLists.txt
@@ -12,6 +12,7 @@ ev_setup_cpp_module()
 target_sources(${MODULE_NAME}
     PRIVATE
         Charger.cpp
+        dc_hlc_limits.cpp
         energy_transfer_modes.cpp
         SessionLog.cpp
         v2gMessage.cpp
--- a/modules/EVSE/EvseManager/Charger.cpp
+++ b/modules/EVSE/EvseManager/Charger.cpp
@@ -2085,6 +2085,11 @@ void Charger::set_hlc_charging_active() {
     shared_context.hlc_charging_active = true;
 }
 
+Charger::HlcTerminatePause Charger::get_hlc_terminate_pause() {
+    Everest::scoped_lock_timeout lock(state_machine_mutex, Everest::MutexDescription::Charger_get_hlc_terminate_pause);
+    return shared_context.hlc_charging_terminate_pause;
+}
+
 void Charger::set_hlc_allow_close_contactor(bool on) {
     Everest::scoped_lock_timeout lock(state_machine_mutex,
                                       Everest::MutexDescription::Charger_set_hlc_allow_close_contactor);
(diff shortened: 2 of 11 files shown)


SebaLukas:
I am sorry but I dont understand this PR. Which issue will be fixed?
If the hlc session is terminated then the EnergyManager should also be able to update the limits for EvseV2G and libiso15118. For the terminated hlc session the new limits are not relevant anymore, but for the new session it could be relevant.

florinmihut:
The why:
The issue: after D-LINK_TERMINATE with the car still plugged in (can be hours in ChargingPausedEV), ev_info is never cleared — it only resets at SessionFinished/session-start. So every enforce_limits tick kept deriving evse_maximum_current_limit from the previous session's target_voltage and kept calling apply_new_target_voltage_current() with that stale target. The PR stops pushing those wrong values, wasting CPU and the MQTT ping/pong.

The how:
If I understand you correctly, the latch (hlc_charging_terminate_pause) only clearing on entering Charging, a plugged-in restart (ChargingPausedEV -> BCB -> PrepareCharging) would reach its ChargeParameterDiscovery before the "gate" reopens, so an EnergyManager limit change during the pause wouldn't reach the new session. 
We can fix that resetting the latch when the new session starts (PrepareCharging) rather than at CurrentDemand, so fresh limits are pushed before parameter discovery completes.

PR #2521, opened 2026-07-27
feat(Huawei_V100R023C10): make Modbus communication timeout configurable

florinmihut:
## Describe your changes

The timeout after which lost Modbus communication with the Huawei power unit raises a `power_supply_DC/CommunicationFault` was hardcoded to 60 s in the fusion-charger dispenser library (`DispenserConfig::modbus_timeout_ms`) and could not be changed via the module configuration. 60 s is a long time to keep charging after the PU has gone silent before declaring the link dead.

This change:

- Adds a new manifest config parameter `modbus_timeout_s` (integer, default **10**, range **1–120**) to the `Huawei_V100R023C10` module and wires it into `DispenserConfig::modbus_timeout_ms` in `init()`.
- Lowers the library's built-in default from 60 s to 10 s so direct library users get the same default.

Note on choosing values: the poll-thread timer resets on *any* Modbus PDU received from the PU (which is the Modbus master), so the practical lower bound depends on the PU's natural register-write cadence; very low values may false-positive.

n/a

### DIFF
--- a/modules/HardwareDrivers/PowerSupplies/Huawei_V100R023C10/Huawei_V100R023C10.cpp
+++ b/modules/HardwareDrivers/PowerSupplies/Huawei_V100R023C10/Huawei_V100R023C10.cpp
@@ -126,6 +126,7 @@ void Huawei_V100R023C10::init() {
     dispenser_config.verify_secure_goose_hmac = config.verify_secure_goose;
     dispenser_config.module_placeholder_allocation_timeout =
         std::chrono::seconds(config.module_placeholder_allocation_timeout_s);
+    dispenser_config.modbus_timeout_ms = std::chrono::seconds(config.modbus_timeout_s);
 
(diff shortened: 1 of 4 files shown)


rolfSp:
Is it ensured that the Modbus master (Huawei_V100R023C10) initiates a communication (PDU) every 10 seconds ?

florinmihut:
No, and we can't control that. Dispenser to PU is controlled by us. PU to dispenser: this is the only thing that resets the timer, and nothing in this codebase ensures its cadence. It depends entirely on the Huawei PU firmware's polling/write behavior.
Since I don't have the HW I can't tell or measure the cadence, you will need to check on a real HW and do the packet capture on an idle charger and it will show the PU's request period. Set the timer then accordingly.

Pietfried (dismissed):

Pietfried (approved):

PR #2524, opened 2026-07-28
feat(libiso15118): Run a single session with an external provided connected socket

SebaLukas:
## Describe your changes
This PR adds a start_session() that runs a single ISO15118 session over an externally-provided, already-connected socket.

Previously the controller could only obtain sessions through the SDP server and/or the listening TCP/TLS socket. This change enables applications that manage their own connection setup, without the SDP server being active.

### DIFF
--- a/lib/everest/iso15118/include/iso15118/tbd_controller.hpp
+++ b/lib/everest/iso15118/include/iso15118/tbd_controller.hpp
@@ -25,6 +25,13 @@
 
 namespace iso15118 {
 
+enum class StartSessionResult : std::uint8_t {
+    KeepFdOpen,
+    FdNotValid,
+    FdClosed,
+    SessionComplete,
+};
+
 struct TbdConfig {
     config::SSLConfig ssl{};
     std::string interface_name;
@@ -32,6 +39,9 @@ struct TbdConfig {
     bool enable_sdp_server{true};
 };
 
+// TbdController is single-threaded. Exactly one driver may run at a time: either loop() or start_session(), driven from
+// a single thread. They must never run concurrently - they share the poll manager and session state, which are not
+// thread-safe. Sequential reuse (one driver returns, then the other is called) is allowed.
 class TbdController {
 public:
     // Creates the connection for sessions when the SDP server is disabled.
@@ -42,7 +52,12 @@ public:
     TbdController(TbdConfig, session::feedback::Callbacks, d20::EvseSetupConfig, ConnectionFactory);
     ~TbdController();
 
+    // Mutually exclusive with the other driver; see the class note. Refuses (logs and returns) if the other driver is
+    // already running.
     void loop();
+    // Mutually exclusive with the other driver; see the class note. Refuses (logs and returns false) if the other
+    // driver is already running.
+    StartSessionResult start_session(int connected_fd);
     void tick();
 
     bool has_active_session() const {
@@ -76,11 +91,23 @@ private:
     std::atomic_bool shutdown_active{false};
     std::atomic_bool terminate_session_requested{false};
 
+    // Guards against loop() and start_session() running concurrently (see class note).
+    std::atomic_bool driver_running{false};
+
     bool shutdown_signaled{false};
     TimePoint next_event{};
 
     // callbacks for sdp server
     void handle_sdp_server_input();
+
+    // Runs one poll_manager.poll() step using next_event; returns false if poll()
+    // threw, so the caller can break out of its loop.
+    bool poll_once();
+
+    // Services the currently active session for one cycle: propagates a pending
+    // shutdown, consumes the terminate request, polls the session and reaps it when
+    // finished. Shared by loop()/tick() and start_session().
+    void service_active_session();
 
     const TbdConfig config;
     const session::feedback::Callbacks callbacks;
--- a/lib/everest/iso15118/src/iso15118/tbd_controller.cpp
+++ b/lib/everest/iso15118/src/iso15118/tbd_controller.cpp
@@ -4,6 +4,8 @@
 
 #include <algorithm>
 #include <chrono>
+#include <fcntl.h>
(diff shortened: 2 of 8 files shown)


lib/everest/iso15118/cmake/local-build.cmake
SebaLukas, line 1: These changes should be in another PR

lib/everest/iso15118/src/iso15118/tbd_controller.cpp
```diff
@@ -111,6 +111,90 @@ void TbdController::tick() {
...
+    shutdown_active.store(false);
+    shutdown_signaled = false;
+
+    next_event = get_current_time_point();
+
+    while (session or not shutdown_active.load()) {
+        if (not poll_once()) {
+            session->close();
+            session.reset();
+            break;
```
mlitre, line 134: Rare edge case, but we could be idle so session is null and shutdown_active is false so we would still enter here and call a nullptr. 

```suggestion
        if (not poll_once() and session) {
            session->close();
            session.reset();
            break;
```

lib/everest/iso15118/src/iso15118/tbd_controller.cpp
```diff
@@ -111,6 +111,90 @@ void TbdController::tick() {
...
+        }
+
+        tick();
+    }
+    logf_info("Exiting TbdController loop gracefully");
+}
+
+// The caller of this function needs to provide a non-blocking, keepalive fd.
+// ConnectionPlain::read() depends on the socket being non-blocking
+bool TbdController::start_session(int connected_fd) {
```
mlitre: The boolean return can be a bit ambiguous because it one case we consume and close the fd if we start the session or if the session didn't start we don't do anything with it. 
We should probably return an enum so that whoever calls this method knows whether they should close or not the fd.

lib/everest/iso15118/src/iso15118/tbd_controller.cpp
```diff
@@ -111,6 +111,90 @@ void TbdController::tick() {
...
+    next_event = get_current_time_point();
+
+    while (session) {
+        if (not poll_once()) {
+            session->close();
+            session.reset();
+            return false;
+        }
+
+        service_active_session();
```
mlitre, line 185: This can block indefinitely as we never set up the  communication_setup_timeout in this path so if the EV never starts the communication we will be stuck here

lib/everest/iso15118/src/iso15118/tbd_controller.cpp
```diff
@@ -111,6 +111,90 @@ void TbdController::tick() {
...
+    if (not fd_is_open(connected_fd)) {
+        logf_error("Passed fd is not open and valid, returning...");
+        return false;
+    }
+
+    if (config.enable_sdp_server) {
+        logf_error("SDP server is active; incompatible with externally-provided-fd mode. Not starting a session and "
+                   "keeping fd open");
+        return false;
+    }
```
mlitre, line 160: We should probably reset `terminate_session_requested`, since this could have been set by setting dlink ready to false and it could cancel the first setup on a new call to start_session.

lib/everest/iso15118/src/iso15118/tbd_controller.cpp
SebaLukas: Good catch :) But I would suggest this:
```cpp
if (not poll_once()) {
    if (session) {
            session->close();
            session.reset();  
    }
    break;
}
```
In the case `poll()` return false but there is no session active.

lib/everest/iso15118/src/iso15118/tbd_controller.cpp
SebaLukas: Yeah I know. Then I will define an enum for that :+1: I thought a bool is enough :sweat_smile:

lib/everest/iso15118/src/iso15118/tbd_controller.cpp
SebaLukas: I see the problem but it should be the responsibility from the user to provide an already connected, open fd when using `start_session()`. Its the same case in the `handle_sdp_server_input()`, libiso resets `communication_setup_timeout`, sends the sdp res and waits for the tcp handshake.
Maybe we need another way to check if the ev starts the communication.

lib/everest/iso15118/src/iso15118/tbd_controller.cpp
SebaLukas: Good point :+1:

lib/everest/iso15118/src/iso15118/tbd_controller.cpp
mlitre: Looks good to me!

lib/everest/iso15118/src/iso15118/tbd_controller.cpp
mlitre: Fair enough

lib/everest/iso15118/include/iso15118/tbd_controller.hpp
```diff
@@ -25,6 +25,13 @@
 
 namespace iso15118 {
 
+enum class StartSessionResult : std::uint8_t {
+    KeepFdOpen,
+    FdNotValid,
+    CloseFd,
```
mlitre: It is a little ambiguous in the wording it sounds like the user should close the fd, but it should actually be FdClosed to indicated that it was closed. Or maybe I misunderstood

SebaLukas: You are right, I will change it to `FdClosed` :+1:

SebaLukas: Changed

mlitre (approved):
Looking good!

PR #2525, opened 2026-07-28
feat(libiso15118): Now it is possible to pass along the vehicle cert hash while using an already open socket

SebaLukas:
## Describe your changes
See title

Before this PR, the vehicle cert hash could not be passed into the v2g session for potential pause/resume.

### DIFF
--- a/lib/everest/iso15118/include/iso15118/io/connection_plain.hpp
+++ b/lib/everest/iso15118/include/iso15118/io/connection_plain.hpp
@@ -12,7 +12,7 @@ namespace iso15118::io {
 class ConnectionPlain : public IConnection {
 public:
     ConnectionPlain(PollManager&, const std::string& interface_name);
-    ConnectionPlain(PollManager&, int connected_fd);
+    ConnectionPlain(PollManager&, int connected_fd, const std::optional<sha512_hash_t>& vehicle_cert_hash);
 
     void set_event_callback(const ConnectionEventCallback&) final;
     Ipv6EndPoint get_public_endpoint() const final;
@@ -22,13 +22,15 @@ public:
 
     void close() final;
 
-    std::optional<sha512_hash_t> get_vehicle_cert_hash() const final {
-        return std::nullopt;
+    [[nodiscard]] std::optional<sha512_hash_t> get_vehicle_cert_hash() const final {
+        return vehicle_cert_hash;
     }
 
     ~ConnectionPlain();
 
 private:
+    ConnectionPlain(PollManager&, int connected_fd);
+
     PollManager& poll_manager;
 
     Ipv6EndPoint end_point;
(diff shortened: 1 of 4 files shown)


mlitre (changes requested):
Everything looks good but your tests will break since the definitions changed and we don't have a default arg for the function calls.

SebaLukas:
> Everything looks good but your tests will break since the definitions changed and we don't have a default arg for the function calls.

I will change that as soon as #2524 is merged

lib/everest/iso15118/src/iso15118/io/connection_plain.cpp
```diff
@@ -65,7 +65,11 @@ ConnectionPlain::ConnectionPlain(PollManager& poll_manager_, const std::string&
 }
 
 ConnectionPlain::ConnectionPlain(PollManager& poll_manager_, int connected_fd) :
-    poll_manager(poll_manager_), fd(connected_fd) {
+    ConnectionPlain(poll_manager_, connected_fd, std::nullopt){};
```
mlitre:
```suggestion
    ConnectionPlain(poll_manager_, connected_fd, std::nullopt){}
```

lib/everest/iso15118/include/iso15118/io/connection_plain.hpp
```diff
@@ -13,6 +13,7 @@ class ConnectionPlain : public IConnection {
 public:
     ConnectionPlain(PollManager&, const std::string& interface_name);
     ConnectionPlain(PollManager&, int connected_fd);
+    ConnectionPlain(PollManager&, int connected_fd, const std::optional<sha512_hash_t>&);
```
mlitre: The name of the std:optional hash disappeared

lib/everest/iso15118/include/iso15118/io/connection_plain.hpp
```diff
@@ -13,6 +13,7 @@ class ConnectionPlain : public IConnection {
 public:
     ConnectionPlain(PollManager&, const std::string& interface_name);
     ConnectionPlain(PollManager&, int connected_fd);
```
mlitre: I am wondering if we should make this private then?

SebaLukas: Yeah why not 👍

lib/everest/iso15118/include/iso15118/io/connection_plain.hpp
SebaLukas: I will add the name because connected_fd is also there.

lib/everest/iso15118/src/iso15118/io/connection_plain.cpp
SebaLukas: 👍

mlitre (approved):

Issue #596, opened 2024-03-22
EvseV2G can not encode ServiceDiscoveryRes if both payment options are disabled


SebaLukas:
### Describe the bug

If in the config both payment options are disabled, then the EvseV2G module can not encode the ServiceDiscoveryRes message. Because no payment is provided.

### EVerest Domain

ISO15118

### Affected EVerest Module

EvseV2G

### To Reproduce

```
evse_manager:
    module: EvseManager
    config_module:
      payment_enable_contract: false
      payment_enable_eim: false
```

PR #2527, opened 2026-07-28
feat(EvseV2G): make the charge service FreeService flag configurable

biubiuzhang:
## Describe your changes

`ServiceDiscoveryRes` always advertises `FreeService=false` for the charge service, regardless of whether the EVSE charges for free. The flag is hardcoded in `v2g_ctx_init_charging_values()`:

```cpp
ctx->evse_v2g_data.charge_service.FreeService = 0;
```

and nothing ever assigns it again. There is no manifest config key and no interface variable for it, so nothing can drive it. `din_server.cpp` then copies that field straight into the response.

The charge service is the only service left out:

| Service | FreeService today |
|---|---|
| Charge service | hardcoded `0`, not settable |
| SAE service | `handle_setup()` sets `1` |
| Contract certificate | set to `1` |
| VAS | from the interface, `service.free_service.value_or(true)` |

That asymmetry is what suggests an oversight rather than a deliberate choice. There is no TODO or comment marking it as unimplemented.

This adds a `free_service` config key and applies it in `EvseV2G::init()` right after `v2g_ctx_create()`. By that point `v2g_ctx_create()` has run the one-shot init via `v2g_ctx_init_charging_session()`, so `initialize_once` is already true and the guarded block does not reset the value on later sessions — which is the behaviour the existing `v2g_ctx_test.cpp` `v2g_ctx_init_charging_values` test already documents:

```cpp
ctx.evse_v2g_data.charge_service.FreeService = 9;
v2g_ctx_init_charging_values(&ctx);
EXPECT_EQ(ctx.evse_v2g_data.charge_service.FreeService, 0);   // first call resets
ctx.evse_v2g_data.charge_service.FreeService = 10;
v2g_ctx_init_charging_values(&ctx);
EXPECT_EQ(ctx.evse_v2g_data.charge_service.FreeService, 10);  // later calls preserve
```

One assignment covers both protocols: `din_server.cpp` copies the field explicitly into `ServiceDiscoveryRes` and `iso_server.cpp` assigns the whole `charge_service` struct, so both pick the value up from the shared context.

The default is `false`, so behaviour is unchanged for deployments that require authorization. Defaulting to `true` would have been a worse regression than the bug.

### How it was found

Observed on a production DC charger running in free vend mode, with a Tesla over DIN 70121. From the field capture:

```
ServiceDiscoveryRes
  ResponseCode: OK
  PaymentOptions
    PaymentOption: ExternalPayment
  ChargeService
    ServiceTag
      ServiceID: 1
      ServiceCategory: EVCharging
    FreeService: false        <-- EVSE was charging for free
    EnergyTransferType: DC_extended
```

Because the value is a compile-time constant, every session on every unit advertises `FreeService: false`.

### A question on the approach

We went with a static manifest config key because it is the smallest change and matches how `supported_DIN70121` / `supported_ISO15118_2` already work. But there is a second option we would be equally happy to implement if you prefer it:

**Expose it through the `ISO15118_charger` interface** instead, the way VAS services already do it (`ISO15118_chargerImpl.cpp` reads `service.free_service`). That would let the flag follow the authorization mode at runtime rather than being fixed at startup, which matters if free vend can be toggled from the CSMS without restarting EVerest. It is a larger change: an interface variable plus a setter on the charge service, and it would make the DIN and ISO-2 response builders read live state.

Trade-off as we see it:

- **Config key (this PR):** 12 lines, no interface change, cannot follow a runtime auth-mode change.
- **Interface variable:** consistent with VAS, follows auth mode at runtime, larger surface and an interface addition.

Happy to rework this as the interface version, or to add it on top so both exist, whichever you would rather carry. Please just say which and we will follow up.

No upstream issue — we searched the tracker and found nothing covering the charge service `FreeService` flag (the nearby ServiceDiscovery items #596 and #1802 concern payment-option encoding and the PnC check). Happy to open one first if you would prefer that order.

- [x] I have made corresponding changes to the documentation — the new key is documented in the module manifest, which is the module's config documentation

### Testing

- `EvseV2G` builds clean; `v2g_ctx_test` passes (5/5).
- Verified end to end on our hardware before porting to `main`: with `free_service` toggled, an ISO 15118-2 session was decoded on the EV side and reported `FreeService` following the config value. The flag is not readable from the EVerest session log because the native cbv2g codec stores raw EXI, so it was read after decode in the EVCC.
- No new unit test: the change is config plumbing applied at module init, and the `initialize_once` behaviour it depends on is already covered by the existing `v2g_ctx_test.cpp` case quoted above. Glad to add one if you can point at the shape you would want.

### DIFF
--- a/interfaces/ISO15118_charger.yaml
+++ b/interfaces/ISO15118_charger.yaml
@@ -48,6 +48,17 @@ cmds:
           Indicates if the vehicle contract may be forwarded to and validated by a CSMS in case
           local validation was not successful
         type: boolean
+  set_free_service:
+    description: >-
+      Set whether the EVCharging service is offered free of charge (the FreeService flag in
+      ServiceDiscoveryRes). The module defaults to false, so only a call with true marks the service
+      as free. Drive this from the source that knows the tariff or free vend state, e.g. the CSMS.
+      The value applies to the next ServiceDiscoveryRes, which is sent before authorization takes
+      place, so call this whenever that state changes rather than at session start.
+    arguments:
+      free_service:
+        description: Set to true when the EVCharging service is offered free of charge
+        type: boolean
   bpt_setup:
     description: >-
       At startup, set the bpt setup for ISO15118-20 at least once. May be updated later on.
--- a/docs/source/reference/EVerest_API/evse_manager_consumer_API.yaml
+++ b/docs/source/reference/EVerest_API/evse_manager_consumer_API.yaml
@@ -406,6 +406,13 @@ channels:
       receive_reply_get_random_delay_countdown:
         $ref: '#/components/messages/receive_reply_get_random_delay_countdown'
 
+  ########## ISO15118_charger
+  send_set_free_service:
+    address: 'm2e/set_free_service'
+    messages:
+      send_set_free_service:
+        $ref: '#/components/messages/send_set_free_service'
+
 
   receive_heartbeat:
     address: 'e2m/heartbeat'
@@ -1008,6 +1015,14 @@ operations:
     action: receive
     channel:
       $ref: '#/channels/receive_reply_get_random_delay_countdown'
+
+  ########## ISO15118_charger
+  send_set_free_service:
+    title: 'Set whether the EVCharging service is offered free of charge'
+    action: send
+    channel:
+      $ref: '#/channels/send_set_free_service'
+    description: 'Direction: Module to EVerest'
 
 
   receive_heartbeat:
@@ -1962,6 +1977,20 @@ components:
           payload:
             42
 
+    send_set_free_service:
+      name: send_set_free_service
+      title: 'Send set free service'
+      summary: >-
+        Set whether the EVCharging service is offered free of charge (the FreeService flag in
+        ServiceDiscoveryRes). The module defaults to false, so only a call with true marks the
+        service as free.
+      payload:
+        $ref: '#/components/schemas/FreeService'
+      examples:
+        - summary: ""
+          payload:
+            true
+
     receive_random_delay_countdown:
       name: receive_random_delay_countdown
       title: 'Countdown of the currently running random delay'
@@ -2475,6 +2504,9 @@ components:
           maximum: 100
     ForceUnlockReply:
       description: "Returns true if unlocking sequence was successfully executed"
+      type: boolean
+    FreeService:
+      description: "Set to true when the EVCharging service is offered free of charge"
       type: boolean
     ForceUnlockRequest:
       description: "Specifies the ID of the connector that should be unlocked"
--- a/modules/API/EVerestAPI/evse_manager_consumer_API/evse_manager_consumer_API.cpp
+++ b/modules/API/EVerestAPI/evse_manager_consumer_API/evse_manager_consumer_API.cpp
@@ -99,6 +99,7 @@ void evse_manager_consumer_API::ready() {
     generate_api_cmd_random_delay_disable();
     generate_api_cmd_random_delay_cancel();
     generate_api_cmd_random_delay_set_duration_s();
+    generate_api_cmd_set_free_service();
 
     helper.generate_api_var_communication_check(&comm_check);
     comm_check.start(config.cfg_communication_check_to_s);
@@ -240,6 +241,19 @@ void evse_manager_consumer_API::generate_api_cmd_random_delay_set_duration_s() {
             }
             return false;
         });
+}
+
+void evse_manager_consumer_API::generate_api_cmd_set_free_service() {
+    if (not r_hlc.empty()) {
+        helper.subscribe_api_topic("set_free_service", [this](std::string const& data) {
+            bool free_service;
+            if (deserialize(data, free_service)) {
+                r_hlc[0]->call_set_free_service(free_service);
+                return true;
+            }
+            return false;
+        });
+    }
 }
 
 void evse_manager_consumer_API::generate_api_var_session_event() {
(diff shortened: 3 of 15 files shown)


SebaLukas:
Hi :)
I would suggest to go straight the interface way.
In my opinion this config option would cause more confusion. For example, in a scenario where the config option says EvCharging is free, but the CSMS or the Auth module says it isn't free. This could lead to unnecessary debug sessions :sweat_smile: 
Therefore, the information about whether it's free or not should come from an external source.
I would go so far as to set the default value in the EvseV2g module for freeService to false (EvCharging service) and only set the value to true when a command is called that sets the value to true. Then Auth/CSMS and the EvCharging service is aligned

biubiuzhang:
Thanks, that argument makes sense — a config key that can silently disagree with Auth/CSMS is exactly the kind of thing that costs an afternoon of debugging. Reworked and force-pushed as a single commit.

**What changed**

- Dropped the `free_service` config key from the `EvseV2G` manifest and the assignment in `EvseV2G::init()`.
- New command `set_free_service(free_service: boolean)` on the `ISO15118_charger` interface.
- `EvseV2G` keeps `FreeService = 0` from `v2g_ctx_init_charging_values()` as its default, and only a call with `true` marks the charge service as free. `din_server.cpp` and `iso_server.cpp` already read the shared context, so DIN 70121 and ISO 15118-2 both follow from that one assignment.
- `IsoMux` forwards to both backends, `DummyV2G` gets the generated stub.
- Added a `din_server_test` case covering both the default and the free case in `ServiceDiscoveryRes`.

**One behavioural note.** The assignment lands outside the `initialize_once` guard, so the value persists across sessions. That is deliberate: `ServiceDiscoveryRes` goes out before authorization happens, so the flag cannot follow a per-session authorization result — it can only reflect free vend / tariff state that is known beforehand. Callers therefore set it when that state changes, not at session start. I documented that in the command description; happy to reset it per session instead if you would rather it be explicit every time.

**ISO 15118-20.** `Evse15118D20` accepts the command but logs a warning and ignores it, because `libiso15118` hardcodes `free_service` to `false` for the energy transfer service list entries in `d20/state/service_discovery.cpp` (and the default in `ServiceDiscoveryResponse`). I can open a small PR there to carry the flag in the session config so `Evse15118D20` can honour it too — just say whether you want that before or after this one.

**On the caller.** I have deliberately left this PR without a caller in everest-core, like `update_supported_app_protocols`. There is no free vend or tariff concept in core today: OCPP 1.6 has no standard field for it at all, and in 2.0.1 the closest thing, `AuthCtrlr.Enabled`, means "no authorization required" rather than "free of charge" and never leaves libocpp. In our own deployment the value comes from the charger's authorization mode, which the CSMS sets over `DataTransfer` — free vend mode calls `set_free_service(true)` at startup, everything else leaves the default. If you would rather `EvseManager` drive it, that needs a source of truth to be surfaced first, and I would be glad to do that as a follow-up once you have a preference on where it should come from.

interfaces/ISO15118_charger.yaml
```diff
@@ -48,6 +48,17 @@ cmds:
           Indicates if the vehicle contract may be forwarded to and validated by a CSMS in case
           local validation was not successful
         type: boolean
+  set_free_service:
```
SebaLukas, line 51: Should be also added in the everest_api

SebaLukas (changes requested):

FaHaGit:
What I'm curious about: Does the free_service parameter affect the Tesla implementation? Or was it just noticed through analysis of the traces?

The question is also: What exactly does “free_service” mean? Does “free” refer to being free of authorization? Or could it also mean that authorization is required, but charging is free?

As a quick solution, I would first pass the “disable_authentication” config key from the EVSEManager to the V2G modules.

biubiuzhang:
Good questions. This was found by analyzing pcap captures from a DIN 70121 session with a Tesla on a DC charger in free vend mode — the flag was wrong on the wire, but from what I remember there was no visible effect on the vehicle. We no longer have the rental car, so I can't re-check its display behavior.

On the semantics: per ISO 15118-2 (Table 105, ServiceType), FreeService means the service is offered *free of charge* — it's about payment, not authorization. "Authorization required but charging is free" is a valid combination, e.g. a fleet site that requires an RFID card but doesn't bill.

That's also why I'd avoid deriving it from `disable_authentication`: that flag describes the authorization side. Coupling them would advertise FreeService=false on sites that require auth but don't charge. The command lets whoever knows the tariff/free-vend state (typically the CSMS) drive the flag, and covers the free vend case too — the same logic that disables authentication can call `set_free_service(true)`.

interfaces/ISO15118_charger.yaml
biubiuzhang: Added in the latest push: evse_manager_consumer_API now has an optional ISO15118_charger requirement and forwards a set_free_service topic, plus the AsyncAPI doc entry.

biubiuzhang:
@SebaLukas gentle ping — the `everest_api` part you asked for is in the current head (`a726c85c1`):

- `modules/API/EVerestAPI/evse_manager_consumer_API/manifest.yaml` — optional `ISO15118_charger` requirement
- `modules/API/EVerestAPI/evse_manager_consumer_API/evse_manager_consumer_API.{hpp,cpp}` — `set_free_service` topic
- `docs/source/reference/EVerest_API/evse_manager_consumer_API.yaml` — AsyncAPI doc entry

Also HIL-verified on two chargers on 2026-08-05 against a local CitrineOS: FreeVend advertises `FreeService=1`, RFID/Autocharge advertises `0` on the wire (pcap + libcbv2g decode).

Could you take another look when you get a chance, so the review thread can be resolved? Happy to rebase if `main` has moved under it.

PR #2528, opened 2026-07-29
Fix(manager): Don't guess CLI options

cburandt:
## Describe your changes

The manager executable does fuzzy matching of CLI arguments (implemented by Boost.Program_options).

This PR disables the fuzzyness: forbids boost::po to guess the best match.

To fix this in a non-breaking way, "--conf" was added as an explicit option (in addition to "--config").

### Reason

The CLI option in question is "--config". However, the auto-generated run-scripts and many examples used "--conf".
Introduction of a "--configuration-api" option (to come in a later PR) created a collision: The fuzzy matcher could not longer match "--conf" to "--config" since "--configuration-api" is an equally good match).

### DIFF
--- a/applications/utils/docker/everest-docker-image/Dockerfile
+++ b/applications/utils/docker/everest-docker-image/Dockerfile
@@ -115,4 +115,4 @@ WORKDIR /opt/everest
 COPY --from=builder /opt/everest ./
 COPY ./scripts/initialize.sh /opt/everest
 
-CMD [ "/opt/everest/bin/manager", "--conf", "/opt/everest/config/config.yaml" ]
+CMD [ "/opt/everest/bin/manager", "--config", "/opt/everest/config/config.yaml" ]
--- a/applications/utils/docker/everest-docker-image/run-config-fallback.sh
+++ b/applications/utils/docker/everest-docker-image/run-config-fallback.sh
@@ -10,7 +10,7 @@ usage() {
 
 export EVEREST_CONFIG=config-fallback.yaml
 export OCPP_CONFIG=ocpp-config.json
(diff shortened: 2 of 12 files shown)


Pietfried (commented):
nit: add a warning if --conf is used.

There is one more occurences of --conf e.g. in applications/utils/docker/everest-docker-image/run-config-fallback.sh (only L13)

Pietfried (dismissed):

Pietfried (approved):

hikinggrass:
I think that qualifies as a "breaking" change, even though the label doesn't quite fix.

Pietfried:
> I think that qualifies as a "breaking" change, even though the label doesn't quite fix.

Isnt it completely backwards compatible?

cburandt:
> > I think that qualifies as a "breaking" change, even though the label doesn't quite fix.
> 
> Isnt it completely backwards compatible?

It's not completely backwards compatible: ALL cli options were "guessed" if no exact match existed. Therefore we break all setups with typos or not fully spelled-out cli options.

hikinggrass:
> > I think that qualifies as a "breaking" change, even though the label doesn't quite fix.
> 
> Isnt it completely backwards compatible?

I should be fine since this is clearly not intended functionality, and not documented. But it could unexpectedly break other (arguably wrong) run configurations not tracked in this repository

Pietfried:
> > > I think that qualifies as a "breaking" change, even though the label doesn't quite fix.
> > 
> > 
> > Isnt it completely backwards compatible?
> 
> I should be fine since this is clearly not intended functionality, and not documented. But it could unexpectedly break other (arguably wrong) run configurations not tracked in this repository

Ah, apart from the fallback handling for `--conf`, that's true. Let's put the label and highlight in the upcoming release notes.

hikinggrass (approved):

PR #2531, opened 2026-07-29
Feature(framework): Allow to set an MQTT LWT

cburandt:
## Describe your changes

This PR allows to set a last-will-and-testament (LWT) via the mqtt_abstraction. This feature is not available from the modules but only from the manager.
In MQTT LWT works such, that for one connection only a single LWT can be defined. The implementation will therefore accept setting the LWT once and fail any further attempts. Setting the LWT must happen before the connection is set up.


Pietfried (approved):
lgtm

hikinggrass (approved):

PR #2533, opened 2026-07-30
feat(YetiSimulator): make simulated AC nominal frequency configurable

hirokiri:
### Motivation

Follow-up to #2523, as suggested by @SebaLukas in [this comment](https://github.com/EVerest/EVerest/pull/2523#issuecomment-5117484919): the YetiSimulator hardcodes the simulated AC grid frequency to 50 Hz (struct defaults in `util/state.hpp`), while several regions of the world run 60 Hz grids (most of the Americas, parts of Asia).

Japan — the motivating region for #2523 — is actually the strongest example: it is one of the very few countries with **both** nominal frequencies in domestic service (50 Hz in East Japan, 60 Hz in West Japan). A SIL setup modeling Japanese charging stations therefore needs this to be configurable even within a single country.

### What changed

Mirrors the approach of #2523 (`ac_nominal_voltage`):

- `modules/Simulation/YetiSimulator/manifest.yaml`: new config option `ac_nominal_frequency` (type `number`, default `50.0`, minimum `50.0`, maximum `60.0`). All public grids worldwide are nominally 50 or 60 Hz, hence the bounds; frequency excursions around the nominal value are already simulated at runtime by `add_noise()` and are not constrained by them.
- `modules/Simulation/YetiSimulator/YetiSimulator.hpp`:
  - `Conf` struct extended with `double ac_nominal_frequency;` (matching the generated code from `ev-cli` for the updated manifest).
  - `reset_module_state()` now applies the configured value to `simdata_setting.frequencies`, `simulation_data.frequencies` and `powermeter_data.freqL1/L2/L3`, so the configured frequency is reported both while the simulation loop is stepping and while it is idle, and survives the runtime module state reset triggered via `ev_board_supportImpl::handle_enable()` — same reasoning as in #2523.

### Depends on #2523

This branch is stacked on `feature/yetisim-configurable-nominal-voltage`, since both PRs touch the same config block and `reset_module_state()`. Until #2523 merges, the diff here also shows its commits — only the last commit is new. I will rebase onto `main` as soon as #2523 lands.

### Backward compatibility

The default remains 50 Hz and the struct defaults in `util/state.hpp` are unchanged, so existing configurations behave exactly as before.

None — follow-up to review discussion on #2523.

- [x] I have made corresponding changes to the documentation (config option description in the manifest; no further docs reference the simulated frequency)

### DIFF
--- a/modules/Simulation/YetiSimulator/YetiSimulator.hpp
+++ b/modules/Simulation/YetiSimulator/YetiSimulator.hpp
@@ -32,6 +32,7 @@ struct Conf {
     std::string dummy_meter_value_blob_start;
     std::string dummy_meter_value_blob_stop;
     double ac_nominal_voltage;
+    double ac_nominal_frequency;
 };
 
 class YetiSimulator : public Everest::ModuleBase {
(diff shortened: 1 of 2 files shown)


SebaLukas (dismissed):

SebaLukas:
There are two conflicts that are need to be resolved before merging

hirokiri:
@SebaLukas Conflicts are resolved — merged `main` into the branch (the conflicts were with the `ac_nominal_voltage` change from #2523 that just landed). Should be mergeable now, thanks!

SebaLukas (approved):
👍 
Before I merge your PR, can you please squash your commits in one commit

hirokiri:
Done — squashed into a single commit (acc0cbb). The branch previously also carried the nominal-voltage commit that was already merged via #2523, so after the squash the PR is only the frequency change. Thanks for the review!

PR #2538, opened 2026-07-30
RsIskraMeter: Change implementation_id to match other meters

dorezyuk:

### DIFF
--- a/docs/source/project/releases/deprecation-index.rst
+++ b/docs/source/project/releases/deprecation-index.rst
@@ -30,3 +30,7 @@ to the corresponding migration guide.
      - 2026.10.0
      - 2027.04.0
      - :ref:`Migrate to the Combined OCPP Module <howto-ocpp-storage-migration>`
+   * - The `RsIskraMeter` deprecates its `meter` implementation_id in favor of `main`.
+     - 2026.10.0
+     - 2027.04.0
+     - Change `implementation_id: meter` to `implementation_id: main` in the config connection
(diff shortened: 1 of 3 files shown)


hikinggrass (dismissed):
Let's track this as a "breaking change" so we can highlight it in the release notes later on

Pietfried:
> Let's track this as a "breaking change" so we can highlight it in the release notes later on

Is this really a required change with this consequence?

dorezyuk:
> Is this really a required change with this consequence?

@Pietfried is there a way we could avoid it? I still would like that the impl id follows the other meters so we can change it easier

Pietfried:
> > Is this really a required change with this consequence?
> 
> @Pietfried is there a way we could avoid it? I still would like that the impl id follows the other meters so we can change it easier

There is currently no auto-migration option for these kind of changes.

dorezyuk:
@Pietfried @hikinggrass maybe this is a better approach - we deprecate the current `meter` interface and add `main` - this should be more inline with our guidelines

Pietfried (changes requested):
The solution looks good avoiding the breaking change.

Please add an entry in the deprecation index for this: https://github.com/EVerest/EVerest/blob/main/docs/source/project/releases/deprecation-index.rst and recommend a migration (e.g "change implementation_id: meter to implementation_id: main in the config connection").

dorezyuk:
done

Pietfried (approved):

PR #2539, opened 2026-07-30
feat(session_storage_consumer_API): add external API module for PersistentSessionStorage

nzbr:
## Describe your changes

Implement the AsyncAPI definition in session_storage_consumer_API.yaml: a new session_storage domain in everest_api_types and the session_storage_consumer_API module bridging the session_storage interface to external MQTT with the three request/reply commands get_sessions, get_session and clear_sessions.

Smoke tests drive the bridge over external MQTT against a probe module providing the session_storage interface.


docs/source/reference/EVerest_API/transaction_storage_consumer_API.yaml
```diff
@@ -0,0 +1,453 @@
...
+        $ref: '#/components/messages/receive_heartbeat'
+  send_communication_check:
+    address: 'm2e/communication_check'
+    messages:
+      send_communication_check:
+        $ref: '#/components/messages/send_communication_check'
+
+
+operations:
+  send_request_get_transactions:
```
marcemmers, line 75: Would having some kind of a filter on the `get_transactions` make sense? Maybe at least an optional to and from date?

Otherwise if chargers are in use for years this reply will potentially get thousands of transactions in the payload.

nzbr: Yes, definitely, but I'd wait to implement that in a separate PR. Adding optional parameters isn't a breaking change after all and we'd need to figure out what kind of filters make sense first

marcemmers: Ok that makes sense, I was wondering if that should be a follow up anyway.

modules/API/EVerestAPI/session_storage_consumer_API/docs/index.rst
```diff
@@ -0,0 +1,44 @@
...
+iteration is complete. An invalid or outdated token starts the iteration from
+the beginning. The payload is optional: an empty object requests the first page
+with the default page size.
+
+Deleting records
+================
+
+``clear_sessions`` cannot be undone. It also deletes records that are still
+open, and when such a session finishes later the finish is dropped and the
+record does not reappear.
```
cburandt, line 38: Maybe add a word concerning concurrently accessing client: All access is stateless and clear_sessions could always be called by client A while client B is getting sessions.

modules/API/EVerestAPI/session_storage_consumer_API/docs/index.rst
```diff
@@ -0,0 +1,44 @@
...
+
+Pagination
+==========
+
+``get_sessions`` returns records oldest first. A reply carries a
+``continuation_token`` while more matching records may exist; pass it in the
+next request to continue the iteration. A reply without a token means the
+iteration is complete. An invalid or outdated token starts the iteration from
+the beginning. The payload is optional: an empty object requests the first page
+with the default page size.
```
cburandt, line 31: If I read the test case from the PR this one is based upon correctly, the last page requested using some continuation token could be empty. Maybe worth pointing out here.

modules/API/EVerestAPI/session_storage_consumer_API/docs/index.rst
cburandt, line 1: Placing these hints here, will probably not help the users. They will usually use the rendered html docs and navigate to "Reference->EVerest API Reference" and select the session_storage_consumer_API there. They will find no link to the handwritten module documentation there.

Maybe this can be included into the AsyncAPI yaml doc?

cburandt (dismissed):
LGTM, however I would not place documentation in the module docs, as it's unlikely that API users will find it there.

modules/API/EVerestAPI/session_storage_consumer_API/session_storage_consumer_API.cpp
```diff
@@ -0,0 +1,103 @@
...
+        }
+        return false;
+    });
+}
+
+void session_storage_consumer_API::generate_api_cmd_clear_sessions() {
+    helper.subscribe_api_topic("clear_sessions", [this](std::string const& data) {
+        std::string reply_to;
+        if (deserialize_request(data, reply_to)) {
+            auto reply = API_types_ext::to_external_api(r_session_storage->call_clear_sessions());
```
cburandt, line 95: This needs an argument (call_clear_sessions(...))

cburandt (changes requested):

cburandt (dismissed):
LGTM

(one could also delete the handwritten part from the module's doc completely, but as long as everything is described in the yaml, it's fine for me)

cburandt (approved):
LGTM

hikinggrass (approved):

PR #2541, opened 2026-07-30
Rust: fix error serialization/deserialization

IvarWithoutBones:
## Describe your changes

The Rust bindings maps each error type in an interface to a strongly-typed enum, which is converted to/from when publishing/receiving an error. This mapping had a small bug: the error type was always converted to snake_case, even though not all interfaces adhere to that convention.

For example:
`power_supply_DC/HardwareFault` was mapped to `power_supply_dc/HardwareFault` (lowercase DC), which causes deserialization to fail and serialization to yield unknown error types.

This commit fixes that by removing the snake_case conversion. Now the error type always follows whichever formatting style the interface chooses.


dorezyuk:
@IvarWithoutBones can you extend the https://github.com/EVerest/EVerest/tree/main/lib/everest/framework/everestrs/tests/modules/RsErrorsCompilationChecks test? I think it would be nice to define a new dummy error ala error_CAP.yaml or similar and add a test case in the main.rs - let me know if you need help with that

IvarWithoutBones:
Done :+1:

lib/everest/framework/everestrs/tests/errors/errors_CAPITALIZED.yaml
dorezyuk, line 1: just a nit - but can we name the errors and the interface the same - one is CAPITALIZED_errors and the other errors_CAPITALIZED. Or is that on purpose?

IvarWithoutBones: I tried to follow what the other definitions use, e.g. `errors_multiple.yaml` and `example_errors.yaml`. Should I still rename it?

dorezyuk: but this file is named CAPITALIZED_errors - and every other file is named erros_<...>.yaml - maybe I dont get what yo try to say , but I would rename it to errors_CAPITALIZED

IvarWithoutBones: I meant that all the other error types are named that way, like `errors/example_errors.yaml` and `errors/more_errors.yaml`. Whereas the interfaces are all named like `interfaces/errors_duplicate` and `interfaces/errors_multiple.yaml`.

Doesn't really matter though, I've renamed it like you suggested.

dorezyuk (approved):

hikinggrass (approved):

PR #2543, opened 2026-07-31
chore(sqlite): Add per-transaction fkey modes and harden transactions

cburandt:
## Describe your changes

Refactor sqlite lib.

DatabaseTransaction gains explicit foreign-key handling:
- begin_transaction() inherits the connection's foreign_keys state and never touches the pragma (unchanged behavior for existing consumers such as libocpp)
- begin_transaction_with_enforced_fkeys() / _deferred_fkeys() enforce constraints for the duration of the transaction (deferred: checked at commit time)
- new begin_transaction_with_disabled_fkeys() disables enforcement, for schema migrations that drop and recreate referenced tables

This refactor is preparing the bigger internal ConfigService changes.

### DIFF
--- a/lib/everest/sqlite/include/everest/database/sqlite/connection.hpp
+++ b/lib/everest/sqlite/include/everest/database/sqlite/connection.hpp
@@ -48,6 +48,8 @@ public:
     virtual bool close_connection() = 0;
 
     /// \brief Start a transaction on the database. Returns an object holding the transaction.
+    ///        The connection's current foreign-key enforcement state is inherited and left
+    ///        untouched.
     /// \note This function can block until the previous transaction is finished.
     [[nodiscard]] virtual std::unique_ptr<TransactionInterface> begin_transaction() = 0;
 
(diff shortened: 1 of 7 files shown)


lib/everest/sqlite/include/everest/database/sqlite/connection.hpp
```diff
@@ -48,9 +48,20 @@ class ConnectionInterface {
...
 
+    /// \brief Same as begin_transaction() but foreign_keys constraints are enforced for the
+    ///        duration of the transaction; the connection's previous foreign-key state is
+    ///        restored when the transaction ends
+    [[nodiscard]] virtual std::unique_ptr<TransactionInterface> begin_transaction_with_enforced_fkeys() = 0;
+
+    /// \brief Same as begin_transaction_with_enforced_fkeys() but constraints are only checked
+    ///        at commit time (deferred)
+    [[nodiscard]] virtual std::unique_ptr<TransactionInterface> begin_transaction_with_deferred_fkeys() = 0;
+
```
Pietfried: According to the comment in L87/L88 you probably also want to move these functions down

Pietfried (approved):

lib/everest/sqlite/include/everest/database/sqlite/connection.hpp
```diff
@@ -92,12 +109,18 @@ class Connection : public ConnectionInterface {
     bool close_connection() override;
 
     [[nodiscard]] std::unique_ptr<TransactionInterface> begin_transaction() override;
```
marcemmers, line 111: Would using just one function with `FkMode` as argument not be simpler? You could use a default argument to not change existing behaviour.

PR #2546, opened 2026-07-31
Feature/management api

cburandt:
## Describe your changes
Add management APIs (configuration/lifecycle) to the manager process.


lib/everest/framework/src/manager.cpp
```diff
@@ -971,6 +1078,117 @@ int Manager::run() {
...
+        }
+    } lifecycle_wakeup_fd_guard{m_lifecycle_wakeup_fd};
+
+    bool lfc_api_rw = false;
+    std::unique_ptr<Everest::api::lifecycle::LifecycleAPI> lifecycle_api;
+    if (lfc_api_active) {
+        const auto lfc_api_value = m_vm["lifecycle-api"].as<std::string>();
+        if (lfc_api_value != "ro" && lfc_api_value != "rw") {
+            EVLOG_error << "Invalid value for --lifecycle-api: '" << lfc_api_value << "'; expected 'ro' or 'rw'.";
+            return EXIT_FAILURE;
```
Pietfried: In this case `ConfigurationAPI` is already constructed but the `ApiTeardownGuard` is not. This can lead to the destruction of the API while MQTT worker threads can still dispatch requests into it, causing a use-after-free . You can validate both cli options before constructing the APIs.

lib/everest/framework/src/manager.cpp
Pietfried: This can leave a retained "running" status forever. ~MQTTAbstractionImpl does a clean mosquitto_disconnect which discards the last will. You could route those error exits through the same shutdown-status publish as the normal path

lib/everest/framework/src/management_api/configuration_type_wrapper.cpp
```diff
@@ -0,0 +1,398 @@
...
+        result.module_implementation_id = val.implementation_id.value();
+    }
+    return result;
+}
+
+ConfigurationParameterIdentifier_External to_external_api(ConfigurationParameterIdentifier_Internal const& val) {
+    ConfigurationParameterIdentifier_External result;
+    result.module_id = val.module_id;
+    result.parameter_name = val.configuration_parameter_name;
+    result.implementation_id = val.module_implementation_id;
```
Pietfried: This leaks the sentinel `!module` into the API, the we should not set the optional `implementation_id` if `val.has_value() and val.value() == "!module"`

lib/everest/framework/src/management_api/configuration_type_wrapper.cpp
```diff
@@ -0,0 +1,398 @@
...
+    return result;
+}
```
Pietfried, line 209: This could be 

```suggestion
ConfigurationParameterIdentifier_Internal to_internal_api(ConfigurationParameterIdentifier_External const& val) {
    ConfigurationParameterIdentifier_Internal result;
    result.module_id = val.module_id;
    result.configuration_parameter_name = val.parameter_name;
    result.module_implementation_id = val.implementation_id.value_or("!module");
    return result;
}

ConfigurationParameterIdentifier_External to_external_api(ConfigurationParameterIdentifier_Internal const& val) {
    ConfigurationParameterIdentifier_External result;
    result.module_id = val.module_id;
    result.parameter_name = val.configuration_parameter_name;
    if (val.module_implementation_id != "!module") {
        result.implementation_id = val.module_implementation_id;
    }
    return result;
}

```

lib/everest/framework/src/management_api/configuration_type_wrapper.cpp
```diff
@@ -0,0 +1,398 @@
...
+    result.config_access = to_external_api(val.access);
+
+    return result;
+}
+
+GetConfigurationResult_External to_external_api(GetConfigurationResult_Internal const& val) {
+    GetConfigurationResult_External result;
+    result.status = to_external_api(val.status);
+    if (val.module_configurations.size() > 0) {
+        result.module_configurations.emplace();
```
Pietfried, line 378: Shouldnt we emplace on every `Success` ? Otherwise `module_configurations` is omitted.

lib/everest/framework/src/management_api/lifecycle_api.cpp
```diff
@@ -0,0 +1,247 @@
...
+            break;
+        case SrcT::FailedToStart:
+            module_status = TarT::FailedToStart;
+            break;
+        case SrcT::RestartTriggered:
+            module_status = TarT::RestartTriggered;
+            break;
+        }
+        if (update.status == m_last_module_status) {
+            return;
```
Pietfried: This dedupe would swallow a second consecutive `FailedtoStart` , although the comment above says "in particular FailedToStart must reach a client that was told "Starting". 

Reset the dedupe (or publish Starting) when a start/restart request is accepted, so a repeated FailedToStart goes out.

Pietfried (changes requested):
Looks pretty good to me. See some comments in line that need fixing in line.

lib/everest/framework/src/manager.cpp
```diff
@@ -666,6 +682,76 @@ void Manager::handle_initiate_graceful_shutdown(const std::chrono::steady_clock:
...
+    // req == LifecycleApiRequest::Restart
+    if (are_modules_started()) {
+        // Also covers a Restart accepted while Idle ("Starting") that raced with the modules coming up.
+        m_shutdown_cause = ShutdownCause::Restart;
+        m_config_service_core->notice_module_restart_triggered();
+        handle_initiate_graceful_shutdown(std::chrono::steady_clock::now(), false,
+                                          "LifecycleAPI requested restart of the modules.", ctx.mqtt_abstraction,
+                                          ctx.ms);
+    } else if (is_idle()) {
+        // Also fulfils a Restart accepted while Running ("Restarting") that a crash-to-idle recovery
```
Pietfried, line 758: While testing I noticed that is idle restart path skips the empty-config guard. When booting with --lifecycle-api=rw and --into-idle with an empty but valid config an std::logic_error is thrown after calling start_modules.

```
2026-09-01 20:17:45.057714 [ERRO] manager         int main(int, char**) :: Main manager process exits because of caught exception:
handle_start_modules() called with an empty module list
```

lib/everest/framework/src/manager.cpp
cburandt: Fixed.

lib/everest/framework/src/manager.cpp
cburandt: Done. Introduced guards in the correct order to publish a final status message in case of "clean" exits if the lifecycle_API is active and added another guard to shutdown any modules which are still running (which could happen in certain situations).

lib/everest/framework/src/management_api/configuration_type_wrapper.cpp
cburandt: Fixed.

lib/everest/framework/src/management_api/configuration_type_wrapper.cpp
cburandt: Fixed.

lib/everest/framework/src/management_api/configuration_type_wrapper.cpp
cburandt: Agree and fixed.

lib/everest/framework/src/management_api/lifecycle_api.cpp
cburandt: Agree, this doesn't behave correctly, but fixed on another level. ConfigServiceCore now adds the "cause" to status updates, such that client (like LifecycleAPI) can decide to publish an update based on the cause. No need to try to heuristically de-duplicate updates.

lib/everest/framework/src/manager.cpp
cburandt: This path was treated differently than other paths (e.g. boot) which was wrong. Fixed and added a test.

lib/everest/framework/src/management_api/lifecycle_api.cpp
cburandt: This is still WIP - the provided fix is incomplete, as I just discovered.

Pietfried (approved):

PR #2547, opened 2026-07-31
feat(libiso15118): Adding skipping sap negotiation when using an already open fd

SebaLukas:
## Describe your changes
See title.
The application that links against libiso15118 can also start_session directly in session_setup. If another application handles the sdp+tcp/tls+sap handshakes.

Sometimes outside of libiso15118 the caller handles the sap negotiation already. Before the changes that was not possible.

### DIFF
--- a/lib/everest/iso15118/include/iso15118/tbd_controller.hpp
+++ b/lib/everest/iso15118/include/iso15118/tbd_controller.hpp
@@ -32,6 +32,11 @@ enum class StartSessionResult : std::uint8_t {
     SessionComplete,
 };
 
+struct StartSessionOptions {
+    std::optional<io::sha512_hash_t> vehicle_cert_hash{std::nullopt};
+    bool skip_app_protocol_negotiation{false};
+};
+
 struct TbdConfig {
     config::SSLConfig ssl{};
     std::string interface_name;
@@ -57,8 +62,7 @@ public:
     void loop();
     // Mutually exclusive with the other driver; see the class note. Refuses (logs and returns false) if the other
     // driver is already running.
-    StartSessionResult start_session(int connected_fd);
-    StartSessionResult start_session(int connected_fd, const std::optional<io::sha512_hash_t>& vehicle_cert_hash);
+    StartSessionResult start_session(int connected_fd, const StartSessionOptions& start_options);
     void tick();
 
     bool has_active_session() const {
--- a/lib/everest/iso15118/src/iso15118/d20/state/session_setup.cpp
+++ b/lib/everest/iso15118/src/iso15118/d20/state/session_setup.cpp
@@ -91,8 +91,12 @@ Result SessionSetup::feed(Event ev) {
         logf_info("Received session setup with evccid: %s", req->evccid.c_str());
         m_ctx.feedback.evcc_id(req->evccid);
         m_ctx.ev_info.evcc_id = req->evccid;
-        m_ctx.feedback.ev_information(m_ctx.ev_info);
 
+        // Only emit ev_information feedback when the sap negotiation is handled here. When SAP is skipped the
+        // application already has the information.
+        if (not skip_app_protocol_negotiation) {
+            m_ctx.feedback.ev_information(m_ctx.ev_info);
+        }
         bool new_session{false};
 
         const auto vehicle_cert_hash = m_ctx.get_new_vehicle_cert_hash();
--- a/lib/everest/iso15118/test/iso15118/session/tbd_controller_start_session.cpp
+++ b/lib/everest/iso15118/test/iso15118/session/tbd_controller_start_session.cpp
@@ -20,9 +20,11 @@
 #include <catch2/catch_test_macros.hpp>
 
 #include <iso15118/d20/config.hpp>
+#include <iso15118/io/logging.hpp>
 #include <iso15118/io/sdp.hpp>
 #include <iso15118/io/sdp_packet.hpp>
 #include <iso15118/io/stream_view.hpp>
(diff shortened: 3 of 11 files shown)


mlitre (dismissed):

lib/everest/iso15118/src/iso15118/d20/state/session_setup.cpp
```diff
@@ -91,8 +91,12 @@ Result SessionSetup::feed(Event ev) {
         logf_info("Received session setup with evccid: %s", req->evccid.c_str());
         m_ctx.feedback.evcc_id(req->evccid);
         m_ctx.ev_info.evcc_id = req->evccid;
-        m_ctx.feedback.ev_information(m_ctx.ev_info);
 
+        // Only emit ev_information feedback when the sap negotiation is handled here. When SAP is skipped the
+        // application already has the information.
+        if (not skip_app_protocol_negotiation) {
+            m_ctx.feedback.ev_information(m_ctx.ev_info);
+        }
```
mlitre, line 99: Don't we get evcc_id from the session setup so we should actually keep this right?

lib/everest/iso15118/test/iso15118/session/tbd_controller_start_session.cpp
```diff
@@ -191,81 +293,63 @@ SCENARIO("session_start functionality") {
...
 
             // start_session() returned true and reaped the session; ConnectionPlain::close() closed the fd.
-            REQUIRE(ok == iso15118::StartSessionResult::SessionComplete);
+            REQUIRE(result.ok == iso15118::StartSessionResult::SessionComplete);
             REQUIRE_FALSE(controller.has_active_session());
             REQUIRE_FALSE(fd_is_open(fds.at(0)));
         }
     }
+
+    close(fds.at(0));
```
mlitre: Is this required since we already check if the fd is closed so it should already be right?

lib/everest/iso15118/src/iso15118/d20/state/session_setup.cpp
SebaLukas: The evcc_id is saved in the ev_info but also send via feedback to the application/module.

lib/everest/iso15118/test/iso15118/session/tbd_controller_start_session.cpp
SebaLukas: No its not required. I copied it over from the other tests. I will remove the close in both tests because the fd will be closed from the session

lib/everest/iso15118/src/iso15118/d20/state/session_setup.cpp
mlitre: Yes exactly so we should keep this feedback signal right?

mlitre: Forgot we had a different signal my bad

mlitre (dismissed):
LGTM!

lib/everest/iso15118/include/iso15118/tbd_controller.hpp
```diff
@@ -58,7 +58,10 @@ class TbdController {
     // Mutually exclusive with the other driver; see the class note. Refuses (logs and returns false) if the other
     // driver is already running.
     StartSessionResult start_session(int connected_fd);
+    StartSessionResult start_session(int connected_fd, bool skip_app_protocol_negotiation);
     StartSessionResult start_session(int connected_fd, const std::optional<io::sha512_hash_t>& vehicle_cert_hash);
+    StartSessionResult start_session(int connected_fd, const std::optional<io::sha512_hash_t>& vehicle_cert_hash,
+                                     bool skip_app_protocol_negotiation);
```
mlitre: I think we should reduce the number of constructors there are quite a few, either we create a struct for the optional args or we use defaults

SebaLukas: Yeah there are too many constructors. First I wanted to have this here and discuss this. I will add a struct with three elements. I think thats the cleanest solution.
Some clang-tidy rules dont like default arguments.

lib/everest/iso15118/test/iso15118/session/tbd_controller_start_session.cpp
```diff
@@ -119,16 +126,114 @@ std::optional<std::vector<uint8_t>> read_v2gtp_frame(int fd, std::chrono::millis
...
+    RoundTripResult result;
+    auto start_config = iso15118::StartSessionConfig{fds.at(0), std::nullopt, false};
+
+    std::thread start_session_thread([&] {
+        if (skip_app_protocol_negotiation) {
+            start_config.skip_app_protocol_negotiation = true;
+            result.ok = controller.start_session(start_config);
+        } else {
+            result.ok = controller.start_session(start_config);
+        }
```
mlitre, line 181:
```suggestion
        if (skip_app_protocol_negotiation) {
            start_config.skip_app_protocol_negotiation = true;
        }
        result.ok = controller.start_session(start_config);
```

lib/everest/iso15118/include/iso15118/tbd_controller.hpp
```diff
@@ -32,6 +32,12 @@ enum class StartSessionResult : std::uint8_t {
     SessionComplete,
 };
 
+struct StartSessionConfig {
+    int connected_fd{-1};
```
mlitre: I would actually not include the fd in the struct and would do it as a StartSessionOptions, so that the connected_fd is "enforced" as mandatory argument

mlitre (approved):

PR #2548, opened 2026-07-31
Chore/cleanup framework changes

cburandt:
## Describe your changes

- Apply C++ coding guidelines to the framwork source files touched by #2412
- recreate autogenerated module code to include shutdown handlers and configuration parameters on_change handlers
- Docs: Move some of the framework docs to the general autogenerated docs folder
- Docs: new section for experimental features, as this PR/its predecessors introduced some of them


docs/source/explanation/configuration-service.rst
```diff
@@ -0,0 +1,355 @@
...
+*******************
+Configuration Slots
+*******************
+
+A configuration slot holds one complete EVerest configuration: the set of active
+modules, their parameters and their connections. The database can hold several
+slots side by side, and exactly one of them is the **active** slot — the one the
+running modules were started from.
+
+Marking a different slot does not switch anything immediately. It records that
```
Pietfried:
```suggestion
Marking a different slot active does not switch anything immediately. It records that
```

docs/source/explanation/manager-lifecycle.rst
```diff
@@ -0,0 +1,329 @@
...
+
+All remaining sections of this page describe the graceful
+(``--graceful-shutdown``) flow. In default mode the MQTT shutdown publish is
+skipped and the force-terminate escalation happens immediately.
+
+********************
+Startup (Happy Path)
+********************
+
+- ``Idle`` (initial C++ object state) → ``Initializing`` when ``run()`` begins.
```
Pietfried:
```suggestion
- ``Idle``→ ``Initializing`` when ``run()`` begins.
```

docs/source/explanation/manager-lifecycle.rst
```diff
@@ -0,0 +1,329 @@
...
+  empty module list is a ``std::logic_error`` — the boot and restart paths
+  reject that case before it gets here, because zero modules would hang the
+  manager in ``StartingModules`` forever (no ready handler would ever fire).
+- ``StartingModules`` → ``Running`` when every non-ignored module has published
+  **ready** on MQTT (and standalone handling rules are satisfied). If
+  ``SIGINT``/shutdown is already in progress when the last ready arrives, the
+  transition to ``Running`` is **skipped** on purpose.
+- On the transition to ``Running`` the manager clears the retained startup
+  topics (unless ``--retain-topics``) and publishes the global ready signal on
+  ``<mqtt_everest_prefix>ready`` (``transition_to_running_and_announce()``).
```
Pietfried:
```suggestion
  ``<mqtt_everest_prefix>ready``.
```

lib/everest/framework/docs/ConfigService.md
Pietfried, line 1: Rewritten implementation notes pin hard-coded source line numbers. Several are already wrong on this branch

docs/source/explanation/manager-lifecycle.rst
Pietfried, line 1: While the configuration-service.rst is mostly at the right altitude the manager-lifecycle docs feel hard to read and leak too many implementation details, including private members, functions, exceptions, etc.

My suggestion is to lift this a little above the implementaiton level so that future changes and refactorings dont bite us here

Pietfried (dismissed):

lib/everest/framework/docs/ConfigService.md
cburandt: Removed the line numbers.

docs/source/explanation/manager-lifecycle.rst
cburandt: Raised the level a bit in this text.

docs/source/explanation/manager-lifecycle.rst
```diff
@@ -0,0 +1,315 @@
...
+
+************************************
+Expected vs. Exceptional Transitions
+************************************
+
+**Expected:** linear startup; clean shutdown after ``SIGINT``; controlled
+restart after an admin request; optional bounded crash recovery restart loop
+when ``--recover-module-crashes`` is set; manager exit after an unexpected
+module exit when that flag is omitted; timeout escalation only when modules miss
+their shutdown deadline.
```
Pietfried: I dont quite understand why this section exists and is helpful. The headline says `Expected vs. Exceptional Transitions` but I only read about expected transitions?

cburandt: Now reworked the text and also adjusted the code a bit (for better consistency).

modules/EVSE/OCPPmulti/OCPPmulti.hpp
```diff
@@ -5,7 +5,7 @@
 
 //
 // AUTO GENERATED - MARKED REGIONS WILL BE KEPT
-// template version 2
+// template version 3
```
florinmihut: This bumps the stamp `2` -> `3`, but the file wasn't actually regenerated: `module.hpp.j2` at version 3 emits `void shutdown();` unconditionally (no manifest-driven `{% if %}` around it) next to `init()`/`ready()`, and here the private section still only has:

```cpp
    friend class LdEverest;
    void init();
    void ready();
```

So the generated `LdEverest::shutdown()` calls `mod_ptr->shutdown()`, which -- non-virtual, nothing found in `OCPPmulti` -- resolves to `Everest::ModuleBase::shutdown()` and logs *"No shutdown handler installed! Please implement shutdown() in your module!"*. Nothing actually breaks today, since OCPPmulti tears down in `~OCPPmulti()` (`m_ocpp.shutdown()` / `m_charge_point.shutdown()`), but the log is misleading and the bumped stamp will make the next regeneration pass skip this file, so it silently stays stale.

Given the PR description says the module code was recreated to include shutdown handlers, could OCPPmulti be re-run through the generator here?

modules/EVSE/OCPPmulti/ocpp_generic/ocppImpl.hpp
```diff
@@ -5,7 +5,7 @@
 
 //
 // AUTO GENERATED - MARKED REGIONS WILL BE KEPT
-// template version 3
+// template version 4
```
florinmihut: Same as in `OCPPmulti.hpp`, and it applies to all five OCPPmulti impl headers -- `auth_token_providerImpl.hpp`, `auth_token_validatorImpl.hpp`, `ocpp_data_transferImpl.hpp`, `ocppImpl.hpp`, `session_costImpl.hpp`. Each one's only change in this PR is the stamp `3` -> `4`, but `interface-Impl.hpp.j2` at version 4 unconditionally emits `void shutdown() override;` right after `init()`/`ready()`, and these files still just have:

```cpp
    virtual void init() override;
    virtual void ready() override;
```

Without the override, `ModuleBase::invoke_shutdown()` lands on the `ImplementationBase::shutdown()` default and logs *"No shutdown handler installed! Please implement shutdown() in your implementation!"* for each of the five. Same ask: re-run the generator over OCPPmulti rather than only bumping the stamps.

florinmihut (commented):
Reviewed the framework cleanup + regeneration. The bulk of the diff is the mechanical `this->x` / `x_` -> `m_x` member rename plus the docs move, and I checked those for shadowing (locals named `config`, `mqtt_settings`, `runtime_settings`, `modules_dir` in `runtime.cpp`), constructor init-list ordering against the new member declaration order, and leftover old identifiers -- all clean. The behavioural bits also look sound to me: the `cleanup_modules_state_if_configured()` null-guard extraction, moving `reinitialize_from_db()` inside the try in `reload_and_update_context()` (`load_and_validate_config` rethrows and the `catch(...)` covers the boost non-`std::exception` path, so `config` can't be null at the deref below), the consolidated boot-failure branch, continuing past a failed DB bootstrap with `--idle-on-failure` (`init_database_bootstrap` sets `bs.db_connection` before any failure return), and the second-SIGINT SIGKILL path.

Two things about OCPPmulti look like it was missed by the regeneration sweep while still getting its version stamps bumped -- see the inline comments.

florinmihut (changes requested):

modules/EVSE/OCPPmulti/OCPPmulti.hpp
cburandt: Did not re-run, as this would require the shutdown() function to be implemented. Rather reset the template version to the original, such that the next generator pass will process the file correctly.

modules/EVSE/OCPPmulti/ocpp_generic/ocppImpl.hpp
cburandt: Same as above.

florinmihut (approved):

PR #2550, opened 2026-08-02
feat(OCPP1.6): support multi-slot network configuration via device model

Pietfried:
## Motivation

OCPP1.6 only allowed a single Network Configuration via the configuration parameters CentralSystemURI, SecurityProfile, AuthorizationKey, etc.

With OCPPmulti and the harmonized adressing of component variables that has been added recently,  a logical follow up is to also harmonize the network connection configuration so that integrators can configure the CSMS connection details for OCPP1.6 just like they can for OCPP2.x. That functionality is provided by this PR.

## Describe your changes

Source OCPP1.6 network connection profiles from device-model slots instead of the legacy single-connection config, enabling multi-slot failover:

- v16 device-model config reads/writes NetworkConnectionProfiles, slot priority ordering and security-profile pruning (confirm-gated)
- connectivity_manager honors the slot list for failover and retries
- OCPPmulti delegates configure_network to the 1.6 stack, reloads profiles on consumer variable writes, and guards edits to the in-use slot the same way 2.x SetVariables does
- allow OCPP16 in the NetworkConfiguration_2 OcppVersion valuesList
- unit tests for device-model connectivity + OCPPmulti variable access, integration test for configure_network in 1.6, docs updates

## Checklist before requesting a review
- [x] I have performed a self-review of my code
- [x] I have made corresponding changes to the documentation
- [x] I read the [contribution documentation](https://everest.github.io/nightly/project/contributing.html) and made sure that my changes meet its requirements

PR #2552, opened 2026-08-03
OCPP: extend the connection status with the connection details

nzbr:
## Describe your changes

Replaces the is_connected variable on the OCPP API and the external ocpp_consumer API with a new, extended connection_status variable. In addition to whether or not there is an active connection to a CSMS, this also carries additional information such as the connection URL, OCPP Version and security profile. For backward-compatibility reasons, the external API still provides the `is_connected` variable, which has been marked as deprecated so that it may be removed in the future. Its status is inferred from the contents of the internal API's connection_status


Pietfried (changes requested):
In general this looks good. For OCPP1.6: Instead of manually requesting CentralSystemURI, SecProfile, etc. on a connection changed callback, I'd recommend to simply change the `connection_state_changed_callback` in v16 to include 

int configuration_slot,
const NetworkConnectionProfile& network_connection_profile,
const OcppProtocolVersion ocpp_version

just like it is done in v2.

nzbr:
Makes sense. I updated the PR accordingly, through I have opted ot only use the network_connection_profile, because the other two variables do not make sense for OCPP 1.6 imo. It does not support multiple configuration slots and the OCPP version is always going to be 1.6 anyway, so it isn't neccessary to pass it around

Pietfried:
> Makes sense. I updated the PR accordingly, through I have opted ot only use the network_connection_profile, because the other two variables do not make sense for OCPP 1.6 imo. It does not support multiple configuration slots and the OCPP version is always going to be 1.6 anyway, so it isn't neccessary to pass it around

Ok with the version! With https://github.com/EVerest/EVerest/pull/2550 we actually supports multiple slots, so the slot becomes relevant.

nzbr:
I rebased onto main, and wired the OCPP 1.6 stack up to pass through the configuration slot. I squashed all previous changes into a single commit, so the actual new changes after the rebase are cleanly separated into their own commit on top of that

lib/everest/ocpp_module_common/src/conversions.cpp
```diff
@@ -1503,6 +1503,46 @@ types::ocpp::StatusInfoType to_everest_status_info_type(const ocpp::v2::StatusIn
...
+        connection_status.csms_url = network_connection_profile.ocppCsmsUrl.get();
+    }
+    connection_status.security_profile = network_connection_profile.securityProfile;
+    if (network_connection_profile.identity.has_value() and
+        not network_connection_profile.identity.value().get().empty()) {
+        connection_status.identity = network_connection_profile.identity.value().get();
+    }
+    connection_status.ocpp_version = to_everest_ocpp_version(protocol_version);
+    connection_status.configuration_slot = configuration_slot;
+    // OCPP 1.6 fills these fields with synthetic data because they are required, so we don't need to pass them on
```
Pietfried: OCPP1.6 also returns real per slot values now.

lib/everest/everest_api_types/tests/expected_types_file_hashes.csv
```diff
@@ -12,7 +12,7 @@ types/iso15118.yaml,ad704529050732e08c2f72d2694b07a6aabbd32fe3fa749bdada052d7c10
 types/isolation_monitor.yaml,45d98b5072fa5d02a476860fe7d45a7b02f8a05eb6be94327f32dbe159d4ec40
 types/money.yaml,4a7001c50216fcf96caac4fe16db17e23ce5149d000be1de4ac1f0135f5bdb0b
 types/network.yaml,5fbb7152570d3b0cf72cabd21359e6dde94b0cec460a3ce1117e4d2d244aea48
-types/ocpp.yaml,9ef0e1b4e44111502813aea59519947a12448ddbfcbf986f17976349e94fd235
+types/ocpp.yaml,453a733eb6c709981d7f94a03b9938dade1c5a3e753788a5967ad79a0bd420b7
```
Pietfried: That hash needs to be updated

Pietfried (changes requested):

lib/everest/ocpp_module_common/src/conversions.cpp
nzbr: Missed that. I removed the condition there and adjusted the corresponding test

Pietfried (approved):

PR #2558, opened 2026-08-05
fix(iso15118): extract the listen socket helper

mlitre:
## Describe your changes

`connection_plain` and `connection_ssl` each built their own listening socket with the same sequence of
`socket`, `setsockopt`, `bind` and `listen` calls, and each cleaned up by hand on every error path. This
extracts that into one `create_tcp_listen_socket` helper in `detail/io/socket_helper` and points both
connections at it.

The extraction is not purely mechanical. Four things changed deliberately:

- The interface name is back in the bind failure message. Losing it makes a bind failure much harder to
  diagnose on a machine with several interfaces, since the message no longer says which one it was.
- Descriptor cleanup is RAII rather than four separate `::close(fd)` calls. Each of those calls was an
  opportunity to return early and leak the descriptor, which is exactly the kind of error path that
  never gets exercised.
- The IPv6 address is passed by value instead of by pointer. It is 16 bytes, so there is nothing to save
  by indirecting, and by value there is no lifetime question at the call site.
- The test loops the descriptor leak check rather than probing once. A single probe cannot distinguish
  "does not leak" from "leaked once and we did not look again".

The helper reports failure through the return value and leaves the caller to decide, rather than
throwing or exiting, which keeps it usable from both connection types.

N/A

---

### Stack

This branch is part of an eight PR series. Three are rooted on `main` (#2558, #2559, #2560);
the rest form a chain that forks at #2562.

| Order | PR | Branch | Base |
|---|---|---|---|
| **1** | **#2558 (this PR)** | **`fix/iso15118-listen-socket-helper`** | `main` |
| 2 | #2559 | `feat/tls-wrap-connecting-fd` | `main` |
| 3 | #2561 | `fix/io-event-handler-truthfulness` | #2559 |
| 4 | #2562 | `feat/io-client-tx-tristate` | #2561 |
| 5 | #2563 | `feat/io-fd-event-client-handshake-phase` | #2562 |
| 6 | #2564 | `feat/io-tls-policies-rebuilt` | #2563 |
| 7 | #2560 | `fix/io-connect-errno-and-backoff` | `main` |
| 8 | #2565 | `refactor/chargebridge-drop-hand-rolled-tx-gates` | #2562 |

Independent of the rest of the series, so it can be reviewed and merged on its own.

### DIFF
--- a/lib/everest/iso15118/include/iso15118/detail/io/socket_helper.hpp
+++ b/lib/everest/iso15118/include/iso15118/detail/io/socket_helper.hpp
@@ -2,6 +2,7 @@
 // Copyright 2023 Pionix GmbH and Contributors to EVerest
 #pragma once
 
+#include <cstdint>
 #include <memory>
 #include <string>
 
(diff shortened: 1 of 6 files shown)


SebaLukas (approved):
👍

PR #2559, opened 2026-08-05
feat(tls): wrap an already connecting descriptor

mlitre:
## Describe your changes

`libtls` could only produce a connection by doing the TCP connect itself inside `Client::connect`. That
forecloses any design where something else owns the connect, for example a caller that wants it driven on
its own event loop rather than blocking a thread.

This adds `Client::wrap_connecting_fd`, which takes a descriptor whose TCP connect the caller has already
started and returns a `ClientConnection` around it without handshaking. Driving the handshake afterwards
is the caller's job.

Alongside it:

- **`Connection::has_pending()`** reports whether decrypted plaintext is still buffered in the record
  layer, which is what lets a reader drain a whole message rather than one chunk per wakeup. It uses
  `SSL_has_pending()` rather than `SSL_pending()`. That is hardening, not a fix: the two only diverge
  under read-ahead, DTLS or kTLS, none of which are enabled here.

- **Descriptor ownership is explicit on the failure path.** When wrapping fails, no `BIO_CLOSE` owner has
  taken the descriptor, so it stays the caller's to close. Both wrap entry points say so, and neither
  closes it, which would be a double close.

- **SNI and optional hostname pinning.** `ClientConnection` now sends the server name in the SNI extension
  and, when `verify_subject_name` is set, pins the peer certificate to that name. Neither existed in
  libtls before, so verification was chain of trust only: a certificate signed by any trusted CA was
  accepted regardless of who it was issued to.

  Pinning is off by default, because enabling it unconditionally would break every PKI whose certificates
  carry no hostname. `Client::init` now **rejects** `verify_subject_name` without `verify_server`: under
  `SSL_VERIFY_NONE` the name check is installed but its result is never enforced, so the combination
  cannot be honored and failing init beats leaving the caller believing it pinned a host. IP literals are
  pinned through the verification parameter and kept out of SNI, which RFC 6066 section 3 does not permit.

`has_pending()` and the SNI/pinning work have no in-tree caller yet; the consumer is #2564. Keeping them
here rather than in the consuming PR avoids widening that PR's scope into `lib/everest/tls`, and a TLS
client with no way to verify who it is talking to is not a complete client, so the capability belongs with
the entry point that exposes it.

N/A

---

### Stack

Root of an eight PR series. #2561 through #2565 all descend from this one, so it wants to land first.

| Order | PR | Branch | Base |
|---|---|---|---|
| 1 | #2558 | `fix/iso15118-listen-socket-helper` | `main` |
| **2** | **#2559 (this PR)** | **`feat/tls-wrap-connecting-fd`** | `main` |
| 3 | #2561 | `fix/io-event-handler-truthfulness` | #2559 |
| 4 | #2562 | `feat/io-client-tx-tristate` | #2561 |
| 5 | #2563 | `feat/io-fd-event-client-handshake-phase` | #2562 |
| 6 | #2564 | `feat/io-tls-policies-rebuilt` | #2563 |
| 7 | #2560 | `fix/io-connect-errno-and-backoff` | `main` |
| 8 | #2565 | `refactor/chargebridge-drop-hand-rolled-tx-gates` | #2562 |

### DIFF
--- a/lib/everest/tls/include/everest/tls/tls.hpp
+++ b/lib/everest/tls/include/everest/tls/tls.hpp
@@ -72,10 +72,18 @@ public:
     TlsKeyLoggingServer(const std::string& interface_name, uint16_t port_);
     ~TlsKeyLoggingServer();
 
-    ssize_t send(const char* line);
+    /**
+     * \brief send one key-log line to the multicast destination
+     * \returns false when the socket is unusable or the line was not sent in full
+     */
+    bool send(const char* line);
 
-    auto get_fd() const {
-        return fd;
+    /**
+     * \brief whether the multicast socket was set up
+     * \returns false when the constructor failed, in which case the object cannot send
+     */
+    [[nodiscard]] bool is_valid() const {
+        return fd != -1;
     }
 
     auto get_port() const {
@@ -138,7 +146,7 @@ public:
 protected:
     std::unique_ptr<connection_ctx> m_context; //!< opaque connection data
     state_t m_state{state_t::idle};            //!< connection state
-    std::string m_ip;                          //!< peer IP address
+    std::string m_peer_host;                   //!< peer host, a DNS hostname or IP literal
     std::string m_service;                     //!< peer port
     std::int32_t m_timeout_ms;                 //!< default operation timeout
     std::string
@@ -225,10 +233,11 @@ public:
     }
 
     /**
-     * IP address of the connection's peer
+     * Host of the connection's peer: a DNS hostname or IP literal. The name is
+     * kept for API stability.
      */
     [[nodiscard]] const std::string& ip_address() const {
-        return m_ip;
+        return m_peer_host;
     }
 
     /**
@@ -261,6 +270,25 @@ public:
      * \returns the underlying socket or INVALID_SOCKET on error
      */
     [[nodiscard]] int socket() const;
+
+    /**
+     * \brief whether the SSL object and its socket BIO were allocated
+     * \returns false when allocation failed, in which case no BIO_CLOSE owner
+     *          adopted the socket fd and the caller still owns it
+     */
+    [[nodiscard]] bool is_valid() const;
+
+    /**
+     * \brief whether the TLS record layer still holds buffered data
+     * \returns true when a further read() can return data without the socket
+     *          becoming readable again
+     * \note SSL_has_pending() is used rather than SSL_pending(): the latter
+     *       counts only the decrypted remainder of the record being processed
+     *       and returns 0 while a complete but still unprocessed record sits in
+     *       the read buffer. A drain loop guarded on that count would stop with
+     *       data stranded in userspace and never be woken again.
+     */
+    [[nodiscard]] bool has_pending() const;
 
     /**
      * \brief obtain the peer certificate
@@ -353,7 +381,8 @@ public:
  */
 class ClientConnection : public Connection {
 public:
-    ClientConnection(SslContext* ctx, int soc, const char* ip_in, const char* service_in, std::int32_t timeout_ms);
+    ClientConnection(SslContext* ctx, int soc, const char* ip_in, const char* service_in, std::int32_t timeout_ms,
+                     bool verify_subject_name);
     ClientConnection() = delete;
     ClientConnection(const ClientConnection&) = delete;
     ClientConnection(ClientConnection&&) = delete;
@@ -699,7 +728,14 @@ public:
         std::int32_t io_timeout_ms{-1};              //!< default socket timeout in milliseconds (recommend > 1 sec)
         //!< minimum TLS protocol version, e.g. TLS1_2_VERSION or TLS1_3_VERSION; 0 means use default
         int min_proto_version{0};
-        bool verify_server{true};      //!< verify the server certificate
+        bool verify_server{true}; //!< verify the server certificate
+
+        //! verify_subject_name: match the peer certificate subject/SAN against the SNI host via
+        //! SSL_set1_host. Requires verify_server, init() rejects the combination without it because the
+        //! host cannot be enforced. The default (false) is chain-of-trust ONLY: a certificate signed by
+        //! any trusted CA is accepted whatever host it was issued for. Leave it false for a PKI whose
+        //! certificates carry no hostname; set true for a hostname-bearing PKI.
+        bool verify_subject_name{false};
         bool status_request{false};    //!< include a status request extension in the client hello
         bool status_request_v2{false}; //!< include a status request v2 extension in the client hello
         bool trusted_ca_keys{false};   //!< include a trusted ca keys extension in the client hello
@@ -710,6 +746,7 @@ public:
 private:
     std::unique_ptr<client_ctx> m_context;                      //!< opaque object data
     std::int32_t m_timeout_ms{-1};                              //!< default operation timeout
+    bool m_verify_subject_name{false};                          //!< applied to new connections
     trusted_ca_keys_t m_trusted_ca_keys;                        //!< trusted CA keys configuration data
     std::unique_ptr<ClientStatusRequestV2> m_status_request_v2; //!< status request extension handler
 
@@ -749,6 +786,19 @@ public:
     [[nodiscard]] inline ConnectionPtr connect(const char* host, const char* service, bool ipv6_only) {
         return connect(host, service, ipv6_only, m_timeout_ms);
     }
+
+    /**
+     * \brief wrap an already-connected or still-connecting TCP socket as a TLS client connection
+     * \param[in] fd TCP socket file descriptor
+     * \param[in] host_for_sni host for the peer-id / SNI slot, may be nullptr
+     * \return nullptr if the SSL_CTX is not initialised or SSL/BIO allocation fails
+     * \note A DNS \p host_for_sni is sent in the SNI extension, an IP literal is not
+     *       (RFC 6066 §3). With config_t::verify_subject_name the peer certificate is
+     *       pinned to \p host_for_sni and a pin failure fails the handshake closed.
+     * \note Ownership: on success the connection owns \p fd and closes it. On a
+     *       nullptr return the caller still owns \p fd; this factory never closes it.
+     */
+    [[nodiscard]] ConnectionPtr wrap_connecting_fd(int fd, const char* host_for_sni);
 
     /**
      * \brief the default SSL callbacks
--- a/lib/everest/tls/src/tls.cpp
+++ b/lib/everest/tls/src/tls.cpp
@@ -11,6 +11,7 @@
 #include <arpa/inet.h>
 #include <array>
 #include <cassert>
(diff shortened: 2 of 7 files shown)


lib/everest/tls/tests/tls_client_wrap_test.cpp
```diff
@@ -0,0 +1,323 @@
...
+
+    server_config.socket = listener.fd;
+    const auto init_state = server.init(server_config, nullptr);
+    ASSERT_TRUE(init_state == state_t::init_complete || init_state == state_t::init_socket);
+
+    std::thread server_side([&]() {
+        auto server_conn = accept_and_wrap(server, listener.fd);
+        if (server_conn) {
+            (void)server_conn->accept(1000);
+            (void)server_conn->shutdown(1000);
```
SebaLukas: Some compilers have issues with this `(void)` casting.

lib/everest/tls/tests/tls_connection_test.hpp
```diff
@@ -154,6 +163,120 @@ inline void run_server(tls::Server& server) {
...
+};
+
+inline LoopbackListener make_loopback_listener() {
+    LoopbackListener listener;
+    const int fd = ::socket(AF_INET, SOCK_STREAM, 0);
+    if (fd < 0) {
+        return listener;
+    }
+    int reuse = 1;
+    (void)::setsockopt(fd, SOL_SOCKET, SO_REUSEADDR, &reuse, sizeof(reuse));
```
SebaLukas: See above. Applies to the entire file.

SebaLukas (commented):
Only some smaller nits

lib/everest/tls/tests/tls_connection_test.hpp
mlitre: Fixed

lib/everest/tls/tests/tls_client_wrap_test.cpp
mlitre: Fixed

lib/everest/tls/include/everest/tls/tls.hpp
```diff
@@ -699,7 +714,12 @@ class Client {
         std::int32_t io_timeout_ms{-1};              //!< default socket timeout in milliseconds (recommend > 1 sec)
         //!< minimum TLS protocol version, e.g. TLS1_2_VERSION or TLS1_3_VERSION; 0 means use default
         int min_proto_version{0};
-        bool verify_server{true};      //!< verify the server certificate
+        bool verify_server{true}; //!< verify the server certificate
+        //!< match the peer certificate subject/SAN against the SNI host via SSL_set1_host, only enforced
```
james-ctc: This comment needs to be separated from `bool verify_server{true};`. 
It isn't clear that `match the peer certificate subject/SAN against the SNI` isn't a continuation of `verify the server certificate`

a blank line or `verify_subject_name matches the peer certificate subject/SAN against …` would help clarify

lib/everest/tls/include/everest/tls/tls.hpp
```diff
@@ -262,6 +263,19 @@ class Connection {
...
+     * \returns false when allocation failed, in which case no BIO_CLOSE owner
+     *          adopted the socket fd and the caller still owns it
+     */
+    [[nodiscard]] bool is_valid() const;
+
+    /**
+     * \brief decrypted bytes buffered in the TLS record layer, readable without
+     *        a socket read
+     */
+    [[nodiscard]] std::size_t pending() const;
```
james-ctc: Perhaps a bool result and implement using `SSL_has_pending()` would be better.
There seem to be some complications around the use of `SSL_pending`
https://docs.openssl.org/master/man3/SSL_pending/

> it is possible for no more bytes to be readable from the underlying BIO (because OpenSSL has already read them) and for SSL_pending() to return 0, even though readable application data bytes are available

lib/everest/tls/src/tls.cpp
```diff
@@ -800,6 +802,16 @@ int Connection::socket() const {
...
 }
 
+bool Connection::is_valid() const {
+    // soc_bio is the fd's BIO_CLOSE owner, null only when the ctor's SSL_new or
+    // BIO_new_socket failed, in which case the fd was never adopted.
+    return m_context != nullptr && m_context->soc_bio != nullptr;
+}
+
+std::size_t Connection::pending() const {
+    return (m_context && m_context->ctx) ? static_cast<std::size_t>(SSL_pending(m_context->ctx.get())) : 0;
```
james-ctc: usage consistency:

```
m_context != nullptr && m_context->soc_bio != nullptr
vs
m_context && m_context->ctx
```

personally I prefer the `!= nullptr` as it is clear what is being tested. Either way a consistent style should be used within the same file.

lib/everest/tls/src/tls.cpp
```diff
@@ -823,18 +835,22 @@ int ssl_keylog_server_index{-1};
...
 void keylog_callback(const SSL* ssl, const char* line) {
 
+    // Registered CTX-wide, so it also fires for connections that skipped the
+    // per-connection key-log server and therefore carry no ex_data.
     auto keylog_server = static_cast<TlsKeyLoggingServer*>(SSL_get_ex_data(ssl, ssl_keylog_server_index));
 
-    std::string key_log_msg = "TLS Handshake keys on port ";
-    key_log_msg += std::to_string(keylog_server->get_port()) + ": ";
-    key_log_msg += std::string(line);
+    if (keylog_server != nullptr) {
```
james-ctc, line 842: should be a check that the keylog_server is configured. The constructor has many failure points

lib/everest/tls/src/tls.cpp
```diff
@@ -823,18 +835,22 @@ int ssl_keylog_server_index{-1};
...
 
-    log_info(key_log_msg);
+        log_info(key_log_msg);
 
-    if (keylog_server->get_fd() != -1) {
-        const auto result = keylog_server->send(line);
-        if (result not_eq strlen(line)) {
-            log_error("key_logging_server send() failed!");
+        if (keylog_server->get_fd() != -1) {
+            const auto result = keylog_server->send(line);
```
james-ctc: this check should be in `TlsKeyLoggingServer::send()`
calling code should need to be checking whether the file descriptor is -1.
Also the length check could be added to send(). As this code stands the length of the string is calculated twice: once here and once in `TlsKeyLoggingServer::send()`

lib/everest/tls/src/tls.cpp
```diff
@@ -864,9 +880,22 @@ ServerConnection::ServerConnection(SslContext* ctx, int soc, const char* ip_in,
...
-            m_keylog_server = std::make_unique<TlsKeyLoggingServer>(std::string(tls_key_interface), port);
-            SSL_set_ex_data(m_context->ctx.get(), ssl_keylog_server_index, m_keylog_server.get());
+            // service_in comes from the accept path and may be empty or
+            // non-numeric, so parse it without throwing.
+            std::uint16_t port{0};
+            bool port_valid{false};
+            if (service_in != nullptr) {
+                const char* end = service_in + std::strlen(service_in);
+                const auto [ptr, ec] = std::from_chars(service_in, end, port);
+                port_valid = (ec == std::errc{}) && (ptr == end) && (end != service_in);
```
james-ctc: if `end == service_in` then there is no need to call `std::from_chars`

lib/everest/tls/src/tls.cpp
```diff
@@ -864,9 +880,22 @@ ServerConnection::ServerConnection(SslContext* ctx, int soc, const char* ip_in,
...
+            }
+            if (port_valid) {
+                m_keylog_server = std::make_unique<TlsKeyLoggingServer>(std::string(tls_key_interface), port);
+                SSL_set_ex_data(m_context->ctx.get(), ssl_keylog_server_index, m_keylog_server.get());
```
james-ctc: it would be safer to only call `SSL_set_ex_data` when TlsKeyLoggingServer was successfully set up.

lib/everest/tls/src/tls.cpp
```diff
@@ -913,11 +942,43 @@ void ServerConnection::wait_all_closed() {
...
+    const bool ip_literal = is_ip_literal(m_peer_host);
+
+    if (!ip_literal) {
+        SSL_set_tlsext_host_name(ssl, m_peer_host.c_str());
+    }
+
+    if (verify_subject_name) {
+        // A failed pin faults the connection so the handshake fails closed
+        // instead of silently downgrading to chain-of-trust only.
+        const int pinned = ip_literal ? X509_VERIFY_PARAM_set1_ip_asc(SSL_get0_param(ssl), m_peer_host.c_str())
```
james-ctc, line 982: from this comment:
> // RFC 6066 section 3 bars IP literals from SNI.

it isn't obvious why an IP check is being used at all.

lib/everest/tls/src/tls.cpp
```diff
@@ -1447,6 +1513,11 @@ bool Client::init(const config_t& cfg, const override_t& override) {
     assert(override.trusted_ca_keys_free != nullptr);
 
     m_timeout_ms = cfg.io_timeout_ms;
+    m_verify_subject_name = cfg.verify_subject_name;
+    if (cfg.verify_subject_name && !cfg.verify_server) {
```
james-ctc, line 1523: this miss-configuration should do more than just log something. Where security is concerned code/configuration needs fixing and not rely on someone spotting the warning.

james-ctc (changes requested):

lib/everest/tls/include/everest/tls/tls.hpp
mlitre: Good idea thanks!

lib/everest/tls/include/everest/tls/tls.hpp
mlitre: Done!

lib/everest/tls/src/tls.cpp
mlitre: Done

lib/everest/tls/src/tls.cpp
mlitre: Good idea, done!

lib/everest/tls/src/tls.cpp
mlitre: I think it's fine since the send won't cause issues if it is not configured

lib/everest/tls/src/tls.cpp
mlitre: Done

lib/everest/tls/src/tls.cpp
mlitre: It used to be only ip based I added domain name support as well now

lib/everest/tls/src/tls.cpp
mlitre: Probably a good idea

james-ctc (approved):

PR #2560, opened 2026-08-05
fix(io): carry the connect errno and name the retry delay

mlitre:
## Describe your changes

Four defects on the client connect path. Three of the four ended with a real errno being reported as
`EBADF`, by different routes.

- **The connect errno was discarded.** A refused connect reported `EBADF` (9) instead of
  `ECONNREFUSED` (111), a timeout `EBADF` instead of `ETIMEDOUT` (110). The errno does reach the
  catch site; what lost it was `get_error()` falling back to the pending error of a descriptor that
  was never assigned. It is now captured in the catch block.
- **The reconnect delay was a second copy of the connect timeout.** A refused connect returns in
  about 0 ms and then slept the full 1000 ms timeout before retrying. A 500 ms timeout against a
  saturated backlog cost 1000 ms per attempt. The retry delay is now a separate named value, and the
  tests bound latency rather than asserting an exact sleep.
- **The recorded errno could outlive its descriptor.** Only `setup()` and a successful `connect()`
  invalidated it, so a failed `connect()`, then a successful `open()`, then a deliberate `close()`
  still reported `ECONNREFUSED`. The descriptor and the reason there is no descriptor now move
  together through `adopt()`, `record_connect_failure()` and `discard()`. On `udp_socket_base` the
  descriptor became private, because `mdns_socket` was assigning the inherited protected member
  directly and reaching around every site a per-site clear would have had to cover.
- **`getaddrinfo` failures surfaced as `EBADF` as well.** All four resolver legs threw a bare
  `std::runtime_error`, recorded by the socket-level catch as error 0, landing back on the same
  unassigned-descriptor probe. They now throw `socket_error` carrying `errno` on `EAI_SYSTEM` and
  `EHOSTUNREACH` otherwise, keeping the `gai_strerror` text. A zero `errno` is floored so it cannot
  reproduce the `EBADF` this removes.

Reading `errno` unconditionally on the resolver path would be wrong: after a non-`EAI_SYSTEM` failure
it holds unrelated leftovers (measured 22/`EINVAL` and 90/`EMSGSIZE`), so the mapping is needed for
correctness, not tidiness.

Two gaps stated rather than hidden. The `EAI_SYSTEM` branch and the `open_udp_server_socket` resolver
leg have no forceable failure mode without fault injection, and are marked untested in a comment. The
resolver tests use a 64-octet DNS label rather than an unresolvable hostname, so the file keeps its
no-DNS, no-egress property.

N/A

---

### Stack

This PR is rooted on `main` and is blocked by nothing.

- `main`
  - **#2560 `fix/io-connect-errno-and-backoff` (this PR)**

### DIFF
--- a/applications/pionix_chargebridge/src/bsp_bridge.cpp
+++ b/applications/pionix_chargebridge/src/bsp_bridge.cpp
@@ -9,8 +9,6 @@
 #include <protocol/evse_bsp_host_to_cb.h>
 
 namespace {
-const int default_udp_timeout_ms = 1000;
-
 const char* cp_state_to_string(CpState state) {
     switch (state) {
     case CpState_A:
@@ -65,7 +63,7 @@ bsp_bridge::bsp_bridge(bsp_bridge_config const& config, everest::lib::io::event:
 }
 
 void bsp_bridge::create_udp_client(std::string const& remote, uint16_t remote_port, std::string const& identifier) {
-    m_udp = std::make_unique<everest::lib::io::udp::udp_client>(remote, remote_port, default_udp_timeout_ms);
+    m_udp = std::make_unique<everest::lib::io::udp::udp_client>(remote, remote_port);
     m_udp_ready = false;
     m_udp_on_error = false;
     m_udp->set_rx_handler([this](auto const& data, auto&) {
--- a/applications/pionix_chargebridge/src/can_bridge.cpp
+++ b/applications/pionix_chargebridge/src/can_bridge.cpp
@@ -10,10 +10,6 @@
 #include <memory>
 #include <protocol/cb_can_message.h>
 
-namespace {
-const int default_udp_timeout_ms = 1000;
-}
-
 namespace charge_bridge {
 using namespace std::chrono_literals;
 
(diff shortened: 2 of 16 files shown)


djchhp:
This looks like a good improvement. I found one issue with the code though.

 m_connect_error is now returned whenever the socket has no FD, but neither tcp_socket::open() nor udp_socket_base::open_as_client() clears or updates that cache.
 
  A caller can:
 
  1. call async setup() / connect() and receive, for example, a device-bind error;
  2. later call the synchronous open() API, which fails for a different reason, such as ECONNREFUSED or DNS resolution failure;
  3. call get_error() while no FD is assigned.
 
  get_error() then returns the error from the earlier async attempt, not the reason the latest open() failed. setup() clears the cache, but the synchronous open() flow does not call setup().
 
  Please at least clear m_connect_error before each synchronous open attempt (tcp_socket::open(), udp_socket_base::open_as_client(), and any other path that can leave the object without an FD). Preferably, make those paths capture and cache
  socket_error too, s

djchhp (changes requested):
so comment above

djchhp (approved):
lgtm

PR #2561, opened 2026-08-05
fix(io): make event handler registration truthful

mlitre:
## Describe your changes

`fd_event_handler`'s registration calls could report success after `epoll_ctl` had failed, so a
caller had no way to tell a live registration from a dead one. This makes those results honest and
gives `fd_event_sync_interface` ownership of the registration it creates.

- **Truthful results.** `register_event_handler` and `unregister_event_handler` now report the
  `epoll_ctl` outcome, not just whether the bookkeeping succeeded. Call sites that already checked
  start seeing real failures, which is the point. `pionix_chargebridge` is the one that checks, so
  wiring mistakes that used to pass silently now surface at startup.
- **`is_registered(fd)`** is added, because a bool return cannot answer whether a descriptor is still
  in the handler.
- **The interface drops its own registration.** An implementor destroyed while still registered left
  its entry in the map. The stale callback cannot fire for the nested-handler case (destruction
  closes the nested epoll descriptor and the kernel drops it), so this is a leak rather than a use
  after free. The real failure is descriptor recycling: a later registration that receives the same
  number is rejected because the map still holds the old entry, failing far from its cause. The
  interface now records the handler and descriptor and unregisters in its own destructor, using the
  recorded number rather than calling the pure-virtual `get_poll_fd()` on an object being destroyed.
  Copy and move are deleted, since a recorded registration has exactly one owner.
- **`register_events` and `unregister_events` are no longer virtual.** Nothing overrides them, and an
  override that skipped the recording would leave the handler calling `sync()` on a destroyed object,
  a use after free rather than a leak. Consumers needing a different registration shape (several
  descriptors, write events) implement `fd_event_register_interface`, whose equivalents stay pure
  virtual.
- **Lifetime documentation corrected.** It claimed the outer handler must outlive the client. It need
  not: the record holds a `std::weak_ptr`, so the destructor skips its unregister step once the
  handler is gone, and both destruction orders are covered by tests. The constraint that does hold is
  now stated instead: while the handler is alive that step removes the client's entry from the
  handler's map, so the client must be destroyed on the handler's thread.

N/A

---

### Stack

This PR is rooted on `main` and is blocked by nothing.

- `main`
  - **#2561 `fix/io-event-handler-truthfulness` (this PR)**

### DIFF
--- a/applications/pionix_chargebridge/src/heartbeat_service.cpp
+++ b/applications/pionix_chargebridge/src/heartbeat_service.cpp
@@ -80,21 +80,19 @@ heartbeat_service::~heartbeat_service() {
 }
 
 bool heartbeat_service::register_events(everest::lib::io::event::fd_event_handler& handler) {
-    // clang-format off
-    return
-        handler.register_event_handler(m_udp.get()) &&
-        handler.register_event_handler(&m_heartbeat_timer, [this](auto&) { handle_heartbeat_timer(); }) &&
-        handler.register_event_handler(&m_error_timer, [this](auto&) { handle_error_timer(); });
-    // clang-format on
+    auto result = true;
+    result = handler.register_event_handler(m_udp.get()) && result;
+    result = handler.register_event_handler(&m_heartbeat_timer, [this](auto&) { handle_heartbeat_timer(); }) && result;
+    result = handler.register_event_handler(&m_error_timer, [this](auto&) { handle_error_timer(); }) && result;
+    return result;
 }
 
 bool heartbeat_service::unregister_events(everest::lib::io::event::fd_event_handler& handler) {
-    // clang-format off
-    return
-        handler.unregister_event_handler(m_udp.get()) &&
-        handler.unregister_event_handler(&m_heartbeat_timer) &&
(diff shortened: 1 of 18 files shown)


djchhp (changes requested):
This PR interface cleanup solves an important ownership issue. Registrations now have a clear owner, teardown removes stale map entries, and registration return values accurately say whether the object is actually monitored.

It also introduces a huge burden on the user, which is the order of members in a client. 
With the changes a client stores a raw pointer to handler. When client is destroyed, it calls through that pointer to unregister itself—but handler has already been destroyed. That is a use-after-free.

Solutions that come to my mind are:
  - have register_event_handler() return an RAII registration token held by the client;
  - make that token refer to shared handler state through weak_ptr, so its destructor becomes a no-op if the handler has already died;
  - or have the handler own connection objects and invalidate them during its own destruction.

djchhp (dismissed):
lgtm

Pietfried (dismissed):

djchhp (approved):
re-approved, still looking good.

Pietfried (approved):

PR #2562, opened 2026-08-05
feat(io): tri-state connection state and bounded tx buffer

mlitre:
## Describe your changes

`generic_error_state` tracked whether a client was up with a single bool, which cannot distinguish
"never connected yet" from "connected and then failed". Those need opposite handling: a payload
written before the first connect resolves is early, the same payload after a failure is
undeliverable. With one bool, both were rejected.

- **`connection_state{fresh, connected, failed}`** replaces the bool. `on_error()` keeps its old
  meaning of "not up", covering `fresh` and `failed`, so existing consumers do not change.
- **`tx()` can buffer during the connect window, but only when asked to.** The bound is
  `max_buffered_tx_payloads`, so an unreachable peer cannot grow the queue without limit, and the
  `bool` return makes a rejected send visible instead of silent. Buffering is **off by default**: a
  policy opts in with `static constexpr bool buffer_tx_before_connect`, and any consumer overrides
  that per instance with a `utilities::tx_buffering` constructor argument. Holding a payload across a
  connect is only correct when its meaning survives the delay, which is false for the periodic
  setpoints and stateful line protocols the in-tree consumers carry: replaying one after a reconnect
  delivers a stale value the peer cannot tell from a fresh one. Deriving that from the transport
  instead would have given every async-connect client replay it never asked for, so the request is
  explicit and the transport trait only decides whether buffering is *available* at all. Policies
  that resolve `open()` synchronously have no pre-connect window and reject until up regardless.
- **A `fresh` client is inert in the error handler.** Introducing a third state made `on_error()`
  true while the stored code was 0, which had never been possible; both chargebridge bridges gate
  their reopen on `if (id not_eq 0)`. Without this, a consumer `reset()` racing a completed I/O left
  the teardown unpaired while the reopen marked the client connected again, so it reported healthy
  forever and accepted into a queue that would never drain.
- **`reset()` applies its state flip at call time.** It sets the state to `fresh` and clears the tx
  buffer synchronously; only the device teardown and reopen stay deferred. This keeps `tx()`'s bool
  truthful across a reset: `true` after `reset()` means buffered for the new connection, not
  discarded with the old one. The clear is unconditional, in both buffering modes.
- **The flush and read paths answer to handle identity, not to connection state.** A user callback
  runs inside the client and can retire the connection mid-dispatch, after which a successful read
  would put the state back to `connected` and let a post-reset payload reach the abandoned peer. Both
  paths now compare a generation stamped when the device is opened, so I/O on a retired handle is
  skipped rather than attributed to its replacement.
- **Error code 0 is an up-edge**, fired exactly once per successful (re)connect, including a first
  connect and a reset that never reported a failure. Documented and pinned; error-only consumers
  filter on a nonzero code. Suppressing the code-0 callback on clean resets would have preserved the
  pre-tri-state observable behavior, but it leaves the callback meaning "an error you may never have
  seen is now gone" and needs its own suppression state machine; the up-edge contract is simpler and
  matches how `can_bridge` already consumes code 0 as its ready signal. Rejecting `tx()` between
  `reset()` and the next `sync()` was also considered and rejected: it breaks the
  reset-inside-error-callback-then-send pattern, which is the natural consumer shape for reconnect
  handling.

Known gap, documented on the error-handler contract rather than fixed here: a read that succeeds
while the client is in `failed` puts the state back to `connected` and arms another code 0, which can
end with the device gone, no reopen, and `tx()` still accepting. It predates this change.

### Consumer-visible changes

Also recorded in `lib/everest/io/README.md` under "Client tx contract", since that outlives the
squash.

1. **`tx()` can now reject on a live connection.** It rejects once `max_buffered_tx_payloads` are
   queued, which a peer that stops reading reaches on its own. This is backpressure, not failure:
   the connection stays up and everything already accepted is still delivered in order. Call sites
   that discard the return turn a stalled peer into silent drops.
2. **`on_error()` is true for the whole reset window.** `reset()` moves the state to `fresh` at call
   time rather than in the queued action, so a consumer polling `on_error()` sees `true` from the
   call until the reopen completes, where it previously stayed `false` until the action ran.
3. **`reset()` is loop-thread only.** It now touches the tx buffer directly, so it carries the same
   threading contract as `tx()`: call it from the thread driving `sync()`, or from a callback that
   thread dispatches. It was previously reachable off-thread via `add_action`.
4. **Pre-connect buffering is opt-in and off by default**, so no existing consumer changes behavior.
   `serial_bridge` keeps dropping during its 1000 ms TCP connect window and the five udp bridges
   keep dropping during theirs, exactly as on `main`. #2565 makes those drops visible in the log
   rather than silent, which was the actual defect. `tls_client_socket` declares the opt-in on
   #2564, keeping the pre-connect buffering `tls_endpoint_base` had before the client moved onto
   `fd_event_client`; without it `TlsE2E.tx_queued_before_handshake_is_flushed_after_ready` fails.

N/A

---

### Stack

This PR is 2 above `main`, so 1 PR (#2561) must merge before it.

- `main`
    - **#2562 `feat/io-client-tx-tristate` (this PR)**

### DIFF
--- a/lib/everest/io/README.md
+++ b/lib/everest/io/README.md
@@ -11,3 +11,37 @@ Currently there are clients for
  - TAP
 
 The clients are single threaded and epoll based. Utilities for file descriptor based event handling are provided and used.
+
+## Client tx contract
+
+`fd_event_client::tx()` returns a `bool` and every call site should check it. It rejects once the
+connection has failed, and once `max_buffered_tx_payloads` are already queued, which a peer that
+stops reading reaches on its own. A rejection on a live connection is backpressure, not a failure:
+the connection stays up and everything already accepted is still delivered in order.
+
+Payloads written before the connection is up are **discarded by default**. Holding one across a
+connect replays it onto a connection it was not written for, which delivers a stale value for the
+periodic setpoints and stateful line protocols most consumers carry.
+
+A consumer whose payload survives the delay asks for the buffer, per policy or per instance:
+
+```cpp
+// Per policy: every client built on it buffers unless told otherwise.
+struct my_socket {
+    static constexpr bool buffer_tx_before_connect{true};
+    // ...
+};
+
+// Per instance, overriding the policy default in either direction.
+tcp::tcp_client client(utilities::tx_buffering::buffer, host, port, timeout_ms);
+```
+
+Only a policy that connects asynchronously has a pre-connect window. Asking a synchronous policy to
+buffer is accepted and does nothing.
+
+`reset()` clears the buffer in both modes, so a `true` from `tx()` after a reset always means
+buffered for the connection the reset opens. `reset()` also moves the connection state at call
+time: `on_error()` reports `true` from the call until the reopen.
+
+`tx()` and `reset()` are loop-thread only: call them from the thread that drives `sync()`, or from
+a callback that thread dispatches.
--- a/lib/everest/io/include/everest/io/event/fd_event_client.hpp
+++ b/lib/everest/io/include/everest/io/event/fd_event_client.hpp
@@ -7,6 +7,7 @@
 
 #include <atomic>
 #include <chrono>
+#include <cstddef>
 #include <cstdint>
 #include <everest/io/event/event_fd.hpp>
 #include <everest/io/event/fd_event_sync_interface.hpp>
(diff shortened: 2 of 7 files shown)


djchhp (commented):
Two things to fix before merge:

    The PR body references docs/adr/0013-io-client-error-callback-up-edge.md, but the file isn't in the branch — no docs/adr/ exists. Add it or drop the reference.
    Please add release notes for the three consumer-visible changes: tx() can now reject while connected (the cap), on_error() is true during a reset window, and reset() is now strictly loop-thread-only.

I see one potential issue with pionix_chargebridge serial bridge. During the 1s connect window messages are queued, rather than dropped. I think that should be fine, but it is a change in behavior.

djchhp (approved):
**feat(io): make pre-connect tx buffering opt-in** is a genuine improvement compared to the last reviewed state. Looking good!

PR #2563, opened 2026-08-05
feat(io): event loop driven handshake phase

mlitre:
## Describe your changes

`generic_fd_event_client` had two states for a connection: not connected, and ready for data. A
protocol that must negotiate before data flows had nowhere to put that negotiation, so it ended up on
its own thread with its own lifetime.

This adds an optional handshake phase between "transport connected" and "ready for data", driven on
the existing event loop. A policy opts in by providing `handshake_step`, `handshake_complete` and
`desired_events`, detected by trait. A policy without them takes exactly the previous path, so
nothing changes for the ten existing clients.

- **The first step runs unsolicited**, as soon as the transport is connected and before any readiness
  event, because the client owes the first protocol message and cannot wait for the descriptor to
  become readable. This is the easiest part of the contract to get wrong. After that the policy is
  stepped only for the single event its `desired_events` asked for.
- **The descriptor is monitored for exactly one of read or write** while a handshake is outstanding,
  never both, because the handshake owns it. Payloads written during that window are queued and
  flushed in order once it completes.
- **A single-shot `timer_fd` bounds the whole handshake**, not each step. Without it,
  `action_status::success` re-armed forever: a peer that completed the TCP connect and then went
  silent left the client in `connected` with `on_error()` reporting false, the ready action never
  fired, `tx()` buffered to its cap and rejected with no diagnostic, and nothing ever called
  `reset()`. Default 10 s; a policy may state its own bound, and a non-positive value falls back to
  the default rather than disarming the timer.
- **A failed arm now fails the connection.** The `set_timeout` result was dropped, so a handshake
  could proceed with no deadline at all, which is the case the timer exists to prevent. Reporting it
  also required an early return, because `maybe_fire_ready()` has no failed-connection guard and
  would otherwise fire ready on the connection the failed arm just killed.
- **A policy carrying `handshake_timeout()` without the trio no longer compiles.** It was silently
  ignored before. `get_error_string` stays independent, and the trio without a timeout still compiles.

Two gaps stated rather than papered over. The expiry path carries a staleness guard so a timer firing
in the same poll batch as the completing event cannot fail a good connection; poll batches have no
defined dispatch order, so the guard is necessary, but it is not reachable from the test harness and
has no test. A failed timer arm is likewise unreachable without fault injection: `timer_fd` takes its
descriptor at construction and exposes no injection seam.

N/A

---

### Stack

This PR is 3 above `main`, so 2 PRs (#2561, #2562) must merge before it.

- `main`
      - **#2563 `feat/io-fd-event-client-handshake-phase` (this PR)**

### DIFF
--- a/lib/everest/io/include/everest/io/event/fd_event_client.hpp
+++ b/lib/everest/io/include/everest/io/event/fd_event_client.hpp
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: Apache-2.0
-// Copyright 2020 - 2025 Pionix GmbH and Contributors to EVerest
+// Copyright 2020 - 2026 Pionix GmbH and Contributors to EVerest
 
 /** \file */
 
@@ -11,6 +11,7 @@
 #include <cstdint>
 #include <everest/io/event/event_fd.hpp>
 #include <everest/io/event/fd_event_sync_interface.hpp>
+#include <everest/io/event/timer_fd.hpp>
 #include <everest/io/event/unique_fd.hpp>
 #include <everest/io/socket/socket.hpp>
 #include <everest/io/utilities/event_client_async_policy.hpp>
@@ -20,11 +21,16 @@
 #include <future>
 #include <memory>
 #include <queue>
+#include <string>
 #include <thread>
 #include <type_traits>
 
 namespace everest::lib::io::event {
 class fd_event_handler;
+
+// Opaque: event/fd_event_handler.hpp defines poll_events and includes this header, so including
+// it back would close a cycle.
+enum class poll_events;
 
 /**
  * @enum action_status
@@ -74,6 +80,38 @@ public:
      * @brief Prototype for error_status
      */
     using error_status = std::function<int()>;
+
+    /**
+     * @var handshake_status
+     * @brief Prototype for the query whether the handshake is still pending
+     */
+    using handshake_status = std::function<bool()>;
+
+    /**
+     * @var handshake_events
+     * @brief Prototype for the query which single event the handshake waits for
+     */
+    using handshake_events = std::function<poll_events()>;
+
+    /**
+     * @var handshake_timeout
+     * @brief Prototype for the query how long the handshake may stay pending
+     */
+    using handshake_timeout = std::function<std::chrono::milliseconds()>;
+
+    /**
+     * @var error_text
+     * @brief Prototype for an error description supplied by the client policy
+     */
+    using error_text = std::function<std::string()>;
+
+    /**
+     * @var default_handshake_timeout
+     * @brief Upper bound of a protocol handshake when the client policy states none.
+     * @details Retransmission with a doubling timeout lets a healthy negotiation over a lossy
+     * link take seconds. Past this the peer stopped talking, and the client reports ETIMEDOUT.
+     */
+    static constexpr std::chrono::milliseconds default_handshake_timeout{10000};
 
     /**
      * @var ready_action
@@ -149,9 +187,9 @@ public:
 
     /**
      * @brief Register a function, that is called once the client is ready
-     * @details This callback is performed immediately, if the client is in a ready state
-     *          or the client is ready. The action is persistent and called whenever the client
-     *          is ready after reset.
+     * @details Registration is deferred to the event loop. Called once per connection, so also
+     *          again after a reset, and immediately when the client is already ready. A client
+     *          policy with a handshake phase counts as ready once the handshake is complete.
      * @param[in] item The callback
      */
     void set_on_ready_action(std::function<void()>&& item);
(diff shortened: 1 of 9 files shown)


lib/everest/io/include/everest/io/event/fd_event_handler.hpp
djchhp, line 1: The staleness guard has a gap: reset() mid-handshake leaves the retired deadline armed. If its expiry lands in the same poll batch as the new connection's connected event and dispatches second (FIFO makes that the likely order),
  start_handshake_deadline() has already recorded the current generation — the disarm() on entry kills the fresh deadline, all three guards pass, and the new connection dies with an unearned ETIMEDOUT. The generation can't close this (the
  expiry event carries no arm identity), but the timerfd can: timerfd_settime clears the tick counter, so a read() finding no expiration marks a stale event. The dispatch wrapper discards that signal. Fix in fd_event_handler.cpp,
  register_event_handler(timer_fd*, ...) (~line 236, outside this diff):

```
  [handler, fd](event_list const& e) {
      // A rearm or disarm since the batch was harvested cleared the tick
      // counter: a read finding no expiration marks a stale event.
      if (fd->read() > 0) {
          handler(e);
      }
  },
```

lib/everest/io/src/event/fd_event_client.cpp
```diff
@@ -1,15 +1,57 @@
...
+    m_get_error(get_error),
+    m_handshake_pending(handshake_pending),
+    m_drive_handshake(drive_handshake),
+    m_handshake_desired_events(handshake_desired_events),
+    m_get_handshake_timeout(get_handshake_timeout),
+    m_get_error_string(get_error_string) {
     m_event_handler = std::make_unique<event::fd_event_handler>();
+    m_handshake_timer.set_single_shot(true);
+    // Registered once and for all: a timerfd notifies only while it is running.
+    m_event_handler->register_event_handler(&m_handshake_timer, [this]() { handshake_deadline_expired(); });
```
djchhp: With the result being ignored every handshake would run unbounded in the case of a failure, which is one thing the deadline is supposed to prevent.

lib/everest/io/include/everest/io/utilities/event_client_async_policy.hpp
djchhp, line 1: Adding this

```cpp
template <typename T>
inline constexpr bool event_client_has_no_handshake_members_v =
    not has_member_handshake_complete_v<T> and
    not has_member_handshake_step_v<T> and
    not has_member_desired_events_v<T>;
```
This helps with the asserts in fd_event_client

lib/everest/io/include/everest/io/event/fd_event_client.hpp
```diff
@@ -303,6 +453,22 @@ class generic_fd_event_client : public ClientInterfacePolicy, public generic_fd_
...
+                  "ClientPolicy must provide all of handshake_complete(), handshake_step() and "
+                  "desired_events(), or none of them");
+
+    // handshake_timeout() bounds the phase the trio implements, so on its own it is read by nothing.
+    static_assert(not utilities::has_member_handshake_timeout_v<ClientPolicy> or
+                      utilities::event_client_handshake_policy_v<ClientPolicy>,
+                  "ClientPolicy provides handshake_timeout() but no handshake to bound: either add "
+                  "handshake_complete(), handshake_step() and desired_events(), or remove "
+                  "handshake_timeout()");
+
```
djchhp:
```suggestion
/ A partially provided handshake trio would silently disable the handshake and release the
  // on_ready action on a socket that never negotiated anything.
  static_assert(utilities::event_client_handshake_policy_v<ClientPolicy> or
                    utilities::event_client_has_no_handshake_members_v<ClientPolicy>,
                "ClientPolicy must provide all of handshake_complete(), handshake_step() and "
                "desired_events(), or none of them");

  // handshake_timeout() bounds the phase the trio implements, so on its own it is read by nothing.
  static_assert(not utilities::has_member_handshake_timeout_v<ClientPolicy> or
                    utilities::event_client_handshake_policy_v<ClientPolicy>,
                "ClientPolicy provides handshake_timeout() but no handshake to bound: either add "
                "handshake_complete(), handshake_step() and desired_events(), or remove "
                "handshake_timeout()");
```

lib/everest/io/include/everest/io/event/fd_event_client.hpp
djchhp, line 1: This PR introduced a new style for the conditional code paths. to keep it consistent we should either go for `enable_if` or `constexpr if ` exclusively.

Going to `constexpr if` would yield a replacement of both  init overloads (fd_event_client.hpp:619-668) 

```
 
  template <class... ArgsT> void init(ArgsT... args) {
      if constexpr (utilities::event_client_async_policy_v<ClientPolicy>) {
          m_open_device = [this, args...]() {
              auto setup_ok = m_handle->setup(args...);
              if (setup_ok) {
                  auto connect_generation = m_connect_generation.load();
                  auto state = m_async_connect_state;
                  auto handle = std::move(m_handle);
                  // The usage of a detached thread is intentional to avoid blocking the event_handler.
                  std::thread([state = std::move(state), handle = std::move(handle), connect_generation]() mutable {
                      bool ok = false;
                      int fd = -1;
                      try {
                          handle->connect([&](bool result_ok, int result_fd) {
                              ok = result_ok;
                              fd = result_fd;
                          });
                      } catch (...) {
                          ok = false;
                          fd = -1;
                      }

                      if (not state or not state->active.load()) {
                          return;
                      }

                      auto inserted = state->connect_results.emplace(
                          async_connect_result{connect_generation, ok, fd, std::move(handle)});
                      if (inserted != 0) {
                          state->ready_event.notify();
                      }
                  }).detach();
              } else {
                  auto generation = m_connect_generation.load();
                  on_client_ready(generation, false, -1);
              }
          };
          // The initial open has no previous peer, so a payload buffered before this action runs
          // belongs to the connection it opens.
          schedule_reset();
      } else {
          m_open_device = [this, args...]() {
              auto result = m_handle->open(args...);
              auto generation = m_connect_generation.load();
              on_client_ready(generation, result, m_handle->get_fd());
          };
          reset_impl();
      }
  }
```
  And the call site in the constructor (fd_event_client.hpp:501) drops the explicit template argument:
```
  init(args...);   // was: init<ClientPolicy>(args...);

```

djchhp (changes requested):

djchhp (dismissed):
All previous comments were addressed. I don't see any further issues.

djchhp (approved):
adaptations limited to tx_buffering changes from previous commit in the stack and documentation.

PR #2564, opened 2026-08-05
feat(io/tls): event loop driven TLS client and server

mlitre:
## Describe your changes

A TLS layer for `lib/everest/io` driven entirely from an existing `fd_event_handler`, so a TLS
connection is polled alongside everything else instead of owning a thread.

- **Client.** `tls_client` is a one-line alias for
  `event::fd_event_client<tls_client_socket>::type`. All behavior comes from `tls_client_socket`, a
  policy satisfying both the async-connect traits and the handshake traits the generic client already
  detects, so user code wires no handshake hooks. The blocking TCP connect reuses the generic
  client's existing short-lived thread; this layer starts none of its own.
- **Server.** `tls_listener` owns the listen socket and an embedded `tls::Server`, yielding a
  `std::unique_ptr<tls_server>` per accepted connection. `tls_server` derives from
  `tls_endpoint_base`, which wires one connection flat onto the handler and drives the accept
  handshake inside its own dispatch callback. The client deliberately does not use
  `tls_endpoint_base`: it nests its own handler and needs a connect, the server is handed a connected
  descriptor.

Points worth a reviewer's attention:

- **Teardown from inside a dispatch callback.** `fail()` cannot remove its own descriptor
  synchronously, since that destroys the `std::function` currently executing. Removal and close are
  deferred together, ordered remove then close so the number stays reserved until it is out of the
  handler; a synchronous close would let an accept in the same batch recycle it and the stale remove
  strand it. The deferred action captures the descriptor and a moved closer, never `this`.
- **Steady-state events are routed by direction.** The post-handshake dispatch ignored
  `desired_events()`, so `SSL_write` returning `WANT_READ` busy-spun POLLOUT and `SSL_read` returning
  `WANT_WRITE` stalled. Routing is now per direction rather than through the exclusive `monitor_for`,
  which would have removed POLLOUT on an ordinary read would-block and stranded the queue. POLLIN
  stays monitored while a send drains: dropping it there went deaf for the whole drain and deadlocked
  a peer that writes before it reads, with no error to notice it by.
- **The server tx queue is bounded** at `max_buffered_tx_payloads`, matching the client's name and
  value. Rejection is backpressure, not failure: the connection stays live and everything already
  accepted is still delivered.
- **`tls::Client::init` failures are recorded.** `setup()` returned false leaving the error at 0, so
  `get_error()` probed an unassigned descriptor and answered `EBADF` with no text. It now reports
  `EINVAL` and names the rejected configuration, since `init` does not say which input it refused.

Known limitation, filed rather than hidden: `tls_socket_base` has a single desired-event slot shared
by both directions, so it cannot express "rx wants write while tx wants read". A would-block is
routed correctly but the blocked operation is not re-driven, which needs per-direction state and a
dispatch that drives it. Reaching that state requires a TLS 1.3 KeyUpdate or a TLS 1.2 renegotiation.

Separately, `tls::Client::init` tolerates an unreadable `verify_locations_file`: it logs and continues
without failing, so a typo'd trust anchor yields a client with an empty trust store whose later
handshakes fail for a reason pointing nowhere near the typo. That is in `lib/everest/tls`, outside
this PR.

### Decoupling the public headers from libtls

Based on a patch from Jan Christoph Hoppe. The layer named `::tls::Connection::result_t` in its own
API; that type is nested, so it cannot be forward declared, and naming it chained libtls and the
OpenSSL headers into anything including these headers.

- `io_result` replaces it, translated in one function. `to_io_result` lists every libtls enumerator
  with no `default:`, so `-Wswitch` breaks the build if libtls adds one.
- The two class templates keep their contract in the public header and their member definitions in
  `*_impl.hpp`, with the library's policies instantiated once each in a `.cpp`. A consumer adding a
  policy of its own includes the impl header instead.
- `tls_client_socket` and `tls_listener` each held a libtls type by value, which is what blocked a
  forward declaration. `Config` is now an incomplete nested type defined in a sibling config header,
  `tls_listener` holds its `Server` by pointer, and the client keeps only the SNI host it reads
  after `setup()`.
- A guard translation unit compiles all eight public headers with libtls off the include path, so
  re-adding an include fails the build and names the header. Verified by injecting one.

The win is in-tree layering and compile time, not installability: a consumer still needs a config
header to construct a client or listener, and those include libtls, so `PATTERN "tls" EXCLUDE`
stays. Making these installable needs libtls exported, which is out of scope here.

Separately, `tls_client_socket` now declares `buffer_tx_before_connect`. #2562 made pre-connect tx
buffering opt in and off by default; a TLS session carries a stream rather than setpoints, and
`tls_endpoint_base` buffered that window before the client moved onto `fd_event_client`. Without the
opt in, `TlsE2E.tx_queued_before_handshake_is_flushed_after_ready` and
`TlsE2E.client_teardown_closes_the_tls_session` fail.

N/A

---

### Stack

This PR is 4 above `main`, so 3 PRs (#2561, #2562, #2563) must merge before it.

- `main`
        - **#2564 `feat/io-tls-policies-rebuilt` (this PR)**

### DIFF
--- a/lib/everest/io/BUILD.bazel
+++ b/lib/everest/io/BUILD.bazel
@@ -43,6 +43,7 @@ cc_library(
     }),
     visibility = ["//visibility:public"],
     deps = [
+        "//lib/everest/tls",
         "//lib/everest/util",
         "@mosquitto",
     ],
--- a/lib/everest/io/README.md
+++ b/lib/everest/io/README.md
@@ -9,8 +9,9 @@ Currently there are clients for
  - PTY
  - TCP
  - TAP
+ - TLS (enabled via `EVEREST_IO_ENABLE_TLS=ON`, default ON)
 
-The clients are single threaded and epoll based. Utilities for file descriptor based event handling are provided and used.
+The clients are driven from a single event loop and are epoll based; the async connect path uses a short-lived detached thread for the blocking connect. Utilities for file descriptor based event handling are provided and used.
 
 ## Client tx contract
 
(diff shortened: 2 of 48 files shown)


djchhp (commented):
tls_endpoint_base::fail() invokes the consumer's error handler synchronously from inside the connection-fd dispatch lambda. If the consumer drops the owning unique_ptr there  fail() resumes on the freed object. heap-use-after-free

The fix is small: don't call the error callback immediately. Queue it as an action instead, so it runs after the connection object has finished its own teardown. Then deleting the connection inside the callback is safe.

djchhp (changes requested):
fail() still invokes m_error synchronously and keeps using the object. It reads m_fd/m_handler and may call m_socket.release_closer() after the callback returns, and the dispatch lambda still reads m_errored  after flush_rx. 
As a consequence dopping the tls_server in the reset callback is not save, but the likely thing that is done. Running the tests with ASAN shows this clearly.

This is a potential fix:
[use_after_free_fix_2564.patch](https://github.com/user-attachments/files/31210670/use_after_free_fix_2564.patch)

djchhp (approved):
looking good.

Pietfried (approved):

PR #2421, opened 2026-07-07
Feat/pionix chargebridge new features

corneliusclaussen:
## Describe your changes

feat(pionix_chargebridge): status TUI, IO bridge, mDNS/IPv6 discovery and FW 0.7.3

Consolidates the ChargeBridge feature work on top of main:

* Status TUI (ftxui) with a print sink that routes all runtime
  diagnostics through it, TTY detection, ANSI suppression when stdout is
  not a terminal, and clamped panel/instance-list layout.
* New io_bridge replaces gpio_bridge, plus telemetry status publishing.
* mDNS discovery gains IPv6 support and endpoint configuration; rdata
  reads are bounds-checked and the A record is dropped for v6-only
  services. Adds mdns unit tests.
* Hardened UDP handling: replies matched to requests by expected type,
  bounded discard loops, tx failure reported on closed sockets, and
  abort-aware firmware-update probes and retries.
* Firmware 0.7.3 image and a --update_only mode that works with
  discovery configs and reports accurate results.
* Consistent EVerest reconnect behaviour: connection status, CP state
  and latched safety errors are replayed for OVM, EV and late-activated
  bridges.
* Config cleanups: shippable CB-EVAL defaults, PLC MTU lowered to 1438,
  leaked IPs/hostnames removed from example configs.

## Checklist before requesting a review
- [ ] I have performed a self-review of my code
- [ ] I have made corresponding changes to the documentation
- [ ] I read the [contribution documentation](https://everest.github.io/nightly/project/contributing.html) and made sure that my changes meet its requirements

PR #2565, opened 2026-08-05
refactor(chargebridge): log rejected serial tx

mlitre:
## Describe your changes

`serial_bridge`'s pty data handler discarded what `tx()` returned, so serial data was dropped with
nothing in the log. It now checks the return and logs on the transition into the rejected state, so a
streak of drops reports once rather than per payload.

The window this covers widened with #2562: pre-connect tx buffering is opt in there and off by
default, and `serial_bridge` does not opt in, so `tx()` rejects for the whole connect window rather
than only once a buffer fills. The latch is indifferent to why a payload was rejected, so it reports
that window the same way, once per streak.

The rejected-state latch is also cleared when the tcp client is recreated, so a streak recorded
against a destroyed client cannot suppress the first rejection log of its replacement. The reset sits
in `create_tcp_client()`, the sole construction site, rather than beside the sibling
`m_tcp_last_error_id` reset in the teardown path, so it does not depend on teardown always being
paired with a recreate.

### Scope reduced on rebase

This PR previously also removed the `m_udp_on_error` tx gates from `gpio_bridge` and
`heartbeat_service` and drove their client reset from the heartbeat tick, on the grounds that their
`handle_error_timer` was never registered with an event handler. That premise no longer holds on
current `main`: #2421 registers `m_error_timer` and resets the client from it, and `m_udp_on_error`
now also feeds the readiness signal and the endpoint reconnect path, so removing it would break live
behavior. `gpio_bridge` no longer exists. Both halves were dropped rather than reapplied.

N/A

---

### Stack

This PR is 3 above `main`, so 2 PRs (#2561, #2562) must merge before it.

- `main`
      - **#2565 `refactor/chargebridge-drop-hand-rolled-tx-gates` (this PR)**

### DIFF
--- a/applications/pionix_chargebridge/include/charge_bridge/serial_bridge.hpp
+++ b/applications/pionix_chargebridge/include/charge_bridge/serial_bridge.hpp
@@ -41,6 +41,7 @@ private:
     utilities::symlink m_symlink;
     std::string m_identifier;
     int m_tcp_last_error_id = -1;
+    bool m_tcp_tx_rejected{false};
     std::uint16_t m_tcp_port{0};
     std::string m_tcp_remote;
     bool m_tcp_ready{false};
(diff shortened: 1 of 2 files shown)


djchhp (dismissed):
looking good and completes the 2561-2564 stack. Merge after that stack is merged.

djchhp (approved):
looking good

PR #2566, opened 2026-08-05
fix(EvseV2G): log responses with their own length in the V2G message log

biubiuzhang:
## Describe your changes

With `debug_mode` enabled, `publish_var_V2G_Message()` sizes the hex/base64 dump of every logged frame with `conn->payload_len + V2GTP_HEADER_LENGTH` — for requests *and* responses. `conn->payload_len` is only written by `V2GTP_ReadHeader()` when a request arrives; the response is encoded into the same buffer and its length lives in the EXI stream position, which only `v2g_outgoing_v2gtp()` reads. The response was also published *before* `v2g_outgoing_v2gtp()` writes the V2GTP header, so the logged frame still carried the request's stale header bytes.

Every logged response is therefore truncated (or padded) to its **request's** length:

| message | req on wire | res on wire | res as logged |
|---|---|---|---|
| ServiceDiscovery | 13 B | 19 B | 13 B — undecodable, `FreeService` flag cut off |
| PaymentServiceSelection | 16 B | 16 B | 16 B — decodes by coincidence |
| Authorization | 13 B | 13 B | 13 B — decodes by coincidence |

Found while verifying #2527 on hardware: the session log's `ServiceDiscoveryRes` hex could not be EXI-decoded, while tcpdump on the same session showed a well-formed 19-byte frame. Wire traffic is unaffected — `v2g_outgoing_v2gtp()` computes its own length — only the published `V2gMessages` debug var is wrong.

### Fix

- `publish_var_V2G_Message()` sizes responses from `exi_bitstream_get_length(&conn->stream)`; requests keep using `payload_len`.
- The two response call sites publish after `v2g_outgoing_v2gtp()` succeeds, so the log contains exactly the frame that was sent (header included). Responses that fail to send are no longer logged as if they had been.

None — found during review/verification of #2527. Can open one if preferred.

- [x] I have made corresponding changes to the documentation (none needed — debug logging only)

### Verification

Reproduced and verified on a DC charger (AM62x) running SIL sessions against PyEvJosev over ISO 15118-2: before the fix, logged `ServiceDiscoveryRes` was 13 bytes with a stale header (undecodable); after the fix the logged frame matches the pcap byte-for-byte and decodes with libcbv2g.

### DIFF
--- a/modules/EVSE/EvseV2G/v2g_server.cpp
+++ b/modules/EVSE/EvseV2G/v2g_server.cpp
@@ -271,9 +271,14 @@ static types::iso15118::V2gMessageId get_v2g_message_id(enum V2gMsgTypeId v2g_ms
 static void publish_var_V2G_Message(v2g_connection* conn, bool is_req) {
     types::iso15118::V2gMessages v2g_message;
 
+    /* conn->payload_len only holds the length of the received request; responses are
+     * encoded into the same buffer and their length lives in the EXI stream position */
+    const size_t msg_len =
+        is_req ? static_cast<size_t>(conn->payload_len) + V2GTP_HEADER_LENGTH : exi_bitstream_get_length(&conn->stream);
+
     u_int8_t* tempbuff = conn->buffer;
     std::string msg_as_hex_string;
-    for (int i = 0; ((tempbuff != NULL) && (i < conn->payload_len + V2GTP_HEADER_LENGTH)); i++) {
+    for (size_t i = 0; ((tempbuff != NULL) && (i < msg_len)); i++) {
         char hex[4];
         snprintf(hex, 4, "%x", *tempbuff); // to hex
         if (std::string(hex).size() == 1)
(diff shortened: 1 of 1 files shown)


SebaLukas (approved):
Thank you very much for detecting and fixing that!
Looking good 👍

barsnick (commented):
I can confirm the original issue.

Not only are the Res message logs truncated, the headers are also broken: Leading `00fe` instead of `01fe`, and using the incorrect length from the Req.

I haven't gotten around to testing the fix. If someone else has, and can confirm that the above mentioned fields are now okay, this fix LGTM.

biubiuzhang:
@barsnick I ran it. SIL DC ISO 15118-2 session on this branch with `debug_mode` on and TLS off, capturing the published `v2g_messages` var.

Every logged frame now carries its own V2GTP header and its own length. Over a full session — 58 frames, 56 ISO-2 plus the two SAP ones — the header is `01fe`, payload type `8001`, and the declared length equals the actual payload length in every single one. No `00fe`, no truncation.

Your example, where the Req is 13 bytes and the Res is 19:

```
ServiceDiscoveryReq   01fe 8001  declared=13  actual=13  total=21
ServiceDiscoveryRes   01fe 8001  declared=19  actual=19  total=27
```

And decoding the logged Res payload with libcbv2g's ISO-2 decoder consumes all 19 bytes and gives the complete message, including the `FreeService` flag that used to be dropped with the tail:

```
decode rv=0 consumed=19 of 19
ServiceDiscoveryRes: ResponseCode=0 ChargeService.FreeService=0 ServiceID=1
                     Category=0 PaymentOptionList.count=1
  EnergyTransferMode[0]=3
```

All 56 ISO-2 frames decode with `rv=0` and consume exactly their declared length.

hikinggrass (approved):

barsnick (approved):

PR #633, opened 2024-04-16
ci: Split up in multiple jobs

andistorm:
## Checklist before requesting a review
- [ ] I have performed a self-review of my code
- [ ] I have made corresponding changes to the documentation
- [ ] I read the [contribution documentation](https://github.com/EVerest/EVerest/blob/main/CONTRIBUTING.md) and made sure that my changes meet its requirements

PR #2568, opened 2026-08-05
fix(ocpp): queue NotifyEvent until registration is accepted

biubiuzhang:
## Describe your changes

`NotifyEvent` requests dispatched before the CSMS has accepted the `BootNotification` are currently silently discarded: `get_message_transmission_priority()` returns `Discard` for non-transaction messages pre-registration unless `QueueAllMessages` is set (it is unset by default).

This permanently loses monitoring/fault events that are generated while the station is not yet registered. The most visible case is a hardware fault that is already active at boot: the OCPP201 module queues errors raised before start and drains that queue right after `connect_websocket()` — before `BootNotificationResponse(Accepted)` — so the resulting `NotifyEvent` calls are dropped. The CSMS only ever sees `StatusNotification(Faulted)` with no cause. A charging station that boots faulted never tells the CSMS why.

Bench evidence (AM62x-based DC charger, DC power stage unpowered at boot): `power_supply_DC/CommunicationFault` raised ~11 s before OCPP start; journal shows the queued error event being processed; zero `NotifyEvent` on the wire.

Note that OCPP 1.6 does not suffer the equivalent information loss: the connector state machine keeps the latest error and `handleBootNotificationResponse(Accepted)` re-sends `StatusNotification` including `errorCode`/`vendorErrorCode`. In OCPP 2.x the fault detail travels only in the fire-and-forget `NotifyEvent`, so discarding it loses the information permanently.

Discarding is allowed by the specification, so this is an information-loss / robustness gap rather than a spec violation. The precedent for "this message type must not be discarded before registration" is `SecurityEventNotification` (#633).

### Change

- `get_message_transmission_priority()` gains an optional `queue_until_accepted` flag mapping to `SendAfterRegistrationStatusAccepted`.
- New `OCPPCommCtrlr` variable `QueueNotifyEventMessages` (boolean, `ReadWrite`, default `false`). When enabled, the v2 `MessageDispatcher::dispatch_call()` sets the flag for `MessageType::NotifyEvent`, so pre-registration NotifyEvents are held in the message queue (`push_call(call, true)`) and flushed once registration is accepted, instead of being discarded. Post-registration behaviour is unchanged (`SendImmediately`).
- v1.6 behaviour is unchanged.

`QueueAllMessages` + `MessageTypesDiscardForQueueing` cannot express this: `MessageTypesDiscardForQueueing` is only evaluated in `MessageQueueConfig::check_queue()` (offline queueing and DB persistence), not in `get_message_transmission_priority()`, so it cannot narrow pre-registration queueing down to a single message type.

Related: EVerest/libocpp#758 (FirmwareStatusNotification dropped pre-registration, same mechanism).

N/A — found during OCPP 2.0.1 fault-reporting verification on our hardware.

### DIFF
--- a/lib/everest/ocpp/config/common/component_config/standardized/OCPPCommCtrlr.json
+++ b/lib/everest/ocpp/config/common/component_config/standardized/OCPPCommCtrlr.json
@@ -226,6 +226,22 @@
       "description": "Comma seperated list of message types that shall not be queued (when offline) even in case QueueAllMessages is true. If QueueAllMessages is false, the configuration of this paramater has no effect.",
       "type": "string"
     },
+    "QueueNotifyEventMessages": {
+      "variable_name": "QueueNotifyEventMessages",
+      "characteristics": {
+        "supportsMonitoring": true,
+        "dataType": "boolean"
+      },
+      "attributes": [
+        {
+          "type": "Actual",
+          "mutability": "ReadWrite",
+          "value": false
+        }
+      ],
+      "description": "When this variable is set to true, NotifyEventRequest messages generated before the CSMS accepted the BootNotification are queued and sent once registration is accepted, instead of being discarded.",
+      "type": "boolean"
+    },
     "ResetRetries": {
       "variable_name": "ResetRetries",
       "characteristics": {
--- a/lib/everest/ocpp/include/ocpp/common/message_queue.hpp
+++ b/lib/everest/ocpp/include/ocpp/common/message_queue.hpp
@@ -99,12 +99,13 @@ class MalformedRpcMessage : public std::runtime_error {
 inline MessageTransmissionPriority get_message_transmission_priority(bool is_boot_notification_message, bool triggered,
                                                                      bool registration_already_accepted,
                                                                      bool is_transaction_related,
-                                                                     bool queue_all_message) {
+                                                                     bool queue_all_message,
+                                                                     bool queue_until_accepted = false) {
     if (registration_already_accepted || is_boot_notification_message || triggered) {
         return MessageTransmissionPriority::SendImmediately;
     }
 
-    if (is_transaction_related || queue_all_message) {
+    if (is_transaction_related || queue_all_message || queue_until_accepted) {
         return MessageTransmissionPriority::SendAfterRegistrationStatusAccepted;
     }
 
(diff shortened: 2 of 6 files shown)


Pietfried (commented):
N07.FR.04 is actually more specific than the claimed in this PR. It is related to triggered monitoring events only, not to all offline NotifyEvents . 

Image

I understand the requirement to report error events though.

What you would like to achieve is already possible if you use the configuration option `QueueAllMessages` in combination with `MessageTypesDiscardForQueueing`, which is arguably a little awkward to configure, because you would need to list all the message types except `NotifyEvent` inside `MessageTypesDiscardForQueueing`.

Instead of queuing all NotifyEvents by default, we could introduce an explicit configuration option like `QueueNotifyEventMessages` . What do you think @biubiuzhang and @mlitre  ?

biubiuzhang:
Fair point on N07.FR.04 — I removed that claim from the description. The motivation is the permanent information loss (1.6 recovers the fault detail via `StatusNotification` on `BootNotification(Accepted)`, 2.x cannot), plus the `SecurityEventNotification` precedent (#633).

One detail on the existing workaround: `MessageTypesDiscardForQueueing` is only evaluated in `MessageQueueConfig::check_queue()` (offline queueing and DB persistence), not in `get_message_transmission_priority()`. So with `QueueAllMessages = true` *every* non-transaction message is stalled until registration is accepted, and the list cannot narrow that down to `NotifyEvent`.

So I went with your suggestion: pushed `QueueNotifyEventMessages` (`OCPPCommCtrlr`, boolean, default `false`) — the hardcoded `NotifyEvent` check is now gated by it. Happy to make it a message type list (`MessageTypesQueueUntilAccepted`) instead if you'd prefer one variable that also covers EVerest/libocpp#758. @mlitre

Pietfried:
This still misses your own scenario: OCPP201/OCPPmulti drains errors right after the async connect_websocket() returns so before on_websocket_connected() un-pauses the queue. While paused, push_call() drops non-transaction messages that don't pass check_queue() and the stall_until_accepted flag is never checked there, so the NotifyEvent is still lost, just one layer deeper.

Suggested fix in push_call():

if (!this->paused || this->resuming || control_message->stall_until_accepted ||
    this->config.check_queue(control_message->messageType) ||
    control_message->messageType == M::BootNotification) {

 A test covering "pushed while paused -> resumed -> accepted -> sent" would be nice to pin this down

Pietfried (changes requested):

PR #2571, opened 2026-08-06
feat(ocpp): report SuspendedEVSE reason changes over OCPP 1.6

mlitre:
# feat(ocpp): report SuspendedEVSE reason changes over OCPP 1.6

## Problem

While a connector sits in `SuspendedEVSE`, the reason can change without the status
changing: a user pause during an energy shortage, or a fault raised on an already
suspended connector. `EvseManager` republishes the reason set, but the OCPP layer
discarded it, so the CSMS kept showing a stale explanation.

## What changes

The reason set is reported in the free-text `info` field of a further
`StatusNotification.req`, with `status` unchanged at `SuspendedEVSE`. Off by default
behind a new device model variable:

| Component | Variable | Type | Mutability | Default |
|---|---|---|---|---|
| `InternalCtrlr` | `ReportSuspendedEVSEReasonChange` | boolean | ReadWrite | `false` |

It doubles as an OCPP 1.6 configuration key of the same name, served by both v16
backends, and is read per suspend event so `ChangeConfiguration` applies without a
restart.

Reasons come from `PauseChargingEVSEReasonEnum` (`UserPause`, `Error`, `NoEnergy`),
sorted alphabetically, deduplicated and comma joined. Worst case
`Error,NoEnergy,UserPause`, 24 of the 50 characters allowed.
`TriggerMessage(StatusNotification)` reports the current reason when no error owns the
`info` field.

## Default-off behavior is unchanged

The main design constraint. With the variable `false` the wire matches the base commit
on every path: a pause on an already suspended connector is discarded before the state
machine (previously it was submitted and dropped for want of a self-transition); a
genuine transition forwards `info` verbatim, so the pre-existing `SwitchingPhases`
literal still reaches the wire; `TriggerMessage` and the boot/reconnect burst report
error state only. `modules/EVSE/OCPP/` has a zero diff against base.

The gate is split deliberately: `OCPPmulti` decides whether a reason exists, since it
owns the wire format; `ChargePointImpl` decides whether a repeat reaches the state
machine and whether `TriggerMessage` may fall back, since both need state the module
cannot see.

## Testing

`libocpp_unit_tests` 1433 run / 1425 passed / 8 pre-existing skips / 0 failed;
`ctest -R OCPPmulti` 1/1; build clean.

New coverage: state machine transitions (changed reason emits, unchanged suppressed,
re-entry after `Charging` emits, no emit while faulted), wire-level tests over a real
`ChargePointImpl` for both variable states and the `TriggerMessage` fallback, key-name
dispatch on both config backends, and encoder cases for empty, single, sorted,
deduplicated and length-bounded input.

Not covered automatically: enabling the variable at runtime against a live CSMS.

## Known limitations

- All energy-related causes collapse into `NoEnergy`, so schedule limits, load balancing
  and missing PV surplus are indistinguishable.
- `MinimumStatusDuration` replaces rather than queues, so a reason flapping inside that
  window collapses into one duplicate message.
- `SwitchingPhases` shares the `info` channel and alternates with the reason string.
- When a fault clears while paused, the notification carries the cleared error's `info`;
  a `TriggerMessage` recovers the reason.
- Turning the variable off at runtime does not clear a reason already reported.

## Scope

OCPP 1.6 only. No OCPP 2.x code path reads the variable; the only v2 files touched are
`ctrlr_component_variables.{hpp,cpp}` (+7 lines) for the `ComponentVariable` that
`MAPPING_INTERNAL` needs to map the 1.6 key onto the device model. Under OCPP 2.x a
suspend reason belongs in a `NotifyEvent.req` against a dedicated component-variable.

### DIFF
--- a/modules/EVSE/OCPPmulti/docs/index.rst
+++ b/modules/EVSE/OCPPmulti/docs/index.rst
@@ -520,6 +520,9 @@ OCPP 1.6 limitations: individual errors cannot be cleared selectively via **Stat
 from MREC, simultaneous errors are reported in one message per error instead of a single semicolon-separated
 message, because of the field length limits.
 
+The ``info`` field has a second, unrelated use: reporting why a connector is suspended. See
+:ref:`suspend reason reporting <handwritten_ocppmulti_suspend-reason-reporting>` below.
+
 OCPP 2.x
 ^^^^^^^^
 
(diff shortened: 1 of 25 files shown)


modules/EVSE/OCPPmulti/v16_chargepoint.cpp
```diff
@@ -78,6 +88,19 @@ inline std::int32_t ocpp_connector_id(const ocpp_multi::GenericChargePointInterf
     return mapping.at(evse_id).at(everest_connector_id);
 }
 
+// wire format owned by the OCPP layer; no default label so -Wswitch flags new enum values
+std::optional<std::string_view> to_pause_reason_string(types::evse_manager::PauseChargingEVSEReasonEnum reason) {
```
Pietfried, line 92: the autogeneration already provides pause_charging_evsereason_enum_to_string so this definition is redundant

Pietfried (dismissed):
lgtm except one small comment in line

modules/EVSE/OCPPmulti/v16_chargepoint.cpp
mlitre: Fixed, let me know if it is ok and then I'll rebase resolve conflicts and we are good to go

Pietfried (dismissed):

mlitre:
@Pietfried I need your approval again since I had to resolve conflicts when rebasing / squashing

modules/EVSE/OCPPmulti/docs/index.rst
```diff
@@ -534,6 +537,100 @@ status **Faulted** for the Inoperative case above). All other errors are reporte
 The variable is constantly set to **Problem** for now; a more fine-grained mapping of errors to component-variable
 combinations may be added in the future.
 
+.. _handwritten_ocppmulti_suspend-reason-reporting:
```
Pietfried, line 540: This is all way too (implementation) detailed compared to the detail of other parts of this module documentation and compared to how other configuration parameters are described. I would cut this part down to its essentials and what integrators need to know. The module documentation should be helpful and shouldnt be too overwhelming for them and leak implementation details.

mlitre: Done

Pietfried (approved):

PR #2572, opened 2026-08-06
docs: add AGENTS.md agent guidelines

mlitre:
## Describe your changes

Adds `AGENTS.md` with build, test, code generation and style guidance for AI coding
agents, a 17-line `CLAUDE.md` that imports it, and one `.gitignore` line. Docs only.

Agents reliably get project conventions wrong: `camelCase` where the C++ guidelines
require `snake_case`, hand-edited `struct Conf` instead of `ev-cli` regeneration, command
calls in `init()` rather than `ready()`. All of it is already decided in the tree, none
of it anywhere an agent reads. `contributing.rst` already states a position on
AI-generated contributions; this is its practical companion.

Two files because Claude Code reads `CLAUDE.md`, not `AGENTS.md`. `AGENTS.md` holds the
content and stays vendor-neutral; the adapter just imports it. A symlink also works but
needs Administrator or Developer Mode on Windows. `CLAUDE.local.md` is gitignored so
personal instructions need no committed reference, which is the only `.gitignore` change.

Policy stays in `contributing.rst`; `AGENTS.md` inlines only the four items agents most
often violate and links out for the rest. No `.claude/` directory or editor config is
committed.

Claims were verified against a clean `main` checkout and anything untraceable was dropped
rather than published. That corrected eight errors, including inverted naming
conventions, `ev-cli if update` and `ty update` (the real actions are
`generate-headers`), and a coverage command that needs `-DEVEREST_ENABLE_COVERAGE=ON`.
Verification was static plus cheap commands; the documented cmake, ninja and pytest runs
were not executed end to end.

Follow-up, deliberately not in this PR: large monorepos commonly split agent guidance
into nested `AGENTS.md` files placed next to the code they describe (Grafana and Sentry
both do, for example `tests/AGENTS.md` or `pkg/storage/unified/AGENTS.md`). Agents read
the nearest file in the directory tree, so subsystem detail such as the ev-cli traps or
per-library quirks could later move into `lib/everest/<lib>/` and `tests/` files, keeping
the root file short. This PR starts with a single root file.

Open questions: is `AGENTS.md` the filename you want, should it live at the root or under
`docs/`, and is the mechanics/policy split drawn where you would draw it?

None. Documentation-only addition.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -121,14 +121,16 @@ requires in `manifest.yaml`, is wired to other modules by a configuration YAML u
 `config/`, and communicates only over MQTT through `everest-framework`. Contracts are
 the source of truth for wiring: read `interfaces/*.yaml`, `types/*.yaml` and
 `errors/*.yaml` before the C++. Lifecycle is `init()` then `ready()`, implemented per
-provided interface in `*Impl.cpp`: subscribe in `init()`; call commands in `ready()` or
-later. Generated headers land under `build/generated/`.
+provided interface in `*Impl.cpp`: subscribe in `init()`; call commands and publish in
+`ready()` or later. Generated headers land under `build/generated/`.
 
 ```cpp
 // modules/EVSE/EvseManager
-r_bsp->subscribe_event([this](auto ev) { ... });  // subscribe: init()
-p_evse->publish_ready(true);                      // publish on a provided interface
-r_bsp->call_enable(true);                         // commands: ready() or later only
+r_bsp->subscribe_event([this](auto ev) {   // subscribe: init()
+    p_evse->publish_ready(true);           // publishing from the callback: fine
+});
+p_evse->publish_ready(true);               // at init() top level: never
+r_bsp->call_enable(true);                  // commands: ready() or later only
 ```
 
 In a configuration YAML, a `connections` entry maps a requirement ID from the consuming
@@ -211,6 +213,8 @@ Full C++ conventions: `docs/source/how-to-guides/c++-coding-guidelines.rst`.
 - Never use `std::cout` or `fprintf` for logging.
 - Never call `mod->r_*->call_*()` inside `init()`. Command calls belong in `ready()` or
   later.
+- Never `publish_*()` or raise an error inside `init()` either. Neither is retained, so
+  subscribers still in their own `init()` miss it.
 - Never skip formatting before proposing C++ changes. CI fails on it, see Code style.
 
 ## Contributing essentials
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -23,8 +23,15 @@ behavior, docs describe intent. Say so when you find the two diverging.
 Docs are rendered at <https://everest.github.io>, starting with the
 [Getting Started Guides](https://everest.github.io/nightly/how-to-guides/getting-started/index.html).
 Sphinx sources are under `docs/source/`: `explanation/` for concepts, `how-to-guides/`
-for recipes. Per-module docs are at `modules/<Category>/<Module>/docs/index.rst`; update
-them when a module's public behavior changes.
+for recipes. Per-module docs are at `modules/<Category>/<Module>/docs/index.rst`.
+
+Docs are updated in the same change as the code they describe. After changing a
+module's public behavior (interface or type YAML, config keys, state machines, error
+semantics), check the module's `docs/index.rst` and any `docs/source/` page describing
+that behavior, and fix what went stale. Libraries document themselves in
+`lib/everest/<lib>/README.md` and their own doc trees (`docs/` usually, `doc/` in
+libocpp); internal library changes count too when they touch documented behavior. A
+change to build, run, test or codegen mechanics means updating this file as well.
 
 ## Build
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -83,9 +83,9 @@ cd build && ctest -R <regex>
 ```
 
 `ctest -R` matches **registered test names, not build targets**, and the two often differ
-in case and spelling. List with `ctest -N` first. Coverage targets exist only under
-`-DEVEREST_ENABLE_COVERAGE=ON`; build `everest-core_create_coverage` and the HTML report
-lands at `build/everest-core_create_coverage/index.html`.
+in case and spelling. List with `ctest -N` first. Coverage needs
+`-DEVEREST_ENABLE_COVERAGE=ON`; the `everest-core_create_coverage` target writes
+`build/everest-core_create_coverage/index.html`.
 
 Integration tests use pytest; the virtualenv is `build/venv`:
 
@@ -155,13 +155,13 @@ Traps:
 - Regeneration is decided by mtime, not content. A target newer than its manifest is
   skipped with `Skipping <name> (up-to-date)`, so a recent edit makes ev-cli silently
   under-generate. Preview with `-d`, force with `-f`, and confirm the change landed.
-- `--disable-clang-format` is in the command above because formatting through ev-cli runs
-  whatever clang-format is on `PATH`, not the version CI uses. Format afterwards, see Code
-  style. `--clang-format-file`, if you do use it, takes a directory, not a file.
-- Editing `types/*.yaml` or `interfaces/*.yaml` fails the ctest
-  `everest-core_API_serialize_tests` until the sha256 pins in
-  `lib/everest/everest_api_types/tests/expected_types_file_hashes.csv` and
-  `expected_interfaces_file_hashes.csv` are updated.
+- `--disable-clang-format` (used above): ev-cli formats with whatever clang-format is on
+  `PATH`, not the version CI uses. Format afterwards (see Code style).
+  `--clang-format-file` takes a directory, not a file.
+- Editing `types/*.yaml` or `interfaces/*.yaml` fails `everest-core_API_serialize_tests`
+  until the sha256 pins in
+  `lib/everest/everest_api_types/tests/expected_{types,interfaces}_file_hashes.csv`
+  are updated.
 
 `ev-cli --help` lists the available actions; `--only which` lists the files an action
 would touch.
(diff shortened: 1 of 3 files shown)


AGENTS.md
```diff
@@ -0,0 +1,205 @@
...
+
+Contributor policy (licensing, DCO, review, and the project's position on AI-generated
+contributions) is in `docs/source/project/contributing.rst`. This file covers mechanics
+only: build, run, test, code generation, and invariants.
+
+## Documentation
+
+Prefer project docs over inference, and cite them. Rendered at
+<https://everest.github.io>, starting with the
+[Getting Started Guides](https://everest.github.io/nightly/how-to-guides/getting-started/index.html).
```
lukas-mertens: This sounds dangerous if docs are outdated? Maybe instead something along the lines of "First look at docs to save on tokens"? Otherwise this could cause problems on feature branches where e.g. config options between the feature branch and the docs diverge

AGENTS.md
```diff
@@ -0,0 +1,205 @@
+# EVerest Core: Agent Guidelines
+
+Guidance for AI coding agents working in `everest-core`: runtime manager, modules,
```
lukas-mertens: The models currently still often don't know that everest-core was merged into a monorepo yet. Wouldn't it make sense to name the repo here the way it is named right now and mention that everest-core was merged into a monorepo? Otherwise I noticed my agents often forget that.

AGENTS.md
```diff
@@ -0,0 +1,205 @@
...
+
+- Sign off every commit (`Signed-off-by`, DCO), enforced by
+  `.github/workflows/job_dco-check.yaml`.
+- New files need copyright and license headers.
+- While review is open, do not rebase or force-push, so reviewers can see that feedback
+  was addressed. Squashing to a single commit once approved is how changes land.
+- Every contribution must be reviewed and understood by a human before submission.
+
+Commit subjects follow Conventional Commits and name the affected module, for example
+`fix(EvseManager): handle unplug during timed charging`.
```
lukas-mertens: I would personally also make this requirement for pull request, as GitHub by default uses the PR title + description for the resulting squash commit if you squash via GitHub

lukas-mertens:
Theoretically you could run this through something like https://github.com/JuliusBrussee/caveman/blob/main/skills/caveman-compress/SKILL.md to save some more tokens while keeping the same info, but not sure if that's wanted.

AGENTS.md
mlitre: Good call I made an improvement

AGENTS.md
mlitre: Fair, I haven't that issue that much since i reset remote but good to have

AGENTS.md
mlitre: Good idea

mlitre:
> Theoretically you could run this through something like https://github.com/JuliusBrussee/caveman/blob/main/skills/caveman-compress/SKILL.md to save some more tokens while keeping the same info, but not sure if that's wanted.

For now I don't think we should just to keep some better human readability, but I did try to tighten the language so we save a little bit.

AGENTS.md
```diff
@@ -0,0 +1,228 @@
...
+
+`<libdir>` is `lib` or `lib64`, following `CMAKE_INSTALL_LIBDIR`. `--config` also takes
+an extensionless name, looked up in the installed config directory. `--check` validates
+a config and exits 0 on success.
+
+Log verbosity is per module in the deployed
+`build/dist/share/everest/modules/<Module>/logging.ini` (template
+`cmake/assets/logging.ini`). Severities: DEBG, INFO, WARN, ERRO, CRIT.
+
+## Testing
```
Pietfried, line 84: We should point to ./run-tests.sh https://github.com/EVerest/EVerest/blob/main/tests/run-tests.sh

Pietfried (changes requested):
We should point to the EVerest Releases and Versioning strategy and sensitize for breaking changes, backwards compatibility, etc. : https://everest.github.io/latest/project/releases/releases-and-versioning.html

AGENTS.md
mlitre: Done!

AGENTS.md
```diff
@@ -0,0 +1,224 @@
...
+```bash
+LD_LIBRARY_PATH=<prefix>/<libdir>:$LD_LIBRARY_PATH \
+  ./build/dist/bin/manager --prefix <prefix> --config <full path to config yaml>
+```
+
+`<libdir>` is `lib` or `lib64`, following `CMAKE_INSTALL_LIBDIR`. `--config` also takes
+an extensionless name, looked up in the installed config directory. `--check` validates
+a config and exits 0 on success.
+
+Log verbosity is per module in the deployed
```
Pietfried: thats not correct, there is no log level verbosity per module

AGENTS.md
```diff
@@ -0,0 +1,224 @@
...
+requires in `manifest.yaml`, is wired to other modules by a configuration YAML under
+`config/`, and communicates only over MQTT through `everest-framework`. Contracts are
+the source of truth for wiring: read `interfaces/*.yaml`, `types/*.yaml` and
+`errors/*.yaml` before the C++. Lifecycle is `init()` then `ready()`, implemented per
+provided interface in `*Impl.cpp`. Generated headers land under `build/generated/`.
+
+```cpp
+mod->r_evse_manager->subscribe_session_event(...);   // required interface, init() ok
+mod->p_main->publish_ready(true);                    // provided interface
+auto result = mod->r_auth->call_validate_token(tok); // command: ready() or later only
```
Pietfried: this wouldnt compile: 
validate_token is a command on auth_token_validator, not auth, and publish_ready on p_main doesn't match any real module (it's p_evse on the evse_manager interface).

AGENTS.md
```diff
@@ -0,0 +1,224 @@
...
+`tests/ocpp_tests/test_sets/validations.py`, and `modules/EVSE/EvseV2G/tests/`.
+
+## Architecture
+
+Each module is an independent process. It declares the interfaces it provides and
+requires in `manifest.yaml`, is wired to other modules by a configuration YAML under
+`config/`, and communicates only over MQTT through `everest-framework`. Contracts are
+the source of truth for wiring: read `interfaces/*.yaml`, `types/*.yaml` and
+`errors/*.yaml` before the C++. Lifecycle is `init()` then `ready()`, implemented per
+provided interface in `*Impl.cpp`. Generated headers land under `build/generated/`.
```
Pietfried: We could add that subscribes should be done in init ; no commands should be called in init;  commands should only be called from within ready.

Pietfried (changes requested):

Pietfried (dismissed):
Looks good from my side now

mlitre:
> Looks good from my side now

Let's wait for a few more approvals to get more input

AGENTS.md
```diff
@@ -0,0 +1,233 @@
...
+
+## Invariants
+
+- Never remove `// ev@<uuid>:v1` markers from `*Impl.cpp` or `CMakeLists.txt`.
+- Never hand-edit ev-cli-managed content outside your own marker-delimited regions,
+  `struct Conf` included. Regenerate instead.
+- Never edit anything under `build/`, including `build/generated/`. It is overwritten.
+- Never use `std::cout` or `fprintf` for logging.
+- Never call `mod->r_*->call_*()` inside `init()`. Command calls belong in `ready()` or
+  later.
```
barsnick, line 230: Isn't this also valid for `publish_*()`?

(Constructing callback lambdas which _contain_ `publish_*()`, on the other hand, is okay.)

barsnick (changes requested):

FaHaGit (dismissed):

barsnick (dismissed):

Pietfried (dismissed):

barsnick (dismissed):

AGENTS.md
```diff
@@ -0,0 +1,250 @@
...
+`make build`, `make test` and `make lint` do not exist here.
+
+```bash
+cmake -S . -B build -GNinja \
+  -DCMAKE_INSTALL_PREFIX=./build/dist \
+  -DCMAKE_BUILD_TYPE=Debug \
+  -DCMAKE_EXPORT_COMPILE_COMMANDS=ON \
+  -DBUILD_TESTING=ON \
+  -DEVEREST_ENABLE_COMPILE_WARNINGS=ON
+ninja -C build install -j$(nproc)
```
djchhp: Going for ninja exclusively is a bit too much, since it is not strictly a dependency of EVerest. Why trigger the build via cmake directly ignore the actual build subsystem?

corneliusclaussen:
One thing I stumbled across a few times: the headers contains a copyright year. Do we want this to update on every change? Or only when the file is created new?

AGENTS.md
mlitre: Fair point. The cmake args as well are quite arbitrary. @Pietfried @barsnick any preferences to this?
This is what I typically use so not sure if other people want to use the same

barsnick: > @Pietfried @barsnick any preferences to this?

I personally don't use ninja. I like a pure CMake based approach (`cmake -B ...; cmake --build ...; cmake --install`) but often fall back to pure `make -C` for compilation, out of habit (and for tab-expansion of rules).

The cmake flags do look relevant for agent builds.

djchhp:
@mlitre  Two more thing I just remembered: 

1.) I would love to see guidelines pointing to the usage of lib/util. Especially the direct usage of std::mutex and conditionvar should be discouraged, _monitor_ for the rescue.  Generally speaking lib/util should be considered as much as possible.

2.) Discourage direct usage of poll/epoll. We have libio for that. Also implement io via the given client policies.

corneliusclaussen:
About the stacked PR: my claude claims there is no native PR stack feature until I point it at it. So maybe we should add a comment that for stacked PR github's native stacked PR tooling shall be used and not manual stacking with references in the PR descriptions

AGENTS.md
mlitre: Done

mlitre:
> One thing I stumbled across a few times: the headers contains a copyright year. Do we want this to update on every change? Or only when the file is created new?

I've added a point about that

mlitre:
> @mlitre Two more thing I just remembered:
> 
> 1.) I would love to see guidelines pointing to the usage of lib/util. Especially the direct usage of std::mutex and conditionvar should be discouraged, _monitor_ for the rescue. Generally speaking lib/util should be considered as much as possible.
> 
> 2.) Discourage direct usage of poll/epoll. We have libio for that. Also implement io via the given client policies.

Done!

mlitre:
> About the stacked PR: my claude claims there is no native PR stack feature until I point it at it. So maybe we should add a comment that for stacked PR github's native stacked PR tooling shall be used and not manual stacking with references in the PR descriptions

I feel like this is more of a personal preference so I think for now I think this should go in the private claude or agents md

hikinggrass (approved):

SebaLukas (approved):

hikinggrass:
lgtm, needs a squash now and additonally a rebase to pass through the updated the ci

PR #2574, opened 2026-08-06
ocpp_consumer_API: Delay the heartbeat until the OCPP module is ready

nzbr:
## Describe your changes

When EVerest starts, there is a brief window where the ocpp_consumer_API is already ready and sending heartbeats, but the underlying OCPP module implementation has not fully started up yet. Clients that use the bridge module will receive errors if they try to perform actions against the OCPP module because they assume that it has started due to the heartbeat from the bridge module. This PR delays the bridge module's heartbeat until the underlying OCPP module has become ready. The OCPP module now published a ready message on successful start up which the bridge module uses to know when the heartbeats should start


cburandt (commented):
LGTM in general

One thought: ocpp_consumer_API.cpp::ready() still setups all the commands before ready. Maybe we should describe in the API yaml what the problem with using the API before "ready" is? (I believe calls don't timeout, they still execute, but are not guaranteed to do the right thing(?))

nzbr:
I added a line to the API definition of the heartbeat endpoint

james-ctc (dismissed):
Have you checked with Jan about this approach?
The stable API heartbeat exists for specific reasons and the description gives me the impression that there is a misunderstanding on its purpose.

> When EVerest starts, there is a brief window where the ocpp_consumer_API is already ready and sending heartbeats, but the underlying OCPP module implementation has not fully started up yet.

The heartbeat ensures that the ocpp_consumer_API module is running and that the application can see it.
If there is a need to synchronise with the underlying EVerest module then the API needs to be expanded.
I'm not sure that re-purposing the heartbeat is a good idea.

e.g. from the application: no heartbeat currently means that the ocpp_consumer_API module is not running, or heartbeat is disabled. Any change to that and it is a breaking change.

> because they assume that it has started 

The documentation needs to be improved to remove that assumption

also there needs to be consideration of module lifetimes. The application can still be running when the ocpp_consumer_API module restarts. Changing the meaning of the heartbeat can lead to issues.

Finally remember that stable API heartbeats are configurable - they can be turned off

cburandt:
> The heartbeat ensures that the ocpp_consumer_API module is running and that the application can see it. If there is a need to synchronise with the underlying EVerest module then the API needs to be expanded. I'm not sure that re-purposing the heartbeat is a good idea.
> 
> e.g. from the application: no heartbeat currently means that the ocpp_consumer_API module is not running, or heartbeat is disabled. Any change to that and it is a breaking change.
 
Reading this, I find it to be very valid. 

@nzbr , an approach could be to provide the ready variable on the API (the "ready" you added to the internal interface). This is non-breaking; the meaning of the heartbeat stays the same for all API modules; it is an even smaller change than the first approach; clients can still get the value if they miss its first publication because it's latched.

nzbr:
Re-publishing the ready variable is actually what I did first. So switching back to that would just be a revert of [this commit](https://github.com/EVerest/EVerest/pull/2574/changes/8fdb97c3847a4681d53cb99dc664391234d7adf6). 
I changed approach because of a comment on this PR I got from @Pietfried via DM. I don't have a strong opinion either way myself

Pietfried:
The real gap in my view is that this race isn't specific to OCPP, but likely also to other API modules. Some are not ready to process requests until a certain point in their own ready function. So rather than solving it once per bridge, we are probably missing a general readiness signal for all API modules. The heartbeat therefore doesn't really provide what consumers need (and expect): That the API is running and can be called. Our docs say

> Although communication check and heartbeat can be disabled with values smaller or equal to zero, this is not recommended in a production environment, since they are the only way to continuously check whether the client and EVerest are online and responsive. [Using the EVerest APIs](https://everest.github.io/nightly/explanation/adapt-everest/apis.html#using-the-everest-api).

so you could argue if we even fullfil this with todays heartbeat implementation.

I missed that the heartbeats can be disabled and agree that we shouldn't repurpose, so to move forward with this PR I would suggest to reintroduce the initial approach. Sorry for the back and forth @nzbr !

james-ctc (commented):
I'm unsure what to suggest, some options

1. error returns until ready
This seems like a good choice. The API module doesn't need to know the OCPP module's state. Calls can fail for many reasons and the API module's life cycle may not match the application's. EVerest is having more functionality to allow modules to restart without the whole of EVerest restarting. The application would need to be robust and know how to manage failures. 

2. caching ready state (in the API module)
This could be implemented just using MQTT:
- retain the ready state in the broker
- add a LWT to clear that state when the API module terminates
- (clear the state if the API module gracefully shuts down)
The main issue is what happens when the OCPP module shuts down, or restarts and the API module doesn't. There is an incorrect cached state.

3. update OCPP module
The OCPP module could be adapted to support querying state, or use retained topics & LWT ... (as in 2. above)
The API module could check on any API call that the OCPP module is still available.
This would detect the OCPP module stopping and restarting and that correct state could then be available to the application

I suspect other modules will have similar issues and a standard approach on how to solve this probably should be adopted. It would be confusing for different APIs to use different approaches.

A generic approach could be:
EVerest modules have a state that is a retained topic, using LWT to reset it on unexpected module termination.
The module would then manage the transitions for that state:
NotRunning->Starting->Ready->Stopping
Every API module could then make that state available to applications, perhaps with a simple "get_state" command - or by replicating the state as a retained topic ...

@Pietfried @nzbr

Pietfried (approved):

PR #2577, opened 2026-08-07
feat(libiso15118): EV stack foundation

mlitre:
Part of the EV-side libiso15118 stack (1/8). Each PR compares against the one below it; review bottom-up.

**Stack (bottom to top):**
1. Foundation — #2577 ⬅️ **this PR**
3. AC charging — #2579
4. AC_DER_IEC — #2580
5. BPT (AC + DC) — #2581
7. Three-phase AC — #2583

## Describe your changes

libiso15118 grows an EV (EVCC) half: an ISO 15118-20 state machine with its own Context and
MessageExchange, the DC states from SessionSetup through SessionStop, an SDP request and response
codec, and a feedback seam for handing EVSE session data upward. Nothing drives it yet. It sits
behind ISO15118_BUILD_EV_SIDE, off by default, so a consumer that only implements a charger
neither builds the EV sources nor links everest::io, which only the EV side needs.


mlitre:
Replaced by #2799, which carries the EV stack foundation as one commit: the state machine base, one `MessageExchange` template, the session feedback channel, and the SDP and data transport clients, all behind the `ISO15118_BUILD_EV_SIDE` gate. The session and controller work that grew on this branch went to #2802 instead, where it is written once rather than accreted. Closing in favor of those two.

PR #2578, opened 2026-08-07
feat(libiso15118): EV DC module and session

mlitre:
Part of the EV-side libiso15118 stack (2/8). Each PR compares against the one below it; review bottom-up.

**Stack (bottom to top):**
1. Foundation — #2577
2. DC module + session — #2578 ⬅️ **this PR**
3. AC charging — #2579
4. AC_DER_IEC — #2580
5. BPT (AC + DC) — #2581
7. Three-phase AC — #2583

## Describe your changes

The EV side becomes drivable. ev::Session pumps V2GTP frames through the FSM on two reactor
timers: a send-delay hold window, so a control event can still replace a pending request before
it is encoded, and a response watchdog that fails the session rather than parking it.
ev::Controller mirrors the SECC TbdController, except its loop returns once the session finishes
or a bounded setup deadline elapses, so tests terminate. The service walk, ScheduleExchange and
the DC charge loop complete the state set, and EvIso15118D20 publishes it as an ISO15118_ev
provider, a C++ SIL alternative to PyEvJosev.


mlitre:
Replaced by #2801, which carries the ISO 15118-20 EV engine, its context and timeout tables, and the DC charging states as one commit. Closing in favor of it.

PR #2579, opened 2026-08-07
feat(libiso15118): EV AC charging

mlitre:
Part of the EV-side libiso15118 stack (3/8). Each PR compares against the one below it; review bottom-up.

**Stack (bottom to top):**
1. Foundation — #2577
2. DC module + session — #2578
3. AC charging — #2579 ⬅️ **this PR**
4. AC_DER_IEC — #2580
5. BPT (AC + DC) — #2581
6. Fixes + test coverage — #2582
7. Three-phase AC — #2583
8. State outcomes refactor — #2584

## Describe your changes

AC energy transfer alongside DC: an AcChargeParams channel from module config into the FSM,
AC_ChargeParameterDiscovery and AC_ChargeLoop, and routing that skips CableCheck and
WeldingDetection. The service walk stops assuming DC, taking its category from config.

Per-state response validation collapses into expect_response, which makes the session-id and
response-code checks structural rather than something every state must remember, and the AC/DC
split, previously spelled out at four sites, becomes one is_ac_family() predicate.

PR #2580, opened 2026-08-07
feat(libiso15118): EV AC_DER_IEC

mlitre:
Part of the EV-side libiso15118 stack (4/8). Each PR compares against the one below it; review bottom-up.

**Stack (bottom to top):**
1. Foundation — #2577
3. AC charging — #2579
4. AC_DER_IEC — #2580 ⬅️ **this PR**
5. BPT (AC + DC) — #2581
7. Three-phase AC — #2583

## Describe your changes

AC_DER_IEC negotiation: energy service 10 with Part20DerIec request and response types through
charge parameter discovery and the charge loop, and DSO directives surfaced to the module through
a der_control callback.

The gating is the substance. The EV declares which IEC DER control functions it supports as a
bitmask pinned to the iec::DERControlName enum tail, ServiceDetail accepts only parameter sets
whose functions are a subset of it, and setpoints for functions never negotiated are stripped in
the charge loop; a mismatch stops the session or continues, per config.


mlitre:
Replaced by #2803, which carries AC DER IEC and its function negotiation together with the rest of the AC family, as one commit. Closing in favor of it.

PR #2581, opened 2026-08-07
feat(libiso15118): EV BPT (AC + DC)

mlitre:
Part of the EV-side libiso15118 stack (5/8). Each PR compares against the one below it; review bottom-up.

**Stack (bottom to top):**
1. Foundation — #2577
3. AC charging — #2579
4. AC_DER_IEC — #2580
5. BPT (AC + DC) — #2581 ⬅️ **this PR**
7. Three-phase AC — #2583

## Describe your changes

Bidirectional power transfer for both AC and DC. Discharge limits ride on the existing
charge-params types, and the BPT request and response variants are handled inside the AC and DC
states already there, keyed on the selected service, so there is no parallel BPT state set.

Two wire-level fixes came out of testing it: the three_phase flag was copying one aggregate into
all three per-phase fields, so those optional fields are no longer emitted, and the mandatory
present voltage and power went out as a constant zero, which a charger reads as measured and uses
to disable its HLC power clamp.

### DIFF
--- a/config/config-sil-ac-bpt-d20-evcpp.yaml
+++ b/config/config-sil-ac-bpt-d20-evcpp.yaml
@@ -1,0 +1,176 @@
+active_modules:
+  iso15118_charger:
+    module: Evse15118D20
+    config_module:
+      device: auto
+    connections:
+      security:
+        - module_id: evse_security
+          implementation_id: main
+  iso15118_car:
+    module: EvIso15118D20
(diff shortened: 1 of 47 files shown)


mlitre:
Replaced by #2803, which carries AC BPT and DC BPT together with the rest of the AC family, as one commit. Closing in favor of it.

PR #2582, opened 2026-08-07
test(libiso15118): EV fixes and test coverage

mlitre:
Part of the EV-side libiso15118 stack (6/8). Each PR compares against the one below it; review bottom-up.

**Stack (bottom to top):**
1. Foundation — #2577
3. AC charging — #2579
4. AC_DER_IEC — #2580
5. BPT (AC + DC) — #2581
6. Fixes + test coverage — #2582 ⬅️ **this PR**
7. Three-phase AC — #2583

## Describe your changes

Present values stop being invented. update_present_values and an EvPresentValues type join the
ISO15118_ev interface and the simulator feeds them, so the charge loops report a measurement
instead of the interim approximation. Every command on that interface binds by name at start-up,
so PyEvJosev needs the handler too or it aborts and takes down every config using it.

Config validation also moves to start_charging, where it checks the merged parameters that reach
the wire rather than the partial objects init saw, and SessionSetup now accepts only
OK_NewSessionEstablished: every other code it used to accept transitioned the EV forward carrying
a zero session id.


mlitre:
Split across three of the new layers rather than replaced by one. Its test coverage follows the code it covers: the ISO 15118-20 DC fixtures and state tests are in #2801, the session and controller suite is in #2802, and the AC family suite is in #2803. Closing in favor of those three.

PR #2583, opened 2026-08-07
feat(libiso15118): EV three-phase AC

mlitre:
Part of the EV-side libiso15118 stack (7/8). Each PR compares against the one below it; review bottom-up.

**Stack (bottom to top):**
1. Foundation — #2577
3. AC charging — #2579
4. AC_DER_IEC — #2580
5. BPT (AC + DC) — #2581
7. Three-phase AC — #2583 ⬅️ **this PR**

## Describe your changes

The EV picked the first Dynamic parameter set without reading its Connector, and EvseManager lists
SinglePhase first even on three-phase hardware, so a three-phase EV selected the single-phase set
and advertised its whole total on one line.

ServiceDetail now prefers a set matching the EV's own ac_phase_count, and the AC states emit
through split_ac_limit, which encodes the three readings ISO 15118-20 gives the base element: one
line's value on SinglePhase, the sum of all three on ThreePhase without peers, L1 alone once peers
are present. A measurement is one aggregate reading, not a capability to divide, so
emit_ac_present stays separate, and EvIso15118D20::shutdown() now joins the session worker.


mlitre:
Replaced by #2803, which carries the three-phase AC split across the selected connector together with the rest of the AC family, as one commit. Closing in favor of it.

PR #2584, opened 2026-08-07
refactor(libiso15118): EV state outcomes

mlitre:
Part of the EV-side libiso15118 stack (8/8). Each PR compares against the one below it; review bottom-up.

**Stack (bottom to top):**
1. Foundation — #2577
3. AC charging — #2579
4. AC_DER_IEC — #2580
5. BPT (AC + DC) — #2581
7. Three-phase AC — #2583
8. State outcomes refactor — #2584 ⬅️ **this PR**

## Describe your changes

Every feed() returned a bare Result whether it had emitted a request, stopped the session or
ignored the event; only Session's consumed-but-idle catch-all told the three apart. Result now
carries the outcome, its default constructor is private, and Session stops a state that declared
an outcome it did not deliver rather than parking the session with no timer armed.

That surfaced a real gap: an EV-initiated stop was honored only inside the charge loops, so one
arriving earlier was latched and ignored while the session closed the contactors and sent
PowerDelivery(Start). Pre-loop states now go straight to SessionStopReq(Terminate), per
[V2G20-2644].


mlitre:
Replaced by #2799, which defines `Event`, `Disposition` and `BasicResult` in the foundation layer, where the state outcomes belong, rather than refactoring them in later. Closing in favor of it.

PR #2585, opened 2026-08-07
feat(ocpp): add monitor_and_get_variables command

Pietfried:
## Describe your changes

Combines monitor_variables and get_variables in one call: registers the monitors first, then returns the current (Actual) values with get_variables result semantics. Lets integrators register monitors at startup and obtain initial values without a race between the first read and the monitor registration.

Implemented in OCPP, OCPP201, OCPPmulti and the BUOcppConsumer stub; exposed via ocpp_consumer_API. The OCPP 1.6 module guards the empty request list, which would otherwise return all configuration keys.

Why? It allows integrators to register monitors and retrieve configuration in one synchronous call, instead of a separate monitor + get_variables . 


modules/EVSE/OCPP/ocpp_generic/ocppImpl.cpp
```diff
@@ -299,6 +299,27 @@ void ocppImpl::handle_monitor_variables(std::vector<types::ocpp::ComponentVariab
...
+    }
+
+    std::vector<types::ocpp::GetVariableRequest> requests;
+    requests.reserve(component_variables.size());
+    for (const auto& cv : component_variables) {
+        types::ocpp::GetVariableRequest request;
+        request.component_variable = cv; // no attribute_type: Actual is the default
+        requests.push_back(request);
+    }
+    return handle_get_variables(requests);
```
cburandt, line 320: The idea of this PR to avoid the race between getting some value initially and monitoring for later updates (the race of missing an update in between). What is the gain of using monitor_and_get_variables compared to first setting up a monitor and then calling get_variables from the other module or API client? Is the intention to give more guarantees or only to provide a convenience function?

As an API user I read the command name as get-me-the-current-value-and-then-push-all-changes-from-there. The race a client could experience: It's not guaranteed that the return value (from get_variables()) is the oldest value.

modules/EVSE/OCPP201/ocpp_generic/ocppImpl.cpp
```diff
@@ -156,6 +156,21 @@ void ocppImpl::handle_monitor_variables(std::vector<types::ocpp::ComponentVariab
...
+    handle_monitor_variables(component_variables);
+
+    std::vector<types::ocpp::GetVariableRequest> requests;
+    requests.reserve(component_variables.size());
+    for (const auto& cv : component_variables) {
+        types::ocpp::GetVariableRequest request;
+        request.component_variable = cv; // no attribute_type: Actual is the default
+        requests.push_back(request);
+    }
+    return handle_get_variables(requests);
```
cburandt, line 171: see above

cburandt (commented):
Not sure how much of a problem it is, what my comment below describes. But it maybe worth at least to change the description of the API.

modules/EVSE/OCPP/ocpp_generic/ocppImpl.cpp
Pietfried: I updated the documentation to be more explicit about the mechanics of the call. It's mainly a convenience function. We observed that consumers require to request the initial value and register the monitor when starting up. The handling for the get_variables and an incoming event for that variable is also similar in many cases.

cburandt (dismissed):
LGTM

hikinggrass (dismissed):

hikinggrass (approved):

PR #2586, opened 2026-08-07
fix(ocpp1.6): make StopTransactionOnEVSideDisconnect writable

bluesandme:
Per the OCPP 1.6 Core spec this key is RW. OCTT fails (TC_005_3_CS) a ChangeConfiguration.req for this key with "Not  able to set config with key 'StopTransactionOnEVSideDisconnect'". Updated the existing unit test to match.

### DIFF
--- a/lib/everest/ocpp/config/v16/profile_schemas/Core.json
+++ b/lib/everest/ocpp/config/v16/profile_schemas/Core.json
@@ -133,7 +133,7 @@
         },
         "StopTransactionOnEVSideDisconnect": {
             "type": "boolean",
-            "readOnly": true
+            "readOnly": false
         },
         "StopTransactionOnInvalidId": {
             "type": "boolean",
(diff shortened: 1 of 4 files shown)


Pietfried:
Hi @bluesandme , thanks for the contribution. According to OCPP 1.6 Errata sheet 2025-04 `StopTransactionOnEVSideDisconnect` is `R or RW. Choice is up to Charge Point implementation.` . EVerest's charger state machine does currently not support to continue transactions when the cable is disconnected on the EV side, which is why the definition in EVerest is ReadOnly and defaults to `false`.

If you want to use the OCPP library standalone, you can try to approach this by using a similar mechanism like for the configuration parameter `CustomDisplayCostAndPrice`, for which the mutability can be optionally overriden in the schema and is then also applied at runtime.

Feel free to open a PR with this new approach.

PR #2588, opened 2026-08-07
fix(OCPP): prevent DataTransfer issues on non-string data

Pietfried:
## Describe your changes

Convert OCPP 2.x json  to the OCPP 1.6 string form explicitly in OCPPmulti instead of relying on nlohmann's implicit conversion, which throws on objects/arrays; harden to_ocpp_data_transfer_response in ocpp_module_common against non-json module responses (affects OCPP201 and both OCPPmulti paths). Add unit tests for both conversion layers.


andistorm (approved):
LGTM 🐧

PR #2589, opened 2026-08-10
fix(ocpp1.6): report every InvalidCentralSystemCertificate failure.

bluesandme:
Previously a logged flag suppressed all but the first InvalidCSMSCertificate SecurityEventNotification until a successful connection, so consecutive bad-certificate attempts after the first were only logged, not reported to the CSMS. OCTT's TC_078_CS walks through several distinct invalid-certificate scenarios back to back without a successful connection in between and expects a notification for each one.

### DIFF
--- a/lib/everest/ocpp/include/ocpp/v16/charge_point_impl.hpp
+++ b/lib/everest/ocpp/include/ocpp/v16/charge_point_impl.hpp
@@ -93,7 +93,6 @@ private:
     ChargePointConfigurationInterface& configuration;
     BootReasonEnum bootreason{BootReasonEnum::PowerUp};
     bool initialized{false};
-    bool InvalidCSMSCertificate_logged{false};
     ChargePointConnectionState connection_state{ChargePointConnectionState::Disconnected};
     std::atomic<RegistrationStatus> registration_status{RegistrationStatus::Pending};
     DiagnosticsStatus diagnostics_status{DiagnosticsStatus::Idle};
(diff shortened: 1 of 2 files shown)


hikinggrass:
This basically reverts https://github.com/EVerest/libocpp/pull/1156 - is that what we want @Pietfried @james-ctc @mlitre ?

mlitre:
> ged flag suppressed all but the first InvalidCSMSCertificate SecurityEventNotification until a successful connection, so consecutive bad-certificate attempts after the first were only logged, not reported to the CSMS. OCTT's TC_078_CS walks through several distinct invalid-certificate scenarios back to back without a

I think a full revert is not correct either because this can pollute the message queue, but I do agree this probably fails the test now

james-ctc:
> This basically reverts [EVerest/libocpp#1156](https://github.com/EVerest/libocpp/pull/1156) - is that what we want @Pietfried @james-ctc @mlitre ?

PR 1156 was a required change to avoid duplicate error logs. Any solution needs to ensure the same problem is not logged multiple times. As it stands this PR doesn't do that.
A proper solution will need to keep a track of the different situations and ensure they are logged only once.

james-ctc:
Looking at  [TC_078_CS](https://openchargealliance.org/wp-content/uploads/2025/08/CompliancyTestTool-TestCaseDocument.pdf)
The steps show a successful connect (step 3) followed by an unsuccessful connect (step 6), followed by a successful connect (step 9).
With step 10 checking for the security event notification. That test doesn't have back-to-back invalid connections.

mlitre:
> Looking at [TC_078_CS](https://openchargealliance.org/wp-content/uploads/2025/08/CompliancyTestTool-TestCaseDocument.pdf) The steps show a successful connect (step 3) followed by an unsuccessful connect (step 6), followed by a successful connect (step 9). With step 10 checking for the security event notification. That test doesn't have back-to-back invalid connections.

In theory but it does back to back checks for different invalid reasons...
But I think your solution of tracking the reason would work. I need to rerun the test to be sure

PR #2590, opened 2026-08-10
feat(ocpp1.6): add extensibility hooks for Custom-profile config keys

bluesandme:
Adds two opt-in, inert-by-default hooks to ChargePointConfigurationInterface: a validation callback that can veto a ChangeConfiguration before it is applied (e.g. based on runtime charge-point state), and a forced-write path that bypasses readOnly for internal/backend-driven updates only (never for CSMS-initiated changes). The validation callback lives in ChargePointImpl::set_configuration_key_internal (mirroring register_configuration_key_changed_callback), so it applies to both config backends (legacy JSON and DeviceModel) and to any key - Custom or Core - without needing key-specific wiring (this is also why LightIntensity no longer needs a hardcoded hook).
There is no read callback: applications should push fresh values in via set_custom_key_forced() directly. An earlier read-callback design re-entered getCustomKeyValue() via setCustomKey() when  refreshing a readOnly key, causing infinite recursion - dropped for that reason. set_custom_key_forced() now locks configuration_mutex and guards against a missing "Custom" section.

Adds tests for veto, forced write on readOnly, and missing keys.

### DIFF
--- a/lib/everest/ocpp/lib/ocpp/v16/charge_point_configuration.cpp
+++ b/lib/everest/ocpp/lib/ocpp/v16/charge_point_configuration.cpp
@@ -3139,7 +3139,7 @@ std::optional<KeyValue> ChargePointConfiguration::getLanguageKeyValue() {
 // Custom
 std::optional<KeyValue> ChargePointConfiguration::getCustomKeyValue(const CiString<50>& key) {
     std::lock_guard<std::recursive_mutex> lock(this->configuration_mutex);
-    if (!this->config["Custom"].contains(key.get())) {
+    if (!this->config.contains("Custom") or !this->config["Custom"].contains(key.get())) {
         return std::nullopt;
     }
 
@@ -4302,6 +4302,15 @@ std::optional<ConfigurationStatus> ChargePointConfiguration::set(const CiString<
     }
 
     return ConfigurationStatus::Accepted;
+}
+
+ConfigurationStatus ChargePointConfiguration::set_custom_key_forced(const CiString<50>& key,
+                                                                    const CiString<500>& value) {
+    std::lock_guard<std::recursive_mutex> lock(this->configuration_mutex);
+    if (!this->config.contains("Custom") or !this->config["Custom"].contains(key.get())) {
+        return ConfigurationStatus::NotSupported;
+    }
+    return this->setCustomKey(key, value, /*force=*/true);
 }
 
 } // namespace ocpp::v16
--- a/lib/everest/ocpp/include/ocpp/v16/charge_point.hpp
+++ b/lib/everest/ocpp/include/ocpp/v16/charge_point.hpp
@@ -641,6 +641,18 @@ public:
     void
     register_generic_configuration_key_changed_callback(const std::function<void(const KeyValue& key_value)>& callback);
 
+    /// \brief registers a \p callback that can veto a ChangeConfiguration before it is applied (e.g.
+    /// based on runtime charge-point state), for both Core and Custom-profile keys
+    /// \param callback returns true to allow the change, false to reject it
+    void register_custom_key_validation_callback(
+        const std::function<bool(const std::string& key, const std::string& value)>& callback);
+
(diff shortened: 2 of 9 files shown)


lib/everest/ocpp/lib/ocpp/v16/charge_point_configuration.cpp
```diff
@@ -4278,4 +4288,21 @@ std::optional<ConfigurationStatus> ChargePointConfiguration::set(const CiString<
...
+    this->custom_key_validation_callback = callback;
+}
+
+void ChargePointConfiguration::register_custom_key_read_callback(
+    const std::function<void(const std::string& key)>& callback) {
+    this->custom_key_read_callback = callback;
+}
+
+ConfigurationStatus ChargePointConfiguration::set_custom_key_forced(const CiString<50>& key, const CiString<500>& value) {
+    if (!this->config["Custom"].contains(key.get())) {
```
Pietfried: This reads this->config without taking configuration_mutex

Pietfried:
```suggestion
    if (!this->config.contains("Custom").contains(key.get())) {
```

lib/everest/ocpp/lib/ocpp/v16/charge_point_configuration.cpp
```diff
@@ -3126,6 +3126,10 @@ std::optional<KeyValue> ChargePointConfiguration::getCustomKeyValue(const CiStri
         return std::nullopt;
     }
 
+    if (this->custom_key_read_callback) {
+        this->custom_key_read_callback(key.get());
+    }
```
Pietfried: This looks a little problematic. Per your PR documentation this is a hook that's expected to side effect the store by calling back into `set_custom_key_forced`. That would result in a deadlock afaics.

Could `custom_key_read_callback` actually return the value?

lib/everest/ocpp/lib/ocpp/v16/charge_point_configuration.cpp
```diff
@@ -3126,6 +3126,10 @@ std::optional<KeyValue> ChargePointConfiguration::getCustomKeyValue(const CiStri
         return std::nullopt;
     }
 
+    if (this->custom_key_read_callback) {
+        this->custom_key_read_callback(key.get());
```
Pietfried: This creates an infinite recursion when used as documented: refreshing a readOnly key requires set_custom_key_forced, but that path re-enters getCustomKeyValue via setCustomKey

Pietfried (changes requested):
Thanks for your contribution!

Heads-up for the OCPP1.6 JSON backend: We added support for all protocol versions for the device model (the 1.6 backend already exists: `ChargePointConfigurationDeviceModel`, with Custom keys mapped to device-model variables via Ocpp16CustomConfigMappings). This will be maintained moving forward and also likely be extended with externally provided telemetry data (which seems to be what you are mostly concerned with). See [#EVerest: Cloud communication > 2026-12-08 Cloud WG: Telemetry Proposal @ 💬](https://lfenergy.zulipchat.com/#narrow/channel/417676-EVerest.3A-Cloud-communication/topic/2026-12-08.20Cloud.20WG.3A.20Telemetry.20Proposal/near/616254259) for more details.

We do not plan to enhance the JSON ChargePointConfiguration beyond the next release, likely 2026.10 . 

Your changes only introduce optional, default no-ops, so it's still fine if we try to integrate it but I'd like to see a few changes additional to the comments in line:
- Please drop the read callback. With set_custom_key_forced in place, the application can push fresh values into the store and it also removes the infinite recursion (see in line comment)
- The veto callback can be moved to ChargePointImpl::set_configuration_key_internal (mirroring the existing register_configuration_key_changed_callback pattern) so it works on both config backends. This would also let you drop the hardcoded LightIntensity wiring
- Tests for e.g. veto, forced write on readOnly, and missing keys.

PR #2594, opened 2026-08-10
EvseManager: Add WaitingForReplug state

dorezyuk:
We analyzed our user data an observed that its seems to be a common issue that especially during the first 30 secs of a session the cp signal can drop for a view seconds from B to A and then to back to B. This right now ends the sessions, leading to a not ideal experience. 

To deal with that we add a new `WaitingForReplug` state (I know that we had a similar naming for a slightly different concept) - during that state we allow users to reconnect their car without stopping the session/transaction. 

Our data also showed that the majority of these toggles happen within the first 5mins of the session. Later toggles are more likely to be legit attempts from users to stop their session. Therefore we also added the `replug_max_transaction_age_s` defaulting to 5min - after this time the replug logic is deactivated

### DIFF
--- a/modules/EVSE/EvseManager/manifest.yaml
+++ b/modules/EVSE/EvseManager/manifest.yaml
@@ -454,6 +454,13 @@ config:
       finished as usual. Set to 0 to finish the transaction immediately on unplug.
     type: integer
     default: 0
+  replug_max_transaction_age_s:
+    description: >-
+      Maximum transaction age in seconds for the `replug_timeout_s` to apply. If the EV is
+      unplugged later than this into the transaction, the `replug_timeout_s` takes no effect.
+      Setting it to 0 or less disables the age limit.
+    type: integer
+    default: 300
 provides:
   evse:
     interface: evse_manager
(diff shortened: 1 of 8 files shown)


modules/EVSE/EvseManager/manifest.yaml
```diff
@@ -446,28 +446,21 @@ config:
...
-      i.e. reported L1/L2/L3 already matches grid L1/L2/L3. STR means reported L1 is
-      grid L2, reported L2 is grid L3, reported L3 is grid L1. TRS means reported L1
-      is grid L3, reported L2 is grid L1, reported L3 is grid L2.
-    type: string
-    enum:
-      - 'RST'
-      - 'STR'
-      - 'TRS'
-    default: 'RST'
+      Grace period in seconds to continue an active transaction after the EV was unplugged.
```
dorezyuk, line 451: mark experimental

modules/EVSE/EvseManager/manifest.yaml
```diff
@@ -446,28 +446,21 @@ config:
...
+      If the EV is plugged in again within this period, the charging session and transaction
+      continue without a new authorization. When the period expires, the transaction is
+      finished as usual. Set to 0 to finish the transaction immediately on unplug.
+    type: integer
+    default: 0
+  replug_max_transaction_age_s:
```
dorezyuk, line 457: mark experimental

Issue #2496, opened 2026-07-21
[OCPPmulti] Improving configuration synchronization in the OCPPmulti module


nsmee:
### Describe the problem

PR #1571 introduces variable mapping between the OCPP 2.0.1 device model and the EVerest configuration, but it is limited to OCPP 2.0.1. It duplicates or implicitly generates device model component variables, still depends on invoking Auth and EVSE Manager interfaces, and does not fully align with the configuration service as the single source of truth. For OCPPmulti, this leads to divergent values between the OCPP device model and EVerest configuration; At init of OCPP201 the configuration parameters get overwritten. reboot requirements of some config parameters are not surfaced back to the backend; This issue aims to bring OCPPmulti up to date by addressing those shortcomings.

### EVerest Domain

OCPP 2.0.1

### Affected Component

Primary impact: modules/EVSE/OCPPmulti and lib/everest/ocpp_module_common. Secondary updates may be required in libocpp, the configuration service integration, and related unit tests.

### Describe your solution

• Extend the mapping approach from PR #1571 so OCPPmulti uses strict 1:1 mappings (e.g., connection_timeout → TxCtrlr/EVConnectionTimeOut) defined in mapping.yaml, without creating new component variables, and allow each map entry to target either a libocpp or internal EVerest CV. Replace build_everest_config_variables with register_everest_configuration_mappings in lib/everest/ocpp_module_common/src/device_model/everest_device_model_storage.cpp.  
• Implement bidirectional runtime sync inside set_variable_attribute_value for ComposedDeviceModelStorage and EverestDeviceModelStorage: SetVariable writes flow through to mapped EVerest configuration parameters, GetVariable reads come from those parameters, and configuration change callbacks update device model storage (without re-triggering the config service).  
• Remove redundant Auth and EVSE Manager interface calls (set_connection_timeout, set_master_pass_group_id, PnCEnabled, CentralContractValidationAllowed, ContractCertificateInstallationEnabled) once the configuration service propagates changes.  
• Ensure the configuration service can signal “reboot required,” add EVLOG_verbose startup dumps of the device model, reuse the single-source-of-truth mechanisms from PR #2412, fetch GetReport data from persistent storage, and update unit tests to cover the new sync path.

### Alternatives considered

Continuing with dynamic creation of device model variables introduced in PR #1571. Maybe this should be configurable.

### Additional context

[...]

PR #2600, opened 2026-08-11
feat(ocpp) add mapping for everest config synchronization

nsmee:
## Describe your changes
- Add debug logs for OCPP
- Make OCPP component exposition configurable; purge device model entries if disabled
- Fetch runtime device-model updates in GetReport
- Add mapping.yaml and mapping_schema.json to sync Everest config with OCPP variables
- Use composed key for std::find
- Add configurable logging and harden merging if composed device model
- Fix initial value of CentralContractValidationAllowed

https://github.com/EVerest/EVerest/issues/2496

- [ ] add functionality of https://github.com/EVerest/EVerest/pull/2412 to this PR


modules/EVSE/OCPPmulti/manifest.yaml
```diff
@@ -73,6 +73,12 @@ config:
       If false (default), the device model is initialized only from the component configs (DeviceModelConfigPath).
     type: boolean
     default: false
+  ExposeFullEverestConfigInDeviceModel:
```
Pietfried, line 76: Technically this is not OCPP2.0.1 only. If its enabled, those variables from the EVerest device model can be get/set via the ocpp_consumer_API and this would work for 2.x as well as for 1.6 . As soon as the EVerest config API is available, accessing this EVerest configuration should be done through this config API of course. Not quite sure if we want to also allow access via the OCPP consumer API as well. 

What do you think? For this PR, just add this in the description

```
OCPP2.x
OCPP1.6
```

and we can have a discussion and potential changes for the access of EVerest vars in a follow up.

Pietfried (commented):
Some first feedback:
- It would be great if we could add some documentation for this mapping as part of the module docs/index.rst as well as a how-to-guide for integrators in docs/source/how-to-guides.
- This needs a rebase once https://github.com/EVerest/EVerest/pull/2596 is merged, which introduces the ComposedDM for 1.6
- Afaics this PR just introduces the mapping format, but doesn't apply it. Is this planned as a follow up?

nsmee:
> * Afaics this PR just introduces the mapping format, but doesn't apply it. Is this planned as a follow up?

I planned as follow up, to not rely on the https://github.com/EVerest/EVerest/pull/2412.


> * It would be great if we could add some documentation for this mapping as part of the module docs/index.rst as well as a how-to-guide for integrators in docs/source/how-to-guides.

Sure. I will add these.

modules/EVSE/OCPPmulti/manifest.yaml
nsmee: I agree. 

> * Not quite sure if we want to also allow access via the OCPP consumer API as well.

I see that as an follow up PR. As a feature this is helpful, i think.

PR #2601, opened 2026-08-11
fix(EnergyNode): Do not call_enforce_limits for all children

Pietfried:
## Describe your changes

route enforce_limits only to the branch containing the target node

Previously every EnforcedLimits was broadcast to all children at each tree level, flooding the whole tree with one command call per node and recipient per optimizer run (resolves the FIXME in handle_enforce_limits).

Record for each cached child subtree which energy_consumer connection it was received from, and deliver limits only to the branches whose subtree contains the target uuid. Unknown uuids (e.g. no energy_flow_request received from a child yet) still fall back to broadcasting to all children.

### DIFF
--- a/modules/EnergyManagement/EnergyNode/energy_grid/energyImpl.cpp
+++ b/modules/EnergyManagement/EnergyNode/energy_grid/energyImpl.cpp
@@ -7,8 +7,10 @@
 #include <chrono>
 #include <date/date.h>
 #include <date/tz.h>
+#include <string>
 #include <string_view>
 #include <utils/date.hpp>
+#include <vector>
 
 namespace module {
 namespace energy_grid {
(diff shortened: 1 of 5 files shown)


djchhp (commented):
The PR improves the efficiency of EVerests energy management significantly. The implementation is correct as far as I can see. 

In order to improve the readability and for a better protection from future regressions I suggest 2 changes, with a possible implementation provided as patch.

1.) Remove the aligned index bookkeeping. 
[001_pr2601.patch](https://github.com/user-attachments/files/31214004/001_pr2601.patch)

2.) Separate the concerns of logic and subscription
[002_pr2601.patch](https://github.com/user-attachments/files/31214019/002_pr2601.patch)

djchhp (dismissed):
--- removed ---

djchhp (approved):
looking good!

PR #2606, opened 2026-08-13
fix(libocpp): use real reason codes on cert install reject

mlitre:
## Describe your changes

`InstallCertificateResponse` rejections carried the invented `statusInfo.reasonCode`
`"UnsecureConnection"` for three unrelated causes. `Security::should_allow_certificate_install`
returned a bare bool, so the handler had one string for all of them.

It now returns the refusal instead of a bool (`check_certificate_install_allowed` ->
`std::optional<StatusInfo>`), with OCPP 2.1 Appendix 5 codes and a cause-specific
`additionalInfo`:

| Cause | reasonCode | additionalInfo |
|---|---|---|
| CSMS root, SecurityProfile <= 1, `AllowCSMSRootCertInstallWithUnsecureConnection` false | `Unspecified` | `CSMSRootCertificateInstallationNotAllowedWithUnsecureConnection` |
| Manufacturer root, same shape with `AllowMFRootCertInstallWithUnsecureConnection` | `Unspecified` | `ManufacturerRootCertificateInstallationNotAllowedWithUnsecureConnection` |
| OEM root, any profile | `UnsupportedRequest` | `OEMRootCertificateInstallationNotSupported` |
| Unhandled certificate type | `UnsupportedRequest` | `CertificateTypeNotSupported` |

`Unspecified` rather than a more specific code because Appendix 5 is informative, Part 3 puts no
`enum` on `reasonCode`, and neither the 2.0.1 Edition 4 nor the 2.1 Edition 2 Part 6 test cases
assert `reasonCode` on `InstallCertificateResponse` (`TC_M_07_CS` / `TC_M_09_CS` check
`status must be Rejected` only). The cause is carried by `additionalInfo`.

Also a functional fix: `OEMRootCertificate` is refused before the `SecurityProfile > 1` early
return. Above profile 1 the request used to reach `evse_security`, where
`from_ocpp(CaCertificateType::OEM)` throws `EnumConversionException`
(`lib/everest/ocpp/lib/ocpp/common/evse_security_impl.cpp:371`), so no CallResult was dispatched
at all. The `// FIXME: Implement OEMRootCertificate` gap itself is untouched.

The previously unreachable post-switch fallthrough now refuses rather than allows, keeping the
old deny-by-default if an enum value is added.

n/a

- [ ] I have made corresponding changes to the documentation - no module docs cover
      `InstallCertificate` reason codes

## How to verify it

- [x] `ctest -R libocpp --output-on-failure` - 7/7 passed, including `libocpp_test_security`
- [x] Full suite of the libs-only build (`EVEREST_INCLUDE_LIBS=ocpp`, `BUILD_TESTING=ON`) - 158/158 passed
- [x] Red-phase check: with the implementation reverted, 4 of the 5 new tests fail
      (`"UnsecureConnection"` vs `"Unspecified"` / `"UnsupportedRequest"`, and the profile-3 OEM
      case shows `install_ca_certificate` called once); the positive-path guard stays green

Five tests added in `lib/everest/ocpp/tests/lib/ocpp/v2/functional_blocks/test_security.cpp`,
covering each refusal cause plus a guard that a CSMS root install at profile 1 with the variable
true still installs. There was no previous coverage of the reject path.

### DIFF
--- a/lib/everest/ocpp/include/ocpp/v2/functional_blocks/security.hpp
+++ b/lib/everest/ocpp/include/ocpp/v2/functional_blocks/security.hpp
@@ -7,6 +7,7 @@
 #include <variant>
 
 #include <ocpp/v2/message_handler.hpp>
+#include <ocpp/v2/ocpp_types.hpp>
 #include <ocpp/v2/ocsp_updater.hpp>
 
 namespace ocpp {
@@ -113,8 +114,8 @@ private:
 
     /// \brief Helper function to determine if a certificate installation should be allowed
     /// \param cert_type is the certificate type to be checked
-    /// \return true if it should be allowed
-    bool should_allow_certificate_install(InstallCertificateUseEnum cert_type) const;
+    /// \return std::nullopt if it should be allowed, otherwise the StatusInfo describing why it was refused
+    std::optional<StatusInfo> check_certificate_install_allowed(InstallCertificateUseEnum cert_type) const;
     void scheduled_check_client_certificate_expiration();
     void scheduled_check_v2g_certificate_expiration();
 };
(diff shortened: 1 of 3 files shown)


Pietfried (dismissed):

mlitre:
@Pietfried this requires a new approval as I had to rebase to resolve conflicts

Pietfried (approved):

PR #2609, opened 2026-08-13
test(util): make FastQuit shutdown assertion deterministic

mlitre:
## Describe your changes

`AsyncWrapperTest.DestructorShutdownPolicies` is flaky: it failed 1 of 7 full-suite runs and 1 of 3
isolated re-runs on the same machine. It fails when it is the only test running, so it is not
load-related.

The failing assertion is `EXPECT_EQ(fast_result, 0)` in the `FastQuitPolicy` half of the test, which
asserts that a task still sitting in the queue is dropped at destruction. The test pushed that task
into an empty queue, so the worker had usually already popped it before `FastQuitPolicy::shutdown`
set the done flag. The worker loop only re-checks the flag between pops, so a popped task always
runs to completion. The old code acknowledged this in a comment: "We rely on the race condition
being won by the destructor for EXPECT_EQ(0) to pass."

This is a test defect, not a production one. `FastQuitPolicy` is documented as potentially lossy and
behaves as documented; the test asserted a guarantee the policy never made.

The fix occupies the worker with a task that blocks until shutdown is signaled, then enqueues the
task under test behind it. It is then provably still queued when the destructor runs: the blocking
task returns as soon as the done flag is set, and the worker loop exits without popping the queued
task. Reading the flag needs a new fixture accessor alongside the existing `get_queue_for_test`; the
fixture is already a friend of the wrapper.

No production code changed.

Testing:
- `everest_util_tests`: 147/147 pass.
- `AsyncWrapperTest.DestructorShutdownPolicies` with `--gtest_repeat=300`: 0 failures (previously
  roughly 1 in 3 in isolation).

n/a

### DIFF
--- a/lib/everest/util/tests/async/wrapper_tests.cpp
+++ b/lib/everest/util/tests/async/wrapper_tests.cpp
@@ -48,6 +48,10 @@ public:
     template <class T> auto& get_queue_for_test(T& wrapper) {
         return wrapper.m_queue;
     }
+
+    template <class T> bool is_shutdown_signaled_for_test(T& wrapper) const {
+        return wrapper.m_done;
+    }
 };
 } // namespace everest::lib::util::testing_interface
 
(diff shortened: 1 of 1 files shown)


lib/everest/util/tests/async/wrapper_tests.cpp
cburandt, line 236: For the sake of symmetry, one could set fast_result to 1 in the first task (and later EXPECT_EQ(..., 1)) to underline, that the first task completes and the second task never starts.

cburandt (dismissed):
LGTM

(see an optional change below)

lib/everest/util/tests/async/wrapper_tests.cpp
mlitre: I added your suggestion

cburandt (approved):
LGTM - thanks for removing another source of flakyness...

PR #2617, opened 2026-08-17
fix(cbv2g): stop the bitstream writer one byte short of the end

mlitre:
## Describe your changes

`exi_bitstream_has_overflow` in `lib/everest/cbv2g/lib/cbv2g/common/exi_bitstream.c` used the wrong bound when rolling the writer over to the next byte. Once the current byte was full (`bit_count == EXI_BITSTREAM_MAX_BIT_COUNT`) it compared `byte_pos < data_size`, so at the last byte of the buffer it happily advanced `byte_pos` to `data_size` and reported no error.

The failure mode is a silent buffer overrun. The caller sizes the output buffer through `exi_bitstream_init`, and after that rollover the next write lands at `data + data_size`, one byte past the end, corrupting whatever the caller placed after the buffer. Nothing reports an error, so the encoder looks like it succeeded.

The fix compares `byte_pos + 1 < data_size` instead, so the writer only advances while the next index is still inside the buffer and otherwise returns `EXI_ERROR__BITSTREAM_OVERFLOW`. No capacity is lost: a buffer of N bytes still accepts exactly 8N bits, and only bit 8N+1 is rejected.

A new test, `lib/everest/cbv2g/tests/common/bitstream_bounds.cpp`, pins the behavior. It places a guard region immediately after the output buffer and asserts that filling the buffer bit by bit is fully accepted, that the very next bit returns `EXI_ERROR__BITSTREAM_OVERFLOW`, that a multi-bit write straddling the end is rejected rather than truncated, that a one byte buffer behaves the same way, that `byte_pos` never points past the buffer, and that the guard bytes are untouched throughout. The `common` test directory is added to `lib/everest/cbv2g/tests/CMakeLists.txt`. Note that these tests only build under `CB_V2G_BUILD_TESTS`, which everest-core does not set.

This is a prerequisite for #2584. That PR drops a defensive struct layout guard that was only safe to remove once this encoder bound was correct, so this one has to merge first.

None. Prerequisite for #2584.

### DIFF
--- a/lib/everest/cbv2g/lib/cbv2g/common/exi_bitstream.c
+++ b/lib/everest/cbv2g/lib/cbv2g/common/exi_bitstream.c
@@ -30,7 +30,10 @@ static int exi_bitstream_has_overflow(exi_bitstream_t* stream)
 {
     if (stream->bit_count == EXI_BITSTREAM_MAX_BIT_COUNT)
     {
-        if (stream->byte_pos < stream->data_size)
+        /* The current byte is full, so advance only if the NEXT index is still inside the
+           buffer. Comparing byte_pos itself lets it reach data_size, after which the caller
+           writes one past the end. */
+        if (stream->byte_pos + 1 < stream->data_size)
         {
             stream->byte_pos++;
             stream->bit_count = 0;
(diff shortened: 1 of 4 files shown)


barsnick:
@mlitre Could you re-open this for the tests (with a fix for the issue mentioned in https://github.com/EVerest/cbexigen/pull/129)? They are valuable.

PR #2618, opened 2026-08-17
fix(iso15118): never send an empty service offer

mlitre:
## Describe your changes

`ServiceDiscoveryRes` answered OK with an empty `EnergyTransferServiceList` when none of the EV's `SupportedServiceIDs` matched the EVSE's energy services, or when the EVSE had none. The element has `minOccurs=1`, so cbv2g failed to encode it (`EXI_ERROR__UNKNOWN_EVENT_CODE`) and the SECC dropped the TCP connection.

- Empty intersection with a configured offer: send the full offer with OK. Table 39 makes the filter optional, the EV decides at `ServiceSelection`.
- No energy service configured: log an error and answer plain `FAILED`. Annex A allows only `OK`, `FAILED`, `FAILED_SequenceError` and `FAILED_UnknownSession` here. The struct's default AC entry is kept so the response stays encodable.

Tests fill the two placeholder scenarios in `test/iso15118/states/service_discovery.cpp` and add the fallback and filtered zero-offer cases.

None

### DIFF
--- a/lib/everest/iso15118/src/iso15118/d20/state/service_discovery.cpp
+++ b/lib/everest/iso15118/src/iso15118/d20/state/service_discovery.cpp
@@ -70,11 +70,14 @@ handle_request(const message_20::ServiceDiscoveryRequest& req, d20::Session& ses
     res.service_renegotiation_supported = false;
     session.service_renegotiation_supported = false;
 
-    // Reset default value
-    res.energy_transfer_service_list.clear();
-
     std::vector<dt::Service> energy_services_list;
     std::vector<dt::VasService> vas_services_list;
+
+    const auto offer_all_energy_services = [&]() {
+        for (auto& energy_service : energy_services) {
+            energy_services_list.push_back({energy_service, false});
+        }
+    };
 
     // EV supported service ID's
     if (req.supported_service_ids.has_value() == true) {
@@ -96,14 +99,28 @@ handle_request(const message_20::ServiceDiscoveryRequest& req, d20::Session& ses
                 ev_energy_services.emplace_back(energy_service);
             }
         }
-    } else {
-        for (auto& energy_service : energy_services) {
-            energy_services_list.push_back({energy_service, false});
+
+        // Filtering by SupportedServiceIDs is optional (Table 39). With no match, offer every energy
+        // service the EVSE has and let the EV decide at ServiceSelection.
+        if (energy_services_list.empty() and not energy_services.empty()) {
+            logf_info("No EV supported service ID matches the offered energy services, offering all of them");
+            offer_all_energy_services();
         }
+    } else {
+        offer_all_energy_services();
         for (auto& vas_service : vas_services) {
             vas_services_list.push_back({vas_service, false});
         }
     }
+
+    if (energy_services_list.empty()) {
+        logf_error("No energy transfer service is configured, rejecting service discovery. Sending the default AC "
+                   "service to avoid encoding issues");
+        return response_with_code(res, dt::ResponseCode::FAILED);
+    }
+
+    // Reset default value
+    res.energy_transfer_service_list.clear();
 
     for (auto& conf_energy_service : energy_services_list) {
         auto& energy_service = res.energy_transfer_service_list.emplace_back();
--- a/lib/everest/iso15118/src/iso15118/d20/state/service_discovery.cpp
+++ b/lib/everest/iso15118/src/iso15118/d20/state/service_discovery.cpp
@@ -70,9 +70,6 @@ handle_request(const message_20::ServiceDiscoveryRequest& req, d20::Session& ses
     res.service_renegotiation_supported = false;
     session.service_renegotiation_supported = false;
 
-    // Reset default value
-    res.energy_transfer_service_list.clear();
-
     std::vector<dt::Service> energy_services_list;
     std::vector<dt::VasService> vas_services_list;
 
@@ -104,6 +101,17 @@ handle_request(const message_20::ServiceDiscoveryRequest& req, d20::Session& ses
             vas_services_list.push_back({vas_service, false});
         }
     }
+
+    // EnergyTransferServiceList is mandatory, so an offer of nothing cannot be expressed: the
+    // encoder rejects the empty list and the session dies on an encode error rather than the EV
+    // receiving a response it can act on. Report the mismatch instead, keeping the default list so
+    // the failure itself stays encodable.
+    if (energy_services_list.empty()) {
+        return response_with_code(res, dt::ResponseCode::FAILED_ServiceIDInvalid);
+    }
+
+    // Reset default value
+    res.energy_transfer_service_list.clear();
 
     for (auto& conf_energy_service : energy_services_list) {
         auto& energy_service = res.energy_transfer_service_list.emplace_back();
--- a/lib/everest/iso15118/test/iso15118/states/service_discovery.cpp
+++ b/lib/everest/iso15118/test/iso15118/states/service_discovery.cpp
@@ -192,7 +192,54 @@ SCENARIO("Service discovery state handling") {
     } // Todo(sl): Fill out
 
     GIVEN("Bad case - EV supported_service_ids do not match with evse supported services") {
-    } // Todo(sl): Fill out
+
+        d20::Session session = d20::Session();
+
+        message_20::ServiceDiscoveryRequest req;
+        req.header.session_id = session.get_id();
+        req.header.timestamp = 1691411798;
+
+        // AC_DER_IEC only, which the EVSE below does not offer.
+        auto& supported_service_ids = req.supported_service_ids.emplace();
+        supported_service_ids.push_back(message_20::to_underlying_value(dt::ServiceCategory::AC_DER_IEC));
+
+        std::vector<dt::ServiceCategory> supported_energy_transfer_services = {dt::ServiceCategory::AC,
+                                                                               dt::ServiceCategory::AC_BPT};
+        std::vector<dt::ServiceCategory> ev_energy_services{};
+
+        const auto res =
+            d20::state::handle_request(req, session, supported_energy_transfer_services, {}, ev_energy_services);
+
+        THEN("ResponseCode: FAILED_ServiceIDInvalid, and no empty energy service list is offered") {
+            // An empty EnergyTransferServiceList cannot be encoded: the element is mandatory, so
+            // cbv2g reports EXI_ERROR__UNKNOWN_EVENT_CODE and the session dies on an encode error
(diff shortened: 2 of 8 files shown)


lib/everest/iso15118/src/iso15118/d20/state/service_discovery.cpp
```diff
@@ -105,6 +102,17 @@ handle_request(const message_20::ServiceDiscoveryRequest& req, d20::Session& ses
...
+    // EnergyTransferServiceList is mandatory, so an offer of nothing cannot be expressed: the
+    // encoder rejects the empty list and the session dies on an encode error rather than the EV
+    // receiving a response it can act on. Report the mismatch instead, keeping the default list so
+    // the failure itself stays encodable.
+    if (energy_services_list.empty()) {
+        return response_with_code(res, dt::ResponseCode::FAILED_ServiceIDInvalid);
+    }
+
+    // Reset default value
+    res.energy_transfer_service_list.clear();
```
a-w50, line 123: Why is `energy_transfer_service_list` in `ServiceDiscoveryResponse` set to AC service in the first place.   This seems to be arbitrary.  Although, the standard defines it should contain at least one item.

a-w50: If it wouldn't, then the clear won't be necessary here.

lib/everest/iso15118/src/iso15118/d20/state/service_discovery.cpp
```diff
@@ -105,6 +102,17 @@ handle_request(const message_20::ServiceDiscoveryRequest& req, d20::Session& ses
         }
     }
 
+    // EnergyTransferServiceList is mandatory, so an offer of nothing cannot be expressed: the
+    // encoder rejects the empty list and the session dies on an encode error rather than the EV
+    // receiving a response it can act on. Report the mismatch instead, keeping the default list so
+    // the failure itself stays encodable.
+    if (energy_services_list.empty()) {
+        return response_with_code(res, dt::ResponseCode::FAILED_ServiceIDInvalid);
```
a-w50: Also note, if the request doesn't set supported service ids, but the EVSE side is configured such that is doesn't offer any services, then FAILED_ServiceIDInvalid would be technically incorrect.

lib/everest/iso15118/src/iso15118/d20/state/service_discovery.cpp
a-w50, line 86: Right now, we're not able to offer free services at all, right (because of `false` everywhere)

lib/everest/iso15118/src/iso15118/d20/state/service_discovery.cpp
```diff
@@ -105,6 +102,17 @@ handle_request(const message_20::ServiceDiscoveryRequest& req, d20::Session& ses
         }
     }
 
+    // EnergyTransferServiceList is mandatory, so an offer of nothing cannot be expressed: the
+    // encoder rejects the empty list and the session dies on an encode error rather than the EV
+    // receiving a response it can act on. Report the mismatch instead, keeping the default list so
+    // the failure itself stays encodable.
```
a-w50: details on the encoder dying probably not necessary here

lib/everest/iso15118/test/iso15118/states/service_discovery.cpp
```diff
@@ -192,7 +192,54 @@ SCENARIO("Service discovery state handling") {
     } // Todo(sl): Fill out
 
     GIVEN("Bad case - EV supported_service_ids do not match with evse supported services") {
-    } // Todo(sl): Fill out
+
+        d20::Session session = d20::Session();
+
+        message_20::ServiceDiscoveryRequest req;
+        req.header.session_id = session.get_id();
+        req.header.timestamp = 1691411798;
```
a-w50, line 200: would be nice to have helper function for this in order to reduce code duplication.  Something like
`auto [session, req] = fresh_request<message20::ServiceDiscoveryRequest>()` or similar.

lib/everest/iso15118/test/iso15118/states/service_discovery.cpp
```diff
@@ -192,7 +192,54 @@ SCENARIO("Service discovery state handling") {
...
+        const auto res =
+            d20::state::handle_request(req, session, supported_energy_transfer_services, {}, ev_energy_services);
+
+        THEN("ResponseCode: FAILED_ServiceIDInvalid, and no empty energy service list is offered") {
+            // An empty EnergyTransferServiceList cannot be encoded: the element is mandatory, so
+            // cbv2g reports EXI_ERROR__UNKNOWN_EVENT_CODE and the session dies on an encode error
+            // instead of the EV receiving a response it can act on.
+            REQUIRE(res.response_code == dt::ResponseCode::FAILED_ServiceIDInvalid);
+            REQUIRE(res.energy_transfer_service_list.empty() == false);
+            REQUIRE(res.service_renegotiation_supported == false);
```
a-w50: requiring this, might be too strict

modules/EVSE/EvseManager/hlc_power_clamp.cpp
a-w50, line 1: I consider it overkill, to create a new compilation unit just for one utility function.  Isn't there another file, where this can be merged into.  Nevertheless, having it as a free function is good.

modules/EVSE/EvseManager/hlc_power_clamp.hpp
```diff
@@ -0,0 +1,32 @@
...
+
+#pragma once
+
+#include <optional>
+
+namespace module {
+
+struct HlcPowerClampResult {
+    double current;
+    bool limit_exceeded;
```
a-w50: using an enum here might lead to better readability (i.e. `Limit::Satisfied / Limit::Exceeded` vs `true/false`)

modules/EVSE/EvseManager/hlc_power_clamp.cpp
```diff
@@ -0,0 +1,24 @@
...
+// Copyright Pionix GmbH and Contributors to EVerest
+
+#include "hlc_power_clamp.hpp"
+
+namespace module {
+
+HlcPowerClampResult clamp_hlc_power(double requested_current, double target_voltage,
+                                    std::optional<double> present_voltage, double maximum_power) {
+    const auto measured = present_voltage.value_or(0.0);
+    const auto voltage = measured > 0.0 ? measured : target_voltage;
```
a-w50: Is this logic allowed, why can I set the voltage to `target_voltage` if I measure 0V?

modules/EVSE/EvseManager/hlc_power_clamp.hpp
```diff
@@ -0,0 +1,32 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Pionix GmbH and Contributors to EVerest
+
+#pragma once
+
+#include <optional>
+
+namespace module {
+
+struct HlcPowerClampResult {
```
a-w50: This types name is misleading, if I clamp power, I would assume to receive clamped power, not a current.

modules/EVSE/EvseManager/hlc_power_clamp.cpp
```diff
@@ -0,0 +1,24 @@
...
+
+    if (voltage <= 0.0) {
```
a-w50: can this branch happen, i.e. can `target_voltage` be 0 or negative?

a-w50: if so, one should also sanity check for a negative requested current ...

modules/EVSE/EvseManager/hlc_power_clamp.hpp
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * @param requested_current [A] what the EV asked for, already clamped against current limits.
+ * @param target_voltage [V] the voltage the session is working towards, used when the EV has
+ *        reported no usable measurement.
+ * @param present_voltage [V] what the EV reports measuring, if anything.
+ * @param maximum_power [W] what the EVSE can deliver.
+ *
+ * A present voltage of zero or below is treated as no measurement rather than as a measured
+ * zero. Taking it literally makes the computed power zero, so the limit can never be exceeded
+ * and the clamp silently stops working for the whole session. With no usable voltage from
```
a-w50: why are we allowed to ignore a present voltage of 0 volts?  What if we measure 0.00001 volts.  How would that be different?

lib/everest/iso15118/src/iso15118/d20/state/service_discovery.cpp
SebaLukas: The standard says in case of a failed message the res message should be so constructed that the exi stream is minimal as possible.
So I added for this case an energy_transfer_list with one element (AC -> Value: 1)

lib/everest/iso15118/test/iso15118/states/service_discovery.cpp
```diff
@@ -8,6 +8,23 @@ using namespace iso15118;
...
+    message_20::ServiceDiscoveryRequest req;
+};
+
+// A request whose session ID matches a new session.
+FreshRequest fresh_request() {
+    FreshRequest fresh;
+    fresh.req.header.session_id = fresh.session.get_id();
+    fresh.req.header.timestamp = 1691411798;
+    return fresh;
+}
```
SebaLukas: Please add session and req back to the unit test.
Its a good idea to refactor this a little bit. But then in one big PR with changes everywhere

SebaLukas (changes requested):
Looking good!
Only one small suggestion

lib/everest/iso15118/src/iso15118/d20/state/service_discovery.cpp
SebaLukas: In general I would not send FAILED_ServiceIDInvalid (According to the standard only ServiceDetailRes can send this ResponseCode). FAILED should be enough.
But it does matter anyway because the car should react to every FAILED_*

lib/everest/iso15118/test/iso15118/states/service_discovery.cpp
mlitre: done

SebaLukas (dismissed):

SebaLukas (changes requested):
Check again the branch, I guess you need to update 😅  Your current changes undo your latest changes in main (DerSetupConfig -> DerIecSetupConfig)

SebaLukas (approved):

Pietfried:
@mlitre Is this fix eligible for a backport?

mlitre:
> @mlitre Is this fix eligible for a backport?

Should be yes

PR #1636, opened 2025-12-10
Integrate libiso15118 v0.9.1

hikinggrass:
## Describe your changes
This contains the libiso15118 source code from its last v0.9.1 release.

This version of libiso15118 can be found at https://github.com/EVerest/libiso15118/releases/tag/v0.9.1

## Checklist before requesting a review
- [x] I have performed a self-review of my code
- [ ] I have made corresponding changes to the documentation
- [x] I read the [contribution documentation](https://github.com/EVerest/EVerest/blob/main/CONTRIBUTING.md) and made sure that my changes meet its requirements

PR #2602, opened 2026-08-12
fix(libiso15118): Rework on secc time for d20

SebaLukas:
## Checklist before requesting a review
- [x] I have performed a self-review of my code
- [ ] I have made corresponding changes to the documentation
- [x] I read the [contribution documentation](https://everest.github.io/nightly/project/contributing.html) and made sure that my changes meet its requirements

PR #2623, opened 2026-08-17
fix(iso15118): encode the selected auth mode

mlitre:
## Describe your changes

`convert(const AuthorizationRequest&, iso20_AuthorizationReqType&)` in
`lib/everest/iso15118/src/iso15118/message/authorization.cpp` set
`EIM_AReqAuthorizationMode_isUsed = true` unconditionally, ignoring `in.authorization_mode`. A
request whose variant held `PnC_ASReqAuthorizationMode` therefore went out on the wire as a
well formed EIM request: the struct said PnC, the bytes said EIM, and nothing reported a
problem.

This code came from main in #1636 and is shared with the SECC, so the fix is split out to land
on its own rather than waiting on EV-side review.

The fix dispatches on the variant through a named `AuthorizationModeVisitor`, modeled on the
sibling visitor in `authorization_setup.cpp`. It is exhaustive by construction: a future
variant alternative becomes a compile error rather than silently routing to one arm.

PnC stays unimplemented, and the PnC arm now throws
`std::runtime_error("PnC authorization mode not implemented")` instead of setting a flag. That
part is deliberate and worth explaining, because simply setting the PnC flag is not safe:
`init_iso20_AuthorizationReqType` clears only the two `_isUsed` bits,
`init_iso20_PnC_AReqAuthorizationModeType` is a no-op, and `serialize_to_exi` declares
`iso20_exiDocument doc;` uninitialized. Setting the flag therefore leaves the encoder reading
indeterminate stack memory for `Id`, `GenChallenge` and `ContractCertificateChain`. Measured
with a standalone harness against the vendored cbv2g: a zeroed stack yields `-150`, a poisoned
stack yields `-111`, and a reachable state makes the encode **succeed** and emit uninitialized
stack bytes into an `AuthorizationReq`. Throwing makes the rejection deterministic and
independent of stack contents.

No production caller constructs a PnC request variant today, on main or in any open EV branch,
so this changes no live behavior. It closes a latent encoder defect and makes the unimplemented
path fail loudly when someone does reach it.

Tests, in `lib/everest/iso15118/test/exi/cb/iso20/authorization.cpp`:
- convert-level assertions pinning both flags in both directions, so "encodes PnC" is
  distinguishable from "sets no flag at all"
- `REQUIRE_THROWS_MATCHES` on the concrete type and message, which rejects both the original
  behavior (no throw) and an earlier attempt at this fix (which threw the emergent
  `Could not encode exi: -150` off uninitialized memory)
- the pre-existing byte-exact EIM serialization test is untouched and still passes

Full iso15118 suite: 219 passed, 0 failed.

None. Prerequisite for the EV-side stack (#2577 through #2584), which depends on this landing
first.

### DIFF
--- a/lib/everest/iso15118/src/iso15118/message/authorization.cpp
+++ b/lib/everest/iso15118/src/iso15118/message/authorization.cpp
@@ -2,7 +2,9 @@
 // Copyright 2023 Pionix GmbH and Contributors to EVerest
 #include <iso15118/message/authorization.hpp>
 
+#include <stdexcept>
 #include <type_traits>
+#include <variant>
 
 #include <iso15118/detail/variant_access.hpp>
 
@@ -35,6 +37,23 @@ template <> void convert(const struct iso20_AuthorizationResType& in, Authorizat
     convert(in.Header, out.header);
 }
 
+struct AuthorizationModeVisitor {
+    AuthorizationModeVisitor(iso20_AuthorizationReqType& out_) : out(out_){};
+    void operator()([[maybe_unused]] const datatypes::EIM_ASReqAuthorizationMode& in) {
+        CB_SET_USED(out.EIM_AReqAuthorizationMode);
+        init_iso20_EIM_AReqAuthorizationModeType(&out.EIM_AReqAuthorizationMode);
+    }
+    void operator()([[maybe_unused]] const datatypes::PnC_ASReqAuthorizationMode& in) {
+        // Todo(mlitre): none of Id, GenChallenge and ContractCertificateChain are converted,
+        // so in.id and in.gen_challenge are discarded. Reject the request until they are,
+        // rather than let a PnC selection go out on the wire as EIM.
+        throw std::runtime_error("PnC authorization mode not implemented");
+    }
+
+private:
+    iso20_AuthorizationReqType& out;
+};
+
 template <> void convert(const AuthorizationRequest& in, iso20_AuthorizationReqType& out) {
     init_iso20_AuthorizationReqType(&out);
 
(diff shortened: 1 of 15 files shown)


lib/everest/iso15118/src/iso15118/message/authorization.cpp
```diff
@@ -35,15 +37,33 @@ template <> void convert(const struct iso20_AuthorizationResType& in, Authorizat
     convert(in.Header, out.header);
 }
 
+struct AuthorizationModeVisitor {
```
corneliusclaussen, line 40: as far as i can see this is a duplicate definition of iso15118::message_20::AuthorizationModeVisitor:
the other one is in lib/everest/iso15118/src/iso15118/message/authorization_setup.cpp:48.

They seem to differ by arguments, but the concept seems very fragile to me, if you change parameters in one place the linker may pick the wrong one. This is even more true for e.g. the destructor that has nor arguments, the linker may choose whatever copy it finds first once it is no longer trivially destructable. This concept was not invented by this PR, but i think it should be fixed in all places (maybe in a separate PR), I consider this dangerous as this is not obvious to developers that are not aware of this split between the two files and the unintended consequences.

My suggestion would be to either give them dedicated names if you need them globally or put them in anonymous namespaces if they are only used in their respective file to make them local. I can make a suggestion in a follow up PR for the other places as well.

corneliusclaussen (dismissed):
approved, but follow up PR is coming to address the double definition issue.

corneliusclaussen:
Rebased onto #2602 (`fix/libiso15118-secc-time-microseconds`) and force-pushed as layer 2 of GitHub stack #2776. Same commit, new head 4991dce13. Base branch is now `fix/libiso15118-secc-time-microseconds`, not main: please rebase onto that branch from now on, a rebase onto main breaks the stack. Local copy: `git fetch && git reset --hard origin/fix/iso15118-authorization-encode-mode`.

corneliusclaussen (dismissed):
lgtm

a-w50 (approved):
Better bail, than silently fail.

The ODR stuff is fixed in the next pr in this stack.  The "todo" is important, so we won't forget to handle this case.  In the long run a proper "exception/error" handling strategy should be worked out for this library.

corneliusclaussen (approved):

Pietfried:
@mlitre Is this fix eligible for a backport?

PR #2628, opened 2026-08-19
feat(EnergyManager): power meter measurement tracking broker

marcnulens-pionix:

### DIFF
--- a/modules/EnergyManagement/EnergyManager/BrokerMeasurementTracking.cpp
+++ b/modules/EnergyManagement/EnergyManager/BrokerMeasurementTracking.cpp
@@ -43,10 +43,34 @@ std::optional<float> BrokerMeasurementTracking::compute_tracking_limit_W() {
         return std::nullopt;
     }
 
+    // Do not arm or consume the per-session state while no session is active. The optimizer
+    // runs continuously and constructs a broker for every EVSE on every run; without this
+    // guard the initial-current flag would be consumed during Unplugged runs and a real
+    // session would start at the minimum current instead of the configured initial current.
+    if (request.evse_state.has_value() and (request.evse_state.value() == types::energy::EvseState::Unplugged or
+                                            request.evse_state.value() == types::energy::EvseState::Finished)) {
+        return std::nullopt;
+    }
+
     const auto& limits = request.schedule_import[0].limits_to_root;
-    const auto phase_count =
+
+    if (not limits.ac_max_current_A.has_value()) {
+        // Tracking is AC-only: the amps-to-watts conversion below has no meaning for a
+        // watt-only (typically DC) node, and a watt budget derived from an AC starting
+        // current would cap it arbitrarily. Leave such nodes on their static limits.
+        return std::nullopt;
+    }
+
+    // The budget is expressed in the watt units the trading engine prices with. Under the
+    // default switch_1ph_3ph_mode of Never, BrokerFastCharging prices every purchase at
+    // ac_max_phase_count phases regardless of how many are active, so the budget must use
+    // the same convention or the minimum-current purchase would not fit into it.
+    const auto pricing_phases =
         limits.ac_max_phase_count.has_value() ? limits.ac_max_phase_count.value().value : ASSUMED_PHASE_COUNT;
-    const float watt_per_ampere = static_cast<float>(phase_count) * local_market.nominal_ac_voltage();
+    const auto active_phases = request.schedule_import[0].limits_to_root.ac_number_of_active_phases.has_value()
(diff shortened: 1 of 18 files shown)


modules/EnergyManagement/EnergyManager/BrokerMeasurementTracking.cpp
mlitre, line 1: I am not quite sure we are going in the right direction here. I think we do need to create a new broker, but I think for now it should mimick what the fast charger broker does and just keeps track of the measurements and is able to notice whether a particular leaf is consuming less than what has been allotted to it. From there we should start playing around with things like if below for x amount of time then start redistributing

modules/EnergyManagement/EnergyManager/manifest.yaml
```diff
@@ -76,6 +76,29 @@ config:
       Set to 0 to disable time based hysteresis.
     type: integer
     default: 600
+  use_power_meter_tracking:
```
mlitre: Probably let's use a string enum that defines which type of broker we want to use and by default uses the current one

modules/EnergyManagement/EnergyManager/manifest.yaml
```diff
@@ -76,6 +76,29 @@ config:
...
+      If true, use a broker that tracks live power meter measurements instead of
+      relying purely on the static fuse limits. The broker requests
+      power_meter_tracking_initial_current_A on the first optimizer run of a
+      session and afterwards limits its request to the measured power plus
+      power_meter_tracking_margin_W, floored at the minimum current the EVSE can
+      signal. Applies per connector.
+      If false (default) the static limit behaviour is used unchanged.
+    type: boolean
+    default: false
+  power_meter_tracking_initial_current_A:
```
mlitre: I think if we change approach as mentioned in my other comment this won't be necessary

modules/EnergyManagement/EnergyManager/manifest.yaml
```diff
@@ -76,6 +76,29 @@ config:
...
+    description: >-
+      Current per phase requested on the first optimizer run of a charging session
+      when use_power_meter_tracking is enabled [A].
+    type: number
+    default: 16.0
+  power_meter_tracking_margin_W:
```
mlitre: I think this will be required once we start adding the redistribution to give some margin. However, I think we should make this some sort of factor so that basically if we get close to a site limit we can more slowly redistribute and the further away from the site limit we are we can more quickly redistribute energy.

modules/EnergyManagement/EnergyManager/tests/measurement_tracking_tests.cpp
mlitre, line 1: Maybe I am mistaken but I didn't see any tests that make sure we never exceed site limit / fuse limit

mlitre (changes requested):

modules/EnergyManagement/EnergyManager/BrokerMeasurementTracking.cpp
mlitre: Maybe a diagram explaining the trading and how it is modified could help here

mnulens: I reworked the broker along these lines. The only thing it adds is that it keeps track of what each    
 connector's power meter reports during a session: total power, and per-phase power and current where the meter provides them. It is measuring only, not yet redistributing the power.

modules/EnergyManagement/EnergyManager/manifest.yaml
mnulens: Done. The boolean is replaced by a broker_strategy option with the values FastCharging (the default, current behaviour) and PowerRedistribution.

modules/EnergyManagement/EnergyManager/manifest.yaml
mnulens: removed.

modules/EnergyManagement/EnergyManager/manifest.yaml
mnulens: removed

modules/EnergyManagement/EnergyManager/tests/measurement_tracking_tests.cpp
mnulens: The new broker measures only, no redistribution of energy yet. so no tests for this yet.

modules/EnergyManagement/EnergyManager/BrokerMeasurementTracking.cpp
mnulens: Made a separate documentation on this.

PR #2629, opened 2026-08-19
feat(EnergyManager): windowed power meter aggregation

marcnulens-pionix:
## Checklist before requesting a review
- [ ] I have performed a self-review of my code
- [ ] I have made corresponding changes to the documentation
- [ ] I read the [contribution documentation](https://everest.github.io/nightly/project/contributing.html) and made sure that my changes meet its requirements

PR #2630, opened 2026-08-19
feat(EnergyManager): measurement based power boosting

marcnulens-pionix:

### DIFF
--- a/modules/EnergyManagement/EnergyManager/docs/index.rst
+++ b/modules/EnergyManagement/EnergyManager/docs/index.rst
@@ -48,7 +48,8 @@ EVSE:
 - ``FastCharging`` (default): allocate as much as the limits allow.
 - ``PowerRedistribution``: currently trades identically to ``FastCharging`` and
   additionally observes the connector's live power meter reading once per optimizer
-  run. Allocations are never modified by the observation.
+  run. Allocations are never modified by the observation. From the observations it
+  infers where power could be redistributed and logs that (see below).
 
 The measurement is taken from the EVSE's own power meter, reported through the
 ``energy_usage_leaves`` field of the energy flow request (with ``energy_usage_root``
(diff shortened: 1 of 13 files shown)


interfaces/energy_manager.yaml
```diff
@@ -1,2 +1,12 @@
 description: This interface defines the global EnergyManager
-vars: {}
+vars:
+  power_can_be_reduced:
```
mlitre, line 3: I am not quite sure I understand why we need this var, this seems to be all internal to the energy manager and doesn't require anybody else to be aware. Am I missing something?

modules/EnergyManagement/EnergyManager/docs/index.rst
```diff
@@ -113,3 +113,83 @@ external limit.
    * - ``power_meter_tracking_margin_W``
      - ``200.0``
      - Margin added to the measured power to form the tracking limit.
+
+Measurement based boosting
```
mlitre, line 117: Overall I am not a fan of calling this boosting in general. The naming is very confusing as this is more of a load balancing or load redistribution.

Also once again the documentation is a little too verbose to easily follow. We should have some basic understanding on how the energy is being rebalanced and then how we can tweak the parameters and what that would do

modules/EnergyManagement/EnergyManager/tests/include/EnergyManagerTestHelpers.hpp
```diff
@@ -33,6 +33,16 @@ inline EnergyManagerConfig make_default_config() {
     c.use_power_meter_tracking = false;
     c.power_meter_tracking_initial_current_A = 16.0;
     c.power_meter_tracking_margin_W = 200.0;
+    // WP1.b's field. It must be set here even though WP1.b introduced it: WP1.b never
+    // touches this helper (the file does not exist on its branch) and WP1.a's version
```
mlitre, line 37: Unrelated comment / context

modules/EnergyManagement/EnergyManager/manifest.yaml
```diff
@@ -109,6 +109,36 @@ config:
...
+      allocation exceeds the measured consumption by more than this much.
+      Only effective when use_power_meter_tracking is enabled.
+      Note for the power_can_be_reduced flag: with measurement tracking active the
+      allocation exceeds consumption by roughly the number of charging connectors
+      multiplied by power_meter_tracking_margin_W, so this value must be set above
+      that product for the flag to carry information. The 500W default suits one or
+      two connectors at the default 200W margin; larger clusters need a larger value.
+    type: number
+    default: 500.0
+  boost_step_A:
```
mlitre, line 126: Same again as previous config, I think this should be a factor instead of a fixed step as we would prefer stepping faster if we have a large margin and slower once we get closer to the limit

modules/EnergyManagement/EnergyManager/manifest.yaml
```diff
@@ -109,6 +109,36 @@ config:
...
+    description: >-
+      Amount by which the tracking limit is widened per boost step [A per phase].
+      The same amount is released again per cycle once the headroom disappears.
+      Set to 0 to disable boosting entirely.
+    type: number
+    default: 2.0
+    minimum: 0.0
+  boost_hysteresis_cycles:
```
mlitre, line 134: Up for debate, but I don't if we want to track cycles or seconds

modules/EnergyManagement/EnergyManager/MeasurementTrackingState.cpp
mlitre, line 1: Not quite sure why we extracted a new class here

mlitre (changes requested):

interfaces/energy_manager.yaml
corneliusclaussen: Not sure what the original intent of this var is, but what would be helpful is the information that more energy would be available but the EV does not want it (i.e. a it is not the evse's fault that this is slow on the display). This however should more be in the evse_manager api per port

modules/EnergyManagement/EnergyManager/docs/index.rst
corneliusclaussen: I don't understand the concept here - When using use_power_meter_tracking in my understanding energy manager basically reduces the requested power to actual consumption + margin. This means the requested / bought power is following the actual consumption up and down. If an EV consumes only 10A even if 32A are the hard wiring limit, it will basically get 12A (with margin) advertised. If it switches to 12A it will get 14A advertised etc. The remaining budget will be automatically redistributed to other EVs already. What is reserved at the grid connection root level? The sum of all allocations (i.e. 14A, with margins included) or is it assumed that EVs do not use the margin and it is ignored when travelling through the tree? I don't really understand what boosting is needed for - the allocation on the grid connection root level point can just be calculated from the actual assignments?

interfaces/energy_manager.yaml
mnulens: Removed the var. The file is back to identical                                                                                                                  
  with main, and the inference result stays inside the energy manager.

modules/EnergyManagement/EnergyManager/docs/index.rst
mnulens: Renamed throughout to PowerRedistribution. No "boost" left in the code,                                                                                                                  
  manifest or docs, and I'll retitle the PR and branch to match.                                                                                                                            
  Trimming this section the way I did the aggregation one on #2629.

modules/EnergyManagement/EnergyManager/tests/include/EnergyManagerTestHelpers.hpp
mnulens: Removed.

modules/EnergyManagement/EnergyManager/manifest.yaml
mnulens: Done. boost_step_A is gone; power_redistribution_gain is a factor on the                                                            headroom beyond the deadband, so the step is large far from the grid limit and vanishes as it is approached.

modules/EnergyManagement/EnergyManager/manifest.yaml
mnulens: Went with seconds: power_redistribution_hold_time_s, measured against the optimizer's start_time so it doesn't shift when update_interval changes. Minimum is 5 s, the IEC 61851-1 duty-cycle response window.

modules/EnergyManagement/EnergyManager/MeasurementTrackingState.cpp
mnulens: Removed the class. The per-session state lives in BrokerContext alongside the rest of it.

PR #2631, opened 2026-08-19
feat(EnergyManager): per phase trading and phase symmetry

marcnulens-pionix:


types/energy.yaml
```diff
@@ -61,6 +82,15 @@ types:
         description: Max current limit (per phase) in Ampere. AC only. Positive number.
         type: object
         $ref: /energy#/NumberWithSource
+      ac_max_current_per_phase_A:
```
mlitre, line 85: Let's edit this object so that it has 3 fields with L1 / L2 / L3 of type NumberWithSource

types/energy.yaml
```diff
@@ -101,6 +131,17 @@ types:
         description: Max current limit (per phase) in Ampere. AC only. Negative numbers mean export to grid direction.
         type: object
         $ref: /energy#/NumberWithSource
+      ac_max_current_per_phase_A:
```
mlitre, line 134: Following up on previous comment we can just create this optional object here to have three fields L1/L2/L3 that reuses NumberWithSource

types/energy.yaml
```diff
@@ -13,6 +13,27 @@ types:
       source:
         description: Source of the value
         type: string
+  NumberWithSourcePerPhase:
```
mlitre, line 16: Instead of creating a new type I think we should just extend the other object to have specific L1/L2/L3 fields.
So we can remove this

modules/EnergyManagement/EnergyManager/docs/index.rst
mlitre, line 1: I think some formatting broke here

modules/EnergyManagement/EnergyManager/EnergyManager.hpp
```diff
@@ -43,6 +43,8 @@ struct Conf {
     double boost_threshold_W;
     double boost_step_A;
     int boost_hysteresis_cycles;
+    bool phase_symmetry_enabled;
+    double max_phase_imbalance_A;
```
mlitre, line 47: I am wondering what unit is best here if we want mA or A

modules/EnergyManagement/EnergyManager/Broker.cpp
mlitre, line 1: I think there is something important missing here telling the broker which phases are actually used by the evse. So I am wondering here if in a way we don't need to actually change the flow request to indicate that we have requests on specific phases so that we know when we should allocate or not.

mlitre (changes requested):

modules/EnergyManagement/EnergyManager/docs/index.rst
```diff
@@ -93,6 +92,80 @@ Three properties of the implementation are worth knowing when reading the code:
...
+freed on one phase is offered to other connectors on that same phase, within the same
+optimizer cycle.
+
+Backwards compatibility is strict. When no node in the tree declares per phase limits,
+nothing changes: the field never appears in a request or a result, and the allocation is
+identical to before. Where only one side of a comparison carries per phase data, the
+missing side is derived from the symmetric limit, and neither side may ever widen the
+other -- the effective limit on a phase is always the smaller of the two.
+
+The enforced result always populates the symmetric ``ac_max_current_A`` as well, set to
```
corneliusclaussen, line 116: I read this a couple of times but I'm unsure if I understand this correctly - set to the "most heavily loaded phase" - does this mean the minimum of the 3 individual abs(phase limits)?

modules/EnergyManagement/EnergyManager/docs/index.rst
```diff
@@ -93,6 +92,80 @@ Three properties of the implementation are worth knowing when reading the code:
...
+The energy types say how *many* phases a connector uses
+(``ac_number_of_active_phases``, ``ac_max_phase_count``) but never *which* ones. A
+connector charging on ``n`` phases is therefore modelled as occupying L1..Ln, with two
+consequences stated plainly:
+
+* **Within one connector the allocation is symmetric.** An EVSE has a single PWM duty
+  cycle -- one current setpoint for all its phases -- so it cannot be given 32 A on L1 and
+  10 A on L2. The binding constraint for a connector is the *minimum* available current
+  across the phases it occupies, applied equally to each. Per phase differentiation exists
+  between connectors and between nodes, never inside a single connector.
```
corneliusclaussen, line 134: Afaik iso-20 supports different limits per phase?

modules/EnergyManagement/EnergyManager/docs/index.rst
```diff
@@ -93,6 +92,80 @@ Three properties of the implementation are worth knowing when reading the code:
...
+* **Two single phase connectors are both modelled on L1.** That is conservative: it can
```
corneliusclaussen, line 135: What about a two phase EV on a 3ph connector? Or a single phase on a 3ph connector?

modules/EnergyManagement/EnergyManager/docs/index.rst
```diff
@@ -93,6 +92,80 @@ Three properties of the implementation are worth knowing when reading the code:
...
+unreachable, so a warning indicates a real problem worth investigating.
+
+Note that symmetry enforcement can prevent a session from *starting*: if the ceiling
+cannot accommodate a connector's ``ac_min_current_A``, the purchase is refused rather than
+granting a current the EVSE cannot signal. That is intended -- it is a hard electrical
+constraint, not a preference.
+
+Phase identity is assumed consistent throughout the tree: L1 at a connector is assumed to
+be L1 at the grid connection. Mapping physically rotated phases is handled separately by
+the power meter driver's phase rotation map.
```
corneliusclaussen, line 168: Shouln't the phase rotation matrix be part of the EnergyNode instead? In realistic installations this will be rotated at different levels to ensure not all single phase ports end up at the L1 of the grid connection?

PR #2634, opened 2026-08-19
perf(LemDCBM400600): reduce HTTP and polling overhead

djchhp:
## Describe your changes

Reduces the HTTP and polling overhead of the LemDCBM400600 driver, and makes the publication rate of
the `powermeter` var independent of the latency of the device.

The rate matters beyond throughput: consumers watch this stream for gaps. `EvseManager` aborts a DC
cable check as soon as it sees no power supply measurement for two seconds, so with a 1 s poll
interval the margin was one second minus whatever the device and network happened to take.

- Reuse HTTP(S) connections across requests via a libcurl share (`http_connection_reuse`, default
  true). Previously every request opened a fresh TCP connection, plus a full TLS handshake when TLS
  is enabled. Requests using the share are serialized, since libcurl does not support using a shared
  connection cache from concurrent transfers - locking callbacks do not change that.
- Run the poll loop at a fixed rate: the request duration is absorbed into the wait instead of being
  added to it, so the period stays at `poll_interval_ms` (default 1000, as before) rather than
  growing to interval + request duration.
- Fetch the fallback OCMF record *after* publishing the live measurements, so its round trip is no
  longer inside the gap between two publications. It can additionally be throttled
  (`transaction_ocmf_fetch_interval_s`, default 0 = every poll as before), and the throttle records
  attempts rather than successes, so a failing OCMF endpoint cannot bypass it.
- Log a warning when the gap between two publications exceeds 1.5x the interval, including the
  measured gap and the request duration, so a slow device or network can be identified from the logs
  of a charger in the field.

All defaults preserve the previous behavior except connection reuse, which is enabled by default.

Co-authored-by: Jan Christoph Habig (@djchhp), who wrote the original version of this PR.

None.

### DIFF
--- a/modules/HardwareDrivers/PowerMeters/LemDCBM400600/main/lem_dcbm_400600_controller.cpp
+++ b/modules/HardwareDrivers/PowerMeters/LemDCBM400600/main/lem_dcbm_400600_controller.cpp
@@ -140,12 +140,7 @@ LemDCBM400600Controller::start_transaction(const types::powermeter::TransactionR
         this->current_transaction_id = value.transaction_id;
         this->need_to_stop_transaction = true;
         // make sure the fallback OCMF record for this transaction is fetched on the next poll
-        {
-            auto ocmf_fetch_state = this->ocmf_fetch_state.handle();
-            ocmf_fetch_state->last_fetch =
-                std::chrono::steady_clock::now() - std::chrono::seconds(this->config.transaction_ocmf_fetch_interval_s);
-            ++ocmf_fetch_state->transaction_generation;
-        }
+        this->ocmf_fetch_state.handle()->fetch_due = true;
     } catch (DCBMUnexpectedResponseException& error) {
         const std::string error_message =
             fmt::format("Failed to start transaction {}: {}", value.transaction_id, error.what());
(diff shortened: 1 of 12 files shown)


modules/HardwareDrivers/PowerMeters/LemDCBM400600/main/lem_dcbm_400600_controller.cpp
```diff
@@ -134,11 +134,13 @@ LemDCBM400600Controller::start_transaction(const types::powermeter::TransactionR
...
             }
         }
-        call_with_retry([this, value]() { this->request_device_to_start_transaction(value); },
+        call_with_retry([this, &value]() { this->request_device_to_start_transaction(value); },
                         this->config.transaction_number_of_http_retries,
                         this->config.transaction_retry_wait_in_milliseconds);
         this->current_transaction_id = value.transaction_id;
         this->need_to_stop_transaction = true;
+        // make sure the fallback OCMF record for this transaction is fetched on the next poll
+        this->last_ocmf_fetch = {};
```
corneliusclaussen: This technically only fetches on the next poll when uptime > interval since the steady clock resets at boot. I think this is acceptable though as it only matters on really long intervals (e.g. if you configure 15 minutes interval but the charger takes only 1 minute to boot and charge, the first poll will be done 14 min into the session). We could make it an optional to have a dedicated state of "not set", but I'm also ok as is.

modules/HardwareDrivers/PowerMeters/LemDCBM400600/main/lem_dcbm_400600_controller.hpp
```diff
@@ -102,6 +106,8 @@ class LemDCBM400600Controller {
     bool need_to_stop_transaction = false;
     std::string current_transaction_id;
     types::units_signed::SignedMeterValue current_signed_meter_value;
+    // when the fallback OCMF record was last fetched; epoch value forces a fetch on the next poll
+    std::chrono::steady_clock::time_point last_ocmf_fetch{};
```
corneliusclaussen: This is not protected by a mutex but accessed from two threads as far as i can see:  live_measure_publisher_thread -> get_powermeter(), read access and command handler thread when starting a transaction. should be fixed.

corneliusclaussen:
```suggestion
    std::chrono::steady_clock::time_point last_ocmf_fetch{};
    std::mutex last_ocmf_fetch_mutex; ///< protects last_ocmf_fetch, accessed by the live measure publisher thread and the command handler thread
```

modules/HardwareDrivers/PowerMeters/LemDCBM400600/main/lem_dcbm_400600_controller.hpp
```diff
@@ -6,6 +6,7 @@
 
 #include "http_client_interface.hpp"
 #include "lem_dcbm_time_sync_helper.hpp"
+#include <chrono>
 #include <functional>
 #include <generated/interfaces/powermeter/Implementation.hpp>
```
corneliusclaussen, line 12: Needed for the `std::mutex` suggested below.

```suggestion
#include <generated/interfaces/powermeter/Implementation.hpp>
#include <mutex>
```

modules/HardwareDrivers/PowerMeters/LemDCBM400600/main/lem_dcbm_400600_controller.cpp
corneliusclaussen:
```suggestion
        {
            std::lock_guard<std::mutex> lock(this->last_ocmf_fetch_mutex);
            this->last_ocmf_fetch = {};
        }
```

modules/HardwareDrivers/PowerMeters/LemDCBM400600/main/lem_dcbm_400600_controller.cpp
```diff
@@ -293,14 +295,17 @@ types::powermeter::Powermeter LemDCBM400600Controller::get_powermeter() {
...
+    if (this->need_to_stop_transaction and
+        now - this->last_ocmf_fetch >= std::chrono::seconds(this->config.transaction_ocmf_fetch_interval_s)) {
         // if there is no ongoing transaction, we do need to fetch the signed meter value to have it available
         // for the upper layers, otherwise we will not have the OCMF value if we lose connection to the device
         try {
             current_signed_meter_value =
                 types::units_signed::SignedMeterValue{fetch_ocmf_result(current_transaction_id), "", "OCMF"};
             current_signed_meter_value.public_key.emplace(public_key_ocmf);
             current_signed_meter_value.timestamp.emplace(powermeter_result.timestamp);
+            this->last_ocmf_fetch = now;
```
corneliusclaussen:
```suggestion
    const auto now = std::chrono::steady_clock::now();
    bool should_fetch_ocmf = false;
    {
        std::lock_guard<std::mutex> lock(this->last_ocmf_fetch_mutex);
        should_fetch_ocmf =
            now - this->last_ocmf_fetch >= std::chrono::seconds(this->config.transaction_ocmf_fetch_interval_s);
    }
    if (this->need_to_stop_transaction and should_fetch_ocmf) {
        // if there is no ongoing transaction, we do need to fetch the signed meter value to have it available
        // for the upper layers, otherwise we will not have the OCMF value if we lose connection to the device
        try {
            current_signed_meter_value =
                types::units_signed::SignedMeterValue{fetch_ocmf_result(current_transaction_id), "", "OCMF"};
            current_signed_meter_value.public_key.emplace(public_key_ocmf);
            current_signed_meter_value.timestamp.emplace(powermeter_result.timestamp);
            {
                std::lock_guard<std::mutex> lock(this->last_ocmf_fetch_mutex);
                this->last_ocmf_fetch = now;
            }
```

corneliusclaussen (approved):
Lgtm now

PR #2635, opened 2026-08-19
perf(EvseManager): make Node-RED debug interface optional

djchhp:
## Describe your changes

Gates all external MQTT topics for the Node-RED debug UI (`everest_external/nodered/<connector_id>/...`) behind the new config option `enable_nodered_interface`, default off. This covers the per-powermeter-sample publishes (including a full JSON serialization of every sample), the state/limit publishes and the five UI command subscriptions.

The SIL configs enable the option explicitly since they are meant to be used with the Node-RED UI, so their behavior is unchanged.

None.

### DIFF
--- a/modules/EVSE/EvseManager/manifest.yaml
+++ b/modules/EVSE/EvseManager/manifest.yaml
@@ -446,6 +446,14 @@ config:
       Maximum ampere per second limit for up/down ramping of current in charging loop.
     type: integer
     default: 25
+  enable_nodered_interface:
+    description: >-
+      Enable the external MQTT interface to the Node-RED debug UI (topics below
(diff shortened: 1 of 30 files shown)


modules/EVSE/EvseManager/manifest.yaml
```diff
@@ -446,6 +446,14 @@ config:
       Maximum ampere per second limit for up/down ramping of current in charging loop.
     type: integer
     default: 25
+  enable_nodered_interface:
+    description: >-
+      Enable the external MQTT interface to the Node-RED debug UI (topics below
```
corneliusclaussen, line 451: Maybe we should add here that it is deprecated. We should move over nodered to stable APIs instead and remove this eventually completely.

corneliusclaussen (approved):

PR #2637, opened 2026-08-19
perf(EVerestAPI): remove hot-path serialization overhead and copies

djchhp:
## Describe your changes

Removes serialization overhead and copies in the EVerestAPI hot path:

- Serialize external API messages as compact JSON by default instead of pretty-printing with 4-space indent (`EVEREST_API_JSON_INDENT` is now -1, still overridable via CMake). This shrinks every published payload and speeds up dump/parse on both sides of the wire.
- Add `deserialize_request()` for single-pass parsing of inbound command messages. Previously the full message was parsed into `generic::RequestReply`, which re-serialized the payload subtree to a string that each handler immediately parsed again (parse -> dump -> parse). Command handlers in ApiHelper and the consumer API modules now convert the payload subtree directly into the target type. `generic::RequestReply` stays for the serialize path and external users.
- `ApiHelper::publish_and_cache_variable` takes the payload by value and moves it into the latch cache, saving a full payload copy per published var.
- `AsyncApiRequestReply`: keep the boost UUID generator in a `thread_local` instead of reseeding it from the entropy source on every request; move the payload into the request json; return the reply from the impl handlers by move instead of copy.

None.

- [ ] I have made corresponding changes to the documentation (no documentation changes needed - internal implementation only; the compact-JSON default is documented via the CMake option)


modules/API/EVerestAPI/evse_security_consumer_API/evse_security_consumer_API.cpp
```diff
@@ -42,51 +43,45 @@ void evse_security_consumer_API::ready() {
...
-                    mqtt_v.publish(msg.replyTo, "true");
-                } else {
-                    mqtt_v.publish(msg.replyTo, "false");
-                }
-                return true;
+        std::string reply_to;
+        API_types_ext::CaCertificateType payload;
+        if (deserialize_request(data, reply_to, payload)) {
+            bool response = r_evse_security->call_is_ca_certificate_installed(to_internal_api(payload));
+            mqtt_v.publish(reply_to, response);
```
corneliusclaussen, line 50: This looks like a double publish, it is later on again published as true or false. 
(not introduced here but could be fixed since that region is rewritten here)

lib/everest/everest_api_types/private_include/everest_api_types/utilities/request_reply.hpp
```diff
@@ -0,0 +1,49 @@
...
+    }
+    return {};
+}
+} // namespace detail
+
+// Single-pass parsing of RequestReply command messages. The payload subtree is
+// converted directly into PayloadT; generic::RequestReply in contrast stores the
+// payload as a string, forcing a serialize/parse round trip on every command.
+// PayloadT's from_json must be visible at the point of instantiation, i.e. the
+// caller needs to include the json_codec.hpp of the payload's type domain.
```
corneliusclaussen, line 25: This header is part of the public API and I think it does not work without the private from_json. Does this header need to be in the public install folder?

corneliusclaussen (changes requested):

modules/API/EVerestAPI/evse_security_consumer_API/evse_security_consumer_API.cpp
```diff
@@ -42,51 +43,45 @@ void evse_security_consumer_API::ready() {
...
+            if (response) {
+                mqtt_v.publish(reply_to, "true");
+            } else {
+                mqtt_v.publish(reply_to, "false");
             }
```
corneliusclaussen: Suggested fix for the double publish: `publish(topic, bool)` already serializes the boolean, so the `if/else` below just republishes the same value redundantly.

```suggestion
            mqtt_v.publish(reply_to, response);
```

djchhp: fixed

lib/everest/everest_api_types/private_include/everest_api_types/utilities/request_reply.hpp
djchhp: moved to private

corneliusclaussen (approved):

PR #2645, opened 2026-08-20
feat(EvManager): optionally pause charging at 100% simulated SoC

hirokiri:
## Describe your changes

The EvManager car simulator keeps drawing full power forever once the simulated battery is full: `simulate_soc()` clamps the SoC at 100 % (and stops accumulating energy into `battery_charge_wh`), but nothing pauses or ends the charging loop, so the simulated EV keeps requesting `dc_target_current` indefinitely while reporting SoC 100 — in HLC sessions the EVCC even reports `charging_complete` while continuing to demand current. No real EV behaves like this, which makes SoC-driven test scenarios (e.g. load management against a CSMS) unrealistic.

This adds an opt-in `pause_charging_at_full_soc` config option (default `false`, so existing long-running/soak simulations keep charging indefinitely as before). When enabled and the simulated SoC reaches 100 %, HLC sessions run the `iso_pause_charging` simulation command through the modify-session command queue (replacing whatever session commands are still running, exactly like an externally injected pause), and basic IEC sessions fall back to the `PLUGGED_IN` sim state like the `pause` command. The EV stops drawing power but stays plugged in, so the EVSE reports a charging pause (OCPP `SuspendedEV`) instead of ending the transaction — matching what a real EV does after charging completes. The trigger re-arms whenever the SoC drops below 100 % again (`set_soc`, discharge), so a later charge in the same plug-in session pauses again.

Verified live in a SIL setup (2× CCS DC EVSE per station, OCPP 2.0.1 CSMS, `soc: 95`, option enabled): the session charges to 100 %, logs `Simulated battery is full, pausing charging`, and performs the regular ISO 15118 DC pause — `PowerDeliveryReq(Stop)` → `WeldingDetectionReq` → `SessionStopReq` → `D-LINK_PAUSE` — while EvseManager transitions `Charging -> StoppingCharging -> Car Paused`, OCPP sends `TransactionEvent` with `chargingState: SuspendedEV`, the periodic meter values drop to 0 W and the transaction stays open.

None — found while running distributed-charging test scenarios against a CSMS. Happy to file an issue first if preferred.

https://claude.ai/code/session_012xcvye1ASFwtXqxpnQTRv3


modules/EV/EvManager/manifest.yaml
```diff
@@ -105,6 +105,13 @@ config:
     description: Force the use of the selected payment option
     type: boolean
     default: false
+  pause_charging_at_full_soc:
+    description: >-
+      Pause charging like a real EV once the simulated battery reaches 100% SoC. The EV stays plugged in and the
```
corneliusclaussen, line 110: Just one comment: Most real EVs stop charging with terminate and not with pause if they are full. Newer versions of Mercedes for example use pause at the end of every session though, so this behaviour also exists but is not the majority.

corneliusclaussen (approved):
lgtm

PR #2651, opened 2026-08-21
fix(EvseManager): AC DER recompute and target forwarding

mlitre:
## Describe your changes

Two independent defects in `EvseManager` that between them stop an ISO 15118-20 AC DER session from
working, plus the small refactor that makes the first one a one-line change. Both defects are
present on `main`; neither was introduced by the AC DER SAE work stacked above.

**1. `ready()` clobbered a pre-`ready()` `set_der_available`.** The AC energy transfer list was
built with `der_available` hardcoded `false` and then stored unconditionally, so a
`set_der_available(true)` landing before that store was discarded until the next `hw_capabilities`
publish. The AC arm now derives from the live flag instead.

**2. AC target values were not forwarded for the DER services.** The forwarding was gated on `AC`
and `AC_BPT` only. In Dynamic control `target_active_power` is a required wire field, so an
`AC_DER_IEC` / `AC_DER_SAE` charge loop substituted zero and advertised **0 kW** for the whole
session while present power was reported correctly.

### One derivation instead of three

The first commit extracts `current_ac_energy_transfers()`. Before it, three call sites each derived
the AC list their own way, and they disagreed on when to publish. Now the derivation lives in one
place and each site states its own publish policy at the call site:

| site | policy |
|---|---|
| `hw_capabilities` subscription | publish only when the list changed |
| `ready()` | store silently; the existing publish after `call_setup` is the only one |
| `setup_AC_mode()` (#2654) | publish unconditionally, since `call_setup` just ran |

This also removes an accidental double publish: `ready()` previously called the publishing helper
before `call_setup`, then published again after it. Harmless on both HLC backends, since
`handle_setup` and `handle_update_energy_transfer_modes` are independent setup steps, but it sent
`update_energy_transfer_modes` twice at boot for no reason.

It pays off further up the stack: #2608 widens `get_supported_ac_energy_transfers()` with a DER
flavor argument, which now lands in one call site rather than two.

### Reachability

Neither defect is reachable in a shipped configuration on `main`, so this is a latent-hazard fix
rather than a live regression. `handle_set_der_available` already calls the recompute helper, so a
post-`ready()` call recovers from the clobber; the only production producer is OCPPmulti, whose
`ready()` waits on `wait_all_ready()` before flushing grid support; and no config under `config/`
wires OCPPmulti to an AC `Evse15118D20` or wires `grid_support` at all.

### Tests

`ctest -R EvseManager` 2/2 pass. The extracted derivation is covered by the existing
`EnergyTransferModesTest` suite through its single call site. No new test for the `ready()` ordering
itself: it is a `ready()`-scoped sequence with no unit seam.


modules/EVSE/EvseManager/EvseManager.cpp
```diff
@@ -981,7 +981,14 @@ void EvseManager::ready() {
 
         r_hlc[0]->call_receipt_is_required(config.ev_receipt_required);
 
-        this->update_supported_energy_transfers(initial_energy_transfers);
+        if (config.charge_mode == "AC") {
+            // Derive from the live der_available flag: a set_der_available landing before this point
+            // must not be clobbered by a list built when der_available was still false. Store only,
+            // the publish below call_setup is the single one.
```
SebaLukas: What kind of sentences are these. Could you maybe write them in a way that's a little easier to understand?

modules/EVSE/EvseManager/EvseManager.cpp
```diff
@@ -1831,11 +1842,13 @@ bool EvseManager::update_supported_energy_transfers(const types::iso15118::Energ
...
     const auto caps = *hw_capabilities.handle();
-    const auto der = der_available.load();
-    const auto energy_transfers = get_supported_ac_energy_transfers(caps, config.supported_iso_ac_bpt, der);
-    if (update_supported_energy_transfers(energy_transfers)) {
+    return get_supported_ac_energy_transfers(caps, config.supported_iso_ac_bpt, der_available.load());
+}
+
+void EvseManager::recompute_and_publish_supported_ac_energy_transfers() {
+    if (update_supported_energy_transfers(current_ac_energy_transfers())) {
         publish_and_update_supported_energy_transfers();
```
SebaLukas, line 1869: In principle okay but in the most cases we add const reference variables and use them in function calls.
For example why is `caps` defined but `der_available.load()` called in the function call?

modules/EVSE/EvseManager/EvseManager.hpp
```diff
@@ -279,6 +279,8 @@ class EvseManager : public Everest::ModuleBase {
     std::atomic_int ac_nr_phases_active{0};
 
     std::atomic<bool> der_available{false};
+    /// The AC energy transfer list implied by the current capabilities and der_available.
+    std::vector<types::iso15118::EnergyTransferMode> current_ac_energy_transfers();
```
SebaLukas: Do we need this function later as a public function?

SebaLukas (commented):
Principally looking good. But look at my comments

modules/EVSE/EvseManager/EvseManager.hpp
mlitre: Yes it is required

SebaLukas (dismissed):

PR #2654, opened 2026-08-21
fix(EvseManager): keep BPT and DER modes on AC mode switch

mlitre:
## Describe your changes

`setup_AC_mode()` hand-built the advertised ISO 15118-20 energy transfer list and published it. Two
commits here: the first fixes what it advertised, the second removes the hand-building.

**Commit 1, the advertisement.** The hand-built list was the plain AC pair only, so it dropped
`AC_BPT` and whichever AC DER service the EVSE advertises, and nothing restored them until the next
capabilities update.

**Commit 2, the derivation.** Replaces the appended list with `current_ac_energy_transfers()`, the
single derivation introduced in #2651, so this path cannot drift from the real capability set again.

That second commit also changes the **phase** set, which the first one does not, and which is worth
review attention:

| board | before | after |
|---|---|---|
| `min_phase=1, max_phase=3` | 1ph, 3ph | 1ph, **2ph**, 3ph |
| `min_phase=3, max_phase=3` | **1ph**, 3ph | 3ph |

The second row is the substantive one: the old code pushed `AC_single_phase_core` unconditionally,
which is wrong on a board whose minimum import phase count is three. The added `AC_two_phase` is
inert on this path, since `EvseV2G` drops it through its `default:` arm.

Also drops a `SetupPhysicalValues` local in that function that was declared and never used.

### Reachability

`setup_AC_mode()` is reached only from `switch_AC_mode()`, subscribed only under
`if (config.ac_with_soc)`, and no shipped DER config sets that key. So this gates nothing the upper
PRs exercise; #2651 is the one that gates DER advertisement on the normal `ready()` path. It sits
here for coherence with #2651, not out of necessity.

### Tests

`ctest -R EvseManager` 2/2 pass. No new test: the defect is in a `ready()`-scoped lambda path driven
by an ISO-2 mode switch, with no unit seam. The fix's correctness is that it now calls the same
already-tested derivation (`EnergyTransferModesTest`) instead of duplicating its logic badly.


SebaLukas (dismissed):
`setup_AC_mode()` is only used for the fake dc mode. But I guess the changes make sense also in this mode

SebaLukas (dismissed):

PR #2659, opened 2026-08-23
ci(ev-cli): report modules that do not regenerate cleanly

corneliusclaussen:
Most modules are out of sync with the ev-cli templates, so running `ev-cli mod update` pulls
unrelated changes into a PR. This adds a check that makes the current state visible.

`tests/ev-cli/check_mod_update.py` force-updates every C++ module one at a time and reports whether
the working tree stayed clean, restoring it after each module. .cpp files are never at risk -
`mod update` leaves them alone even with `--force`.

| Finding | Check |
| --- | --- |
| every module regenerates to a zero diff | pass |
| only a template refresh pending (version bump plus the `shutdown()` handlers the new template adds) | pass with a warning |
| code outside the `ev@<uuid>` blocks would be lost, or ev-cli refuses to update a module | fail |

Runs as its own PR check, with the per-module table in the job summary and as an artifact. The job
is `continue-on-error` and is not among the required status checks of `main`, so it cannot block a
merge while the backlog is worked off.

State on main when this was written: 3 clean, 8 would lose hand-written code, 99 need a template
refresh, 1 fails outright, 7 skipped (Rust/Python modules).


djchhp (approved):

a-w50 (approved):

tests/ev-cli/check_mod_update.py
```diff
@@ -0,0 +1,331 @@
...
+def run(cmd, cwd, check=False):
+    return subprocess.run(cmd, cwd=cwd, check=check, capture_output=True, text=True)
+
+
+def git(repo, *args):
+    return run(['git', *args], cwd=repo)
+
+
+def discover_modules(repo: Path):
+    """Return (cpp_modules, skipped) as lists of (name, reason)."""
```
andistorm, line 90: This is not true for cpp_modules, which is a list of str

.github/workflows/job_ev-cli-mod-update.yaml
```diff
@@ -0,0 +1,79 @@
...
+
+on:
+  workflow_call:
+    inputs:
+      runner:
+        description: 'Which runner to use'
+        required: false
+        default: 'ubuntu-24.04'
+        type: string
+      build_env_base_image_tag:
```
andistorm, line 11: The worklfow should use the build kit that is built by the build-build-kit job, so the unified image is used that can be overwritten by making changes in .ci/build-kit/docker

.github/workflows/job_ev-cli-mod-update.yaml
```diff
@@ -0,0 +1,79 @@
...
+        description: 'Tag of the build-env-base image to run in. Keep in sync with the image used by
+          the run-clang-format action of everest-ci, so that generated files are formatted with the
+          same clang-format version the lint job checks against.'
+        required: false
+        default: 'v1.5.4'
```
andistorm, line 16: I suggest to omit the default and to require always a tag

.github/workflows/on_pr.yaml
```diff
@@ -110,6 +110,13 @@ jobs:
       deploy_docs: false
       instance_name: ${{ format('{0} @ {1}', github.ref_name, github.sha) }}
 
+  ev-cli-mod-update:
```
andistorm, line 113: depend on `setup-env` to get the build kit tag and depend on `build-kit-kit` to ensure image is built before starting

.github/workflows/on_pr.yaml
```diff
@@ -110,6 +110,13 @@ jobs:
...
+    name: Call ev-cli Module Update Check workflow
```
andistorm, line 114:
```suggestion
    name: Modules (ev-cli)
```

andistorm: Match new naming scheme

.github/workflows/job_ev-cli-mod-update.yaml
```diff
@@ -0,0 +1,79 @@
...
+        type: string
+
+jobs:
+  ev-cli-mod-update:
+    name: ev-cli mod update
```
andistorm, line 21: Match new naming scheme
```suggestion
    name: check
```

.github/workflows/job_ev-cli-mod-update.yaml
```diff
@@ -0,0 +1,79 @@
+name: ev-cli Module Update Check
```
andistorm, line 1: In general the name of this check is a bit clunky, but OI have no better idea so let's keep this for know

tests/ev-cli/check_mod_update.py
andistorm, line 1: I suggest to move this file to .ci/build-kit/scripts to match the struture of the existing workflows. This scripts directory is usually mounted to the build-kit then and used from there.

andistorm: Alternatively it could be moved to applications/\<somewhere>

andistorm (changes requested):

andistorm:
Is the plan to make this a required CI job? Then we would need to fix some modules first.
Or to introduce it as non-required for now?

.github/workflows/job_ev-cli-mod-update.yaml
hikinggrass: done

.github/workflows/on_pr.yaml
hikinggrass: done

tests/ev-cli/check_mod_update.py
hikinggrass: changed

.github/workflows/job_ev-cli-mod-update.yaml
hikinggrass: I have no better idea either 🤷

.github/workflows/on_pr.yaml
hikinggrass: done

applications/utils/scripts/check_mod_update.py
hikinggrass: moved the python script itself to `applications/utils/scripts` and a `.sh` wrapper to `.ci/build-kit/scripts`

hikinggrass (approved):

PR #2660, opened 2026-08-23
ci: update all GitHub Actions to Node 24 runtimes

corneliusclaussen:
Every workflow run warns that Node.js 20 is deprecated and is being forced onto Node 24. 48 action references across 12 workflow files were still on `node20` (the DCO job on `node16`). All are bumped to the latest major, all of which declare `runs.using: node24`.

| action | before | after |
| --- | --- | --- |
| `actions/checkout` | v3, v4, v4.2.2, v6 | v7 / v7.0.1 |
| `actions/cache` | v4, v5 | v6 |
| `actions/upload-artifact` | v4, v4.4.3, v4.6.2 | v7 / v7.0.1 |
| `actions/download-artifact` | v4, v4.1.8, v5, v5.0.0 | v8 / v8.0.1 |
| `actions/setup-python` | v4 | v7 |
| `actions/create-github-app-token` | v1 | v3 |
| `docker/build-push-action` | v6 | v7 |
| `docker/metadata-action` | v5 | v6 |
| `docker/setup-buildx-action` | v3 | v4 |
| `korthout/backport-action` | v4.5.1 | v4.6.0 |
| `pmeier/pytest-results-action` | v0.7.1, v0.7.2 | v0.9.0 |
| `prince-chrismc/check-actor-permissions-action` | v3.0.2 | v4.0.0 |

Pin granularity is kept per line: bare majors stay bare, exact pins move to the exact latest patch.

These majors require Actions Runner >= 2.327.1. All jobs run on `ubuntu-24.04` or `large-ubuntu-24.04-xxl`, so the floor is met. No input or output renames apply — every input still passed is a current one.

Two behavioural changes:

- `download-artifact@v8` defaults `digest-mismatch` to `error` instead of a warning. `digest-mismatch: warn` restores the old behaviour.
- `checkout@v7` refuses to check out a fork PR head under `pull_request_target`. `on_label.yaml` is the only such workflow and its checkout passes no `ref:`, so it takes the base repo.

## Validation

`on_pr` covers 10 of the 12 files and is green. The other two were dispatched on this branch: docs build with `deploy_docs: true` ([32649145976](https://github.com/EVerest/EVerest/actions/runs/32649145976)) and instance deletion ([32649907192](https://github.com/EVerest/EVerest/actions/runs/32649907192)), both successful, and the test docs instance was deleted again. No Node 20 warnings remain in any log.

The docs job logs a pre-existing `Failed to save` cache warning: the Josev PKI keys are root-owned inside the build-kit container, so `tar` cannot read them. Identical on `on_main` run [32527441484](https://github.com/EVerest/EVerest/actions/runs/32527441484) with `cache@v4`, so it is not a `cache@v6` regression.

`job_create-coverage-badge.yml` is bumped but not validated: it has no caller while the coverage steps in `job_build-cmake-gcc.yml` stay commented out.


andistorm (approved):
🐧

PR #2673, opened 2026-08-24
fix(GenericPowermeter): raise CommunicationFault when meter is unreachable

florinmihut:
## Problem
When Modbus communication with the meter fails, the driver only logs a one-time warning and keeps republishing the last known values with fresh timestamps. Consumers cannot tell the data is stale. In the field this caused a spurious `MREC4OverCurrentFailure`: an EMS lowered the current limit while the meter was frozen, and EvseManager's soft over current check compared the new limit against stale phase currents.

## Changes
- Raise `powermeter/CommunicationFault` (severity High) after 2 consecutive failed read cycles; clear it once a full cycle succeeds again. The debounce avoids ending a charging session on a single transient bus disturbance, since EvseManager treats powermeter errors as fatal.
- Fail fast within a read cycle: on the first request timeout, skip the remaining registers — a silent device will not answer them either. With default SerialCommHub settings (500ms initial timeout, 2 retries) and a 13-register model this cuts dead-meter cycle time from ~19.5s to ~1.5s, so the fault is raised after ~5s instead of ~41s, and other devices on a shared bus are not blocked.
- Non-timeout failures (CRC errors, Modbus exception frames) still continue the cycle, since the device is alive and other registers may read fine.

## Testing
- Logical walkthrough against field traces (frozen values republished for seconds after contactor opened).
- No behavior change while communication is healthy.


modules/HardwareDrivers/PowerMeters/GenericPowermeter/main/powermeterImpl.cpp
```diff
@@ -335,12 +335,24 @@ void powermeterImpl::read_powermeter_values() {
...
+            break;
+        }
     }
     if (all_pm_registers_success) {
         pm_values_are_complete = true;
     }
 
+    this->update_communication_fault(all_pm_registers_success);
+
     if (not pm_values_are_complete) {
```
corneliusclaussen: pm_values_are_complete never resets to false as far as I can see. I do not really understand the logic here. It only gates the very first read needs to complete once after boot, then it publishes a value every time, even if stale with a new timestamp right?

florinmihut: True. I added a new mechanism without really checking the old one. Fixed now. The pm_values_are_complete is a member now and gets cleared when the CommunicationFault is raised, so publishing is suspended while the meter is unreachable and resumes after a full successful cycle. A single failed cycle still publisheshe flag is cleared on the same 2-cycle threshold that raises the error, so the two can't contradict.

PR #2674, opened 2026-08-24
feat(EvseManager): warn about stale power meter data in soft OC check

florinmihut:
## Problem
The soft over current check compares `current_drawn_by_vehicle` against the current limit without knowing how old those values are. If the power meter stops publishing, the check operates on stale data and can raise a spurious `MREC4OverCurrentFailure` after a limit reduction. A field trace showed identical phase currents being evaluated for seconds after the contactor had opened.

## Changes
Diagnostics only — the values are still used unchanged:
- Track when `current_drawn_by_vehicle` was last updated by a powermeter publication.
- If the data is older than 5s, log a warning (rate-limited to one per 5s) and append the data age to the soft overcurrent session log messages and the raised error string, so field logs are self-diagnosing.
- No warning before the first powermeter publication arrives, so setups without per-phase current data stay silent.
- Initialize `ac_with_soc_timer`/`ac_with_soc_timeout` in `SharedContext`: they were only assigned in `setup()`, leaving them indeterminate in unit tests. The layout change exposed this via an early return in `run_state_machine()`.

With the 7s soft OC timeout, a stale-data warning appears in the log before a stale-data-induced error triggers.

Companion PR on the driver side: #2673


modules/EVSE/EvseManager/Charger.cpp
```diff
@@ -1941,9 +1941,16 @@ float Charger::get_max_current_signalled_to_ev_internal() {
 void Charger::set_current_drawn_by_vehicle(float l1, float l2, float l3) {
     Everest::scoped_lock_timeout lock(state_machine_mutex,
                                       Everest::MutexDescription::Charger_set_current_drawn_by_vehicle);
+    if (l1 == shared_context.current_drawn_by_vehicle[0] and l2 == shared_context.current_drawn_by_vehicle[1] and
+        l3 == shared_context.current_drawn_by_vehicle[2]) {
+        shared_context.current_drawn_by_vehicle_identical_count++;
```
corneliusclaussen: Not sure if the identical count should really be here in main, always enabled. I think for a debugging case maybe yes, but it is quite likely in normal operation that this outputs a false warning. Not all meters report a lot of digits, so e.g. a meter that reports only one decimal is quite likely to stay contant at e.g. 15.9A during AC charging for 5 s while everyhing is good. If the driver is publishing stale data with new timestamps, I would regard that as a bug in the driver that should be fixed. I think it should be removed and reduced to timestamp checking only, or if really needed, hidden behind a default off config option.

corneliusclaussen (changes requested):

modules/EVSE/EvseManager/Charger.cpp
florinmihut: Valid point - I was thinking about that too but decided to go with it nevertheless 'cause I've not seen a meter with only one digit after comma. 
Removed that feature, if dangerous better out.

PR #2675, opened 2026-08-24
fix(iso15118): give TU-local message helpers internal linkage

corneliusclaussen:
## What

Five helper visitor type names in `iso15118::message_20` are each defined in more than one translation unit, with different contents:

| Name | Defined in |
|---|---|
| `AuthorizationModeVisitor` | `authorization.cpp`, `authorization_setup.cpp` |
| `ControlModeVisitor` | `ac_charge_loop.cpp`, `ac_der_charge_loop.cpp`, `dc_charge_loop.cpp` |
| `ModeRequestVisitor` | `ac_charge_loop.cpp`, `ac_charge_parameter_discovery.cpp`, `dc_charge_parameter_discovery.cpp`, `schedule_exchange.cpp` |
| `ModeResponseVisitor` | `ac_charge_parameter_discovery.cpp`, `dc_charge_parameter_discovery.cpp`, `schedule_exchange.cpp` |
| `ReqControlModeVisitor` | `ac_der_charge_loop.cpp` |

Two definitions of one class type in a single program is an ODR violation — ill-formed, no diagnostic required. Compiling the two `AuthorizationModeVisitor` TUs with `-flto` shows it:

```
authorization.cpp:40:8: warning: type 'AuthorizationModeVisitor' violates the C++ One Definition Rule [-Wodr]
note: a different type is defined in another translation unit
note: the first difference of corresponding definitions is field 'out'
```

This PR wraps them in anonymous namespaces, matching what `tbd_controller.cpp`, `socket_helper.cpp` and `ac_der_sae_charge_loop.cpp` already do. Internal linkage gives each file a distinct type, so there is nothing left to collide.

## Why it matters, given nothing is broken today

Nothing miscompiles right now, but only by accident. The only functions these classes emit are the constructor and `operator()`, whose mangled names carry their differing parameter types, so no two symbols collide:

```
_ZN...24AuthorizationModeVisitorC1ER26iso20_AuthorizationReqType        # authorization.cpp
_ZN...24AuthorizationModeVisitorC1ER31iso20_AuthorizationSetupResType   # authorization_setup.cpp
```

Destructors, copy/move members, and any no-argument member mangle from the class name alone, so they collide outright. All these symbols are weak/COMDAT, which the linker folds *silently* — keeping one definition for the whole program. Give any of these classes a member that owns a resource and you get a clean build plus one destructor running against two different layouts.

`-Wodr` is the only thing that detects this, and it only exists under `-flto`, which this project does not enable — so the compiler never had a chance to tell us.

## Scope

`io::SSLContext` is deliberately **not** changed: `connection_ssl.hpp:17` forward-declares it for a `unique_ptr` pimpl, so it needs external linkage. Three TU-local helpers that do not collide today (`SupportedEnergyModes`, `RequestControlModeVisitor`, `ParameterValueVisitor`) are wrapped too, so the rule is uniform.

No functional change.

## Verification

- All **61** libiso15118 translation units recompiled with the project's real flags plus `-Werror` — all clean.
- Full library links with **zero** undefined `iso15118::` symbols, confirming no cross-TU reference relied on the old external linkage.
- Symbols confirmed changed from `W` (weak global) to `t` (local) in the emitted objects.
- The `-Wodr` diagnostic above: **1 before, 0 after**.
- `clang-format` 15.0.6 from `ghcr.io/everest/everest-clang-format:v1.3.3` — no changes.

### DIFF
--- a/lib/everest/iso15118/src/iso15118/d20/state/supported_app_protocol.cpp
+++ b/lib/everest/iso15118/src/iso15118/d20/state/supported_app_protocol.cpp
@@ -25,14 +25,13 @@ constexpr std::array AcNamespaces = {ISO20_AC_NAMESPACE, ISO20_DER_IEC_NAMESPACE
 
 using ResponseCode = message_20::SupportedAppProtocolResponse::ResponseCode;
 
+namespace {
 struct SupportedEnergyModes {
     bool ac{false};
     bool dc{false};
     bool acdp{false};
     bool wpt{false};
 };
-
-namespace {
 
 message_20::SupportedAppProtocolResponse handle_request(const message_20::SupportedAppProtocolRequest& req,
                                                         const SupportedEnergyModes& modes,
--- a/lib/everest/iso15118/src/iso15118/message/ac_charge_loop.cpp
+++ b/lib/everest/iso15118/src/iso15118/message/ac_charge_loop.cpp
@@ -337,6 +337,7 @@ void convert(const datatypes::BPT_Dynamic_AC_CLResControlMode& in,
     convert(static_cast<const datatypes::Dynamic_AC_CLResControlMode&>(in), out);
 }
 
+namespace {
 struct ControlModeVisitor {
     using ScheduledCM = datatypes::Scheduled_AC_CLResControlMode;
     using BPT_ScheduledCM = datatypes::BPT_Scheduled_AC_CLResControlMode;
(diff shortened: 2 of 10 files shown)


a-w50 (dismissed):
That is alright

mlitre (dismissed):

a-w50 (approved):
An anonymous namespace changes the game,
It gives the symbol a local name.
The Translation Unit holds it tight,
Hidden completely from the linker's sight.

PR #2678, opened 2026-08-25
Improve compiler cache (ccache) performance

a-w50:
## Describe your changes
There were some autogenerated header files (containing absolute paths) and other cmake compile definitions, that were pretty much incompatible with proper compiler cache.  This PR fixes them and makes ccache for the everest build pretty fast again.


lib/everest/project_info/src/project_info.cpp
```diff
@@ -0,0 +1,26 @@
+// SPDX-License-Identifier: Apache-2.0
```
corneliusclaussen, line 1: I think this may break out of tree module builds as this lib is not installed but required by the auto generated module code?

lib/everest/framework/lib/CMakeLists.txt
```diff
@@ -107,6 +107,20 @@ target_compile_definitions(framework
...
 
+
+if (CMAKE_INSTALL_PREFIX)
+    set_source_files_properties(runtime.cpp PROPERTIES
+        COMPILE_DEFINITIONS "EVEREST_INSTALL_PREFIX=\"${CMAKE_INSTALL_PREFIX}\""
+    )
+endif()
+
+if (CMAKE_INSTALL_LIBDIR)
+    set_source_files_properties(runtime.cpp PROPERTIES
```
corneliusclaussen: This removes the EVEREST_INSTALL_PREFIX from the binary which will result in a broken installation:

strings lib/everest/framework/lib/libframework.so.0.25.0 | grep -E '^(/home/.*dist|/usr|lib|lib64)$'

main-based build (/home/cornelius/checkout/EVerest/build/dist/lib64/libframework.so.0.25.0):
/home/cornelius/checkout/EVerest/build/dist    ← EVEREST_INSTALL_PREFIX
/usr                                           ← the "/usr → /etc" special-case compare

This branch (.claude/worktrees/increase-compile-cache-perf/build/lib/everest/framework/lib/libframework.so.0.25.0):
lib64                                          ← EVEREST_INSTALL_LIBDIR
/usr                                           ← now also the *prefix fallback*

corneliusclaussen (changes requested):

lib/everest/project_info/src/project_info.cpp
a-w50: What do you mean with installed?  It is statically linked in. Maybe I don't get your point?

lib/everest/framework/lib/CMakeLists.txt
a-w50: Good catch, was assuming it appends ...

lib/everest/project_info/src/project_info.cpp
corneliusclaussen: the header is not installed as a usable header so that modules living in a separate repository can also use it? Or am I wrong here?

a-w50: You're probably right, in non EDM mode an out-of-order module will still require the installed artifacts of everest(core), i.e. all the cmake cruft.  But the project_version library will be missing as a target, it should be built again within the out-of-order context (the header file is also generated there in this context).  So either ship the library (2 files) with the everest install, or have some compile time check resorting to the auto generated header file instead of the static library.
Any (other) suggestions?

corneliusclaussen: I'm not sure what is the best way to fix this

corneliusclaussen (dismissed):

lib/everest/project_info/include/everest/project_info.hpp
```diff
@@ -0,0 +1,17 @@
...
+// Copyright Pionix GmbH and Contributors to EVerest
+#pragma once
+
+#include <string_view>
+
+namespace everest::project_info {
+
+std::string_view name() noexcept;
+
+std::string_view description() noexcept;
```
Pietfried: Unlike the other functions, `description()` doesnt have any caller, so we could remove it here and in th cpp

Pietfried (dismissed):

Pietfried (approved):

corneliusclaussen (approved):

PR #2702, opened 2026-09-01
feat(evse_manager): add configured evse_id and evse_id_din to get_evse

Pietfried:
## Describe your changes

The Evse type gains optional evse_id and evse_id_din fields (eMI3 and DIN SPEC 91286 format), filled by EvseManager from its configuration. The evse_id var is deprecated in favor of the get_evse command. Descriptions are clarified throughout: the connector_id config parameter is actually the EVSE id in the OCPP 3-tier model, and the EVSE ID belongs to the EVSE, not a connector. The external API types, the AsyncAPI spec (with a deduplicated EvseId schema) and the file hash guards are updated accordingly.


docs/source/project/releases/deprecation-index.rst
```diff
@@ -33,4 +33,11 @@ to the corresponding migration guide.
...
-     - Change `implementation_id: meter` to `implementation_id: main` in the config connection
\ No newline at end of file
+     - Change `implementation_id: meter` to `implementation_id: main` in the config connection
+   * - The `evse_id` topics of the `evse_manager_consumer_API` AsyncAPI
+       (`e2m/evse_id` and `m2e/evse_id/get`), superseded by the `get_evse`
+       command
+     - 2026.10.0
+     - 2027.04.0
+     - Use the `get_evse` command, which provides both the eMI3-format EVSE ID
+       (ISO 15118-2 Annex H) and the DIN 70121 EVSE ID (DIN SPEC 91286)
```
cburandt, line 43: Did you consider logging a deprecation warning in the evsemanager_consumer_API module when get_evse command is called? (not a trivial fix however, since the responsible code is burried in generic subscribe_latched_value_request(...))

modules/EVSE/EvseManager/manifest.yaml
```diff
@@ -4,18 +4,27 @@ description: >-
   Will be used for billing if present.
 config:
   connector_id:
-    description: Connector id of this evse manager
+    description: >-
+      ID of this EVSE within the charging station, according to the 3-tier model of OCPP
+      (charging station > EVSE > connector). The parameter is named connector_id for
+      backwards compatibility reasons only; it does not identify a connector.
```
cburandt, line 10: Shouldn't we deprecate this setting and introduce a new one? (could be optional and be ignored if left to the default; finding a good name is not easy, as "evse_id" is already taken; maybe "evse_id_3tier"?)

cburandt, line 10: Should there be a warning in the log if a configuration defines a mapping for this EvseManager module which is different?

cburandt (commented):
LGTM in general.

Find some thought below.

docs/source/project/releases/deprecation-index.rst
Pietfried: Added a warning in e42fc984d51b802e035a964068d47d2fd128d4cc

modules/EVSE/EvseManager/manifest.yaml
Pietfried: Added a warning in 2a52e33

Pietfried: Yes we should deprecate, but rather in favor of the framework wide module mapping, otherwise we keep having this definition in multiple places. I'd rather do this in a follow up PR though.

Pietfried: https://github.com/EVerest/EVerest/issues/2712

cburandt (approved):

modules/EVSE/EvseManager/manifest.yaml
cburandt: Agree. That's even better.

mlitre (approved):
lgtm!

PR #2703, opened 2026-09-01
fix(EvseManager): correct the AC target path for discharge

mlitre:
## Describe your changes

Two defects on the ISO 15118-20 AC target path, both reachable whenever an EV discharges. Both are
present on `main`; neither comes from the AC DER work stacked above.

**1. The export target was scaled by the import phase count.** `Charger::set_max_current` encodes
direction in the sign of the current (`Charger.cpp:1318-1322`), so a discharging EV arrives as a
negative `ampere`. The target was computed as `ampere * ac_nominal_voltage * max_phase_count_import`
regardless of sign. On a board with fewer export phases than import phases the station over-reports
its export capability to the EV: a 3-phase-import, 1-phase-export board offers 11 kW of discharge
when it can deliver 3.7 kW. `max_phase_count_export` is a required field on `HardwareCapabilities`,
so the correct value was already in hand.

**2. A discharge request before ISO-20 service selection threw.** The ISO-2 early return was gated
on `not selected_d20_energy_service.has_value() and ampere >= 0.0`, so a negative current with no
service selected fell through to `selected_d20_energy_service.value()` and raised
`std::bad_optional_access`. `selected_d20_energy_service` is unset until
`subscribe_selected_service_parameters` fires, so any discharge request in that window killed the
module. The direction of the current has no bearing on which protocol is in use, so the guard now
tests only whether a service was selected, and the non-negative check moved inside the early return.
ISO-2 still never receives a negative target, exactly as before, but the empty optional is no longer
reachable.

### Why here

This PR sits below the AC DER stack rather than beside it because it edits the same block #2651
widens to the DER services. Landing it underneath keeps that block conflict-free and puts the whole
negative-current story in one commit. Only #2715, which is pure `lib/everest/iso15118` prep, sits
below it.

### Tests

`ctest -R EvseManager` passes. No new test: the block is a `ready()`-scoped lambda over
`charger->signal_max_current` with no unit seam, and both changes are local corrections to
expressions inside it.


modules/EVSE/EvseManager/EvseManager.cpp
```diff
@@ -1255,16 +1255,24 @@ void EvseManager::ready() {
...
-            if (not selected_d20_energy_service.has_value() and ampere >= 0.0) {
+            // No selected service means ISO-2, whatever the direction. Gating this on a non-negative
+            // current would fall through to the value() calls below on an empty optional.
+            if (not selected_d20_energy_service.has_value()) {
                 r_hlc[0]->call_update_ac_max_current(ampere); // ISO-2
                 return;
             }
 
             if (selected_d20_energy_service.value() == types::iso15118::ServiceCategory::AC or
                 selected_d20_energy_service.value() == types::iso15118::ServiceCategory::AC_BPT) {
```
SebaLukas, line 1279: With your changes it is now possible that ISO-2 gets a negative current value. I dont know if an EV likes that.
I would add again the check if the current is equal/bigger then 0A. And add `selected_d20_energy_service.has_value()` check before accessing `selected_d20_energy_service.value()` in line 1265

SebaLukas (changes requested):

SebaLukas (dismissed):
👍

PR #2705, opened 2026-09-01
feat(evse_manager): report charge mode and HLC capability per connector

mlitre:
## Describe your changes

`get_evse`'s `Connector` gains two fields, both static and settled in `init()`:

- **`charge_mode`** (`{AC, DC}`) - the AC or DC power topology of that connector's power path.
- **`hlc_capable`** - whether an ISO 15118 session can ever occur on that connector.

OCPP needs an EVSE's topology before its device model is built, and neither existing source works:
supported energy transfer modes arrive asynchronously over MQTT and are not waited on, and reading
`charge_mode` through the config service is default-deny with an `"AC"` default indistinguishable
from "could not read anything". Reporting the fact on the interface makes it available before any
session.

The fields sit on `Connector` rather than `Evse` because that is where the fact lives: topology is a
property of a plug and its power electronics. An EVSE-level field would state it at the wrong tier
and bake in the single-connector assumption the OCPP 2.x three-tier model intends to shed.

`hlc_capable` is `EvseManager`'s own effective boot answer (`is_hlc_enabled()`: an SECC wired, a SLAC
provider wired, and configuration permitting HLC), not merely "an SECC is connected". A connector
whose SECC is wired but whose HLC is switched off reports `false`.

Both in-tree `evse_manager` providers fill the fields. `EvseManagerSimulator` reports `AC` and
`hlc_capable: false`, consistent with the `NoHlc` it already returns from every HLC-gated command.

### Required internally, optional on the external contract

`types/evse_manager.yaml` marks both fields `required`, so a consumer never has to model absence.
Module interfaces and types are internal and may change without notice, so that is not a breaking
change.

The public `evse_manager_consumer_API.yaml` adds them as **optional**, even though EVerest always
populates them, so a client validating against that schema is not newly constrained. Adding
`required` constraints there would be breaking.

`slac_enabled` and `hlc_enabled` gain default initializers: `handle_get_evse` can now read
`hlc_enabled`, and both were previously left uninitialized.


lib/everest/everest_api_types/include/everest_api_types/evse_manager/API.hpp
```diff
@@ -213,9 +213,16 @@ enum class ConnectorTypeEnum {
...
+enum class ChargeMode {
+    AC,
+    DC,
+};
+
 struct Connector {
     int32_t id;
     std::optional<ConnectorTypeEnum> type;
+    std::optional<ChargeMode> charge_mode;
+    std::optional<bool> hlc_capable;
```
djchhp, line 225: those two can be conflicting. I propose to add additional members to the ChargeMode. Also why the optional? That could be an Unknown in the enum.

mlitre: Optional is to avoid making it a breaking change for the api.
What do you mean they can be conflicting?

hikinggrass: I think the explanation is found further up the stack in lib/everest/ocpp_module_common/include/everest/ocpp_module_common/device_model/everest_device_model_storage.hpp / https://github.com/EVerest/EVerest/pull/2706/changes#diff-c979e94bf97fc8f81bbc3103a4c787e7518059b694e27f92836765f146cd13d3
```
/// Ac additionally requires that some connector is hlc_capable, because AC DER exists only as an
/// ISO 15118-20 relay. Dc does not check it: EvseManager refuses to start a DC EVSE without HLC, so a DC
/// connector reporting no HLC could not have booted.
```

mlitre: Yes that's the right explanation

hikinggrass (approved):

PR #2709, opened 2026-09-01
feat(io): rx pause/resume, tx drain hooks and coalescing for fd_event_client

corneliusclaussen:
## Describe your changes

Flow control for `fd_event_client`, needed to bridge a pty into TCP in `pionix_chargebridge`: since #2562 bounded the tx buffer, overflow dropped silently.

- `pause_rx()` / `resume_rx()` / `rx_paused()`: stop reading, the kernel buffer fills and the writer blocks. A paused stream socket reports the peer closing (EPOLLRDHUP, `poll_events::read_hungup`). Rejected without a monitored connection; a reconnect starts unpaused.
- `tx_queue_depth()` / `set_tx_drained_action()`: pause the source at a watermark, resume on drain. `reset()` and a failed connection do not drain; resume from `set_on_ready_action`.
- `tx_coalescing()`: append to the newest waiting payload within a bound. Only for policies declaring `supports_tx_coalescing` (`tcp_socket`, `pty_handler`); others do not compile.
- `fd_event_handler::modify_event_handler(fd, enable, disable)`: one `EPOLL_CTL_MOD`.
- `pty_handler`: non blocking master; would-block and partial writes return false with `get_error() == 0`. A slave that stops reading no longer stalls the loop; undrained data is rejected once the tx buffer is full (README).

No behaviour change for existing users unless they call the new functions; all in-tree consumers compile unchanged.

Tests: `fd_event_client_backpressure_test`, `pty_handler_test`, event swap in `fd_event_handler_test`, alias gate for `tx_coalescing`.

Follow-up to #2562. Base of #2723 and #2718.

### DIFF
--- a/lib/everest/io/include/everest/io/event/fd_event_client.hpp
+++ b/lib/everest/io/include/everest/io/event/fd_event_client.hpp
@@ -640,7 +640,8 @@ public:
      * @tparam P SFINAE hook, always \p ClientPolicy. Do not pass.
      * @param[in] payload The data to be transmitted
      * @param[in] max_payload_size Upper bound for a merged payload; a larger \p payload is queued on
-     * its own
+     * its own. Peer dependent: use the peer's receive window or the MSS, not larger than the socket
+     * send buffer
      * @return True if merged or buffered. A merge is accepted even with the buffer at
      * \ref max_buffered_tx_payloads; a new entry is rejected as in \ref tx
      */
@@ -653,10 +654,6 @@ public:
         if (not m_tx_buffer.empty()) {
             auto& newest = m_tx_buffer.back();
             if (newest.size() + payload.size() <= max_payload_size) {
-                // Size for the bound once instead of growing per merge.
-                if (newest.capacity() < max_payload_size) {
-                    newest.reserve(max_payload_size);
-                }
                 newest.insert(newest.end(), payload.begin(), payload.end());
                 // A waiting payload already holds the write registration or a wakeup is in flight.
                 return true;
--- a/lib/everest/io/include/everest/io/event/fd_event_client.hpp
+++ b/lib/everest/io/include/everest/io/event/fd_event_client.hpp
@@ -194,6 +194,34 @@ public:
      */
     void set_on_ready_action(std::function<void()>&& item);
 
+    /**
+     * @brief Stop monitoring the connection for readability.
+     * @details Backpressure towards the writer behind the descriptor: the kernel buffer fills and
+     *          the writer blocks. Error, hangup and write monitoring are unaffected. A stream socket
+     *          is watched for the peer closing its writing side instead (EPOLLRDHUP), reported
+     *          through the error handler; unread data is lost with the connection. Rejected while
+     *          no connection is monitored (before the first \ref sync, during a reopen or a
+     *          handshake). A new connection starts unpaused, re-apply from \ref set_on_ready_action.
+     *          Idempotent. Loop thread only, like \ref sync.
+     * @return True if reads are paused
+     */
+    bool pause_rx();
+
+    /**
+     * @brief Resume monitoring the connection for readability after \ref pause_rx.
+     * @details Level triggered: data queued while paused is dispatched on the next poll. Idempotent,
+     *          same constraints as \ref pause_rx.
+     * @return True if reads are resumed
+     */
+    bool resume_rx();
+
+    /**
+     * @brief Whether reads are paused on the current connection.
+     * @details False without a monitored connection and for every new connection until paused.
+     * @return True if paused by \ref pause_rx
+     */
+    bool rx_paused() const;
+
 protected:
     /**
      * @brief Constructor.
@@ -303,14 +331,31 @@ protected:
     /**
      * @brief Consume the error code of an error or hangup notification on \p fd.
      * @details Prefers the error of the client policy, then SO_ERROR, then the notification kind
-     *          itself (ENOTCONN for a hangup, EIO otherwise) for a descriptor that is not a
-     *          socket. Never resolves to 0, which would read as a cleared error. Reading SO_ERROR
-     *          clears it, so call this exactly once per connection.
+     *          itself (ENOTCONN for a hangup, ECONNRESET for the peer shutting its writing side,
+     *          EIO otherwise) for a descriptor that is not a socket. Never resolves to 0, which
+     *          would read as a cleared error. Reading SO_ERROR clears it, so call this exactly
+     *          once per connection.
      * @param[in] fd The file descriptor the notification arrived on
-     * @param[in] kind poll_events::hungup for a hangup, poll_events::error otherwise
+     * @param[in] kind poll_events::hungup for a hangup, poll_events::read_hungup for a peer that
+     *            shut its writing side while reads were paused, poll_events::error otherwise
      * @return A nonzero error code
      */
     int consume_poll_error(int fd, poll_events kind);
+
+    /**
+     * @brief The descriptor of the current connection.
+     * @return The descriptor, -1 without a connection
+     */
+    int connected_fd();
+
+    /**
+     * @brief Switch monitoring the connection for readability off or on.
+     * @details Implements \ref pause_rx and \ref resume_rx: swaps readability against the peer close
+     *          watch in one event handler modification, so on failure nothing changed.
+     * @param[in] paused True to stop reads, false to resume them
+     * @return True if the requested state is in force
+     */
+    bool set_rx_paused(bool paused);
 
     static poll_events default_desired_events();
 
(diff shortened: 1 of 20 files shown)


lib/everest/io/include/everest/io/event/fd_event_client.hpp
```diff
@@ -586,6 +607,65 @@ class generic_fd_event_client : public ClientInterfacePolicy, public generic_fd_
...
+     * an empty buffer sends immediately. Appending to the front payload mid partial-write is
+     * safe: the policy's tx replaces it with exactly the bytes still to send. Requires a
+     * contiguous byte-sequence PayloadT (e.g. `std::vector<std::uint8_t>`). Shares the
+     * threading contract of \ref tx.
+     * @param[in] payload The data to be transmitted
+     * @param[in] max_payload_size Upper bound for a merged payload; a \p payload larger than
+     * this is still accepted as its own entry
+     * @return True if the payload was buffered or merged, false as in \ref tx
+     */
+    bool tx_coalescing(ClientPayloadT const& payload, std::size_t max_payload_size) {
```
djchhp, line 649:
```suggestion
  template <class P = ClientPolicy, std::enable_if_t<utilities::policy_tx_trims_partial_writes_v<P>, int> = 0>
  bool tx_coalescing(ClientPayloadT const& payload, std::size_t max_payload_size) {
      static_assert(std::is_same_v<P, ClientPolicy>);
```
This function must be guarded by a policy constraint. The current code only implicitly fails on the datatype, which is not enough. E.g. TLS will certainly break. TAP also likely does not want this.

In event_client_async_policy.hpp mirror the existing pattern for enable/disable

The feature must then be enabled for the respective compatible policies.

lib/everest/io/include/everest/io/event/fd_event_client.hpp
```diff
@@ -586,6 +607,65 @@ class generic_fd_event_client : public ClientInterfacePolicy, public generic_fd_
...
     }
 
+    /**
+     * @brief Send data, coalescing it with the newest buffered payload when possible.
+     * @details Like \ref tx, but when payloads are already waiting, \p payload is appended to the
+     * newest one instead of queued on its own, as long as the merged payload stays within
+     * \p max_payload_size. For a stream transport this batches many small writes into fewer,
+     * larger segments — essential against a peer with a tiny receive window (Nagle plus a
+     * delayed ACK per mini-segment collapses throughput otherwise) — while adding no latency:
+     * an empty buffer sends immediately. Appending to the front payload mid partial-write is
```
djchhp: ... sends with the on writable event.

lib/everest/io/include/everest/io/event/fd_event_client.hpp
```diff
@@ -194,6 +194,27 @@ class generic_fd_event_client_impl : public fd_event_sync_interface, protected u
...
 
+    /**
+     * @brief Stop monitoring the connection for readability.
+     * @details Flow control towards the writer feeding this client's descriptor: while reads are
+     *          paused the kernel buffer behind the descriptor fills up and a writer into it (e.g.
+     *          on the slave side of a pty) blocks instead of the consumer having to drop. Only
+     *          readability is affected; error, hangup and on-demand write monitoring stay as they
+     *          are. A no-op without a connection, and rejected while a handshake is pending (the
+     *          handshake owns the monitored events). A new connection always starts unpaused.
+     *          Shares the threading contract of \ref sync: call it from the loop thread only.
```
djchhp:
```suggestion
     *          Call from the thread that drives sync, either directly or from a callback that thread dispatches
```

djchhp (changes requested):
See the inline comments. 

One additional note. Paused TCP will not notice a peer closed connection, or maybe not event any closing reason.

A set of tests for the new feature would be appreciate.

lib/everest/io/include/everest/io/event/fd_event_client.hpp
mlitre: This also breaks raw_socket I believe

lib/everest/io/src/event/fd_event_client.cpp
```diff
@@ -309,6 +309,37 @@ int generic_fd_event_client_impl::consume_poll_error(int fd, poll_events kind) {
...
+    return m_event_handler->modify_event_handler(fd, poll_events::read, event_modification::remove);
+}
+
+bool generic_fd_event_client_impl::resume_rx() {
+    if (m_handshake_pending()) {
+        return false;
+    }
+    int fd = -1;
+    {
+        auto client_status = m_client_status.handle();
```
a-w50: is this completely the same code as above?  If so, could be refactored into a single named and descriptive helper function

corneliusclaussen: I will push  a new version soon.

lib/everest/io/include/everest/io/event/fd_event_client.hpp
corneliusclaussen: now protected by policy

lib/everest/io/src/event/fd_event_client.cpp
```diff
@@ -102,19 +102,24 @@ bool generic_fd_event_client_impl::setup_error_event_handler() {
...
     m_local_error_text.clear();
     m_event_handler->register_event_handler(
         fd,
         [this, fd](auto events) {
             // A terminal notification wins over a pending handshake: the connection it would be
-            // negotiated on is gone.
-            if (events.count(poll_events::error) or events.count(poll_events::hungup)) {
+            // negotiated on is gone. read_hungup is monitored only while paused, see pause_rx.
+            if (events.count(poll_events::error) or events.count(poll_events::hungup) or
+                events.count(poll_events::read_hungup)) {
```
djchhp: Since this is touched here, maybe now is the moment to go for 
```suggestion
            if (utils::exists_any(events, poll_events::error,  poll_events::hungup, poll_events::read_hungup)) {
```

and adding this template to lib/util/include/everest/util/misc/container.hpp, where there are already optimized look-ups implemented.

```

  template <typename Container, typename... T>
  bool exists_any(const Container& c, const T&... vals) {
      return (exists(c, vals) || ...);
  }

```

lib/everest/io/src/event/fd_event_client.cpp
```diff
@@ -102,19 +102,24 @@ bool generic_fd_event_client_impl::setup_error_event_handler() {
...
                 // Level triggered, so the fd keeps notifying until the queued teardown removes it,
                 // and reading SO_ERROR clears it.
                 if (not m_connection_failed) {
                     m_connection_failed = true;
-                    auto const kind = events.count(poll_events::error) != 0 ? poll_events::error : poll_events::hungup;
+                    auto const kind = events.count(poll_events::error) != 0    ? poll_events::error
+                                      : events.count(poll_events::hungup) != 0 ? poll_events::hungup
+                                                                               : poll_events::read_hungup;
```
djchhp: Same here, maybe a good moment to cleanup. But I am also happy to do a follow up.
```suggestion
                     auto const kind = first_present(events, poll_events::error, poll_events::hungup)
                        .value_or(poll_events::read_hungup);
```

with this added to container.hpp

```
 template <typename Container, typename T, typename... Ts>
  std::optional<T> first_present(const Container& c, const T& first, const Ts&... rest) {
      if (exists(c, first)) {
          return first;
      }
      if constexpr (sizeof...(rest) == 0) {
          return std::nullopt;
      } else {
          return first_present(c, rest...);
      }
  }
```

lib/everest/io/include/everest/io/event/fd_event_client.hpp
```diff
@@ -586,6 +629,67 @@ class generic_fd_event_client : public ClientInterfacePolicy, public generic_fd_
...
+            return false;
+        }
+        if (not m_tx_buffer.empty()) {
+            auto& newest = m_tx_buffer.back();
+            if (newest.size() + payload.size() <= max_payload_size) {
+                // Size for the bound once instead of growing per merge.
+                if (newest.capacity() < max_payload_size) {
+                    newest.reserve(max_payload_size);
+                }
+                newest.insert(newest.end(), payload.begin(), payload.end());
```
djchhp, line 657: I think this introduces an easy to fall in trap. This code path triggers the first time a merge is attempted. If the user supplies an unreasonably high max_payload_size, let's say 1GB this throws and likely crashed the program mid session. What about just dropping 656 - 659?

I am also wondering, if a reasonable default should go into the policy, not sure? It would give some orientation and also a value for testing.

djchhp (changes requested):
Most things are just comments, but it is necessary to look into the max_payload_size issue.

lib/everest/io/include/everest/io/event/fd_event_client.hpp
corneliusclaussen: I think removing this is the correct choice. I would not put a default in the policy, because setting this needs some thinking.

lib/everest/io/src/event/fd_event_client.cpp
corneliusclaussen: ok, will clean this up

lib/everest/io/src/event/fd_event_client.cpp
corneliusclaussen: will cleanup

djchhp (approved):
looking good now.

mlitre (approved):

PR #2710, opened 2026-09-01
fix(EvseManager): request energy again after a pause by the EVSE

Pietfried:
## Describe your changes

With request_zero_power_in_idle enabled, EvseManager requested 0A in ChargingPausedEVSE. The energy manager enforced 0A, so power_available() stayed false and the NoEnergy pause reason kept the state machine paused even after resume_charging() had cleared the user pause. Charging could only continue after an unplug. A pause for missing energy could not recover at all.

Only request 0A while a pause reason without energy demand is active (UserPause or Error). Once only NoEnergy is left, request the local limits again immediately so the energy manager can hand the budget back and charging resumes.

Track the active pause reasons in energyImpl via the Charger's ChargingPausedEVSE event and document the behaviour in the manifest. Parametrize test_pwm_ac_session_paused_by_evse over both values of request_zero_power_in_idle; the true variant fails without this fix.

### DIFF
--- a/modules/EVSE/EvseManager/energy_grid/energyImpl.cpp
+++ b/modules/EVSE/EvseManager/energy_grid/energyImpl.cpp
@@ -4,6 +4,7 @@
 #include "energyImpl.hpp"
 
 #include <everest/helpers/phase_rotation.hpp>
+#include <everest/util/misc/container.hpp>
 
 #include <chrono>
 #include <cstdint>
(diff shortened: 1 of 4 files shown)


hikinggrass (dismissed):

modules/EVSE/EvseManager/energy_grid/energyImpl.cpp
```diff
@@ -142,6 +143,31 @@ void energyImpl::ready() {
...
+
+    mod->charger->signal_charging_paused_evse_event.connect(
+        [this](const types::evse_manager::ChargingPausedEVSEReasons& paused) {
+            using types::evse_manager::PauseChargingEVSEReasonEnum;
+            const auto has_reason = [&paused](const PauseChargingEVSEReasonEnum reason) {
+                return std::find(paused.reasons.begin(), paused.reasons.end(), reason) != paused.reasons.end();
+            };
+            paused_by_user_or_error =
+                has_reason(PauseChargingEVSEReasonEnum::UserPause) or has_reason(PauseChargingEVSEReasonEnum::Error);
+        });
```
djchhp, line 152: This collapses in line count using lib/util/misc/containher.hpp:exists_any

modules/EVSE/EvseManager/energy_grid/energyImpl.cpp
```diff
@@ -142,6 +143,31 @@ void energyImpl::ready() {
...
+    case Charger::EvseState::PrepareCharging:
+    case Charger::EvseState::WaitingForAuthentication:
+    case Charger::EvseState::ChargingPausedEV:
+        return true;
+    case Charger::EvseState::ChargingPausedEVSE:
+        // a pause for missing energy alone keeps requesting so that charging can resume
+        return not paused_by_user_or_error;
+    default:
+        return false;
+    }
```
djchhp: much easier to read than before. i suggest adding the last remaining or of the if() as well.

djchhp (changes requested):
The implementation is correct to the best of my knowledge. I still suggest readability and maintainability improving changes.

djchhp (dismissed):

djchhp (approved):

PR #2718, opened 2026-09-02
feat(chargebridge): host-session ownership, end-to-end backpressure and vcan bus emulation (protocol v5)

corneliusclaussen:
## Describe your changes

Host side of firmware cb-session-v1 (`chargebridge-software`, `feat/session-ownership`). `shared/protocol/` is synced from the firmware repo: `CB_CONFIG_VERSION` 5, `CbHeartbeatPacket` carries `session_id`/`session_flags`, the reply carries the ownership status, a CAN datagram carries several records. Firmware image 0.7.6 included.

> **Deployment:** heartbeat and reply layouts changed; a peer on the old layout is ignored by packet size. Deploy both sides together.

### Session ownership
- Random nonzero `session_id` per tool instance in every heartbeat; the MCU treats an unknown id as a takeover, so a restarted tool takes over at once.
- `CBSS_RejectedBusy` is not liveness; reported once per episode with owner IP and session id. `heartbeat.force_takeover` (default `false`) steals a busy MCU.

### Serial
- Pause the pty at 64 queued TCP payloads, resume on drain and on the up edge; stays paused across outages so nothing is lost. Before: pty drained at memory speed (230 MB/s accepted, 2.4 kB/s delivered).
- `TCP_NODELAY` and coalescing to 512 B against the MCU's window; Nagle plus delayed ACK gave 1-3 kB/s.
- Keepalive 10 s idle + 4 x 2 s, user timeout 60 s: liveness is owned by the heartbeat session, and the tx path runs against a closed window by design.

### CAN
- Datagrams carry 1..12 frames; a 3 ms hold-off batches a smooth flood (opportunistic batching alone: 1.0-1.2 frames/datagram).
- Token bucket pacing at the heartbeat bitrate; the CAN socket is paused while waiting, so bursts become latency instead of `tx_ring_drops` in the CB's 35 ms TX ring.
- With `CAP_NET_ADMIN` a tbf on the vcan emulates the bus. tbf charges 16 B per frame, the wire 55..150 bits, so rate = bitrate * 128 / avg_wire_bits (EWMA of forwarded frames) minus 5 % headroom; decreases apply at once, increases once per second. The in-bridge pacer stands down while the qdisc is installed.
- The bridge's socket asks for a 1 MiB `SO_RCVBUF` and the shaper bypass priority.
- CAN keep-alive after 4 s without a datagram, checked every second. The old 10 s timer with a 10 s condition left gaps of up to 20 s, which is the MCU's CAN slot timeout; a receiver-only CB lost bus-to-host batches after each trip.

### Diagnostics
Silent in normal operation: every log line carries the wall clock with milliseconds; a warning when the event loop stalls for more than 500 ms (a 1 s watchdog timer firing late); a warning when a CAN keep-alive follows a gap of more than 10 s.

Depends on #2709 and #2723 (base).


applications/pionix_chargebridge/src/serial_bridge.cpp
```diff
@@ -43,20 +56,33 @@ void serial_bridge::create_tcp_client(std::string const& remote, uint16_t remote
...
         }
         m_tcp_tx_rejected = not accepted;
+        // Past the watermark stop reading the pty until the buffer drained; a rejected tx pauses too.
+        // The on_ready action resumes.
+        if (not m_pty_paused and (not accepted or m_tcp->tx_queue_depth() >= pty_pause_watermark)) {
+            m_pty_paused = m_pty.pause_rx();
+        }
     });
+    m_tcp->set_tx_drained_action([this]() { resume_pty(); });
     m_tcp->set_rx_handler([this](auto const& data, auto&) { m_pty.tx(data); });
```
djchhp, line 86: maybe a comment like: // Local RS-485 semantics: bytes nobody reads are lost, silently and without a backlog.

djchhp (dismissed):
Consider adding a comment to make intention clear for the tcp rx handler

djchhp (commented):
Two things on the CAN side (both small):

applications/pionix_chargebridge/src/can_bridge.cpp
```diff
@@ -198,6 +275,145 @@ void can_bridge::send_can_to_udp(cb_can_message const& msg) {
...
+void can_bridge::update_vcan_rate_limit() {
+    if (not m_vcan_shaped or m_pace_bitrate_bps == 0) {
+        return;
+    }
+    auto const now = std::chrono::steady_clock::now();
+    auto const wanted = vcan_rate_for(m_avg_wire_bits);
+    auto const installed = m_vcan_rate_bps;
+    if (wanted < installed) {
+        // Admitting more than the bus carries loses frames at the CB: lower the rate at once.
+        if ((installed - wanted) * 100 < installed * vcan_rate_decrease_hysteresis_pct) {
```
djchhp, line 323: Decreases apply on every flush with only a 1 % hysteresis, and each applied change is a netlink call plus an info line. When the payload mix shifts (e.g. 8-byte to 0-byte frames) the EWMA (1/64) moves more than 1 % per 12-frame flush for about 20 flushes in a row, so ~20 netlink calls and ~20 log lines within ~60 ms. That contradicts "silent in normal operation".

Suggest a minimum period for decreases as well, e.g.

```cpp
constexpr auto vcan_rate_decrease_period = std::chrono::milliseconds(100);

if (wanted < installed) {
    if ((installed - wanted) * 100 < installed * vcan_rate_decrease_hysteresis_pct) {
        return;
    }
    if (now - m_vcan_rate_updated < vcan_rate_decrease_period) {
        return;
    }
} else { ... }
```

If the shaper should keep reacting at once, the alternative is to leave the netlink call as is and only rate-limit the `print_info` (one line per second).

applications/pionix_chargebridge/src/can_bridge.cpp
```diff
@@ -65,25 +106,53 @@ can_bridge::can_bridge(can_bridge_config const& config, everest::lib::io::event:
...
-        everest::lib::io::udp::udp_client::ClientPayloadT pl;
-        cb_can_message msg;
-        msg_host_to_cb(data, msg);
-        send_can_to_udp(msg);
+        // Opportunistic batching alone yields 1.0-1.2 frames/datagram on a smooth flood; a 3 ms
+        // hold-off cuts the MCU's per-datagram cost up to 12x for at most 3 ms latency. Queued
+        // frames drain in the same pass (rx returns false on EAGAIN).
+        // Pacing may leave the batch full and the socket paused; the backlog waits in the kernel
+        // queue.
+        append_to_can_batch(data);
```
djchhp, line 144: When pacing holds a full batch, `flush_can_batch` pauses the socket and returns. If that `pause_rx()` fails (e.g. reopen in progress, no fd registered) this handler keeps running and appends before checking fullness, so the batch grows past 12 records and the datagram exceeds the MCU's 1280 B buffer.

Guard on entry:

```cpp
auto const full = can_udp_batch_max * sizeof(cb_can_message);
if (m_can_batch.buffer.size() >= full) {
    // Pacing holds a full batch but pause_rx() did not stick (e.g. reopen in progress):
    // a 13th record would overflow the MCU's datagram buffer.
    flush_can_batch();
    if (m_can_batch.buffer.size() >= full) {
        return; // still paced: drop, as the kernel queue would have
    }
}
append_to_can_batch(data);
```

The re-check is needed because `flush_can_batch` returns without sending while the bucket is empty.

applications/pionix_chargebridge/src/serial_bridge.cpp
corneliusclaussen: Added, with the queueing corrected: bytes queue in the pty and then the client's tx buffer (max_buffered_tx_payloads); beyond that tx() rejects and they are dropped without a log line. aed746db6

applications/pionix_chargebridge/src/can_bridge.cpp
corneliusclaussen: Confirmed, and worse at steady state: on a bus with mixed 0/8 byte payloads the EWMA jitters ~2 % per flush, so the 1 % threshold re-applied the rate on roughly a third of the flushes. 100 ms is too long for decreases: 2.4x over-admission for 100 ms overflows the CB's 256-frame TX ring (~35 ms). Now decreases apply after 10 ms, increases after 1 s, both on a 3 % move; the info line is throttled to one per second. aed746db6

applications/pionix_chargebridge/src/can_bridge.cpp
corneliusclaussen: Confirmed: 13 * 100 B > 1280 B. Guard applied as suggested. aed746db6

djchhp (approved):

PR #2723, opened 2026-09-02
feat(io): SocketCAN socket options and vcan transmit shaping for bus emulation

corneliusclaussen:
## Describe your changes

`lib/everest/io` support for emulating a CAN bus on a vcan (`pionix_chargebridge`, #2718).

- `socket_can_handler::open(dev, filters, socket_can_options)`: `SO_RCVBUF` and `SO_PRIORITY` on request; defaults set nothing, existing drivers unchanged.
- `vcan_netlink_manager::set_transmit_rate_limit()`: tbf over rtnetlink. prio root, pfifo band for priority 6/7 (`unshaped_socket_priority()`, the bridge's own bus-to-host writes), tbf band for the rest. An update sends only the tbf; the tree is installed on ENOENT; a failure past the root removes the root, so `false` always means unshaped. `clear_transmit_rate_limit()` removes it. `burst >= CANFD_MTU` enforced, bucket arithmetic overflow free.

Tests: `vcan_netlink_manager_test` (argument checks; installing needs `CAP_NET_ADMIN`).

Stacked on #2709, base of #2718.

### DIFF
--- a/lib/everest/io/include/everest/io/netlink/vcan_netlink_manager.hpp
+++ b/lib/everest/io/include/everest/io/netlink/vcan_netlink_manager.hpp
@@ -71,6 +71,42 @@ public:
     bool destroy(std::string const& interface_name);
 
     /**
+     * @brief Shape what producers write into the vcan with a token bucket (tbf), so a writer feels
+     *        the bus rate as on real hardware: a small SO_SNDBUF blocks at bus rate, the default
+     *        buffer gets ENOBUFS when the queue is full.
+     * @details Qdisc tree over rtnetlink (RTM_NEWQDISC): a 2-band prio root; skb->priority 6/7
+     *          (\ref unshaped_socket_priority, the bridge's own bus-to-host writes) -> pfifo,
+     *          everything else -> tbf. Idempotent; the qdiscs disappear with the interface. tbf
+     *          meters skb bytes (16 per classic frame, 72 per CAN FD frame), derive \p rate_bps from
+     *          the wire cost of the traffic. Without IFF_ECHO (vcan default) local readers get a
+     *          clone before the qdisc, so only the writer is paced; a reader that needs bus rate
+     *          paces its reads.
+     * @param interface_name The interface (must exist).
+     * @param rate_bps Token refill rate in bit/s of skb bytes.
+     * @param burst_bytes Bucket in bytes, at least the CAN FD MTU (72); should be >= rate/HZ.
+     * @param limit_bytes Queue capacity in bytes.
+     * @return True on success. False otherwise (reported through the error handler). A shaper
+     *         installed by an earlier call stays in place with its previous parameters; a failed
+     *         first installation leaves the interface unshaped, unless removing the half
+     *         installed tree fails as well (also reported).
+     */
+    bool set_transmit_rate_limit(std::string const& interface_name, std::uint64_t rate_bps, std::uint32_t burst_bytes,
+                                 std::uint32_t limit_bytes);
+
+    /**
+     * @brief Remove the shaper of \ref set_transmit_rate_limit. Idempotent.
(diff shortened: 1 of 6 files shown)


lib/everest/io/src/can/socket_can_handler.cpp
```diff
@@ -185,6 +188,15 @@ int socket_can_handler::open_device() {
 
     socket::set_non_blocking(can_fd);
     socket::set_socket_send_buffer_to_min(can_fd);
+    // Best effort, see socket_can_options.
+    if (m_options.receive_buffer_bytes != 0) {
+        int rcvbuf = static_cast<int>(std::min<std::uint32_t>(m_options.receive_buffer_bytes, INT32_MAX));
+        (void)::setsockopt(can_fd, SOL_SOCKET, SO_RCVBUF, &rcvbuf, sizeof(rcvbuf));
```
djchhp, line 194: why is this intentionally discarded? Shouldn't we fail or notify at least?

lib/everest/io/src/can/socket_can_handler.cpp
```diff
@@ -185,6 +188,15 @@ int socket_can_handler::open_device() {
...
+    }
+    if (m_options.socket_priority.has_value()) {
+        int prio = *m_options.socket_priority;
+        (void)::setsockopt(can_fd, SOL_SOCKET, SO_PRIORITY, &prio, sizeof(prio));
```
djchhp, line 198: same here

lib/everest/io/include/everest/io/netlink/vcan_netlink_manager.hpp
```diff
@@ -70,6 +70,40 @@ class vcan_netlink_manager {
...
+     *          meters skb bytes (16 per classic frame, 72 per CAN FD frame), derive \p rate_bps from
+     *          the wire cost of the traffic. Without IFF_ECHO (vcan default) local readers get a
+     *          clone before the qdisc, so only the writer is paced; a reader that needs bus rate
+     *          paces its reads.
+     * @param interface_name The interface (must exist).
+     * @param rate_bps Token refill rate in bit/s of skb bytes.
+     * @param burst_bytes Bucket in bytes, at least the CAN FD MTU (72); should be >= rate/HZ.
+     * @param limit_bytes Queue capacity in bytes.
+     * @return True on success. False otherwise (reported through the error handler); then nothing
+     *         is left installed, also when an earlier call had shaped the interface.
```
djchhp, line 89: This claim is false for any error but ENOENT. A previously successfully installed tbf is not uninstalled.

lib/everest/io/src/netlink/vcan_netlink_manager.cpp
```diff
@@ -253,6 +276,202 @@ bool vcan_netlink_manager::bring_down(std::string const& interface_name) {
...
+            return false;
+        }
+        // No parent 1: to hang the tbf under: install the tree.
+    } catch (std::exception& e) {
+        report_error(interface_name, caller, e.what());
+        return false;
+    } catch (...) {
+        report_error(interface_name, caller, "Unexpected exception");
+        return false;
+    }
```
djchhp, line 400: No uninstall attempt in contrast to what the header claims

djchhp (changes requested):

lib/everest/io/src/can/socket_can_handler.cpp
corneliusclaussen: I think we should notify but not fail. The socket option may or may not be available on the kernel.

djchhp: ok, that is fine with me.

djchhp (dismissed):

djchhp (approved):

PR #2730, opened 2026-09-02
feat(PersistentSessionStorage): add persistent session storage module

Pietfried:
## Describe your changes

Persist one record per EvseManager charging session in SQLite: created at SessionStarted, with the EVerest transaction as an optional nested object and an Open/Finished/Stale lifecycle. Records hold start/stop timestamps, import energy readings, a SHA256 id token hash, optional signed meter values (config gated), session cost, and the OCPP transaction id with the OCPP transaction start and stop timestamps, which can differ from the EVerest ones because OCPP 2.x transaction start and stop points are configurable.

The new session_storage interface provides paginated, filterable read access and a clear command. Records survive restarts; sessions with a running transaction interrupted by a power loss are closed during recovery on the next boot, the oldest records are pruned beyond the configurable maximum.


modules/Misc/PersistentSessionStorage/PersistentSessionStorage.cpp
```diff
@@ -0,0 +1,104 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Pionix GmbH and Contributors to EVerest
+#include "PersistentSessionStorage.hpp"
+
+#include <algorithm>
+#include <atomic>
```
cburandt, line 6:
```suggestion
```
Already in the hpp

modules/Misc/PersistentSessionStorage/PersistentSessionStorage.cpp
```diff
@@ -0,0 +1,104 @@
...
+#include <cstddef>
+#include <cstdint>
```
cburandt, line 8:
```suggestion
#include <cstdint>
#include <filesystem>
```
Used for the migrations file path.

modules/Misc/PersistentSessionStorage/PersistentSessionStorage.cpp
```diff
@@ -0,0 +1,104 @@
...
+#include <utility>
+
+#include <everest/exceptions.hpp>
```
cburandt, line 11:
```suggestion
#include <everest/database/sqlite/connection.hpp>
#include <everest/exceptions.hpp>
```
Maybe, better don't rely on transitive include here.

modules/Misc/PersistentSessionStorage/docs/index.rst
```diff
@@ -0,0 +1,133 @@
...
+    and will be converted to HTML and PDF by Sphinx.
+    This index.rst file is the entry point for the module documentation.
+
+..  Use underlined-only headlines inside this document (highest-level
+    sub-section headline should use "=" characters)
+
+..  The content of this file will be included in the auto-generated HTML
+    page for the module. You can link to it using the following
+    reference: everest_modules_PersistentSessionStorage.
+
```
cburandt, line 19:
```suggestion
```

modules/Misc/PersistentSessionStorage/docs/index.rst
```diff
@@ -0,0 +1,133 @@
...
+``get_sessions`` is paginated: the caller passes an optional page size and filter
+(state, EVSE id, started-after timestamp compared against the session start), and
+iterates by passing the ``continuation_token`` of a reply into the next call until a
+reply carries no token. The token is opaque and held only by the caller; the module
+keeps no iteration state. Records stored during the iteration appear at its end, pruned
+records are skipped, and a token that is invalid or refers to a replaced database file
+restarts the iteration from the oldest record, so a stale token can re-deliver records
+but never silently skip them. Page sizes are additionally capped by a byte budget, so a
+page may contain fewer records than requested.
```
cburandt, line 133: Didn't read the implementation yet: Maybe it's rather academic, but calling get_sessions() often enough to get all session records and then clear_sessions() leaves the small gap, where another transaction could start in between and immediately be deleted. So I client could miss a session without noticing.
Possible cure: Allow clear_sessions to be called with the session_id/transaction_id up to which the records shall be deleted.

modules/Misc/PersistentSessionStorage/tests/session_recorder_tests.cpp
```diff
@@ -0,0 +1,765 @@
...
+        SessionEventEnum::Disabled,         SessionEventEnum::AuthRequired,     SessionEventEnum::PrepareCharging,
+        SessionEventEnum::ChargingStarted,  SessionEventEnum::ChargingPausedEV, SessionEventEnum::ChargingPausedEVSE,
+        SessionEventEnum::StoppingCharging, SessionEventEnum::ChargingFinished, SessionEventEnum::ReservationStart,
+        SessionEventEnum::ReservationEnd,   SessionEventEnum::PluginTimeout,    SessionEventEnum::SwitchingPhases,
+    };
+
+    for (const auto event : events) {
+        recorder.on_session_event(make_session_event(event, SESSION_ID));
+    }
+}
```
cburandt, line 655: There are no EXPECTs here: The idea is, that it doesn't crash?

modules/Misc/PersistentSessionStorage/tests/session_store_tests.cpp
```diff
@@ -0,0 +1,1623 @@
...
+    ASSERT_TRUE(first.continuation_token.has_value());
+    const auto second = store->get_sessions(make_request(2, first.continuation_token));
+    ASSERT_TRUE(second.continuation_token.has_value());
+    const auto third = store->get_sessions(make_request(2, second.continuation_token));
+
+    EXPECT_EQ(first.sessions.size(), 2);
+    EXPECT_EQ(second.sessions.size(), 2);
+    EXPECT_TRUE(third.sessions.empty());
+    EXPECT_FALSE(third.continuation_token.has_value());
+}
```
cburandt, line 1299: Is there a specific reason to serve this last empty page, instead of not returning a continuation_token with the second page?

modules/Misc/PersistentSessionStorage/tests/session_store_tests.cpp
```diff
@@ -0,0 +1,1623 @@
...
+    const auto page = store->get_sessions(make_request(std::nullopt, std::nullopt, filter));
+    EXPECT_TRUE(page.sessions.empty());
+}
+
+TEST_F(SessionStoreTest, the_started_after_filter_accepts_nanosecond_timestamps) {
+    ASSERT_TRUE(store->store_session_started(make_session("s1", 1, "2026-08-21T10:00:00.000Z")));
+    ASSERT_TRUE(store->store_session_started(make_session("s2", 2, "2026-08-21T12:00:00.000Z")));
+
+    SessionFilter filter{};
+    filter.started_after = "2026-08-21T11:00:00.123456789Z";
```
cburandt, line 1485: Shouldn't this rather filter for ...T10:00:00.123456789Z in order to prove, that the nanoseconds are used?

cburandt (commented):
Looks pretty good to me.

Find some comments below.

modules/Misc/PersistentSessionStorage/PersistentSessionStorage.hpp
```diff
@@ -0,0 +1,108 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Pionix GmbH and Contributors to EVerest
+#ifndef PERSISTENT_SESSION_STORAGE_HPP
+#define PERSISTENT_SESSION_STORAGE_HPP
+
+//
+// AUTO GENERATED - MARKED REGIONS WILL BE KEPT
+// template version 2
+//
+
```
cburandt, line 10: Maybe development was started before the shutdown branch was merged. However, it could be considered best practice to use the latest template and implement the shutdown logic from the beginning, when introducing new modules.

cburandt (commented):
Just another thought. Could be a follow-up as well, but maybe it's not a lot of work for this module.

modules/Misc/PersistentSessionStorage/PersistentSessionStorage.hpp
Pietfried: Agreed, shutdown implemented as part of 553f77e2a

modules/Misc/PersistentSessionStorage/tests/session_store_tests.cpp
Pietfried: added an explicit test in 7d1a4237f

modules/Misc/PersistentSessionStorage/tests/session_store_tests.cpp
Pietfried: The query now looks one row ahead, so a full last page carries no token, implemented in e9b5d394d

modules/Misc/PersistentSessionStorage/tests/session_recorder_tests.cpp
Pietfried: Yes, the StrictMock is the assertion. Made that explicit in 9d64db915

modules/Misc/PersistentSessionStorage/docs/index.rst
Pietfried: clear_sessions now takes an optional up_to_session_id; newer records are kept, an unknown id deletes nothing, implemented in c1fb93acf

modules/Misc/PersistentSessionStorage/docs/index.rst
Pietfried: fixed with a0082a529

modules/Misc/PersistentSessionStorage/PersistentSessionStorage.cpp
Pietfried: fixed with 44f387033

modules/Misc/PersistentSessionStorage/PersistentSessionStorage.cpp
Pietfried: fixed with 44f387033

modules/Misc/PersistentSessionStorage/PersistentSessionStorage.cpp
Pietfried: fixed with 44f387033

cburandt (dismissed):
LGTM

cburandt (dismissed):
LGTM

modules/Misc/PersistentSessionStorage/storage/record_conversions.cpp
```diff
@@ -0,0 +1,54 @@
...
+/// \brief Encodes \p digest as lower case hex string
+std::string to_hex(const openssl::sha_256_digest_t& digest) {
+    std::string result{};
+    result.reserve(digest.size() * 2);
+    for (const auto byte : digest) {
+        result.push_back(HEX_DIGITS.at(byte >> 4U));
+        result.push_back(HEX_DIGITS.at(byte & 0x0FU));
+    }
+    return result;
+}
```
hikinggrass, line 27: something for a follow-up, lets try to consolidate all of our multiple to_hex(string) implementations. this would be a good example to put into one of the helper/util libraries

hikinggrass (dismissed):

cburandt (approved):
LGTM

hikinggrass (approved):

PR #2741, opened 2026-09-04
fix(ocpp): inject OCPP16LegacyCtrlr for OCPP201/OCPPmulti 2.x device model

Pietfried:
## Describe your changes

The three-argument DeviceModelStorageSqlite constructor used by OCPP201 and OCPPmulti disabled the OCPP16LegacyCtrlr default fallback. A DeviceModelConfigPath copied from 2026.02.0 lacks OCPP16LegacyCtrlr.json, so NumberOfConnectors was missing, the integrity check threw and the module never became ready. An existing database even had the component removed.

Enable the fallback in that constructor. Add the built-in component to the config map whenever the config directory lacks it, independent of the database contents, so it is inserted and updated like every other component and picks up variables added to the built-in schema on upgrade. Run the missing-config-file warning on the augmented map so an existing database no longer warns about the injected component on every boot.

Mention the injected file in the required-variable error, cover the fresh database, the upgraded database and the re-merge of new variables in unit tests, and document that the component is required for every OCPP version because the shared device model enables switching between 1.6 and 2.x.

### DIFF
--- a/docs/source/how-to-guides/ocpp-storage-migration.rst
+++ b/docs/source/how-to-guides/ocpp-storage-migration.rst
@@ -300,9 +300,13 @@ What to keep in mind
   individual ``MeterPublicKey[N]`` variables, and
   ``ChargingScheduleAllowedChargingRateUnit`` values ``Current``/``Power`` are
   converted to ``A``/``W``.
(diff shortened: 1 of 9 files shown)


mlitre (approved):
lgtm

PR #2742, opened 2026-09-04
fix(OCPPmulti): gate configure_network delegation behind DelegateNetworkConfigurationToSystem

Pietfried:
## Describe your changes

OCPPmulti asked the system provider to configure the network before every CSMS connection attempt. With system_API and an external client that does not answer e2m/configure_network, each attempt blocked until the framework command timeout (300 s) and the charge point never connected.

Add the boolean config key DelegateNetworkConfigurationToSystem (default false). When false, no configure_network request is sent, the configure_network_status subscription is skipped and the connection is established as with the OCPP201 module (success, no interface address). When true, the behaviour is unchanged.

- unit tests: existing delegation tests enable the flag; new tests cover the disabled path (immediate success, provider never called, status publishes ignored)
- integration tests opt in via OCPPMultiModuleConfigStrategy
- docs: describe the flag, the integrator obligations before enabling it (provider must answer, NotSupported suffices, system_API clients must reply) and which timeout bounds a non-responding provider

### DIFF
--- a/modules/EVSE/OCPPmulti/generic_ocpp.cpp
+++ b/modules/EVSE/OCPPmulti/generic_ocpp.cpp
@@ -585,17 +585,20 @@ void GenericOcpp::init_subscribe() {
     mv_requires.system.subscribe_firmware_update_status([this](auto arg) { cb_firmware_update_status(arg); });
     mv_requires.system.subscribe_log_status([this](auto arg) { cb_log_status(arg); });
 
-    mv_requires.system.subscribe_configure_network_status([this](const types::network::ConfigureNetworkStatus status) {
-        if (status.status != types::network::ConfigureNetworkFinalStatusEnum::Ready) {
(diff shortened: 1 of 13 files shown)


modules/EVSE/OCPPmulti/generic_ocpp.cpp
```diff
@@ -926,6 +929,14 @@ std::future<ocpp::ConfigNetworkResult> GenericOcpp::cb_configure_network_connect
         return future;
     }
 
+    if (!mv_config.getDelegateNetworkConfigurationToSystem()) {
```
mlitre, line 932: Instead of just handling this in the callback should we just not register the callback at all?

Pietfried: I think it's fine. Not would require a slightly larger change because v16 and v2 wrappers are built with only the callbacks object and the security interface, and they have no view of the module config.

mlitre (approved):
Lgtm

PR #2766, opened 2026-09-10
fix(EvseManager): restart matching on the BCB wake path

djchhp:
## Describe your changes

Two small EvseManager changes.

1. **Charger: restart matching on the BCB wake path.** The state machine deliberately stays in `ChargingPausedEV` after an HLC session ended with Terminate, so the EV can restart a session with a BCB toggle. The restart re-enters `PrepareCharging` and re-applies PWM, but nothing restarts matching: `dlink_terminate` reset the slac provider, EvseSlac then sits in Idle where only `enter_bcd` starts matching again (an EV's SLAC_PARM.REQ is ignored there), and a B→C→B toggle raises none of the CP events that issue `enter_bcd`. `signal_slac_start()` is now raised on that path, as the `ChargingPausedEVSE` resume path already does. After a `dlink_pause` the provider is still Matched and ignores `enter_bcd`, so the call is a no-op there.
   Found on an MCS data link (ISO 15118-10, `McsDataLink`, #2696), where the EV always wakes by BCB toggle and the SDP server stayed deaf to every second session.

2. **IECStateMachine: change-gated diagnostics** for the `allow_power_on` request and for its forward to the BSP. The forward is edge-triggered, so a lost edge was invisible in the logs (bench: a cable check that never closed the contactors, with every permissive set upstream). Logs only on change, silent in steady state.

**Tests:** EvseManager unit tests 83 pass / 2 pre-existing skips, Charger tests 25/25.
**Bench:** two-ChargeBridge MCS setup (EvseManager + Evse15118D20 + McsDataLink against EvManager + EV stack), wake cycles (session end → EV BCB wake → second session) pass with 1.; CCS EVAL hardware sessions (DIN, ISO 15118-2, ISO 15118-20 incl. pause/resume) unaffected.

—

- [x] I have made corresponding changes to the documentation (n/a: no interface or config change)


corneliusclaussen (approved):
lgtm

corneliusclaussen (approved):

Pietfried:
@djchhp is this eligible for a backport?

PR #2770, opened 2026-09-10
fix(CMake): export external dependencies correctly

### LINKED ISSUE #2750: Non-EDM builds: dependency issue (circular dependencies)
### Describe the bug

When building repositories with out-of-tree module sources against current everest-core, we cannot successfully resolve dependencies. When building without EDM, as in Yocto, we are getting a dependency failure coming from everest-core:

```
| -- Configuring incomplete, errors occurred!
| See also "/home/dev/yocto/build/tmp/work/cortexa7t2hf-neon-poky-linux-gnueabi/everest-chargebyte/0.32.0+gitAUTOINC+0f5525071c-r0/build/CMakeFiles/CMakeOutput.log".
| CMake Error at CMakeLists.txt:32 (find_package):
|   Found package configuration file:
| 
|     /home/dev/yocto/build/tmp/work/cortexa7t2hf-neon-poky-linux-gnueabi/everest-chargebyte/0.32.0+gitAUTOINC+0f5525071c-r0/recipe-sysroot/usr/lib/cmake/everest-core/everest-core-config.cmake
| 
|   but it set everest-core_FOUND to FALSE so package "everest-core" is
|   considered to be NOT FOUND.  Reason given by package:
| 
|   The following imported targets are referenced, but are missing:
|   everest::ocpp everest::framework
```

This happens identically in all out-of-tree repos.

### Expected behavior

_No response_

### EVerest Domain

Build System

### Affected Component

CMake
and others

### EVerest Version

EVerest main 2026.10.0-pre, 2827316bd2b597c81d0faf7142794d29f9fb2e78

### To Reproduce

_No response_

### Anything else?

After various attempts at solving the issue, I had to get OpenAI Codex (gpt-6-astra) to do an in-depth analysis. I have not asked for or attempted a fix:

The root cause is a circular dependency **between CMake export sets**, introduced in everest-core’s non-EDM packaging. A missing Yocto `DEPENDS` does not explain this failure: the component packages are present, but CMake loads their exports in an order that cannot satisfy their checks.

The current checkout also corrects one earlier finding: `find_dependency(everest-util)` **already exists** in OCPP’s package configuration.

The export relationships explain why the previous fixes [edit: by myself and Codex] kept moving the error:

| Export set | Defines | Requires from another export set |
|---|---|---|
| `everest-core-targets` | `util`, `yaml`, `io`, OCPP conversions, `ocpp_module_common`, others | `ocpp`, `framework`, `log` |
| OCPP | `ocpp` | `util`, `log`, `timer`, `evse_security`, `sqlite` |
| Framework | `framework` | `yaml`, `helpers`, `util`, `log`, `sqlite` |

These are imported targets, rather than aliases in the consumer.

The problematic grouping mixes low-level libraries with libraries that depend on OCPP/framework. Loading core first makes its conversion targets reference missing OCPP/framework targets. Loading the components first makes them reference missing core targets.

In [project-config.cmake.in](https://github.com/EVerest/EVerest/blob/2827316bd2b597c81d0faf7142794d29f9fb2e78/cmake/project-config.cmake.in#L16), core’s export is included before framework and OCPP. Its generated validation immediately sets `everest-core_FOUND=FALSE`. Configuration continues and can successfully import both components afterward, but **that earlier failure flag remains set**. This explains the seemingly contradictory log: substantial configuration succeeds, yet the outer `find_package(everest-core)` ultimately fails.

There is a second defect in `everest-cmake`: its `evc_setup_package()` template includes the target export **before** executing the supplied `find_dependency()` calls. Therefore, merely adding the missing dependency declaration does not ensure it exists when the export checks it.

Loading `everest-util` independently before core can cause another failure: core also exports `everest::util`, and CMake rejects an export set when only some of its targets already exist. This is another reason simple reorderings are insufficient.

The relevant history is precise:

- **[`0796e370`](https://github.com/EVerest/EVerest/commit/0796e3703ebf7b37cc169caa0f2772dbbb7cdf7b), 17 March 2026, [#1975](https://github.com/EVerest/EVerest/pull/1975):** introduced standalone util packaging while util remained in core’s export set. This created overlapping target ownership.
- **[`e05a9322`](https://github.com/EVerest/EVerest/commit/e05a93225c2dd0068ed60ed128e6282d9eebe526), 11 June 2026, [#2243](https://github.com/EVerest/EVerest/pull/2243):** introduced the direct cause of your current error. It added `ocpp_evse_security` and `ocpp_conversions` to `everest-core-targets`, although both link OCPP/framework, and moved OCPP discovery after core’s export. The same commit added the previously missing OCPP util dependency.
- **[`fddc9283`](https://github.com/EVerest/EVerest/commit/fddc928363d8bb9ee7c95956f6733571e0673361), 3 July 2026, [#2396](https://github.com/EVerest/EVerest/pull/2396):** added `ocpp_module_common` to that same export set, adding another OCPP/framework dependency.

The missed adjustment was therefore in **everest-core’s export-set organization**, particularly the conversion-library installation added by #2243. EDM builds create targets together in the build tree and do not exercise this installed-package import sequence.

My recommended fix is to break the export-set cycle:

1. Keep the existing core export for foundational targets such as util, YAML and IO.
2. Move `ocpp_evse_security`, `ocpp_conversions` and `ocpp_module_common` into a separate export set.
3. Import foundational core targets first, then TLS/helpers/framework/OCPP, then the new conversion/module export.
4. Fix `everest-cmake` to discover dependencies before importing targets. Move `include(CMakeFindDependencyMacro)` before its first use in core’s template as well.

This preserves the real exported link interfaces and requires no workaround in the out-of-tree projects. Standalone component packaging also deserves cleanup: `evc_setup_package_no_export()` constructs reduced imported targets without preserving their link interfaces, so it should not be used to conceal the cycle.

I verified the mechanism with an isolated, compilation-free CMake reproducer: core-first fails with missing OCPP; component-first fails with missing util; separating the dependent adapter export succeeds. This validates the proposed structure, but is not yet a full EVerest integration test.

barsnick:
## Describe your changes
Change the order and method of exporting external dependencies, to avoid circular dependencies, e.g. in out-of-tree non-EDM builds.
Possibly fixes https://github.com/EVerest/EVerest/issues/2750.

Requires companion PR in everest-cmake https://github.com/EVerest/everest-cmake/pull/36.


barsnick:
I have bumped the everest-cmake dependency to 0.8.0, and modified the Yocto recipes for this PR.

Those two changes could also go into a separate PR (to be merged before this PR).

PR #2777, opened 2026-09-11
fix(tests): ignore stale startup connection-status payloads

corneliusclaussen:
## Describe your changes

`test_connection_status_is_forwarded_with_legacy_is_connected` republishes the connected status until both external topics respond. Responses to those startup publications can remain queued or in flight when the test publishes the disconnect. The disconnect assertion then compares a stale connected payload with the expected disconnected payload and fails depending on delivery timing.

Allow the queue reader to skip explicitly supplied stale values, and skip only the exact connected payloads already validated by this test when checking the disconnect. The first changed payload must still match the complete disconnected status or the legacy `False` value. The existing deadline remains in force if no transition arrives. This also covers startup responses delivered after waiting for the disconnect has begun.

Validation:
- Both tests in `tests/async_api_tests/test_ocpp_consumer.py` pass against the local EVerest installation: **2 passed in 2.93s**.
- A deterministic bridge-delivery harness invokes the actual test coroutine, forces a startup retry and queued plus in-flight connected responses: the original test fails and the revised test passes.
- The same harness verifies that an incorrect changed status payload or legacy value still fails, even if the correct payload follows, and that a missing disconnect still times out.
- `git diff --check` passes.

Observed failure: https://github.com/EVerest/EVerest/actions/runs/32976906771/job/98206963494

- [x] I have made corresponding changes to the documentation (queue-reader docstring and synchronization comment)
- [ ] Human review and verification before requesting review

### DIFF
--- a/tests/async_api_tests/test_ocpp_consumer.py
+++ b/tests/async_api_tests/test_ocpp_consumer.py
@@ -80,15 +80,18 @@ def _subscribe_to_queue(handler: AsyncApiMqttHandler, topic: str) -> Queue:
     return queue
 
 
-async def _get_from_queue(queue: Queue, deadline: float = 10.0):
+async def _get_from_queue(queue: Queue, deadline: float = 10.0, *, ignore_values=()):
+    """Return the next payload other than an explicitly allowed stale value."""
     loop = asyncio.get_event_loop()
     end = loop.time() + deadline
     while loop.time() < end:
         try:
-            return await loop.run_in_executor(None, lambda: queue.get(timeout=0.5))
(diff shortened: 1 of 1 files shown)


a-w50 (approved):
In my words: due to multiple connection retries (because the async api brigde module might not be listening yet on the first connect event), multiple connect events might be in flight on the mqtt bus.  When waiting for the proper disconnect event, we still might get one of these initial connect events and hence fail.  Therefore this test was patched, so these initial ones get filtered away and only the corresponding disconnect event should be fetched.  This is by design brittle and might also give synchronization issues in real world.

Right now, having non-flaky tests is important for development progress.

FIXME (aw): revisit synchronization strategy of bridge modules

PR #2778, opened 2026-09-11
fix(PyEvJosev): preserve starts across session teardown

corneliusclaussen:
## Describe your changes

PyEvJosev clears its start event after a session finishes. A new start request received during teardown or in response to `v2g_session_finished` can therefore be discarded, leaving the next session waiting indefinitely.

Consume the start event before running the session so a subsequent request remains pending for the next iteration.

Validation: Python syntax and `git diff --check` pass. Full SIL validation remains pending.


a-w50 (dismissed):
The comment doesn't really explain the issue to me.  It shouldn't be in this code at all.  Should go to the commit message and explain it like:
Clear the event we were waiting for.  If we do it in "teardown" i.e. "finally", we might clean another / parallel happened event.

FIXME(aw): depending on parallelism, a queue or semaphore might be more appropriate.

a-w50 (approved):

PR #2779, opened 2026-09-11
fix(YetiSimulator): retain EV state changes during forced CP F

corneliusclaussen:
## Describe your changes

Plug & Charge CI can leave a connector occupied after the simulated EV unplugs during the EVSE's CP state F recovery interval. YetiSimulator currently drops `set_cp_state(A)` while F is active, so releasing F restores the old plugged-in state and the OCPP test times out waiting for `Available`.

Remember EV-requested A–D physical states while the EVSE drives F. Keep EV-side event publication suppressed during F; the existing voltage simulation continues to drive -12 V until F is released, then observes the latest stored EV state. Preserve the existing handling of state E during F.

All three failed OCPP 2.0.1 Plug & Charge tests in [the September 11 scheduled run](https://github.com/EVerest/EVerest/actions/runs/34552166327/job/103131810451) contain the same sequence: reinitialization enters CP F, Yeti logs `Cannot change CP state, because PWM F is active`, and the EV simulator finishes. The unplug never reaches the EVSE. This is a simulator fix; no retries or timeout increases are added.

Validation:
- Added two C++ regression tests invoking the actual EV board-support command handler: unplug during F and multiple EV transitions during F. They also verify that F remains active and premature EV-side events remain suppressed.
- Both tests pass with the fix and fail against the original handler.
- Configured and built the new tests/CMakeLists.txt target in an isolated CMake project importing existing generated headers and framework/GTest libraries; CTest passes. Full OCPP SIL suite has not been rerun locally; this draft needs CI confirmation of the three original failures.

CI failure linked above.

- [x] I have made corresponding changes to the documentation (handler and regression-test comments)
- [ ] Full OCPP SIL CI validation and human review


modules/Simulation/YetiSimulator/tests/EvBoardSupportTest.cpp
```diff
@@ -0,0 +1,80 @@
...
+    adapter.events.clear();
+    simulator.module_state->pwm_error_f = true;
+
+    set_cp_state(EvCpState::A);
+
+    // The EVSE must continue driving F, while the physical unplug is retained for
+    // the simulation to observe after the EVSE releases F.
+    EXPECT_TRUE(simulator.module_state->pwm_error_f);
+    EXPECT_TRUE(adapter.events.empty());
+    EXPECT_DOUBLE_EQ(simulator.module_state->simdata_setting.cp_voltage, 12.0);
```
a-w50: shouldn't use hard coded values here, rather CP_VOLTAGE_X

a-w50 (dismissed):
The title is a bit misleading.  Not really the state, but cp_voltage is retained

a-w50 (approved):

PR #2782, opened 2026-09-11
fix(slac_API): Fix slac api mismatch and missing types

marcemmers:
## Describe your changes

The slac_API documentation and the implementation did not match up completely.
To not change the actual API I renamed the SLACState type in the slac_API documentation.
Also I added the ErrorEnum and Error type in the same way it's used in all the other APIs instead of using the generic error type here.

### DIFF
--- a/docs/source/reference/EVerest_API/slac_API.yaml
+++ b/docs/source/reference/EVerest_API/slac_API.yaml
@@ -229,7 +229,7 @@ components:
       summary: Provides the state enum.
       contentType: application/json
       payload:
-        $ref: '#/components/schemas/SLACState'
+        $ref: '#/components/schemas/State'
       examples:
         - summary: "Example state MATCHED"
           payload: "MATCHED"
@@ -322,7 +322,7 @@ components:
     Enabled:
       type: boolean
       description: 'true: start SLAC after reset, false: stop SLAC'
-    SLACState:
+    State:
       description: Provides the state enum.
       type: string
       enum:


lib/everest/everest_api_types/src/everest_api_types/slac/json_codec.cpp
```diff
@@ -44,4 +44,59 @@ void to_json(json& j, State const& k) noexcept {
...
+void to_json(json& j, const Error& k) noexcept {
+    j = json{
+        {"type", k.type},
+    };
+    if (k.sub_type) {
+        j["sub_type"] = k.sub_type.value();
+    }
+    if (k.message) {
+        j["message"] = k.message.value();
+    };
```
cburandt, line 89:
```suggestion
    }
```

modules/API/EVerestAPI/slac_API/slac_API.cpp
```diff
@@ -112,9 +112,9 @@ void slac_API::generate_api_var_clear_error() {
     });
 }
 
-std::string slac_API::make_error_string(API_generic::Error const& error) {
+std::string slac_API::make_error_string(API_types_ext::Error const& error) {
     auto error_str = API_generic::trimmed(serialize(error.type));
-    auto result = "generic/" + error_str;
+    auto result = "slac/" + error_str;
```
cburandt, line 117: The make_error_string(...) function creates strings for the internal error raising mechanism. There are no slac specific errors (see the top-level errors/*.yaml - there is no slac.yaml). The slac_API module provides the slac interface which in turn references the generic errors. Therefore this must remain "generic" as all other string would lead to the errors being dropped by the framework.

docs/source/reference/EVerest_API/slac_API.yaml
cburandt, line 1: I think this is the only change required.

Concerning the other files: The api type conversion routines for the error type already exist in ".../generic/...". And that's also the actual conversion routines for the internal generic type which the slac module implements.

cburandt (changes requested):
As there are no slac-specific errors, the C++ changes are not required.

docs/source/reference/EVerest_API/slac_API.yaml
marcemmers: Ok that makes sense. I was looking at all the other API's and this was the only API that reused the generic errors and thought it was just from copying during implementation. That's what I get for just assuming stuff about how it works haha.

I will revert the C++ changes

cburandt (approved):
LGTM

hikinggrass (approved):

PR #2791, opened 2026-09-14
fix(io): de-flake pty_handler write-retry test

corneliusclaussen:
`pty_handler_test.a_write_the_slave_has_not_drained_is_a_retry_not_an_error` failed on an unrelated PR:

```
pty_handler_test.cpp:44: Failure
Value of: handler.tx(payload)
  Actual: true
Expected: false
```

Nothing in the test reads the slave, so the assertion assumes the pty stays full. It does not: `pty_write` fills the slave's flip buffer and a `flush_to_ldisc` kworker moves up to 4096 bytes of it into the line discipline buffer asynchronously, so a retry can fit without any reader.

Reproduced out of tree (posix_openpt + TIOCPKT + EXTPROC + raw slave + non-blocking master, as `serial::openpty`):

| retry timing | full write succeeded |
|---|---|
| immediate | 0 / 2900 |
| after 1 ms | 62 / 200 |

CI hits the immediate case on a loaded runner.

Fix: retry until it stalls again, bounded by the existing 1000-write helper, instead of asserting on a single write. The test's contract - a retry `false` leaves `get_error()` at 0 - is unchanged.

300 local runs of the fixed test, 0 failures.


djchhp (approved):

a-w50 (commented):
Don't see what the value of this test is?  It only shows, that the kernel buffer is finite (actually twice).  But what for?

PR #2792, opened 2026-09-14
fix(iso15118): complete PnC certificate installation on the EV side

mlitre:
## Describe your changes

Make the EV side able to complete an ISO 15118-2 Plug and Charge certificate installation.
Two defects blocked it, one behind the other; neither is specific to the tests.

1. **Inbound V2GTP buffer was 2048 bytes.** A conformant `CertificateInstallationRes` runs
   about 4.1 kB, so the EV rejected every response as malformed and tore down the
   connection before `PaymentDetailsReq`. `EvseV2G` and the EV outbound path already use
   8192.
2. **The eMAID signature reference was coded as a complex type.** ISO 15118-2 declares
   `eMAID` four times across two type names differing only in case, and a fragment has no
   parent to select a declaration by, so EXI 1.0 section 8.5.3 requires the element
   fragment grammar. Three of four signed digests matched byte for byte; the fourth did
   not, so every response failed verification.

Fixed in the generator (EVerest/cbexigen#135, `fix/exi-element-fragment-grammar` at
`0d81841`) rather than in hand-edited output. ISO 15118-2 regenerated alone; DIN 70121 is
untouched and its diff against the base is empty.

### Reading the diff

Six files are machine output, not to be read as a diff:
`lib/everest/cbv2g/{include,lib}/cbv2g/iso_2/iso2_msgDef{Datatypes,Decoder,Encoder}.{h,c}`,
11,396 of the 11,824 insertions. Most of that is inherent to regenerating this library:
against a no-fix regeneration, roughly 16,800 line movements are reordering and 1,037 are
grammar renumbering. The fix's own effect is 233 lines across four files, no new files.

**The reviewable content is the other eleven files: 428 insertions, 30 deletions.** Two
tests make the generated half checkable, both red before and green after on identical
source:

- `tests/iso2/emaid_fragment.cpp` pins the fragment against twelve vectors from an
  independent EXI codec, `Id` lengths one to six including those breaking byte alignment.
- `ev/d2/certificate_installation_interop.cpp` verifies a real response signed by another
  implementation. The existing signature test builds the response with the same encoder it
  verifies with, so it agreed with itself regardless of the bytes.

`fix(iso15118): assign V2G fragments instead of memcpy` carries no behavior change alone.
Its purpose is visible here: the regeneration changes the member's type and the build stops
with `no match for operator=` at both call sites. Under `memcpy` it would have compiled in
silence and emitted 12 bytes where 27 were required. That is also why the regeneration and
its call site conversion are one commit.

### Results

Full OCPP suite 972 collected, 0 failed, 922 executed, 50 skipped, identical to the base
measured the same way. The regenerated codec is linked into `EvseV2G`, whose binary changed
in the rebuild. libiso15118 542 of 542, libcbv2g 7 of 7.

Based on `feat/ev-simulator-e2e-swap`, the head branch of #2256, not `main`; inherits that
PR's dependencies and check state.

None.


mlitre:
Not replaced by a single pull request, so this one names all three. Its `cbv2g` half is in #2798, the eMAID element-fragment codec. Its Plug and Charge fixes are in #2804, where the ISO 15118-2 engine lives. Its harness and configuration changes are in #2256. Closing in favor of those three together; no one of them is its successor.

PR #2794, opened 2026-09-15
fix(cmake): make the Rust module build work, and keep it working

mlitre:
`-DEVEREST_ENABLE_RS_SUPPORT=ON` does not build on Debian, Ubuntu, or in this
repository's own build-kit image. Nothing in CI would have said so: the image
ships no Rust toolchain, so that path has never run.

Four defects, each reproduced:

- The link-dependency file is written with a `$<LIST:JOIN,...>` generator
  expression, which needs CMake 3.27; the build-kit image has 3.25.1. It is also
  piped through `echo -e`, which dash prints as data. `file(GENERATE)` replaces
  both and needs no shell.
- Installed Rust modules carry no `RUNPATH`, so they cannot find
  `libframework.so` outside the generated development scripts.
- Editing an interface or type YAML does not regenerate the Rust code, and the
  build can reach a state no rebuild clears.
- `cargo install cxxbridge-cmd` runs at configure time, over the network,
  without saying so.

The CI commit turns the Rust lane on inside `build-cmake-gcc` rather than adding
a job. A separate job would configure and build everest-core a second time,
because `build_rust_modules` carries order-only dependencies across the whole
tree and there is no narrow Rust target today. Turning it on in place costs
about 40 seconds of compile.

Verify:

```
cmake -B build -G Ninja -DEVEREST_ENABLE_RS_SUPPORT=ON
ninja -C build && ninja -C build install/strip
readelf -d <prefix>/libexec/everest/modules/RsExample/RsExample | grep RUNPATH
```

Six ctests cover the regressions, including one that fails if a CMake `COMMAND`
shells out to `echo` again, and one that asserts the runpath on a built module.

No C++ changes.


cmake/lint/cmake-file-generation.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/sh
+# Guards how CMake writes files that later stages read.
+#
+# Usage: cmake-file-generation.sh <source-root> [<build-root>]
+#
+# Check 1 is static, so its verdict does not depend on what /bin/sh is on the
+# machine running it. That matters: with bash as /bin/sh a shelled-out
+# `echo -e` looks correct, and the breakage only shows up where /bin/sh is
+# dash, which is the platform EVerest ships on.
```
hikinggrass: That doesn't seem to be correct, might be true for the ci, but EVerest certainly doesn't ship on just one specific platform requiring dash

cmake/lint/gitignored-sources.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/sh
+# Fails if a source file the build needs is ignored by git.
+#
+# Usage: gitignored-sources.sh <source-root>
+#
+# .gitignore's leading `*build*` matches every Rust module's build.rs, so
```
hikinggrass: Since this is changed in this PR (.gitignore having `!build.rs` added to it) isn't this whole script already obsolete?

tests/CMakeLists.txt
```diff
@@ -44,3 +44,16 @@ add_subdirectory(async_api_tests)
...
+    COMMAND
+        ${PROJECT_SOURCE_DIR}/cmake/lint/cmake-file-generation.sh
+        ${PROJECT_SOURCE_DIR}
+        ${PROJECT_BINARY_DIR}
+)
+
+add_test(
+    NAME gitignored_sources
+    COMMAND ${PROJECT_SOURCE_DIR}/cmake/lint/gitignored-sources.sh ${PROJECT_SOURCE_DIR}
+)
```
hikinggrass: Same as above, probably already obsolete by the fix to `.gitignore`

hikinggrass (changes requested):

cmake/lint/cmake-file-generation.sh
mlitre: Done fixed

hikinggrass (approved):

.ci/build-kit/scripts/install.sh
```diff
@@ -7,3 +7,21 @@ if [ $retVal -ne 0 ]; then
     echo "Installation failed with return code $retVal"
     exit $retVal
 fi
+
+# The Rust modules are the only ones whose runpath nothing else checks: cargo
```
dorezyuk, line 11: i somewhat cannot parse this sentence

.ci/build-kit/scripts/install.sh
```diff
@@ -7,3 +7,21 @@ if [ $retVal -ne 0 ]; then
...
+# emits none and install(PROGRAMS) does no RPATH rewriting, so a regression here
+# is invisible until a module is started outside the development run scripts.
+m="$EXT_MOUNT/dist/libexec/everest/modules/RsExample/RsExample"
+readelf -d "$m" | grep -q 'RUNPATH.*\$ORIGIN' || {
```
dorezyuk, line 15: in the bazel x86 build we're linking statically to libframework.so to avoid that. If we want to keep the dynamic linking I think the cleanest way should be to include the runpath in the binary - this can be emitted in the build.rs -> there we have already code which detects if the binary is build with cmake or not

.ci/build-kit/docker/Dockerfile
```diff
@@ -20,3 +20,21 @@ RUN npm install -g corepack@0.33.0 \
     && apt-get -y remove npm \
     && corepack enable npm --install-directory /usr/bin \
     && corepack prepare npm@10.9.8 --activate
+
+# Rust toolchain for EVEREST_ENABLE_RS_SUPPORT. Pinned rather than floating, and
+# kept in sync with RUST_VERSION in MODULE.bazel so the two lanes agree;
+# Debian's packaged rustc is 1.63 and far too old. rustfmt is not part of the
+# minimal profile, and everestrs-build ignores a failing rustfmt, so leaving it
+# out ships unformatted generated code without saying so.
+ARG RUST_VERSION=1.86.0
```
dorezyuk, line 29: is this also ending up in the dev container?

cmake/everest-generate.cmake
```diff
@@ -315,54 +315,113 @@ if (EVEREST_ENABLE_RS_SUPPORT)
...
+# The everestrs crates are members too, by way of the symlinks below. Only
+# the modules are built.
+default-members = [
+  \"$<JOIN:$<TARGET_PROPERTY:generate_rust,RUST_MODULE_LIST>,\", \">\",
+]
+
+[workspace.dependencies]
+everestrs = { path = \"everestrs\" }
+everestrs-build = { path = \"everestrs-build\" }
+everestrs-derive = { path = \"everestrs-derive\" }
```
dorezyuk, line 336: do we need this?

cmake/everest-generate.cmake
```diff
@@ -315,54 +315,113 @@ if (EVEREST_ENABLE_RS_SUPPORT)
...
+        OUTPUT ${RUST_LINK_DEPENDENCIES_FILE}
+        CONTENT "$<JOIN:${RUST_LINK_DEPENDENCIES},\n>\n"
     )
 
-    add_custom_target(generate_rust
-        DEPENDS
-            ${RUST_WORKSPACE_CARGO_FILE}
-            ${RUST_LINK_DEPENDENCIES_FILE}
-    )
+    # Both files above are written at configure time, so they are inputs to
```
dorezyuk, line 361: that is cargo.lock and everest-link-dependencies.txt right?

cmake/everest-generate.cmake
```diff
@@ -315,54 +315,113 @@ if (EVEREST_ENABLE_RS_SUPPORT)
...
+    # cargo rather than build-time outputs; listing them as DEPENDS would turn a
+    # deleted file into a ninja error with no rule to recreate it.
+    add_custom_target(generate_rust)
+
+    # Cargo counts a path dependency living inside the workspace directory as a
+    # member, and the lock is a lock over that member set. Linking the crates in
+    # rather than pointing at them keeps the two sets equal, which is what lets
```
dorezyuk, line 368: which 2 sets?

lib/everest/framework/everestrs/everestrs-build/src/codegen.rs
```diff
@@ -132,6 +134,9 @@ struct YamlRepo {
     interfaces: HashMap<String, Interface>,
     data_types: HashMap<String, DataTypes>,
     error_types: HashMap<String, ErrorList>,
+    /// Every YAML file this repo actually read, so that a build script can
```
dorezyuk, line 137: reads

lib/everest/framework/everestrs/everestrs-build/src/codegen.rs
```diff
@@ -132,6 +134,9 @@ struct YamlRepo {
...
+    /// declare them and an edit to one is not invisible.
```
dorezyuk, line 138: what?

lib/everest/framework/everestrs/everestrs-build/src/codegen.rs
```diff
@@ -706,12 +729,28 @@ fn emit_config(config: BTreeMap<String, ConfigEntry>) -> Vec<ArgumentContext> {
...
 
 pub fn emit(manifest_path: PathBuf, everest_core: Vec<PathBuf>) -> Result<String> {
+    Ok(emit_with_inputs(manifest_path, everest_core)?.0)
+}
+
+/// Like [`emit`], but also returns every YAML file the generator read.
+pub fn emit_with_inputs(
+    manifest_path: PathBuf,
+    everest_core: Vec<PathBuf>,
+) -> Result<(String, Vec<PathBuf>)> {
```
dorezyuk, line 739: maybe we should add here a type instead of carrying through the tuple . Something like 
```rs
struct Emitted {
      generated_out: String,
      dependencies: Vec<PathBuf>
}
```

lib/everest/framework/everestrs/everestrs-build/src/codegen.rs
```diff
@@ -132,6 +134,9 @@ struct YamlRepo {
     interfaces: HashMap<String, Interface>,
```
dorezyuk, line 134: it might be cleaner to extend the hashmap to hold both, the deserialzed content of the file and the file path.

something like 

```rs

#[derive(Hash)
struct ParsedYamlBlob<T> {
      content: T,
      path: Pathbuf
}

...

fn lazy_load<'a, T: DeserializeOwned>(
    storage: &'a mut HashMap<String, ParsedYamlBlob<T>>,
    everest_root: &Vec<PathBuf>,
    prefix: &str,
    postfix: &str,

```
then interfaces, data_types and error_types just hold both, their content and the path

.ci/build-kit/scripts/install.sh
mlitre: reworded

.ci/build-kit/scripts/install.sh
mlitre: good call I made a fix in my latest commit you can take a look

.ci/build-kit/docker/Dockerfile
mlitre: I added it

cmake/everest-generate.cmake
mlitre: Yes everestrscargo.toml declares everestrs-derive with workspace = true so we have to provide it

cmake/everest-generate.cmake
mlitre: I improved the comment to make it less ambiguous as we meant cargo toml and everestrs link dependencies txt

cmake/everest-generate.cmake
mlitre: workspace member set and the cargo lock, I changed the wording

lib/everest/framework/everestrs/everestrs-build/src/codegen.rs
mlitre: reworded

lib/everest/framework/everestrs/everestrs-build/src/codegen.rs
mlitre: reworded

lib/everest/framework/everestrs/everestrs-build/src/codegen.rs
mlitre: Done

lib/everest/framework/everestrs/everestrs-build/src/codegen.rs
mlitre: I achieved the same goal of having a file unable to be read without it's path being recorded but I put them all in one place instead of in each map. I can still change it to your suggestion if you find any issues with it

PR #2808, opened 2026-09-16
fix(iso15118): cancel contactor timeout at stop

mlitre:
`PowerDelivery` returns early on every `ClosedContactor{false}` control event, before it
cancels the contactor close timeout armed by an earlier `PowerDeliveryReq(Start)`.
`EvseManager` publishes that event on `CPEvent::PowerOff` and `Evse15118D20` forwards it
during teardown, so a session that stops while still waiting for the contactor leaves the
timeout running. It then fires, logs an error, sends a FAILED response code on the wire and
sets `session_stopped` on a session that was already stopping.

This cancels the timeout once a shutdown has been requested, and drops the "waiting until
the contactor is closed" warning, which is untrue once the session is going away.

The contactor opens either way, driven by `EvseManager`, so this is a correctness and
log-honesty fix rather than a safety one.

The regression test drives the real timeout the way `Session::loop` does, and is red
before the change and green after.


lib/everest/iso15118/src/iso15118/d20/state/power_delivery.cpp
```diff
@@ -66,8 +66,15 @@ Result PowerDelivery::feed(Event ev) {
...
-                logf_warning(
-                    "Got ClosedContactor event, but contactor is not closed.  Waiting until the contactor is closed");
+                if (not m_ctx.shutdown_requested()) {
+                    logf_warning("Got ClosedContactor event, but contactor is not closed.  Waiting until the "
+                                 "contactor is closed");
+                } else if (previous_req.has_value()) {
+                    // The session is stopping, so the contactor will never close. Cancel the timeout armed
+                    // together with previous_req by PowerDeliveryReq(Start), which would otherwise fail a
+                    // session that is already going away.
+                    m_ctx.stop_timeout(d20::TimeoutType::CONTACTOR);
```
SebaLukas, line 75: I am wondering, is there the possibilty that the ClosedContactor event is sent before the PowerDeliveryReq is sent?

mlitre: Could be a race, but here we are just cancelling the contactor timeout if the ev stops the session or there is a session teardown. Basically more gracefully stopping

lib/everest/iso15118/test/iso15118/fsm/power_delivery.cpp
```diff
@@ -0,0 +1,83 @@
...
+            REQUIRE(result.transitioned() == true);
+            REQUIRE(fsm.get_current_state_id() == d20::StateID::AC_ChargeLoop);
+
+            const auto response_message = ctx.get_response<message_20::PowerDeliveryResponse>();
+            REQUIRE(response_message.has_value());
+            REQUIRE(response_message.value().response_code == dt::ResponseCode::OK);
+        }
+    }
+
+    GIVEN("The session tears down before the contactor closes") {
```
SebaLukas, line 63: I dont get the test.
Here the state machine should be in the AcChargeLoop state. How can the state_id == PowerDelivery.
In AcChargeLoop state after powerdelivery the state machine goes to SessionStop

lib/everest/iso15118/src/iso15118/d20/state/power_delivery.cpp
SebaLukas: So a session teardown before entering AcChargeLoop?

SebaLukas: If a session is shutdown before getting the PowerDeliveryReq, then the state machine is sending a Res and dont start the Contactor timeout -> thats fine
So I see a problem here. In the rare case that shutdown was triggered after the PowerDeliveryReq was sent and the timout was started, then there is no way right now for the state to send the PowerDeliveryRes back. Because the changes here stopped the timeout. So in this case it should be okay to send the res back

PR #2812, opened 2026-09-16
chore(cbv2g): regenerate at cbexigen bdf082f4c

mlitre:
## Describe your changes

Layer 1 of 4 in a stack that walks `lib/everest/cbv2g/` forward through four
cbexigen checkpoints. Each layer is based on the one below it and carries exactly
one thing: the vendored tree regenerated at that checkpoint, plus the provenance
line in `CMakeLists.txt`.

**Checkpoint:** cbexigen `bdf082f4cadafca846c3cbc474e3c76d01ebc068`,
*"SchemaAnalyzer: ensure identical order of elements"*, three commits on from the
`871fcb3` this tree recorded. Generated with `xmlschema 4.1.0`, the pin that
commit's own `requirements.txt` names.

### How the diff was produced

Everything under `include/` and `lib/` is `python src/main.py` output from a
clean checkout of that commit. The only hand edit in the whole stack is the one
`message(STATUS ...)` line at `CMakeLists.txt:34`. Between the generator and this
repository sits one scripted mechanical step, the same one the original vendoring
did: split the single output folder per protocol into `include/cbv2g/<p>/` and
`lib/cbv2g/<p>/`, and prefix a header include with `cbv2g/<p>/` when it crosses a
directory. That script was validated by regenerating at the *recorded* pin
`871fcb3`: 45 of 57 files come back byte identical to what is on `main`.

### What this layer actually changes

11 generated files, **12,744 insertions and 12,744 deletions**. Nothing is added
or removed; it is pure reordering of elements and grammar ids in ISO 15118-2 and
in the ISO 15118-20 AC and DC codecs.

What it buys is **reproducibility**. At `871fcb3` the generator emitted two
different ISO 15118-2 trees depending only on whether Python found a
`__pycache__` to load. Each was reproducible on its own, and they were about
11,300 lines apart:

| generator ref | cold runs (no `__pycache__`) | warm runs | agree? |
| --- | --- | --- | --- |
| `871fcb3` | 3/3 identical | 5/5 identical | **no**, 4 files / ~11,300 lines apart |
| `bdf082f4c` and later | 3/3 identical | 5/5 identical | **yes**, byte identical |

So from this checkpoint on, what this repository vendors can be reproduced from
the generator version the `CMakeLists.txt` records. That is what made the rest of
the stack measurable: every layer above was diffed against its own two-arm churn
first, and every one of them has zero churn to subtract.

### What is deliberately not regenerated

- **DIN 70121.** Its schemas are not publicly distributable and are not on the
  machine that ran this, so the six `din_msgDef*` files stay exactly as `871fcb3`
  emitted them. This is the one artifact the stack cannot produce.
- **ISO 15118-20 AC-DER.** Absent from this checkpoint's generator config
  entirely; it arrives in layer 4.

### The wire format does not move

12,000 reordered lines in ISO 15118-2 deserve more than a compile. Nothing in
this repository tests ISO 15118-2 at all: the six `cbv2g` tests cover
appHandshake, DIN and ISO 15118-20 only. So I encoded seven ISO 15118-2 messages
through each arm and compared the bytes:

`SessionSetupReq`, `ServiceDiscoveryReq` (optional enum), `ChargingStatusReq`
(empty body), `SessionStopReq` (enum), `PreChargeReq` (nested `DC_EVStatus` and
two `PhysicalValueType`), `PowerDeliveryRes` with and without its `EVSEStatus`
branch.

**Every one is byte identical to `origin/main`**, and every one round trips.
This holds for layers 2 and 3 as well.

### Verification

All 37 translation units compile warning free under
`gcc -std=c99 -Wall -Wextra`, and the six `cbv2g` tests pass.


SebaLukas:
It would be better to have also the din messages regenerated

SebaLukas (approved):

PR #2818, opened 2026-09-16
fix(LemDCBM400600): send plain Content-Length requests, not chunked

florinmihut:
## Problem

The LemDCBM400600 HTTP client feeds PUT/POST bodies through `CURLOPT_READFUNCTION` without declaring their size. libcurl therefore sends them as `Transfer-Encoding: chunked` and adds an `Expect: 100-continue` header.

The LEM DCBM tolerates this, but other meters implementing the same DCBM REST API with a minimal embedded HTTP server do not. Concrete case: the **AST DC650 with the K220031 detached display unit** rejects chunked requests with `400` and the `Expect` header with `405`. In practice this breaks the driver's time synchronization (`PUT /v1/settings` fails on every retry) and, since `start_transaction` runs a time sync first, would make **every transaction start fail**.

Verified against the device: the identical JSON payloads succeed with a plain `Content-Length` request and fail when sent chunked or with `Expect: 100-continue`.

## Fix

In `setup_connection()`:
- declare the request body size (`CURLOPT_INFILESIZE` for PUT, `CURLOPT_POSTFIELDSIZE` for POST) so libcurl sends a plain `Content-Length` request,
- suppress the `Expect` header.

Both changes are no-ops for the real LEM DCBM, which accepts plain requests as well.

## Testing

- Bench with an AST DC650 + K220031 display unit (LEM-compatible REST API): before the fix, `Failed to sync time settings: ... 400` every retry; after the fix, zero sync warnings, device time kept in sync, transactions startable.
- `GET` paths unaffected (size options only set for non-empty bodies).


hikinggrass (approved):
looks good, just needs a sign-off

PR #2832, opened 2026-09-17
Adding the missing ac contactor check if the session is shutdown

SebaLukas:
## Describe your changes
Add almost the same pattern for checking the status from PowerDelivery to AcChargeLoop

Right now the status of the contactor is checked while starting the session. Not while stopping.

Tested on the SIL and checked in the logs


SebaLukas:
Need to align with #2808

lib/everest/iso15118/include/iso15118/d20/state/ac_charge_loop.hpp
```diff
@@ -26,6 +27,9 @@ struct AC_ChargeLoop : public StateBase {
     AcPresentPower present_powers{};
 
     bool first_entry_in_charge_loop{true};
+
+    bool ac_connector_closed{true}; // In this state, we assume that the contactor is closed
```
mlitre, line 31: Just wondering, but this might be a future improvement, should we make this part of the constructor for the state or something in the context so that we don't have to assume anything

lib/everest/iso15118/src/iso15118/d20/state/ac_charge_loop.cpp
```diff
@@ -206,6 +250,20 @@ Result AC_ChargeLoop::feed(Event ev) {
 
         const auto shutdown_requested = m_ctx.shutdown_requested();
 
+        // If the car wants to stop the session and the contactor is closed
+        if (req->charge_progress == dt::Progress::Stop and ac_connector_closed) {
+            previous_req = *req;
+            // Open the AC contactor
+            m_ctx.feedback.signal(session::feedback::Signal::AC_OPEN_CONTACTOR);
+            m_ctx.feedback.signal(session::feedback::Signal::CHARGE_LOOP_FINISHED);
```
mlitre, line 258: Is this signal a bit premature since we are still in charge loop?

SebaLukas: In this case we are in PowerDelivery. The libiso state is only AcChargeLoop

lib/everest/iso15118/include/iso15118/d20/state/ac_charge_loop.hpp
SebaLukas: Yeah maybe

lib/everest/iso15118/src/iso15118/d20/state/ac_charge_loop.cpp
```diff
@@ -190,12 +194,52 @@ Result AC_ChargeLoop::feed(Event ev) {
...
+            ac_connector_closed = *control_data;
+
+            if (ac_connector_closed) {
+                logf_warning(
+                    "Got ClosedContactor event, but contactor is not opened. Waiting until the contactor is opened");
+                return {};
+            }
+
+            const auto* active_timeout = m_ctx.get_active_timeout();
+            if (active_timeout != nullptr and *active_timeout == d20::TimeoutType::CONTACTOR) {
```
mlitre, line 207: I just realized that this check only checks the type of the timeout not if the timeout was actually reached

lib/everest/iso15118/src/iso15118/d20/state/ac_charge_loop.cpp
```diff
@@ -215,15 +273,13 @@ Result AC_ChargeLoop::feed(Event ev) {
             return {};
         }
 
-        // V2G20-3210 -> state machine direct transition (skipped PowerDelivery)
-        if (req->charge_progress == dt::Progress::Stop or shutdown_requested) {
-            m_ctx.feedback.signal(session::feedback::Signal::CHARGE_LOOP_FINISHED);
-            m_ctx.feedback.signal(session::feedback::Signal::AC_OPEN_CONTACTOR);
+        if (req->charge_progress == dt::Progress::Stop and not ac_connector_closed) {
```
mlitre, line 276: Is this supposed to be not closed or closed? Because if we are supposed to wait for the contactor to open before moving to session stop that's not happening here

lib/everest/iso15118/src/iso15118/d20/state/ac_charge_loop.cpp
```diff
@@ -206,6 +250,20 @@ Result AC_ChargeLoop::feed(Event ev) {
...
+            m_ctx.feedback.signal(session::feedback::Signal::AC_OPEN_CONTACTOR);
+            m_ctx.feedback.signal(session::feedback::Signal::CHARGE_LOOP_FINISHED);
+
+            m_ctx.start_timeout(d20::TimeoutType::CONTACTOR, AC_OPEN_CONTACTOR_TIMEOUT);
+
+            logf_info("Waiting for contactor is opened"); // [V2G20-863]
+            return {};
+        }
+
+        // The contactor is already opened
```
mlitre, line 266: Should call the feedback signal to let the higher layers that charge loop is finished?

mlitre (commented):
I also just wondered if this should also apply to ac der iec

lib/everest/iso15118/src/iso15118/d20/state/ac_charge_loop.cpp
SebaLukas: Yeah thats correct. It could happen that the car is too fast going into b and the charger opens contactor before libiso signals that.
In that case m_ctx.stop_timeout() is printing a warning that the timeout was not started.
So this check for an active_timeout is only for preventing the warning

lib/everest/iso15118/src/iso15118/d20/state/ac_charge_loop.cpp
SebaLukas: ?

lib/everest/iso15118/src/iso15118/d20/state/ac_charge_loop.cpp
SebaLukas: In this case the contactor is already opened and the state machine can go directly in SessionStop. In this case the timeout and the signal to open the contactor is not sent

SebaLukas:
> I also just wondered if this should also apply to ac der iec

Yeah ofc. Good catch

mlitre (approved):
lgtm

PR #2850, opened 2026-09-19
fix(iso15118): reduce schedule exchange stack usage

corneliusclaussen:
The main issue here was to have a 1024 element std::array, which was used multiple times and also had nested arrays and hence created a pretty large stack element.  Now we're using std::vector for this max 1024 large element to fix that (the library is anyway far from no-heap-alloc).

Furthermore the `respond_with_code` function was retired, as it forbid NRVO.  Now we just use `set_response_code` and return the response by ourself.


a-w50 (dismissed):
The main issue here was to have a 1024 element std::array, which was used multiple times and also had nested arrays and hence created a pretty large stack element.  Also fixed the NRVO handling (response_with_code -> set_response_code)

SebaLukas (changes requested):
The change from `response_with_code()` to `set_response_code()` has nothing to do with reducing the schedule exchange stack usage.
So please split the PR into two PRs

mlitre (commented):
I would have preferred a fix where we made a dynamically allocated version of the fixed vector instead of just replacing it by an std vector. I feel like that would be more homogeneous like that

lib/everest/iso15118/test/exi/cb/iso20/schedule_exchange.cpp
```diff
@@ -353,4 +353,91 @@ SCENARIO("Se/Deserialize schedule_exchange messages") {
             REQUIRE(control_mode.departure_time == 2000);
         }
     }
+
+    GIVEN("Se/Deserialize schedule_exchange_res - scheduled mode with an absolute price schedule") {
```
SebaLukas: An exi stream is here missing.

lib/everest/iso15118/test/iso15118/fsm/schedule_exchange_dc.cpp
```diff
@@ -0,0 +1,51 @@
...
+// Copyright 2026 Pionix GmbH and Contributors to EVerest
+#include <catch2/catch_test_macros.hpp>
+
+#include "helper.hpp"
+
+#include <iso15118/d20/state/schedule_exchange.hpp>
+
+using namespace iso15118;
+
+SCENARIO("ISO15118-20 scheduled DC schedule exchange state transitions") {
```
SebaLukas, line 15: Check the other fsm tests for that. We use for each test `GIVEN` and `THEN`

lib/everest/iso15118/include/iso15118/message/schedule_exchange.hpp
```diff
@@ -50,17 +49,19 @@ struct PriceRule {
...
     RationalNumber service_fee;
 };
 
-using TaxRuleList = std::array<TaxRule, TAX_RULE_LENGTH>;
-using PriceRuleStackList = std::array<PriceRuleStack, PRICE_RULE_STACK_LENGTH>;
-using AdditionalServiceList = std::array<AdditionalService, ADDITIONAL_SERVICE_LENGTH>;
+using TaxRuleList = everest::lib::util::fixed_vector<TaxRule, TAX_RULE_LENGTH>;
+// Inline storage is not used here: the standard allows 1024 stacks, which would add over a megabyte
+// to every ScheduleExchangeResponse. The encoder bounds this against the codec's own limit of 64.
+using PriceRuleStackList = std::vector<PriceRuleStack>;
```
SebaLukas, line 65: Is the length of the vector checked against the max length of cbv2g PriceRuleStackList array? A fixed_vector with a max length of 64 would be better

SebaLukas (changes requested):
Overall looking good!
Only some small stuff you need to change

lib/everest/iso15118/test/exi/cb/iso20/schedule_exchange.cpp
a-w50: Do we have a real world exi stream containing a `ScheduleExchangeRes`, filled with charging schedules, price schedules, tax rules etc?  If so, I'm happy to add it.  Otherwise we could only generate the exi by ourself, i.e. what the test is doing right now.

lib/everest/iso15118/include/iso15118/message/schedule_exchange.hpp
a-w50: Yes, this has been done before already with the std::array, in `message/schedule_exchange.cpp` line 398 and it just throws if it is out of bounds.  This should be one the todo list for the future, i.e. how to handle decode/encode errors.

lib/everest/iso15118/test/exi/cb/iso20/schedule_exchange.cpp
a-w50: Ok, used exificient to compute the raw exi

SebaLukas (approved):
Looking good!

PR #2852, opened 2026-09-19
fix(ocpp): use OcspRequestInterval default if missing

### LINKED ISSUE #1991: Configured OCSPRequestInterval is not used at startup

Tarekchehahde:
Hi,

#1991 is still open after #2428. That PR already reads `OcspRequestInterval` in the timer, but I ran into a couple leftovers while looking at it.

The fallback comment said 12 hours, while the value is 7 days (same as the Internal.json default). And if the key is missing, `operator[]` inserts null and nlohmann throws `type_error`, which is not a `std::runtime_error`, so the catch never ran and the fallback was dead.

Getter now returns the schema default when the key is missing or not an integer. The timer catches `std::exception`. I also added tests for the default, the 86400 minimum, and `config.json` without the key.

## Describe your changes
Use the OcspRequestInterval schema default when the key is missing, and catch json errors in the OCSP timer.

Fixes #1991


Pietfried (changes requested):
Thanks for the cleanup. The comment fix, named constant and `std::exception` catch are good.

The missing-key bug does not occur, though: `ChargePointConfiguration` patches schema defaults into the config at construction, and Internal.json has had `"default": 604800` since the key was introduced. Your new `ConfigWithoutOcspRequestInterval` test also passes with main today and `Config/Configuration.OcspRequestInterval/json` already covers `config.json` without the key.

Please reword the description as a cleanup of #2428 leftovers and drop the duplicate test.

PR #2854, opened 2026-09-21
fix(util): make FixedSizeScalingGrowsAtLimit deterministic

a-w50:
- The test pushed a blocking task and a second task back to back, assuming both were still queued when the second push evaluated the scaling policy.
- When the sole worker popped the blocking task first, the second push saw queue_size 1, FixedSizeScaling<2> declined to grow, and with no supervisor tick nothing ever re-evaluated, so the task never ran.
- Wait until the blocking task is confirmed running, then push both tasks, so the queue depth the policy sees is fixed rather than racing the worker.
- Also assert that depth 1 does not grow, covering both sides of the limit.
- Failed 8/100 locally before, 0/400 after (0/150 under CPU load).

Assisted-by: Claude Opus 5 (1M context) <noreply@anthropic.com>


mlitre (approved):
lgtm

PR #2855, opened 2026-09-21
Fix carlo gavazzi em580 live measurement interval

MichaelSaur-chargeBIG:
### Bug:
The powermeter module for CarloGavazzi EM580 has a config value to define the interval of measurment updates. 
`` live_measurement_interval_ms ``
This interval was used as a delay between reads instead of actually defining the interval of updates published to the broaker.
### Fix:
I changed the logic to actually push new measurement values with the exact interval specified.
If the interval can't be held (bus overload,...) the module logs a warning every 60s containing the minimum time needed.

- None

- [x] I have performed a self-review of my code -> Tested on hardware
- [x] I have made corresponding changes to the documentation -> Code now does what the docs already said it would do


modules/HardwareDrivers/PowerMeters/CarloGavazzi_EM580/main/powermeterImpl.cpp
```diff
@@ -578,10 +580,25 @@ void powermeterImpl::ready() {
...
+                auto time_to_wait = measurement_interval - (now - measurement_start);
+                if (time_to_wait > std::chrono::milliseconds{0}) {
+                    last_interval_overrun_warning = std::chrono::steady_clock::time_point{};
+                    std::unique_lock<std::mutex> lock(stop_mutex_);
+                    stop_cv_.wait_for(lock, time_to_wait, [this] { return stop_requested_.load(); });
+                } else {
+                    if (last_interval_overrun_warning == std::chrono::steady_clock::time_point{} ||
+                        (now - last_interval_overrun_warning) >= std::chrono::seconds{60}) {
+                        last_interval_overrun_warning = now;
+                        EVLOG_warning << "Measurement took longer than configured interval, consider increasing "
```
barsnick: I consider it useful to also print the actual configured interval.
```suggestion
                        EVLOG_warning << "Measurement took longer than configured interval " << measurement_interval << "ms, consider increasing "
```

